PR #3, opened 2017-01-13
Fix parser

daejunpark:
@grosu please review. Several parsing errors/bugs are fixed, which is reflected in the unit tests.


daejunpark:
Codeship build test: https://app.codeship.com/projects/130795/builds/21679165

grosu:
@daejunpark the first 3 tests fail with minor differences in the output (there could only be white-space differences, like differences between end-of-line in windows vs linux).

also, for some reason, my IntelliJ setting shows all the outputs of the parseTestFail tests when I click on parseTest4.  not sure if this is my misunderstanding, or my misconfiguration, or a bug in intelliJ.

grosu:
Some recommendations:

1) It feels that lookahead, nextWithSpaces(), next(), putback() belong to the Scanner not to the parser.  Or maybe Scanner needs to be promoted to some more important class, something in charge of the "head" of the parser.

2) I would rename "expect(str)" to "parse(str)" or "consume(str)", because you actually move the head of the parser as well, not only check an expectation.

3) Similarly, I would add a function "parseWhiteSpaces()" or "consumeWhiteSpace()" that consumes all the whitespaces moving the head to the first non-white-space.  This is intuitive and thus helps understanding the parser and later on migrate it to other languages.  This way you can eliminate the non-intuitive "putback(next())" on line 112 in TextToMini as a bonus.  In the process, this function can also call readLine() for you, so it gives a nice abstraction.

daejunpark:
All the comments are addressed in PR #4.

PR #4, opened 2017-01-14
Fix parser

daejunpark:
@grosu please review. Addressed was your previous comments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,6 +146,7 @@
           </execution>
         </executions>
       </plugin>
+      <!--
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-gpg-plugin</artifactId>
@@ -160,6 +161,7 @@
           </execution>
         </executions>
       </plugin>
+      -->
     </plugins>
   </build>
 </project>
--- a/src/test/scala/org/kframework/minikore/TextToMiniTest.scala
+++ b/src/test/scala/org/kframework/minikore/TextToMiniTest.scala
@@ -445,4 +445,18 @@ class TextToMiniTest {
     }
   }
 
+  @Test def errorTest(): Unit = {
+    new TextToMini().error('a', "abc") match {
+      case ParseError(msg) =>
+        assertEquals(
+          """
+            |ERROR: Line 0: Column 0: Expected 'a', but abc
+            |null
+            |^
+            |""".stripMargin.stripPrefix(System.lineSeparator()).stripSuffix(System.lineSeparator()),
+          msg)
+      case _ => assert(false)
+    }
+  }
+
 }
--- a/src/test/scala/org/kframework/minikore/TextToMiniTest.scala
+++ b/src/test/scala/org/kframework/minikore/TextToMiniTest.scala
@@ -21,7 +21,7 @@ class TextToMiniTest {
 
   @Test def parseTest4(): Unit = {
     val s =
-      """
+      strip("""
         |[]
         |module A
         |  import B []
@@ -32,25 +32,25 @@ class TextToMiniTest {
         |  axiom ` f o o `() []
         |  axiom ` _,_( ) `() []
         |endmodule []
-        |""".stripMargin
+        |""")
     parseFromString(s)
   }
 
   @Test def parseTest5(): Unit = {
     val s =
-      """
+      strip("""
         |[]
-        |""".stripMargin
+        |""")
     parseFromString(s)
   }
 
   @Test def parseTest6(): Unit = {
     val s =
-      """
+      strip("""
         |[]
         |module A
         |endmodule []
-        |""".stripMargin
+        |""")
     parseFromString(s)
   }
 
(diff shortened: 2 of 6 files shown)


src/test/scala/org/kframework/minikore/TextToMiniTest.scala
```diff
@@ -237,45 +243,103 @@ class TextToMiniTest {
...
+      assert(false)
+    } catch {
+      case ParseError(msg) =>
+        assertEquals(
+          """
+            |ERROR: Line 4: Column 11: Expected \true, \false, \and, \or, \not, \implies, \exists, \forall, \next, \rewrite, or \equal, but 't\n'
+            |  axiom \t
+            |          ^
+            |""".stripMargin.stripPrefix(System.lineSeparator()).stripSuffix(System.lineSeparator()),
+          msg)
```
grosu: We should have 100% code coverage in minikore, but all coverage criteria.  So we'll need more tests, as we only have 92% coverage in some classes.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
```diff
@@ -70,7 +70,7 @@ class TextToMiniTest {
             |ERROR: Line 4: Column 8: Expected 't', but ' '
             |  impor t B []
             |       ^
-            |""".stripMargin.stripPrefix("\n").stripSuffix("\n"),
+            |""".stripMargin.stripPrefix(System.lineSeparator()).stripSuffix(System.lineSeparator()),
```
daejunpark: Thanks for spotting this issue, but this change is still not correct. The newline characters put by the multiline string don't depend on the platform but on the text editor. I can still put `\r\n` even in mac, where you can still put `\n` in windows, depending on the setting of your text editor.

Just to make sure, did your tests fail without this change?

daejunpark: Sorry, now I see why it failed. I'll fix it.

grosu:
> Just to make sure, did your tests fail without this change?

Yes, the tests failed with just "\n".  That's why I changed the code.

> The newline characters put by the multiline string don't depend on the platform but on the text editor.

Aren't the text editors consistent with the platform's conventions?

> I can still put \r\n even in mac, where you can still put \n in windows, depending on the setting of your text editor.

Are you sure this is not too fragile?  I would say that if my change does not break the tests on your platform, then we should go with it.

It would be nice to have a clean solution here.  Working explicitly with "\n" and "\r" seems fragile to me.

pom.xml
```diff
@@ -160,6 +161,7 @@
           </execution>
         </executions>
       </plugin>
+      -->
```
daejunpark: Deploying to the maven central requires gpg singing, while deploying to snapshot repository doesn't. For now, we deploy only to the snapshot repository. I'll think about a better way to handle this later.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
daejunpark: Ok, I can go with the current fix, since it will work mostly unless one has weird text editor settings.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
daejunpark: I added more tests to increase the test coverage, but still not 100%. Having 100% is impossible or nonsense, partly due to scala-specific issues of intellij coverage measurement, and conservative error handling codes that are not possible to reach for now. If you can find uncovered codes that should be covered, let me know.

daejunpark:
@grosu more tests are added. Now every line is covered except ones with the explicit comment.

grosu:
thanks, @daejunpark.

i found a bug, when <EOF> is encountered before the entire definition is parsed.  I added a test for it, parseTestFailUnfinishedModule, but will let you fix the code.

also, i found that some of error messages were not precise wrt that follows next, so I added a couple more tests with the desired error message.

in the process, i also thought that it would be a good idea to add a strip function that does all we need with these multistrings.  the code is a bit more compact now.  sorry i didn't put that in a separate commit.

grosu:
for the sake of completeness, i think we should also have tests where `<EOF>` can occur in the middle of parsing everything.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
```diff
@@ -61,347 +61,443 @@ class TextToMiniTest {
...
+  @Test def parseTestFailWhiteSpacesPrefix(): Unit = {
+    val s =
+      """
+
+
+
+        |[]
+        |module A
+        |endmodule [}
+        |""".stripMargin
```
daejunpark: This is not a valid multiline string in Scala. You missed the bar `|` in the beginning. Is that your intention? Also, you missed applying `strip`; is this also your intention?

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
```diff
@@ -61,347 +61,443 @@ class TextToMiniTest {
...
+    try {
+      parseFromString(s)
+      assert(false)
+    } catch {
+      case ParseError(msg) =>
+        assertEquals(
+          strip("""
+                  | add corresponding error message
+                  |"""),
```
daejunpark: This is expected. Indeed, either `ParseError` or `EOFException` will be caught in `krun`. But, I think it is a good idea that the parser catches `EOFException` and uniformly throws the `ParseError` exception.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
grosu: I skipped the strip(...) call on purpose, to also have tests with white lines at the beginning, as he name of the test says.

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
grosu: Yes, it makes sense to throw an error if the EOF is reached before a complete definition is parsed.  It would also be nice to give precise error messages, as I suggested in the tests, so there is no doubt about what the problem was.

daejunpark:
First see my previous two comments.

Regarding the error messages, it is not simple to generate such error messages that you suggested. Yes, I can add an ad-hoc support for that cases, but you will always find another case. In general, providing a high-quality parsing error message is a challenging problem, for which ones still write paper. The general support for more kind error messages should be a separate project, which I'll not implement in this pull request.

Also, this pull request became already very long, and now it is hard to keep track of the changes properly. You should merge this pull request when it is in a reasonable shape and request another pull request for remaining issues. Otherwise, it will never be merged, and grow too bigger to be able to keep track of.

grosu:
We are very close to a perfect solution here, aren't we?  I am OK to merge this PR asap if you think it is getting too long.  But at least we should have only passing tests.  Are you going to catch the EOF exception and generate a ParseError in this PR, or you want to do it in the next one?

daejunpark:
I'm working on the EOF exceptions. I'll disable or comment out the enhanced parsing error messages. Also, I'll enhance the strip function to support the cross-platform editing environment.

grosu:
Looks good.

PR #5, opened 2017-01-17
Fix newline

grosu:


src/main/scala/org/kframework/minikore/Scanner.scala
```diff
@@ -49,7 +49,7 @@ class Scanner {
           input.next()
         } else if (!yieldEOL) { // end of line
           yieldEOL = true
-          '\n'
+          ' '   // safer than '\n' (platform independent)
```
daejunpark: I can understand why you did this, but it could be misleading for the new members. I have no better idea though. Maybe, more detailed comments may help..

src/test/scala/org/kframework/minikore/TextToMiniTest.scala
```diff
@@ -281,7 +281,7 @@ class TextToMiniTest {
       case ParseError(msg) =>
         assertEquals(
           strip("""
-            |ERROR: Line 3: Column 11: Expected \true, \false, \and, \or, \not, \implies, \exists, \forall, \next, \rewrite, or \equal, but 't '
+            |ERROR: Line 3: Column 11: Expected \true, \false, \and, \or, \not, \implies, \exists, \forall, \next, \rewrite, or \equal, but '\t '
```
daejunpark: This error message seems misleading. `\t` could be seen as a tab character, instead of the two characters. That' why I didn't put the leading backslash..

grosu: I actually found it confusing the other way, without a `\`.  The next line shows the exact position anyway.  I am OK either way.

src/main/scala/org/kframework/minikore/Scanner.scala
grosu: I just tried to eliminate any explicit references to `\n`, because that seems to be platform-specific.  Ideally, I would have liked to replace that with System.lineSeparator(), but that returns a string, because on some platforms the end of line is `\r\n`.

PR #6, opened 2017-01-18
Javadoc

daejunpark:


daejunpark:
@grosu please review. The codeship build failed because the change made in K for this PR hasn't been merged into master. I guess we need to have a better way of the integration tests; I have an idea though.

grosu:
In TextToMini I can only see 2 members in the generated scaladoc html.  I think we should see all of them.  We want to even put a link to this scaladoc on the webpage of kore later on, so people know what it provides without looking at the code.

daejunpark:
All the other methods other than the two are private. You cannot call those private methods, and will be never shown in the scaladoc. If you're talking about the grammar, I think we can add a file for the grammar in the resource directory and put its link in the scaladoc.

PR #7, opened 2017-01-30
minikore interface proposal

daejunpark:
do not merge. just for a review.


src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  trait Pattern
+
+  trait Variable extends Pattern {
+    def name: String
+    def sort: String
+  }
+
+  trait Application extends Pattern {
+    def label: String
```
grosu: symbol not label anymore, right?

daejunpark:
@grosu please take a look. we can talk more in person.

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
daejunpark: Yes. I'll fix it in a separate PR.

daejunpark:
@msaxena2 I just updated. Please take a look again.

daejunpark:
@grosu @msaxena2 Here is 2nd proposal. Please take a look.

PR #9, opened 2017-01-31
Minikore refactoring

msaxena2:
WIP. Do not merge yet.


src/main/scala/org/kframework/minikore/Constructors.scala
```diff
@@ -0,0 +1,97 @@
+package org.kframework.minikore
+
+import org.kframework.minikore.MiniKore._
+
+
+// Constructor Factory.
```
msaxena2: Constructor abstract class to create the kore data structures. The implementation is expected to implement the methods in this abstract class. The advantage of having a constructor interface is that tools (like the debugger), will just need a reference to the constructor's implementation to create new Kore data structures. A default implementation of this class is provided.

src/main/scala/org/kframework/minikore/Constructors.scala
```diff
@@ -0,0 +1,97 @@
...
+  def symbolDeclaration(sort: String, label: String, args: Seq[String], att: Attributes): SymbolDeclaration
+
+  def rule(p: Pattern, att: Attributes): Rule
+
+  def axiom(p: Pattern, att: Attributes): Axiom
+
+}
+
+// Provided Constructors over default objects.
+
```
msaxena2: Default constructor for the data structures. It's used in the tests.

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,310 @@
 package org.kframework.minikore
 
-import scala.collection._
+import org.kframework.minikore.MiniKore._
 
-/** Algebraic data type of MiniKore. */
+
+// The main Kore Interface.
 object MiniKore {
```
msaxena2: New Traits for AST-like properties of Kore data structures.

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,310 @@
...
+  }
+
+
   //
-  case class And(p: Pattern, q: Pattern) extends Pattern
-  case class Or(p: Pattern, q: Pattern) extends Pattern
-  case class Not(p: Pattern) extends Pattern
-  case class Implies(p: Pattern, q: Pattern) extends Pattern
-  case class Exists(v: Variable, p: Pattern) extends Pattern
-  case class ForAll(v: Variable, p: Pattern) extends Pattern
```
msaxena2: And extends both pattern and Node. So suppose patterns x, y, and p = Pattern(x, y). Now, you can write stuff like - 
p match {case Pattern(A:Pattern, B: Pattern) => ..} or

p match {case Node("\and", Seq(A: AST, B: AST)) => ...}.

src/main/scala/org/kframework/minikore/TextToMini.scala
```diff
@@ -38,7 +38,8 @@ class TextToMini {
   private def parseDefinition(): Definition = {
     val att = parseAttributes()
     val modules = parseModules(Seq())
-    Definition(modules, att)
+
```
msaxena2: Kore data structures cannot be directly created anymore, hence, the parser used the default constructor implementation.

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,310 @@
...
+
+  trait Rule extends Sentence {
+    val p: Pattern
+    val att: Attributes
+  }
+
+  object Rule {
+    def unapply(arg: Rule): Option[(Pattern, Attributes)] = Some((arg.p, arg.att))
+  }
+
```
msaxena2: Original case classes have been replaced by traits.

src/main/scala/org/kframework/minikore/Constructors.scala
grosu: I personally do not like this idea to group all the constructors into one place, separate from the destructors.  Since abstract classes allow constructors, I would rather have an abstract class corresponding to all the (mini)kore language/data-structures and then traits for other views of them.  In other words, instead of having traits for all interfaces, including the basic kore syntax/language/constructs as well as the Ast view, I would use an abstract class for the "main trait", with constructors in it which must be therefore enforced to be implemented, and traits for anything else.  This should also facilitate interfacing with Java (which seems to have problems with traits providing some default implementations).  Or is it exactly what you are doing and I'm missing something?

msaxena2: @grosu At the moment I'm using only traits, and no abstract classes. I like the idea of having the "main" classes as abstract classes instead of traits. However, most of the java backend Kil classes that'll extend the main kore abstract classes already extend other (unrelated to kore) abstract classes. A Java class can only extend one abstract class, but can implement multiple traits, so it'll be easier to make the kil classes implement (mini)kore if we use traits everywhere. As far as interoperability with traits goes, @cos suggested the issue was fixed in java8. 

As far as constructors are concerned, I'm ok with them being in the same place as the destructors.

daejunpark:
@msaxena2 could you close this PR and delete the branch, if not needed any more?

PR #10, opened 2017-02-16
Interface changes

msaxena2:
WIP. Some Fixes + Cleanup Pending


msaxena2:
This is ready for review. Note that the build, however, is marked as failed. This happens since the current version of K needs slight changes to build with these changes. Once we're close to the merge, I'll make the required changes to K, and that'll make the build pass.

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,162 @@
...
+
+  case class True() extends i.True {
+    override def build = DefaultBuilders.TrueBuilder
+  }
+
+  case class False() extends i.False {
+    override def build: Node0Builder[i.Pattern] = DefaultBuilders.FalseBuilder
+  }
+
+  case class And(override val p: i.Pattern, override val q: i.Pattern) extends i.And {
```
daejunpark: what does `override val` mean?

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,162 @@
...
+  case class Application(label: String, args: Seq[i.Pattern]) extends i.Application {
+    override def build = DefaultBuilders.ApplicationBuilder(label, _)
+  }
+
+
+  case class DomainValue(label: String, value: String) extends i.DomainValue {
+    override def build = DefaultBuilders.DomainValueBuilder
+  }
+
+  case class True() extends i.True {
```
cos: a `case class` with no arguments may as well be an `object`

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,162 @@
...
+    override def build: Node2Builder[i.Pattern] = DefaultBuilders.RewriteBuilder
+  }
+
+  case class Equals(p: i.Pattern, q: i.Pattern) extends i.Equals {
+    override def build: Node2Builder[i.Pattern] = DefaultBuilders.EqualsBuilder
+  }
+
+}
+
+object DefaultBuilders {
```
cos: is there a particular reason to keep the `Builder` implementations separate from the case classes above?

cos: is there a particular reason to keep the `Builder` implementations separate from the case classes above?

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+
+
+  sealed trait Node[T <: AST[T]] extends AST[T] with Product {
+    def children: Seq[_ <: T]
+
+    def build: NodeBuilder[T]
+  }
+
+  object Node {
+    def unapply[T <: AST[T]](arg: AST[T]): Option[Seq[_ <: T]] = arg match {
```
cos: seem like `arg` should be `T`

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  trait Node0Builder[T <: AST[T]] extends NodeBuilder[T] {
+
+    def apply(): T
+
+    override def apply(children: Seq[_ <: T]) = {
+      assert(children.isEmpty)
+      apply()
+
```
cos: extra new line

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+
+  sealed trait Leaf[T <: AST[T], C] extends AST[T] {
+    def contents: C
+
+    def build: LeafBuilder[T, C]
+  }
+
+  object Leaf {
+    def unapply[T <: AST[T], C](arg: AST[T]): Option[C] = arg match {
```
cos: `arg: T`

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+  trait LabelledNode[L, T <: AST[T]] extends Node[T] {
+    def label: L
+  }
+
+  trait LabelledNodeBuilder[L, T <: AST[T]] {
+    def apply(x: L, c: Seq[_ <: T]): T
+  }
+
+  object LabelledNode {
+    def unapply[L, T <: AST[T]](arg: AST[T]): Option[(L, Seq[_ <: T])] = arg match {
```
cos: `arg: T`

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+
+  trait VariableBuilder extends LeafBuilder[Pattern, (String, Sort)] {
+
+    def apply(name: String, sort: Sort): Variable
+
+    override def apply(contents: (String, Sort)) = apply(contents._1, contents._2)
+
+  }
+
+  object Variable {
```
cos: We touched upon this last week. Having objects with `unapply` declared at this level will introduce an ambiguity between them and the implementations of the `Builder` objects. E.g., writing `case And(x)` will fail at compile time because `And` is declared both here and as a `Builder`.

cos: I would move all these `unapply` methods into the `*Builder` traits.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,97 @@
...
+  @Test def mapTest1(): Unit = {
+    val e1: Pattern = t.b.Application("IntContainer", Seq(t.int1, t.int2))
+    val intSort: String = "Int"
+
+    def doubleInt: (Pattern) => Pattern = { (p) =>
+      p match {
+        case DomainValue(intString: String, `intSort`) => t.b.DomainValue((intString.toInt * 2) toString, intSort)
+        case other => other
+      }
+    }
```
cos: Some tests for destructors will highlight the `Builder`-object ambiguity I mentioned above.

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
cos: I think `Labelled` should have only one `l` in US english.

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -0,0 +1,463 @@
...
+  trait LeafBuilder[T <: AST[T], C] {
+    def apply(contents: C): T
+  }
+
+
+  trait LabelledNode[L, T <: AST[T]] extends Node[T] {
+    def label: L
+  }
+
+  trait LabelledNodeBuilder[L, T <: AST[T]] {
```
cos: Why doesn't this extend `NodeBuilder`?

src/main/scala/org/kframework/minikore/MiniKore.scala
grosu: They are compiled the same way, aren't they?  If yes, I'd rather keep a case class here, for uniformity.

src/main/scala/org/kframework/minikore/MiniKore.scala
grosu: Typo?  I don't even think that `val` is needed.

src/main/scala/org/kframework/minikore/MiniToText.scala
```diff
@@ -49,7 +50,7 @@ object MiniToText {
     case ForAll(v, p) => "\\forall(" + apply(v) + "," + apply(p) + ")"
     case Next(p) => "\\next(" + apply(p) + ")"
     case Rewrite(p, q) => "\\rewrite(" + apply(p) + "," + apply(q) + ")"
-    case Equal(p, q) => "\\equal(" + apply(p) + "," + apply(q) + ")"
+    case Equals(p, q) => "\\equal(" + apply(p) + "," + apply(q) + ")"
```
grosu: Either `Equals` and `equals`, or `Equal` and `equal`.

src/test/scala/org/kframework/minikore/NodeTests.scala
grosu: We need 100% code coverage anyway.

grosu:
[Update Kore Pull Request + address comments to allow merging.](https://trello.com/c/2tecOnjI/8-update-kore-pull-request-address-comments-to-allow-merging)

src/main/scala/org/kframework/minikore/MiniKore.scala
cos: No, they are compiled differently. To use the case class one would have to write `val x = True()` which would create a new, identical instance of the case class at each invocation. Using `object` would create a singleton, with the usual property that only one instance would ever be created, and that instance would be available by calling `True`. So it would be: `val x = True`.

My preference is for using object as it encodes the constant nature of True at the language level. Also, I think it should be called `Top` to distinguish it from the `True` in the defined language.

src/main/scala/org/kframework/minikore/MiniKore.scala
cos: Yes, it's mostly not needed. The `override` part helps a bit to ensure the field is not named differently by mistake. I would go without.

src/main/scala/org/kframework/minikore/MiniKore.scala
grosu: `Top` and `Bottom` is fine, we can do that change.

This discussion is reminding me that @hreada and I discussed that we may want to include the sorts of all patterns in their signature.  Here are the reasons:

1) Sorts are part of the semantics of ML, so all provers and backends should need them.

2) Frontends already compute the sorts of everything as part of their consistency checking, so why throw that information away and then have the backends re-infer it?

3) It is not nice to put that sort information in attributes, considering that ALL backends should need it.

Anybody can think about any argument against not putting the sort info in each pattern?

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -1,39 +1,162 @@
...
+    override def build: Node2Builder[i.Pattern] = DefaultBuilders.OrBuilder
+  }
+
+  case class Not(p: i.Pattern) extends i.Not {
+    override def build: NodeBuilder[i.Pattern] = DefaultBuilders.NotBuilder
+  }
+
+
+  case class Implies(p: i.Pattern, q: i.Pattern) extends i.Implies {
+    override def build: Node2Builder[i.Pattern] = DefaultBuilders.ImpliesBuilder
```
grosu: I would remove the type of `build` above, that is, the `: Node2Builder[i.Pattern]` part.  Scala can infer it on the one hand, and it makes the users be aware of "nodes" and the "ast" view.  We'd like to offer the "ast view for free" to people who are interested in it and appreciate its benefits; but we'd like the others to not have to pay any cost whatsoever, not even to have to be aware of it, if they chose to.

src/main/scala/org/kframework/minikore/MiniKore.scala
cos: We cannot infer precise sorts but lower/upper bounds. Should we somehow encode this?

msaxena2: I'd rather have this this way for the sake of uniformity, throughout the code. I'm ok with either.

src/main/scala/org/kframework/minikore/MiniKore.scala
msaxena2: discussed. Agreed that this works.

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
msaxena2: irrelevant

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
msaxena2: Fixed

src/main/scala/org/kframework/minikore/MiniKore.scala
cos: > Anybody can think about any argument against not putting the sort info in each pattern?

Is there a decision on this already?

src/main/scala/org/kframework/minikore/MiniKore.scala
cos: We can of course choose to disregard this, but the convention in Scala is to put types on public methods and fields, even if it can infer them.

msaxena2:
Please refer to #12 #13 for other comments.

msaxena2:
@daejunpark Please review the added documentation. @cos Please have a look at the changes to the interface - this should be in line with our discussions.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,484 @@
...
+    override def build(_1: Pattern, _2: Pattern): Equals
+  }
+
+  object Equals {
+    def unapply(arg: Equals): Option[(Pattern, Pattern)] = Some(arg._1, arg._2)
+  }
+
+}
+
+object build {
```
cos: woud could do without this wrapper

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,201 @@
...
+    def getLeaf2Count(p: Pattern): Int = p match {
+      case Leaf2(_ , _) => 1
+      case Node(c: Seq[Pattern]) => c.map(getLeaf2Count) sum
+      case _ => 0
+    }
+
+    def renameVariable(p: Pattern): Pattern = p match {
+      case n@BinderNode(v: Variable, p: Pattern) => {
+        val freshVar: Pattern = v.build(("#" + v._1, v._2))
+        n.build(Seq(freshVar, renameVariable(p)))
```
cos: does `n.build(freshVar, renameVariable(p)` work? i.e., without `Seq`

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -453,7 +454,7 @@ object build {
 
     def Variable(_1: Name, _2: Sort): Variable
 
-    def DomainValue(_1: Label, _2: Sort): DomainValue
+    def DomainValue(_1: Label, _2: Value): DomainValue
```
cos: this is a bug the type alias alone didn't catch -- it's good now

src/main/scala/org/kframework/minikore/implementation/MiniKore.scala
```diff
@@ -0,0 +1,122 @@
+package org.kframework.minikore.implementation
+
+import org.kframework.minikore.interfaces.{build, pattern => i}
+
+import scala.collection._
+
+/** Algebraic data type of MiniKore. */
+object MiniKore {
+
+  /** A collection of classes that serve as the default implementation of the [[org.kframework.minikore.interfaces.pattern]] **/
```
daejunpark: I prefer to put the namespace `i.` to make them explicit.

src/main/scala/org/kframework/minikore/implementation/MiniKore.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  def Equals(_1: i.Pattern, _2: i.Pattern): i.Equals = m.Equals(_1, _2)
+
+  def Exists(_1: i.Variable, _2: i.Pattern): i.Exists = m.Exists(_1, _2)
+
+  def ForAll(_1: i.Variable, _2: i.Pattern): i.ForAll = m.ForAll(_1, _2)
+
+  def Rewrite(_1: i.Pattern, _2: i.Pattern): i.Rewrite = m.Rewrite(_1, _2)
+
+  def Application(_1: i.Symbol, args: Seq[i.Pattern]): MiniKore.Application = m.Application(_1, args)
```
daejunpark: Why `MiniKore.` instead of `i.`?

daejunpark:
It looks good to me.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+      build(children.head.asInstanceOf[Variable], children(1))
+    }
+  }
+
+  object BinderNode {
+    def unapply(arg: BinderNode): Option[(Variable, Pattern)] = Some(arg._1, arg._2)
+  }
+
+}
+
```
ehildenb: As far as I can tell, `Variable` is the only dependency from `object tree` to `object pattern` (other than the type-dependency on `Pattern`).

The only way I can think to remove this dependency is to add a `trait Variable` to the `tree._` object, and have the `pattern.Variable` extend that `tree.Variable`. Or we could just not have a `pattern.Variable` and have `Variable` be defined at the `AST` level (rather than at the `Pattern` level). We would still have a `case class Variable ...` in the implementation.

Thoughts? No immediate benefits of either jump out to me, other than being able to remove a dependency of the `AST` view on the `Pattern` view.

cos: Everything but `AST`, `Node`, and `Leaf` are just helpers for making it easier to create AST nodes, and could be moved somewhere else.

msaxena2: I think this is fine, if we don't look at object tree as something that's completely generic. Since we're looking the the tree interface as a view to the pattern interface, I think it's ok to have some links to stuff from the pattern interface in the tree interface.

ehildenb: Sounds good to me, if we do come up with a compelling reason later to move `Variable` around, I think the change will probably be minimal and easy.

src/main/scala/org/kframework/minikore/implementation/MiniKore.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  case class Module(name: i.Name, sentences: Seq[Sentence], att: Attributes)
+
+  sealed trait Sentence
+
+  case class Import(name: i.Name, att: Attributes) extends Sentence
+
+  case class SortDeclaration(sort: i.Sort, att: Attributes) extends Sentence
+
+  case class SymbolDeclaration(sort: i.Sort, symbol: i.Symbol, args: Seq[String], att: Attributes) extends Sentence
```
ehildenb: Should it be `args: Seq[i.Sort]` instead of `args: Seq[String]`?

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
+package org.kframework.minikore.interfaces
+
+object tree {
```
grosu: scaladoc for each object, too

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+
+}
+
+
+object pattern {
+
+  import tree._
+
+  /* ML Pattern Type */
+  sealed trait Pattern extends AST
```
grosu: The above is not proper Scaladoc, I cannot see it in the generated doc.  You need two *.  Also, you may want to be a bit more explicit in comments, they are quite concise.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+  type Value = String
+
+  type Name = String
+
+
+  /**
+    * Matching Logic DomainValue.
+    *
+    * Provides (Implementations for members)
+    *    - contents of type Product2[Label, Value].
```
grosu: We do not call them "labels" anymore, we call them symbols now.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case n@_ => n
+    }
+
+    val changedVar: Variable = b.Variable("#A", Sort("Int"))
+
+    val changedExists: Exists = b.Exists(changedVar, b.Equals(changedVar, TestPatterns.int1))
+
+    assert(TestFunctions.map(changeVar)(TestPatterns.simpleExists) == changedExists)
+  }
+
```
grosu: We should also add some tests where we pattern match on ML constructs: And, Or, etc.  For example, some implementation of some of the functions above, like size or map where we do matching over pattern constructs and not AST ones.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+  object TestFunctions {
+
+    def size(p: Pattern): Int = p match {
+      case Node(c: Seq[Pattern]) => c.map(size).sum
+      case Leaf(_) => 1
+    }
+
+    def getNode0Count(p: Pattern): Int = p match {
+      case Node0() => 1
+      case Node(c: Seq[Pattern]) => c.map(getNode0Count) sum
```
cos: Postfix dotless notation, like ` sum` is discouraged by Scala conventions lately -- it introduces some tricky ambiguities. In this case, I would write: `(c map getNode0Count).sum`

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+  trait Variable extends Pattern with Leaf2[Name, Sort] {
+    def build(_1: Name, _2: Sort): Variable
+  }
+
+  object Variable {
+    def unapply(arg: Variable): Option[(Name, Sort)] = Some(arg._1, arg._2)
+  }
+
+
+  case class Symbol(symbol: String)
```
cos: `name` could also be a good option.

src/test/scala/org/kframework/minikore/NodeTests.scala
grosu: @cos, I do not see `(c map getNode0Count).sum` as a big improvement.  So `c` is a sequence object and map is a member of sequence objects, so on Earth to say `c map f`?  What is wrong with the good old `c.map(f)`?  Also, in all the functional world, `a b c` means `(a b) c`, so why on Earth does Scala need to have `a b c` mean `a.b(c)`?

grosu: BTW, I'm not trying to be ironic here, I literally want to understand this.  So if you know any any good reading on this issue, please let us know @cos .

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+
+  object Node2 {
+    def unapply(arg: Node2): Option[(Pattern, Pattern)] = Some(arg._1, arg._2)
+  }
+
+
+  /**
+    * Extends [[Node2]], and only allows [[pattern.Variable]] as the first element, and [[pattern.Pattern]] in its args list. Extended by [[pattern.Exists]], [[pattern.ForAll]].
+    */
+  sealed trait BinderNode extends Node2 {
```
grosu: I guess this is OK for now, although note that we want to allow users to define their own symbols to be binders, and those symbols may have more than two arguments and the binder may act in only some of the arguments.  Consider, for example, the `let x = e in e'` construct.  The first argument is bound in the third only.

grosu: But please add a comment to the code, to make sure we do not forget about this and also show our users that we are aware of it.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+
+  }
+
+  object TestPatterns {
+
+    val b: build.Builders = DefaultBuilders
+
+    private val symbol = interfaces.pattern.Symbol("Int")
+
+    val int1: Pattern = b.DomainValue(symbol, "1")
```
grosu: can we avoid using `b.` everywhere?  something like `import b._` ?

src/test/scala/org/kframework/minikore/NodeTest.scala
```diff
@@ -0,0 +1,207 @@
...
+    case Node(c: Seq[Pattern]) => c.map(getLeafCount).sum
+  }
+
+  def getLeaf2Count(p: Pattern): Int = p match {
+    case Leaf2(_, _) => 1
+    case Leaf(_) => 0
+    case Node(c: Seq[Pattern]) => c.map(getLeaf2Count).sum
+  }
+
+  def renameAllVariable(p: Pattern): Pattern = p match {
```
grosu: `renameAllVariables` (note the `s`)

msaxena2: Oh! missed typo. I'm still working through rest of the changes.

src/test/scala/org/kframework/minikore/NodeTests.scala
msaxena2: I've avoided using ` `, and have stuck to `.` everywhere.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+
+    def getLabelledNodesCount(p: Pattern): Int = p match {
+      case LabeledNode(_, c: Seq[Pattern]) => c.map(getLabelledNodesCount).sum + 1
+      case Node(c: Seq[Pattern]) => c.map(getLabelledNodesCount).sum
+      case _ => 0
+    }
+
+    def getNode1Count(p: Pattern): Int = p match {
+      case Node1(p: Pattern) => 1 + getNode1Count(p)
+      case Node(c: Seq[Pattern]) => 0 + (c.map(getNode1Count) sum)
```
msaxena2: Removed 0.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case Leaf(_) => 0
```
msaxena2: Used Leaf(_) everywhere, for consistency.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    def getNode2Count(p: Pattern): Int = p match {
+      case Node2(x: Pattern, y: Pattern) => 1 + (getNode2Count(x) + getNode2Count(y))
```
msaxena2: Yes. Removed the unnecessary parenthesis.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case Node(c: Seq[Pattern]) => 0 + (c.map(getNode1Count) sum)
```
msaxena2: fixed.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case _ => 0
+    }
+
+    def getLeafCount(p: Pattern): Int = p match {
+      case Leaf(_) => 1
+      case Node(c: Seq[Pattern]) => c.map(getLeafCount) sum
```
msaxena2: fixed.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case _ => 0
```
msaxena2: Yeah, it's not needed. Removed the default case.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    def getLeaf2Count(p: Pattern): Int = p match {
+      case Leaf2(_, _) => 1
+      case Node(c: Seq[Pattern]) => c.map(getLeaf2Count) sum
+      case _ => 0
+    }
+
+    def renameVariable(p: Pattern): Pattern = p match {
```
msaxena2: done

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+      case n@BinderNode(v@Variable(name, _), p: Pattern) => {
+        val freshVar: Pattern = v.build("#" + name, v._2)
```
msaxena2: Yes, restructured the function.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+    val simpleForAll: ForAll = b.ForAll(simpleIntVar, b.Equals(simpleIntVar, b.Variable("Y", Sort("Int"))))
+
+    val simpleEquals: Equals = b.Equals(simpleOr, top)
+
+    val simpleList: Seq[Pattern] = Seq(top, bottom, simpleAnd, simpleOr, simpleNot, simpleImplies,
+      simpleIntVar, simpleExists, simpleForAll, simpleEquals, plusApp, e1, e2)
+  }
+
+
+  def b: build.Builders = DefaultBuilders
```
msaxena2: Restructured the test file.

src/test/scala/org/kframework/minikore/NodeTests.scala
```diff
@@ -0,0 +1,205 @@
...
+    assert(TestFunctions.getLabelledNodesCount(t.b.Equals(t.intVar, t.plusApp)) == 1)
+  }
+
+  @Test def node1CountTest(): Unit = {
+    assert(TestFunctions.getNode1Count(t.simpleNot) == 1)
+  }
+
+  @Test def node2CountTest(): Unit = {
+    val n2Pattern: Pattern = b.And(t.simpleAnd, b.And(t.simpleAnd, b.Bottom()))
+    println(TestFunctions.getNode2Count(n2Pattern))
```
msaxena2: done

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
```diff
@@ -0,0 +1,493 @@
...
+    def build(contents: C): Pattern
+  }
+
+  object Leaf {
+    def unapply[C](arg: Leaf[C]): Option[C] = Some(arg.contents)
+  }
+
+
+  /**
+    * A Leaf with Product2[CC1, CC2] as its contents. [[pattern.DomainValue]], [[pattern.Variable]] extend this trait.
```
msaxena2: Appended doc for Leaf2.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
msaxena2: Using str everywhere seems a good way of avoiding confusion. Changed to str in Symbol and Sort

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
msaxena2: Appended Documentation.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
msaxena2: Changed.

src/test/scala/org/kframework/minikore/NodeTests.scala
msaxena2: We have complete code coverage. The tests for unparsing cover all the ML constructs. Since I wrote NodeTest.scala to only cover Node behavior, I'll have to add a different set of tests that mixes both Node and Pattern behavior. I think that will be well covered in the next iteration, when we'll have utils + meta level, and tests for them.

src/main/scala/org/kframework/minikore/interfaces/Interfaces.scala
msaxena2: I've added a comment regarding this. I agree, that this would be nice, but how we do this would require discussion.

src/test/scala/org/kframework/minikore/NodeTest.scala
msaxena2: Fixed typo

src/test/scala/org/kframework/minikore/NodeTests.scala
cos: @grosu, there is nothing wrong with `c.map(f)` as opposed `c map f`. I think the idea for the dotless syntax is to make it clear/easier to read. Other languages, including Haskell, have taken the same approach. 

Regarding its particular meaning, Scala is not only functional but also object-oriented, so I'm not sure the conventions from other languages can apply directly. The way I see it, in Haskell, function application is the most common operation, so it chose to use space for this operation. In Scala, method call is the most common operation so it chose to use space for it. 

Still, if seeing an object as a function taking as parameters the method name and something else, one could see Scala's syntax fit into the functional paradigm as well. `c map` means apply function `c` to argument `map`, then apply to result to argument `f`.

PR #12, opened 2017-02-24
Interface changes 2

msaxena2:
Follows From #10. Opening it here so that both versions of code can be compared. I've been working on cleaning up some of the code in #10 .


src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -24,14 +24,14 @@ object TreeInterface {
...
-    def apply(children: Seq[_ <: T]): Node[T]
+    def apply(children: Seq[T]): Node[T]
   }
 
   trait Node0Builder[T <: AST[T]] extends NodeBuilder[T] {
 
     def apply(): Node0[T]
 
-    override def apply(children: Seq[_ <: T]) = {
+    override def apply(children: Seq[T]) = {
```
grosu: Common, @msaxena2, push the throttle all the way to floor and get rid of `T` completely.  It is just fine to have `seq[Pattern]`.

cos:
@msaxena2, the extra PR makes it harder to review/track changes, especially as it is made against `master`, not against `interface_changes`.

cos:
This may help:
https://github.com/kframework/kore/pull/13

Personally, I would leave everything on the original branch/PR.

msaxena2:
@cos The cleanup made it hard to view what was going on, so I'd like this PR to be viewed against master. I've made a few changes -
1) Getting rid of the Generics in AST.
2) Simplifying the builder hierarchy.  

Please have a look now.

src/main/scala/org/kframework/minikore/MiniKore.scala
```diff
@@ -29,7 +29,7 @@ object MiniKore {
   case class Axiom(pattern: i.Pattern, att: Attributes) extends Sentence
 
   case class Variable(_1: i.Name, _2: i.Sort) extends i.Variable {
-    def apply(_1: i.Name, _2: i.Sort): Variable = Variable(_1, _2)
+    def apply(_1: i.Name, _2: i.Sort): Variable = new Variable(_1, _2)
```
cos: Why does this class define an `apply` method? This method belongs to the *companion object*, not every instance of the class, and is already generated there by the case class.

src/main/scala/org/kframework/minikore/interfaces/MiniKoreInterface.scala
```diff
@@ -0,0 +1,303 @@
...
+
+  object Leaf {
+    def unapply[C](arg: Leaf[C]): Option[C] = Some(arg.contents)
+  }
+
+
+  sealed trait Leaf2[CC1, CC2] extends Leaf[Product2[CC1, CC2]] with Product2[CC1, CC2] {
+    override def contents: Product2[CC1, CC2] = (_1, _2)
+
+    def apply(_1: CC1, _2: CC2): Pattern
```
cos: All of these `apply` methods defined on the instance, as opposed to the companion object, look very strange to a Scala developer? The convention in Scala is for the `apply` method to belong to the companion object. And the convention makes very much sense. It seems to me that we're moving towards a contrived, unconventional solution just for the sake of avoiding the dreaded `*Builder` hierarchy.

src/main/scala/org/kframework/minikore/parser/TextToMini.scala
```diff
@@ -160,63 +162,63 @@ class TextToMini {
         val c1 = scanner.next()
         val c2 = scanner.next()
         (c1, c2) match {
-          case ('t', 'r') => consume("ue"); consumeWithLeadingWhitespaces("("); consumeWithLeadingWhitespaces(")")
-            True()
-          case ('f', 'a') => consume("lse"); consumeWithLeadingWhitespaces("("); consumeWithLeadingWhitespaces(")")
-            False()
+          case ('t', 'o') => consume("p"); consumeWithLeadingWhitespaces("("); consumeWithLeadingWhitespaces(")")
+            b.Top()
```
cos: Is the `b.` prefix required? Does it work to `import b._` at the top?

src/main/scala/org/kframework/minikore/interfaces/MiniKoreInterface.scala
msaxena2: @cos - I think the apply is most commonly used in the companion object. Given my limited understanding of scala, I don't see why it'd be the convention. From what I've seen, it seems that given an object x, x(y) desugars to x.apply(y). In the case of a class, having a companion object with the same name and apply method is useful, as it avoids having to create an object with the new keyword. However, if I have the right idea, it doesn't matter where apply actually is. The only thing that matters is x(y) desugars to x.apply(y), which works well in our case. If the scala community is enforcing the convention that the apply must be in the companion object, then isn't it limiting it's usage?

cos: The convention is not documented in the official style guide (which is very short) but it does appear in one very popular style guide: https://github.com/databricks/scala-style-guide#apply-method.

And it makes sense. Data structures are usually **not** constructors, at least not in our case. Let's say someone writes:
`And(a, b)(c, d)`, would it be reasonable to expect the result to be the conjunction of `a`, `b`, `c`, and `d`? Don't you find it surprising that it actually means the conjunction of just `c` and `d`? Is it good to find it surprising?

I'm not saying a non-companion object should never "apply", but it needs to make sense semantically. E.g., `Rewrite` could have the ability to apply, rewriting the target term. So one could write `Rewrite(x, Not(x))(a)` to apply the rewrite rule `x => Not(x)` to constant a, yielding `Not(a)`. But then, given we also have the weird `apply`, we can also do `Rewrite(x, Not(x))(a, a)`, which means `a => a`. So:
`Rewrite(x, Not(x))(a)` yields `Not(a)`
`Rewrite(x, Not(x))(a, a)` yields `a => a`

I find this extremely confusing.

msaxena2: Even with the builders, is it more explicit that `And(a, b).build(c, d)`is the conjunction of c and d?

So, if it's confusing, then let's just rename apply to build, to give us the liberty of having the apply method reserved for something else. Everything should still work fine without the builder hierarchy.

ehildenb: I agree with Cosmin that naming it `apply` is confusing (especially for the example that he gave with `Rewrite`), but I think that renaming it to `build` is a fine solution. Though for the implementations it would be nice if the `apply` method could be used to actually build them (so it's easier to construct them), which means we'll just be adding the `apply` back in at that level.

I'm still opposed to the builder hierarchy because we were creating a whole hierarchy just to wrap the types `Seq[Pattern] => _`, which is already a covariant functor supported directly by the `Function1` trait. So there's no need to re-create some specific wrapper of `Function1` for our needs when it already exists.

Also, when looking at the diff, all you see is red, so we know that this design is more compact. I would also argue it's much easier to understand, but that's subjective.

cos: I am ok with @ehildenb suggestion of renaming that function to `build`. I know that in the end, it's the same, but the non-standard name will suggest to developers that something must be different here.

@ehildenb, why do you say the `apply` methods need to be added back in the implementations?

If you like seeing the `build`ers as `Seq[Pattern] => Pattern`, let's make them that. I'm ok with that. Right now they are actually methods on the `Node`s, which restricts us from ever attaching any extra information to them, e.g., the Kale backend could not be made compatible.

Regarding red/green, which diff has more red than green? I see this one as +912 −127, and 1->2 as +838 −714.

ehildenb: Hmm, that does seem to be the case. I'm not sure what I was looking at earlier, perhaps I was just looking at one file's diff and for that one I saw mostly red.

By "adding back in the `apply`", I mean when we define the `case class And ...` and friends for implementations. I was under the impression that this created a companion object with an associated apply method?

cos: Yes, but that apply method is added on the companion, as it should, not on the instances of the class.

```
case class Foo(x: Int)
```

generates:

```
object Foo {
  def apply(x: Int) = new Foo(x)
}

class Foo(val x: Int) {
  def equals ...
  def hashcode ...
}
```

The class `Foo` does not have an `apply` method, the companion object `Foo` does. So `Foo(1)` works by calling the `apply` on the companion object, while `new Foo(1)(2)` because the instance itself does not have an `apply` method.

msaxena2: The last commit renames the apply methods in the interface to build. I guess this is a better solution since it gives us the leeway to introduce apply methods in the future. Also note that the lines of code from 1->2 is not a good metric, since 2 has many lines of documentation that 1 didn't.

msaxena2:
Closing this PR, since all the changes have been ported over to #10

PR #13, opened 2017-02-25
Interface changes 1 ... 2

cos:


grosu:
Do objects take more memory when they extend more traits?  For example:
```
  trait Or extends Pattern with Node2 {
    def p: Pattern
    def q: Pattern
    override val _1 = p
    override val _2 = q
    override def apply(p: Pattern, q: Pattern): Or
  }
```
I guess the overrides of `_1` and `_2` are artifacts of extending `Product2`.  Which makes me wonder: does that mean that instances of `Or` will now have to carry two additional fields with them pointing to the same `p` and `q`?  That's almost like taking twice as much memory for very little we get in return.

grosu:
@msaxena2 please eliminate the compilation warnings (two of them), too.

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -258,145 +197,119 @@ object PatternInterface {
...
 
     def p: Pattern
 
     override val _1 = v
 
     override val _2 = p
+
+    def apply(v: Variable, p: Pattern): Exists
+
+    override def apply(p: Pattern, q: Pattern): Exists = apply(v, q)
```
grosu: This `apply` does not look right: takes `p` and `q` and uses only `q` (and `v`?).  Fix and add a test that would catch this error.

cos:
They take up memory as `val`s, but they can be replaced by `def`s which makes them functions taking no extra memory per object.

grosu:
@msaxena2 , @cos : what do we get from extending `Product`?  All I see is that we add repetitive code for no reason.  Also, @cos, you should have reacted to the `val _1`, etc., taking memory.  Doesn't it bother you to just double the memory so gratuitously?  Or is Scala so memory inefficient that wasting memory like that is a minor aspect?

grosu:
@msaxena2, there are two pending things: 1) test for the bug you just fixed; and 2) rename True and False to Top and Bottom.

msaxena2:
Test for bug fix + renaming done. I'm going to continue to add more tests before we merge, so the tests are still a wip. 


As far as the extra 'vals' are concerned, I'm not entirely sure how much of a memory burden they cause.  

@cos  you mentioned 
`val`s, but they can be replaced by `def`s which makes them functions taking no extra memory per object.

Could you elaborate on what the difference between 'val _1: Pattern' and 'def _1:Pattern' will be?  I'm actually not sure about how scala works. In java for instance - 

```Java
class And {
    Pattern p;
    Pattern q;
    Pattern _1;
    Pattern _2;
}
```
p and q are references to Pattern Objects, so the memory footprint of the 'p' itself is minimal. One can think of it as a pointer to the actual p object. So, having copies of 'p' and 'q' is hardly a problem in '_1' and '_2' is not a big memory hit, since 'p' is just a reference, and occupies very little memory itself. 

Now, let's replace this with scala - 
```scala
class And {
def p: Pattern
def q: Pattern

override  val _1 = p
override val _2 = q
```

My guess is that _1 and _2 are still just references to pattern objects. The val makes sure that the references themselves are immutable. So, the memory footprint should still not be something to be worried about. But then, how does changing _1 to be a def make any difference? I'll have to read up more on defs to come up with a concise answer, but I don't think it'll make any difference memory-wise. I may be wrong though, given my limited experience with scala. @cos what's your take?

msaxena2:
Also, there are no warnings on building anymore. The current solution doesn't fix the problem though. The warnings have to do with type erasure, which is a problem with the way the jvm works. There seem to be solutions for the issue, and I'm reading up on them. Will have a fix out soon.

grosu:
@msaxena2 : the `val _1` have two problems, as I already mentioned.  First, they DOULBE the size of ALL the Node2 nodes.  4 references instead of 2.  Second, they are a LOT of code repetition.  Basically you do the same thing in each Pattern that instantiates a Node2.  It is very ugly and repetitive.  PLEASE stay away from code duplication!  It is almost always a bad idea to duplicate code.

I am not sure if Product is OK with it, but making `_1` and `_2` into defs, they become (constant) methods in the JVM, instead of fields.  So they are stored once and for all with the class itself.  Then instances have access to them through the type of their class.  You pay with runtime overhead now, because you have one more level of indirection.  But I do not recommend doing this.

I actually suggest that we eliminate all these `Product` extensions and all the `_1` and `_2`.   And in general, we should stop adding more generality than we need.  We waste a lot of time with such unnecessary generalizations.

msaxena2:
I'm actually ok with removing Product from Nodes. We will however, lose the ability to do stuff like
```scala
p match {
   n @ Node1 => n._1
}
```
Instead to access _1 and _2, we'll have to go through the children, like - 
```scala
p match {
   n @ Node1 => n.children.head
}
```
which I'm ok with.

grosu:
Yes, no big deal.  Please remove Products.  We can add them later if really really needed.

cos:
> Doesn't it bother you to just double the memory so gratuitously? 

I try to focus on the aspects that I suspect will be larger issues in the future.

> Or is Scala so memory inefficient that wasting memory like that is a minor aspect?

Scala is as memory-efficient as Java. This particular performance fix (`val` => `def`) could have been applied at a later time with no (or at most minimal) refactoring of client code.

Regarding `Product`, extending it is the standard Scala way of saying something is a product type, e.g., case classes are by default also `Product`s. Going down the hierarchy, it enforces consistent names for children (`_1`, `_2`,...). In my opinion, it is nice and we get it mostly for free, but I am ok with the decision to skip it.

grosu:
> Scala is as memory-efficient as Java.

Maybe in principle, but many people complain about it in forums that it has lots of defaults that end up eating lots of resources.  Remember the parsing performance issue that we had, where it just took 1s to "initialize" its collections.  And even now, take this Product thing.  You said that case classes are by default Product.  Does that mean that we are going to have to pay the memory price of Product anyway, so we can just as well keep it, so we at least get some coding convenience?

cos:
> Maybe in principle, but many people complain about it in forums that it has lots of defaults that end up eating lots of resources. Remember the parsing performance issue that we had, where it just took 1s to "initialize" its collections. 

Regarding the 1s initialization, Scala, and to a good degree Java, are mostly used for long running/server applications. The JVM itself is also tuned to work well after the performance-sensitive code paths have been hit many times. So, when running K for a few seconds, we are not hitting the use case Scala/Java and the JVM have been optimized for. This is why nailgun improved performance so much, and why I have been advocating for a K *console*. Yes, Scala can be more heavy-weight in initialization than Java so we'll have to be careful to not hit those cases.

> You said that case classes are by default Product. Does that mean that we are going to have to pay the memory price of Product anyway, so we can just as well keep it, so we at least get some coding convenience?

`Product` and its `Product*` subtraits only define methods so there is no performance hit. In Scala there is the option of overriding a method with a field, but can choose not to do that.

grosu:
I've already lost my hope for performance.  I just hope that at least the code will be clean and easy to understand and maintain, although the fact that we are spending so much time on these issues makes me wonder if even that will be possible.

> why I have been advocating for a K console.

We do not design K around Scala's limitations.  If Scala does not give us what we need, then we drop it and pick another language.  But so far you were a convinced preacher of Scala and we followed, and it is too late to switch.  So we'll live with the 1s penalty, but we should still do our best to not unnecessarily waste more resources.

> Product and its Product* subtraits only define methods so there is no performance hit. In Scala
> there is the option of overriding a method with a field, but can choose not to do that.

So `_1`, etc., are methods by default?  Previously you said that we did have a performance hit with `val _1`.  Are you saying that we didn't use the `Product` subtrait properly?  Can you @cos please be less vague in your answers?  You are our Scala expert and the reason we are all implementing K in Scala.  We expect you to tell us straight, and authoritative as an expert, what works best and what does not.  Giving us options on top of options does not really help.  For example, how can this help us at all here:

> In Scala there is the option of overriding a method with a field, but can choose not to do that.

msaxena2:
@grosu @cos Please have a look now. I'll continue to add tests until we merge. Let me know if you spot something that seems problematic.

msaxena2:
[Update Kore Pull Request + address comments to allow merging.](https://trello.com/c/2tecOnjI/8-update-kore-pull-request-address-comments-to-allow-merging)

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -5,436 +5,312 @@ import org.kframework.minikore.PatternInterface._
...
+
+    def build(contents: C): Pattern
   }
 
   object Leaf {
-    def unapply[T <: AST[T], C](arg: AST[T]): Option[C] = arg match {
-      case l: Leaf[T, C] => Some(l.contents)
-      case _ => None
-    }
+    def unapply(arg: Leaf[_]): Option[_] = Some(arg.contents)
```
lucaspena: Shouldn't this have explicit types?

src/main/scala/org/kframework/minikore/MiniKoreInterface.scala
```diff
@@ -5,436 +5,312 @@ import org.kframework.minikore.PatternInterface._
...
+    override def build(contents: Product2[CC1, CC2]): Pattern = apply(contents._1, contents._2)
   }
 
-  object LabelledNode {
-    def unapply[L, T <: AST[T]](arg: AST[T]): Option[(L, Seq[_ <: T])] = arg match {
-      case p: LabelledNode[L, T] => Some(p.label, p.children)
-      case _ => None
-    }
+  object Leaf2 {
+    def unapply(arg: Leaf2[_, _]): Option[(_, _)] = Some(arg._1, arg._2)
```
lucaspena: Same thing here, should have explicit types as discussed in the K meeting.

src/main/scala/org/kframework/minikore/MiniToText.scala
```diff
@@ -40,8 +40,8 @@ object MiniToText {
     case Variable(name, sort) => apply(name) + ":" + apply(sort)
     case Application(label, args) => apply(label) + "(" + args.map(apply).mkString(",") + ")"
     case DomainValue(label, value) => apply(label) + "(\"" + StringEscapeUtils.escapeJava(value) + "\")"
-    case True() => "\\true()"
-    case False() => "\\false()"
+    case Top() => "\\true()"
+    case Bottom() => "\\false()"
```
lucaspena: True and False were changed to Top and Bottom, but in the text we still use "\\true()" and "\\false()". Is there a reason we're not using "\\top()" and "\\bottom()" for the text as well?

PR #19, opened 2017-04-18
Main + Extended split

msaxena2:
Parser Tests Passing. WIP on making K work with restructuring. Do not merge yet.


src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+    val q: Pattern
+  }
+
+  object Rewrite {
+    def unapply(arg: Rewrite): Option[(Pattern, Pattern)] = Some(arg.p, arg.q)
+  }
+
+  trait Equals extends Pattern {
+    val p: Pattern
+    val q: Pattern
```
cos: Scala usually uses `_1` and `_2` for projections -- e.g., `Product2` has getters `_1` and `_2`. We had like this before for KORE, and having it that way also matches Kale better. We can always make it work but we'd introduce indirection.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+  trait Name {
+    val str: String
+  }
+
+  object Name {
+    def unapply(arg: Name): Option[String] = Some(arg.str)
+  }
+
+  trait Sort {
+    val str: String
```
cos: We could name this field "name", as in "sort name".

src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  object Sort {
+    def unapply(arg: Sort): Option[String] = Some(arg.str)
+  }
+
+  trait Symbol {
+    val str: String
```
cos: Same here, "symbol name"

src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  object Symbol {
+    def unapply(arg: Symbol): Option[String] = Some(arg.str)
+  }
+
+  trait Value {
+    val str: String
```
cos: We could leave the trait empty, and rely on the `Object.toString` method

cos: Or even not have the trait to begin with. We started with creating a special wrapper for symbols as it ensured we mess up the order when calling the `DomainValue` constructor, i.e., when writing `DomainValue("foo", "bar")` it's not clear which one is the symbol and which one is the value, and it's very easy to introduce bugs this way (I did). But we could very well live with `DomainValue(Symbol("foo"), "bar")`, or with a `Value` trait that has a String-based builder in `Builders` but relies on the `toString` method here.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+    val args: Seq[Pattern]
+  }
+
+  object Application {
+    def unapply(arg: Application): Option[(Symbol, Seq[Pattern])] = Some(arg.symbol, arg.args)
+  }
+
+  trait DomainValue extends Pattern {
+    val symbol: Symbol
+    val value: Value
```
cos: It would make my life much easier if `value` would be of type `Any` here. Otherwise, I'd have to wrap the actual value every time, wasting both memory and characters. Safety-wise, we're not gaining much from having this type. For deconstruction, `str` can be replaced with `toString` (which is available in `Any`). For construction, we can have: `DomainValue(Symbol("Int"), "5")`. I don't see `DomainValue(Symbol("Int"), Value("5"))` as providing much/any safety. It gives us nothing at the type level and the `DomainValue` builder itself can do any safety checks on the String.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
msaxena2: Ok. I've changed it to _1, and _2 everywhere except the quantifiers, where v, and p remain, representing Variable and Pattern respectively. Given that this doesn't have any relation to the Kore language itself, I'm ok with this decision, if it helps make things easier in Kale.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
msaxena2: Sort name makes sense, and I'm perfectly fine with changing this. However, we use str everywhere, and this breaks consistency. So, I'd let this remain, unless of course, it makes it hard for Kale to handle.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
msaxena2: This I disagree with. What is symbol.name supposed to mean? IMO, symbol.str is more intuitive. I'm open to making this change, given that we can come up with something consistent across Symbol, Name, Sort, ModuleName.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
msaxena2: I'm not sure how to address this. Value is in the Kore language. We'll have to change https://github.com/kframework/kore/wiki/KORE-Language-Syntax here. 

I actually like `DomainValue(Symbol("Int"), Value("3"))`?

grosu: I also like to wrap all strings with their meaning, that is, `DomainValue(Symbol("Int"), Value("3"))`.  But what is really important to me is uniformity and consistency in conventions.  When these get broken we waste time.  People are always ok to write a few more characters for the sake of clarity and uniformity.  They waste more time having to remember special cases.

src/main/scala/org/kframework/kore/Default.scala
```diff
@@ -0,0 +1,122 @@
+package org.kframework.kore
+
+object default {
```
cos: Let's not expose these case classes at all. We can make then all private so that users will be required to use the `default.DefaultBuilders` (which we could rename to `default.Builders`) and kore interfaces. This way we don't end up with client code depending directly on the default implementations.

cos:
Thanks @msaxena2, this look good. I'll remerge into my branch and see how it goes.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
cos: Ok, I adjusted Kale to match.

cos:
I adjusted Kale to the latest version. The only change I still rely on is that Pattern `Builders` return `Pattern`, not something more specific. You can pick up the change from here:

https://github.com/kframework/kore/tree/kale-integration

src/main/scala/org/kframework/kore/interfaces/Interface.scala
```diff
@@ -0,0 +1,303 @@
...
+
+  def Application(p: Symbol, args: Seq[Pattern]): Application
+
+  def DomainValue(symbol: Symbol, value: Value): DomainValue
+
+  def Top(): Top
+
+  def Bottom(): Bottom
+
+  def And(p: Pattern, q: Pattern): And
```
cos: I mentioned this in the message below, and I need this to be addressed. Otherwise, Kale will not be able to implement Kore.

grosu: Hm, interesting.  One would expect that having the builders be more specific is better, but in this case it is the opposite.  But what's the general rule to follow here?  All builders return the least concrete type?  Why "pattern" and not "Any" then?

One could also argue the decision to apply rewrites/deduction at build time, instead of after.

Do we get any real advantage from having the return type `And` instead of `Pattern`?

In case we decide to go with `def And(p: Pattern, q: Pattern): Pattern` then we should change the KORE language back to:
```
  syntax Pattern = \and(Pattern,Pattern)
```

cos: Regarding the rule, I think it's a case by case decision, and I would definitely not go as high as `Any`. For `Pattern`s, we know they can be reduced, and at most to a `Pattern`. In other cases, e.g., `Module`, it's unlikely a more general type will be needed, so we could leave the specific type.

grosu: There should be a rule here.  The point of the KORE interface is to be an exact correspondent to the KORE language, not a case-by-case adhoc decision.  As we modified the KORE language now, we are basically asking the builder to return `And` because the production results in an `And`.  Let's think a bit more for a rule.

msaxena2:
To keep track - This has to be merged at the same time as https://github.com/kframework/k/pull/2295, to ensure K works with our changes to Kore.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
grosu: Ok, I think we can do what @cos says, and at second thoughts it seems to be the right thing to do.  But, we need to change the syntax of KORE back to:
```
...
Pattern =  Variable
         | \and(Pattern,Pattern)
...
Variable = Name:Sort
```
Here is "meaning" of the translation to Scala:

1) Each production `N = prod(N1,...,Nk)` yields an extractor trait extending the result sort and a companion object for matching purposes:
```
trait Prod extends N {
  val _1 : N1
  ...
  val _k : Nk
}
object Prod {
    def unapply(arg: Prod): Option[(N1,...,Nk)] = Some(arg._1, ..., arg._k)
}
```
as well as a builder:
```
  def Prod(n1: N1, ..., nk: Nk): N
```

2) Each subsort production `N = M` yields a trait extension relation:
```
trait M extends N
```

With these, 
```
Pattern = \and(Pattern,Pattern)
```
yields:
```
  trait And extends Pattern {
    val p: Pattern
    val q: Pattern
  }
  object And {
    def unapply(arg: And): Option[(Pattern, Pattern)] = Some(arg.p, arg.q)
  }
...
def And(p: Pattern, q: Pattern): Pattern
```
Similarly, 
```
  Pattern = Variable
  Variable = Name:Sort
```
yields:
```
  trait Variable extends Pattern
  trait SortedVariable extends Variable {
    val name: Name
    val sort: Sort
  }
  object SortedVariable {
    def unapply(arg: Variable): Option[(Name, Sort)] = Some(arg.name, arg.sort)
  }
...
def SortedVariable(n: Name, s: Sort): Variable  // Variable, not Pattern!
```
One could now argue that we can merge `SortedVariable` with `Variable` because there is only one constructor for `Variable`, but if we had another one, say `UnsortedVariable`, due to some extended KORE syntax `Variable = var(Name)`, then we may want to keep a separation between `Variable` and the two constructs for them.

msaxena2: Another slightly confusing thing would be, let's say, someone uses the builders as - 
```
val a: Pattern =...
val b: Pattern = ...
val build: Builders
val myAnd: Pattern = build.And(a, b)
```
But then you may or may not be able to call the destructor, depending on how the build happened  - 
```
myAnd match {case And(x, y) => .. } 
```
Which makes using the destructors tricky/confusing.

grosu: This is true, @msaxena2, and seems to be a problem inherent and thus unavoidable if we decide to allow builders to implicitly apply rewrite/simplification rules.  One could also think of it as a feature, because this way one can check if the `builder.And` simplified the pattern or not.

So the decision we have to make here is whether we want to allow rewrites/simplifications in builders.  I think @cos will say yes, and I suspect @daejunpark will also say yes.  It makes sense, for performance reasons.  In fact, we can imagine super-smart builders that decide the best representation of a pattern based on many heuristics and/or properties; for example, a builder that works with a SAT solver may keep everything in a conjunctive or in a disjunctive normal form, or can push the negations as down as possible, etc., etc.  Do we want to disallow such implementations of KORE?  Probably not.  And in fact, they would not be in conflict at all with the trait structure generated from the grammar modified as I explained above.

cos: @msaxena2, in your example, as we already have `a` and `b` in hand, we can simply use them instead of calling the destructor. We call a destructor when we need to ask for information that we might not have, e.g., "is this an And?". If the `And` can self-reduce or is passed through some external reducer, the question makes a lot of sense.

src/main/scala/org/kframework/kore/Default.scala
```diff
@@ -0,0 +1,122 @@
+package org.kframework.kore
+
+object implementation {
```
cos: "default" sounded better. "implementation" is move vague -- there are many other implementations. Also, you haven't replied to my comment on making the implementation classes private.

src/main/scala/org/kframework/kore/interfaces/Interface.scala
grosu: @msaxena2 , @cos , everybody:  do you guys agree with the explanation of the meaning/translation of the KORE language syntax to Scala traits I gave above?  If yes, then I think we should go ahead with it asap, that is, modify the KORE language as explained and add an explanation in the documentation of the KORE interface relating it to the KORE language syntax, and probably also in the KORE language syntax wiki page.  I can take of the wiki page if you agree with the above; I will do it as soon as I head from @msaxena2 and @cos (others: if you disagree then please say so asap).

cos: I agree.

cos: The translation gives us a coherent way to translate to a language with subtyping. This could be useful for other purposes as well, e.g., we could compile arbitrary definitions (not only KORE's) to Scala code to be executed directly.

grosu: The only little issue is the discrepancy between `Variable` and `SortedVariable`.  Normally, to follow the translation mechanically, we would have to introduce `SortedVariable`, but my feeling is that you guys do not want to.  So that deviation would need to be explained if we go with it.

cos: Deciding what we want to generate for `Application = Symbol(Pattern,*)` might clarify this as well...

grosu: They are different issues.  Once we change the KORE language syntax back, the application will become
```
Pattern = ...
         | Symbol(Pattern,*)  // Application
```
so there is no doubt now.  But we cannot do the same with `Variable`, because it is used in other productions ... which is what started the discussion.

cos: I am ok with having a `SortedVariable`. It will give us flexibility for the future. And I don't mind the `Variable` constructor returning a more specific type.

src/main/scala/org/kframework/kore/Default.scala
msaxena2: I had it as default earlier, but we can't use it since it's a keyword in java. I'm ok with any other word that works

src/main/scala/org/kframework/kore/interfaces/Interface.scala
msaxena2: I'm ok with Variable and SortedVariable. I agree with @cos that having flexibility for unsorted variables in the future would also be nice.

src/main/scala/org/kframework/kore/Default.scala
msaxena2: Ok. I agree.

msaxena2:
@cos , I've made the changes from our discussion. You can use the last commit. @grosu I can make the changes to the syntax file as well, if you'd like, to ensure the ordering between code and text is maintained.

grosu:
@msaxena2 :
> @grosu I can make the changes to the syntax file as well, if you'd like, to ensure the ordering between code and text is maintained.

That would be great!  Thanks, @msaxena2

msaxena2:
@grosu This PR now works with K, and can be merged. Please have a look. We can starting adding to extended in subsequent PRs, and enrich the wiki further.

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -0,0 +1,297 @@
...
+  def sentences: Seq[Sentence]
+  def att: Attributes
+}
+
+object Module {
+  def unapply(arg: Module): Option[(ModuleName, Seq[Sentence], Attributes)] = Some(arg.name, arg.sentences, arg.att)
+}
+
+trait Sentence {
+  def att: Attributes
```
grosu: This deviates from the translation rules at https://github.com/kframework/kore/wiki/KORE-Language-Syntax...

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -0,0 +1,297 @@
...
+}
+
+object DomainValue {
+  def unapply(arg: DomainValue): Option[(Symbol, Value)] = Some(arg.symbol, arg.value)
+}
+
+trait Top extends Pattern
+
+object Top {
+  def unapply(arg: Top): Boolean = true
```
grosu: Why `Boolean`?  This does not respect the rules at https://github.com/kframework/kore/wiki/KORE-Language-Syntax.  `void`?

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -0,0 +1,297 @@
...
+}
+
+trait Bottom extends Pattern
+
+object Bottom {
+  def unapply(arg: Bottom): Boolean = true
```
grosu: Same as for `Top`.

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -0,0 +1,297 @@
...
+  def _1: Pattern
+  def _2: Pattern
+}
+
+object Implies {
+  def unapply(arg: Implies): Option[(Pattern, Pattern)] = Some(arg._1, arg._2)
+}
+
+trait Exists extends Pattern {
+  def v: Variable
```
grosu: `_1` and `_2`?

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -0,0 +1,297 @@
...
+
+trait ModuleName {
+  def str: String
+}
+
+object ModuleName {
+  def unapply(arg: ModuleName): Option[String] = Some(arg.str)
+}
+
+trait Name {
```
daejunpark: The order is different with the syntax. You can change either of them though.

daejunpark:
I went through all the discussions and changes so far. They look good to me as well.

src/main/scala/org/kframework/kore/Interface.scala
msaxena2: The unapply only denotes match success/failure. 
So, this is scala way of allowing 
```Scala
s match {
  case Top() =>... //Success
}
```

src/main/scala/org/kframework/kore/Interface.scala
msaxena2: Same here as for Top. We want to be able to match on Bottom().

PR #32, opened 2017-12-21
Small fixes.

virgil-serbanuta:
Some of the fixes are about math correctness. I'm not sure that my understanding of matching logic is good enough, so I may be wrong. If that's the case, I would appreciate feedback on what I did wrong, as it will improve my understanding of these issues.

* fixed substitution for \sigma(L)
* fixed the alpha-equivalence axiom
* fixed the usage of #wellFormed when defining #getSort
* fixed the meta-variable's name as being '#s' instead of 's' when naming stuff in a theory T_L

I am not a native English speaker, so I may be wrong, but I tried to fix the language a bit, especially in sections 8 and 9. At some point I realized that it's probably not a good use of my time, so I tried to fix only the section titles. I can roll those changes back if needed.

Of course, there are places where I tried to fix both kinds of issues.


grosu:
@hreada can you please merge this?  Seems to have conflicts, though

hreada:
I've merged this PR to the branch `document` and I'm closing this PR.

PR #35, opened 2017-12-29
Document

hreada:
This is the PR that adds a new directory `/docs` which contains the document _Semantics of K_.

In correspondence to that, it also updates the `.gitignore` file so that auxiliary files under the `/docs` directory are not tracked.

The codeship build test was failed because of some tests needed a longer time to run. I have let @grosu know about it, and we decided to merge this PR even if we got the codeship test failure because this PR does not change any existing code at all.


grosu:
We discussed to merge this nevertheless.  We should look into why tests take a lot to execute; this PR didn't touch any code.

PR #42, opened 2018-01-15
pom.xml: Version 2.12-1.1-SNAPSHOT => 2.12-0.1-SNAPSHOT

msaxena2:
- Downgrade to version 0.1, depicting a BETA release.
- Used in the current version of UIUC-K.
- See kframework/k#2396 for related PR in K.

Do not merge.


hreada:
This branch is intended not to be merged to master, but rather give us a way of keeping track of it, so we can recover it whenever we need to.

PR #44, opened 2018-01-16
Haskell kore parser

virgil-serbanuta:
We'll add more documentation tomorrow.


src/haskell/parser/src/CharDict.hs
```diff
@@ -0,0 +1,24 @@
...
+import           Data.Maybe (fromMaybe)
+
+newtype CharDict a = CharDict { getCharDict :: Array.Array Char a }
+
+memoize :: (Char -> a) -> CharDict a
+memoize f =
+    CharDict $ Array.array ('\0', '\255') [(c, f c) | c <- ['\0'..'\255']]
+
+make :: [(Char,a)] -> a -> CharDict a
+make dict defaultValue =
```
raduom: There are a lot of 'make' functions that get exported. Maybe it would be a better idea to name them differently (like mkCharDict in this case) so they do not conflict when importing multiple packages.

src/haskell/parser/src/KoreAST.hs
```diff
@@ -0,0 +1,229 @@
...
+
+newtype Id a = Id { getId :: String }
+    deriving (Show, Eq, Typeable)
+
+newtype StringLiteral = StringLiteral { getStringLiteral :: String }
+    deriving (Show, Eq)
+
+data SymbolOrAlias a = SymbolOrAlias
+    { symbolOrAliasConstructor :: !(Id a)
+    , symbolOrAliasParams      :: ![Sort a]
```
raduom: Are all params supposed to have the same sort?

src/haskell/parser/src/KoreAST.hs
```diff
@@ -0,0 +1,229 @@
+module KoreAST where
```
raduom: An exports list would help to discern what is public and what is internal. If everything is public then maybe write a comment about it?

src/haskell/parser/src/KoreParserImpl.hs
```diff
@@ -0,0 +1,339 @@
...
+sortParser x = do
+    identifier <- idParser x
+    c <- Parser.peekChar
+    case c of
+        Just '{' -> actualSortParser identifier
+        _        -> return (SortVariableSort $ SortVariable identifier)
+  where
+    actualSortParser identifier = do
+        sorts <- inCurlyBracesParser (sortListParser x)
+        when (metaType x == MetaType) (checkMetaSort identifier sorts)
```
raduom: Isn't illegal for a non-meta sort to start with `#`?

src/haskell/parser/src/KoreAST.hs
traiansf: we thought everything here would be public.  But you are right about a comment. Will do

src/haskell/parser/src/KoreParserImpl.hs
traiansf: a sort is an identifier, so this check is done in the identifier parser

ehildenb:
any chance we can break this PR into two? One just for the KoreAST (so that we can discuss that independently), then another for the Parser?

Another benefit would be that we can get the KoreAST in sooner, and then other Haskell K projects can get off the ground.

PR #45, opened 2018-01-22
Kore AST

virgil-serbanuta:


grosu:
@hreada @daejunpark can you please also take a look at this?  Are these data-structures in synch with what you had in mind when you implemented the Scala ones?

src/main/haskell/ast/src/Data/Kore/AST.hs
```diff
@@ -0,0 +1,563 @@
+{-|
+Module      : Data.Kore.AST
+Description : Data Structures for representing the Kore language AST
+Copyright   : (c) Runtime Verification, 2018
+License     : UIUC/NCSA
+Maintainer  : traian.serbanuta@runtimeverification.com
+Stability   : experimental
+Portability : POSIX
```
grosu: So this is not going to work on Windows natively?

traiansf: We should probably change that line... it was the default in some example, then we started copying it around.  Thanks for pointing it out!

src/main/haskell/ast/src/Data/Kore/AST.hs
```diff
@@ -0,0 +1,563 @@
...
+Please refer to Section 9 (The Kore Language) of the
+<http://github.com/kframework/kore/blob/master/docs/semantics-of-k.pdf Semantics of K>.
+
+
+-}
+module Data.Kore.AST where
+
+import Data.Typeable
+
+data MetaType
```
hreada: It seems that here we use the name `MetaType` to refer to the "object-level" or "meta-level" in the Semantics of K document. I'm OK with it but it looks a bit confusing. Probably using something like "Level"?

hreada:
@grosu @virgil-serbanuta These data structures are in sync with what I have in mind and what we had in the scala parser.

daejunpark:
I'm not familiar with Haskell, but it looks fine to me.

A question: how to achieve the different views of the data structure?

When we designed the Scala data structure, one of the main goal is to support at least two views of the data structure: trees and patterns. The tree view helps to traverse the data structure when the exact pattern type is not relevant (e.g., renaming variables). Can you achieve the same thing in an efficient way?

traiansf:
@daejunpark we are going to write  type classes to offer various views on the ast.

daejunpark:
@traiansf thanks for the clarification.

PR #48, opened 2018-01-30
Patterns parameterized to make them traversable.

traiansf:


ehildenb:
@traiansf this looks a-la the unification-fd package that we discussed? is that correct?

traiansf:
@ehildenb yes, we were aiming towards recovering some of that genericity; that is why I though that you reviewing it might be a good thing.  Will try to add some tests.

ehildenb:
Just quickly playing with these in GHCi, it seems like you have to write a whole bunch of stuff just to build simple patterns. Not sure if this is speaking to some design issues, or if we should supply some smart constructors or something.

ehildenb:
Hmmmm, I think we are missing the `Fixed` construction. For example, consider (in GHCi):

```
> let mySort = SortActual (Id "testSort") []
> :t And (SortActualSort mySort)
p -> p -> And a p
```

But then we go for a two deep one:

```
> :t And (SortActualSort mySort) (And (SortActualSort mySort) "test" "test")
And a1 [Char] -> And a2 (And a1 [Char])
```

So, the type grows bigger as the terms get bigger, which we *do not* want. Not only that, both sides of the `And` are now required to have the *same* depth (otherwise they will not have the same type), which means we can only construct depth-balanced terms. For example, the following does not type:

```
> :t And (SortActualSort mySort) (And (SortActualSort mySort) "test" "test") "test"
```

ehildenb:
The above example is using the data-structures incorrectly, here is a correct usage:

```
> let mySort = SortActual (Id "test") []
> let and a b = ObjectPattern (AndPattern (And mySort a b))
> let top = ObjectPattern (TopPattern (Top (SortActualSort mySort)))
> :t and top top
and top top :: UnifiedPattern
> :t and top (and top top)
and top (and top top) :: UnifiedPattern
```

So my remaining question is, why the two layers `AndPattern (And ...`, can we directly do `data Pattern = And ...` instead of `data Pattern = AndPattern !And`?

traiansf:
Jenkins: test this please

traiansf:
Jenkins: test this please

bmmoore:
Adding `tasty-jenkins-xml` breaks `stack init`. I don't see it in any recent stackage snapshots.

traiansf:
@bmmoore on my machine it suggested (and it worked) to add it as an `extra-deps` in `stack.xml`.  

I'll see if I cam make it work, but I'll extract the code in a separate PR anyway.

bmmoore:
I've never heard of `stack.xml`, do you mean `stack.yaml`? Listing it under extra-deps in a stack.yaml file definitely works and is probably the best way to go.
The problem is that stack.yaml is currently automatically generated from package.yaml rather than checked in, `stack init` will complain and refuse to work if you have both a manually edited `stack.yaml` and a `package.yaml` file, and there is nothing you can put in a `package.yaml` file which will make it generate a stack.yaml with an extra-deps list.

 I recommend committing the `language-kore.cabal` and `stack.yaml` files and just dropping the `package.yaml`.

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -9,7 +9,7 @@ import           Control.Monad.Writer
 
 import           Data.Kore.AST
 import           Data.Kore.Parser.CString (escapeCString)
-import           Data.List (intersperse)
+import           Data.List                (intersperse)
 
 {-# ANN module "HLint: ignore Use record patterns" #-}
```
raduom: Why not use a record pattern match? I think would have the same safety effect.

src/main/haskell/language-kore/src/Data/Kore/AST.hs
```diff
@@ -209,19 +218,61 @@ data Variable a = Variable
...
-    = MetaVariable !(Variable Meta)
-    | ObjectVariable !(Variable Object)
-    deriving (Eq, Show)
+data UnifiedVariable v
+    = MetaVariable !(v Meta)
+    | ObjectVariable !(v Object)
+
+asUnifiedVariable'
+    :: (Show (v Object), Show (v Meta))
+    => Maybe (v Object)
```
raduom: Isn't this better represented using Either?

src/main/haskell/language-kore/src/Data/Kore/AST.hs
```diff
@@ -209,19 +218,61 @@ data Variable a = Variable
...
+    -> Maybe (v Meta)
+    -> UnifiedVariable v
```
raduom: Isn't this better represented using Either (v Object) (v Meta)?

src/main/haskell/language-kore/src/Data/Kore/AST.hs
```diff
@@ -615,69 +680,19 @@ data Definition = Definition
...
-    asPattern = OrPattern
-
-instance AsPattern Top where
-    asPattern = TopPattern
-
-instance AsPattern Variable where
-    asPattern = VariablePattern
-
 {-|'MLPatternClass' offers a common interface to ML patterns
   (those starting with '\', except for 'Exists', 'Forall', and 'Mem')
```
raduom: Why exclude exists, forall and mem?

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -255,7 +255,7 @@ instance PrettyPrint UnifiedSortVariable where
     prettyPrint flags (MetaSortVariable sv)   =
         writeOneFieldStruct flags "MetaSortVariable" sv
 
-instance PrettyPrint UnifiedVariable where
+instance PrettyPrint (UnifiedVariable Variable) where
     prettyPrint flags (ObjectVariable sv) =
         writeOneFieldStruct flags "ObjectVariable" sv
```
raduom: The Writer monad is also frowned upon. A good alternative would be something like a 'streaming' library. I think conduit and pipes can do the job. This is however a pretty big change and maybe someone else, better versed in Haskell could pitch in with an opinion. Relevant: https://twitter.com/gabrielg439/status/659170544038707201?lang=en, https://www.fpcomplete.com/blog/2017/06/readert-design-pattern.

Not that I don't think this is a big issue, but more opinions would be welcomed.

traiansf: It seems from the links you've sent that a writer monad with a strict tell would be ok. And that's fairly easy to write, and a super easy change. So I would wait until we see some space leaks from it to decide if it's worth the change.

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
virgil-serbanuta: I think it doesn't, and I thought that the comment below explains it, please correct me if I am wrong. Consider the following case:
```
prettyPrint p@(And _ _ _) =
   stuff (andSort p) (andFirst p) (andSecond p)
```
if I add a field to 'And', then I'll get a compilation error for the `(And _ _ _)` part. However, the following code:
```
prettyPrint And {andSort=as, andFirst=af, andSecond ad} =
   stuff as af ad
```
will happily compile and run in the same case. This means that one may add a field and forget to update the pretty-printer.

traiansf:
@ehildenb @bmmoore @raduom @virgil-serbanuta @mattoxb: I think this is now ready for review. Please take another look and suggest improvements.

traiansf:
Also, w.r.t. Pattern Synonyms and Pattern Views, I started looking into them, but they don't seem mandatory for this PR, so I would like to have this merged and gather for more evidence for how they should look like and how they could help.

src/main/haskell/language-kore/src/Data/Kore/Substitution/Class.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE FlexibleContexts      #-}
```
virgil-serbanuta: Could you add a test with a substitution `x -> something` and a pattern `and(exists x:x, x)`?

traiansf:
@virgil-serbanuta please take a look at my latest commit
@kframework/k-team if there are no further comments, I'd like to merge this tomorrow.

docs/semantics-of-k.tex
```diff
@@ -2,13 +2,13 @@
 
   % Package for using the full page.
   \usepackage{fullpage}
-  
+
```
virgil-serbanuta: I would have put the semantics-of-k files into a different pull request.

traiansf: Oups! UI thought I did `git commit -a` .... :-(
I'll make sure to remove those, as they were partial anyway, and some I think were done by my editor automatically.

PR #49, opened 2018-01-30
PrettyPrinting for AST

virgil-serbanuta:


src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -0,0 +1,471 @@
...
+    => String -> a -> m()
+writeOneFieldStructNewLine name content = do
+    write "("
+    write name
+    withIndent 4 (prettyPrint content)
+    write ")"
+
+writeFieldOneLine
+    :: (PrettyPrintOutput w m, PrettyPrint a)
+    => String -> (b->a) -> b -> m()
```
traiansf: here, and below:  `(b->a)` should probably be `(b -> a)`

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -0,0 +1,471 @@
...
+writeAttributesField fieldName (Attributes []) = do
+    write fieldName
+    write " = Attributes []"
+writeAttributesField fieldName attributes = do
+    write fieldName
+    write " ="
+    withIndent 4
+        (betweenLines >> prettyPrint attributes)
+
+writeStructure :: PrettyPrintOutput w m => String -> [m()] -> m()
```
traiansf: also, you should be consistent with `m()` vs. `m ()`.  Probably the latter is better.

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -0,0 +1,471 @@
...
+writeStructure name fields =
+    write name >> inCurlyBracesIndent (printableList fields)
+
+printableList :: PrettyPrintOutput w m => [m()] -> [m()]
+printableList = intersperse (betweenLines >> write ", ")
+
+instance (IsMeta a) => PrettyPrint (Id a) where
+    prettyPrint id'@(Id name) = do
+        write "(Id "
+        write (
```
traiansf: do you need to break the line here?

PR #61, opened 2018-03-02
Quantified vars match quantifier type in Haskell

traiansf:


traiansf:
@hreada please review the changes I've made to the scala code (commit 
b777c6d)

src/main/haskell/language-kore/src/Data/Kore/AST.hs
```diff
@@ -368,6 +368,43 @@ data MLPatternType
...
+    , ImpliesPatternType
+    , InPatternType
+    , NextPatternType
+    , NotPatternType
+    , OrPatternType
+    , RewritesPatternType
+    , TopPatternType
+    ]
+
+patternString :: MLPatternType -> String
```
virgil-serbanuta: Could you also use this in unparse?

traiansf:
@hreada @xiaohong.chen I've constrained the KORE parser to only accept `#\...` identifiers if the string following `\` corresponds to one of `and`, `or`, `exists`, ....

Please consider updating the Semantics-Of-K to reflect this constraint.

hreada:
@traiansf I've updated the Semantics-Of-K to reflect your constraint that to only accept `#\...` identifiers if the strings correspond to one of the logic connectives. See this commit: https://github.com/kframework/kore/commit/93fe7d2494151b9913d3b4958b6f5dfcd1b81c79

PR #66, opened 2018-03-09
Meta object interactions

virgil-serbanuta:
Hi, this is a bit early since it does not have tests and code comments, but I wanted to get feedback from Xiaohong about the kore tests changes and errors - do they make sense?

I don't know how to fix the remaining ones, feel free to add a commit that fixes them (or explain to me how to do it, but it would probably be slower).

To test, run
```
stack test --no-keep-going --ta --accept
```

and look at a git diff of the golden files.


src/test/resources/expected/test-attribute-1.kore.golden
```diff
@@ -1,28 +1 @@
...
-                                })
-                            ]
-                    }
-                ]
-            , moduleAttributes = Attributes []
-            }
-        ]
-    }
\ No newline at end of file
+Parse error: Error in module 'TEST-ATTRIBUTE-1' -> sort 'Nat' declaration -> attributes -> symbol or alias '#hook': Symbol '#hook' not defined..
```
traiansf: I'm not sure this test was supposed to fail... maybe the original .kore file needs to be changed? @hreada

src/test/resources/expected/test-attribute-2.kore.golden
```diff
@@ -1,79 +1 @@
...
-                                })
-                            ]
-                    }
-                ]
-            , moduleAttributes = Attributes []
-            }
-        ]
-    }
\ No newline at end of file
+Parse error: Error in module 'TEST-ATTRIBUTE-2' -> sort 'Nat' declaration -> attributes -> symbol or alias '#hook': Symbol '#hook' not defined..
```
traiansf: I'm not sure this test was supposed to fail... maybe the original .kore file needs to be changed? @hreada

src/test/resources/expected/imp.kore.golden
```diff
@@ -1,3228 +1 @@
...
-                                                    { applicationSymbolOrAlias =
-                                                        SymbolOrAlias
-                                                            { symbolOrAliasConstructor = Id Object "divAExp"
-                                                            , symbolOrAliasParams = []
-                                                            }
-                                                    , applicationChildren =
-                                                        [ ObjectPattern (ApplicationPattern Application
-                                                            { applicationSymbolOrAlias =
-                                                                SymbolOrAlias
-
```
traiansf: I'm not sure this test was supposed to fail... maybe the original .kore file needs to be changed? @hreada

src/test/resources/expected/lambda.kore.golden
```diff
@@ -1,692 +1 @@
...
-                            })
-                    , sentenceAxiomAttributes = Attributes []
-                    }
-                ]
-            , moduleAttributes = Attributes []
-            }
-        ]
-    }
\ No newline at end of file
+Parse error: Error in module 'LAMBDA' -> axiom declaration -> \equals -> symbol or alias '#isExp' -> symbol or alias 'lambda': Expected 2 operands, but got 1..
```
traiansf: I'm not sure this test was supposed to fail... maybe the original .kore file needs to be changed? @hreada

src/test/resources/expected/test-attribute-1.kore.golden
hreada: @traiansf I've examined the original .kore file, and there's no problem there. This test should not fail. I want to know if there was an error in parsing it, or in checking its well-formedness. We should not check the well-formedness of patterns in attributes using the declarations in the module where the attributes are in. For now, it is OK to not checking well-formedness of patterns that are in attributes.

virgil-serbanuta: Fixed by not checking attribute welformednees.

src/test/resources/expected/test-attribute-2.kore.golden
virgil-serbanuta: Fixed by not checking attribute welformednees.

hreada:
@virgil-serbanuta We shouldn't allow variable patterns to be seen as also being of sort `#variable`. We should stick to the lifting function in the document.

virgil-serbanuta:
We are not doing that, see the `verifySameSort (MetaSort expectedSort) (ObjectSort actualSort)` branch of the `verifySameSort` function.

src/test/resources/expected/lambda.kore.golden
hreada: @traiansf @virgil-serbanuta There seems to be a mismatch of parentheses in `lambda.kore`. I've made a commit to `master` that fixes the problem. That fix should make this failing test pass.

virgil-serbanuta: @hreada Thanks, that seems to have fixed something, but the next error is:
Parse error: Error in module 'LAMBDA' -> axiom declaration -> \implies -> symbol or alias '#provable' -> \equals -> symbol or alias '#substitute' -> variable 'X': Expecting meta sort '#Variable{}' but got object sort 'Exp{}'.

hreada:
@virgil-serbanuta I've fixed a few bugs in `lambda.kore` and `imp.kore`. They should now pass. Let me know if there's anything failing. If now, we should merge this PR.

virgil-serbanuta:
@hreada Thanks, it seems to work now!
I cannot merge until @traiansf accepts my changes.

PR #72, opened 2018-03-15
Set things up for testing parsing and verification performance.

virgil-serbanuta:


src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -1,17 +1,96 @@
...
 import           Data.Kore.Parser.Parser
 
+import           Control.Exception                        (evaluate)
+import           Control.Monad                            (when)
+import           System.Clock                             (Clock (Monotonic),
+                                                           diffTimeSpec,
+                                                           getTime)
 import           System.Environment                       (getArgs)
 
+data CommandLineFlags = CommandLineFlags
```
virgil-serbanuta: Traian: Maybe add a TODO to use a command line parsing library.

virgil-serbanuta: Done.

PR #73, opened 2018-03-16
Fix random errors.

radumereuta:
Please review doc changes/fixes.


docs/semantics-of-k.tex
```diff
@@ -6350,7 +6350,7 @@ \subsubsection{Wellformedness}
 \item In any pattern that is a head being applied to a list of argument 
 patterns, the number and sorts of argument patterns should match with the head 
 declaration;
-\item Every sorts, symbols, and aliases should be declared to be used;
+\item All sorts, symbols, and aliases should be declared before used;
```
virgil-serbanuta: "before being used", "before they are used", "before usage" or something else instead of "before used" (used is a participle or an adjective so it should be composed with something before attaching "before" to it).

radumereuta:
@virgil-serbanuta added your suggestion. Please re-review.

docs/semantics-of-k.tex
```diff
@@ -2236,7 +2236,7 @@ \subsubsection{Fake Subsort Relation}
 		s'$ \\
 		$\inj{s}{s}(x) = x$
 		& \doubleslash (identity) axiom \\
-		$x = y \to \inj{s}{s'}(x) = \inj{s}{s'}$
+		$x = y \to \inj{s}{s'}(x) = \inj{s}{s'}(y)$
```
virgil-serbanuta: @hreada Shouldn't this implication go the other way?

traiansf:
Jenkins: add to whitelist

virgil-serbanuta:
I'll merge this since it is an improvement over the existing state and I'll ask Xiaohong on Slack to check the implication.

PR #83, opened 2018-04-06
Convert to megaparsec

mattoxb:
Here is Megaparsec conversion of the parser.  The test cases that test error messages will still need to be updated, but all the other tests seem to pass now.


traiansf:
Jenkins: add to whitelist

src/main/haskell/language-kore/src/Data/Kore/Parser/LexemeImpl.hs
```diff
@@ -37,8 +40,9 @@ import qualified Data.ByteString.Char8        as Char8
 import           Data.Char                    (isHexDigit, isOctDigit)
 import           Data.Maybe                   (isJust)
 import qualified Data.Trie                    as Trie
-import qualified Text.Parsec.Char             as Parser (char, string)
-import           Text.Parsec.String           (Parser)
+import           Text.Megaparsec              (takeWhileP)
+import qualified Text.Megaparsec.Char.Lexer   as L
```
virgil-serbanuta: Could you call this 'Lexer'? I think it will make the code easier to follow.

traiansf:
Apparently the comments parser needs to be tuned to fully support C-style comments:
see `src/test/resources/test-comment-5.kore` and several unit tests, including https://github.com/kframework/kore/blob/master/src/main/haskell/language-kore/test/parser/Data/Kore/Parser/LexemeTest.hs#L50

bmmoore:
Rebased and tests fixed in #92

PR #87, opened 2018-04-17
Attributes verification

virgil-serbanuta:
Please first check the .kore files, especially kore.kore, to see if I'm doing the right thing. I'm especially worried about the object symbols that I'm including in _every_ kore definition.


hreada:
The definition `BUILTIN-attributes` in `kore.kore` looks good to me. @virgil-serbanuta, I am fine with including symbols in `BUILTIN-attributes` in every Kore definitions, (if that's what you mean), but I don't think we need to do that, right? We don't need them when parsing any Kore definitions, except for parsing attributes. So basically, the well-formedness checker (verifier) of Kore definitions will only look for `BUILTIN-attributes` when it parses attributes.

virgil-serbanuta:
@hreada I can make the changes needed to make these visible only in attributes, but I'm not sure that's better, since that allows people to define their own symbols, with the same name.

Let's say that someone defines their own `hook` symbol. They can use that in normal patterns, but it's unclear what happens when they attempt to use `hook` in attributes: do we assume that it's the implicit one, the user defined one, or do we resolve it in a context-dependent way? This is actually more tricky than it seems because all of our code assumes that a symbol name points to a unique thing, and in attributes that would not true anymore. For additional fun, consider that the user-defined hook is of sort `Attribute`.

Now that I think more about this, how about allowing backticks in ids only for kore internal stuff and using that as a prefix for the attribute symbols and sorts, or something similar?

hreada:
Currently backticks in ids are meant for (meta-level) symbols that correspond to something in the object-level theories. For example, `#'Nat` corresponds to `Nat`, `#'succ` corresponds to `succ`. The reason is that we want a separate name space for those (meta-level) symbols which correspond to object-level things. In the meantime, the users can define whatever meta-level symbols they need, without worrying about accidentally using names that are reserved for other purposes. 

However, I don't think, quote: "that allows people to define their own symbols, with the same name", would be a problem. My understanding of attributes is that they do not carry semantic meaning and they are just a hint for users and backend tools to do tasks more efficiently, without re-engineering the axioms. In other words, we shouldn't expect anything from attributes appear in any proof objects. All patterns appear in a proof object should only use symbols and sorts that are declared in the module (that is indicated at the top of the proof object). In the same time, all patterns appear in attributes should only use symbols and sorts that are declared in the `BUILTIN-Attributes` module. We will never mix these two up, as they really live in different world. 

Another way to think of it is that, we could just as well use string literals to write attributes. We could set up some conventions (e.g., "strict" must follow by a parenthesis with a comma-separated list of positions in it) so that backend tools understand them in case they want to use them as hints. We didn't do that because we realize it would be convenient if we provide a simple syntax to write them so that they look more structural and easy to process. And then we choose the Kore syntax, just because we have designed and implemented it, and it is good enough for the purpose of writing attributes, too. The fact that we use Kore syntax to write attributes doesn't mean we treat them as a part of the actual Kore module definitions or adding something new to the built-in modules that every other Kore module will import, unless there's a potential scenario where we do need that.

hreada:
@virgil-serbanuta Did you remove `kore.kore` file? I cannot find it anymore.

virgil-serbanuta:
@hreada I moved all the changes to kore.kore to attributes.kore (it was cleaner to have a different definition if these definitions were visible only in attributes)

src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -300,6 +300,19 @@ object Mem {
   def unapply(arg: Mem): Option[(Sort, Sort, Pattern, Pattern)] = Some(arg.s, arg.rs, arg.p, arg.q)
 }
 
+/**
+  * Membership, denoted as \in(P, Q) === \ceil(P and Q)
```
traiansf: This comment should be updated to reflect the trait below.

traiansf:
Before we merge, @hreada are you ok with the changes to `.kore` files, i.e., the proposed syntax for specific attributes?

PR #85, opened 2018-04-16
Unified

traiansf:

PR #93, opened 2018-04-26
Unification

traiansf:
This is based on PR #85, so it should be reviewed and merged only upon merging that PR.

It offers a sample of unification, addressing one level (meta or object) only, and not considering object patterns lifted to meta patterns.

It does not produce proofs.


traiansf:
Relevant files:
Unifier.hs : https://github.com/kframework/kore/pull/93/files#diff-494a0eca29f21c9df6eec063ebe9c1a2 
UnifierTest.hs : https://github.com/kframework/kore/pull/93/files#diff-caac7a0af0c3ce57df244a236c66450b

PR #95, opened 2018-05-01
Unification, take 2

traiansf:
Ready for review:  Note that it does not yet solve unification subproblems arising from constraints. I plan to work on that (and additional tests) in the next iteration.


src/main/haskell/language-kore/src/Data/Kore/Unification/Unifier.hs
```diff
@@ -0,0 +1,241 @@
...
+
+-- |'unificationSolutionToPurePattern' packages an unification solution into
+-- a 'CommonPurePattern' by transforming the constraints into a conjunction of
+-- equalities and conjoining them with the unified term.
+unificationSolutionToPurePattern
+    :: MetadataTools level
+    -> UnificationSolution level
+    -> CommonPurePattern level
+unificationSolutionToPurePattern tools ucp =
+    foldr
```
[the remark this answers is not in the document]
traiansf: I wasn't sure what we would really prefer here... My first thoughts were the same as yours, but then I thought one could also prefer having the unification constraint at the top. I'll revise it to your suggested way, nevertheless.

In any case, if we follow the suggestions Xiaohong made in the doc ( https://docs.google.com/document/d/12JudMIqeRfPaewzdyY6LIcjwC4c1UqWRYeyP0X7TyAg/edit#  ) we might not even need to compute the fully combined term as part of the unification process.

src/main/haskell/language-kore/src/Data/Kore/Unification/Unifier.hs
```diff
@@ -0,0 +1,241 @@
...
+    | NonFunctionalHead (SymbolOrAlias level)
+    | NonFunctionalPattern
+    | UnsupportedPatterns
+  deriving (Eq)
+
+-- checks whether a pattern is functional or not
+isFunctionalPattern
+    :: MetadataTools level
+    -> PureMLPattern level Variable
+    -> Either (UnificationError level) [FunctionalProof level]
```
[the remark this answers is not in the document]
traiansf: It is actually the collection of all symbols in a pattern, not just a path.  All of them being functional imply that the pattern itself is functional.

src/main/haskell/language-kore/src/Data/Kore/Unification/UnifierImpl.hs
```diff
@@ -0,0 +1,386 @@
...
+solveEqualityConstraint tools (v, p, ecProof) = do
+    (solution, proof) <- simplifyAnd tools p
+    return
+        ( (v, unificationSolutionTerm solution)
+          : unificationSolutionConstraints solution
+        , CombinedUnificationProof [ecProof, proof]
+        )
+
+-- iteratively finds ocurrences of x = t1 /\ x = t2, solves them, and
+-- computes the new substitution
```
virgil-serbanuta: Could you explain (as a code comment) why you can't just split the substitution by variable, then foldr with a combination of normalizeSubstitution1 and solveEqualityConstraint? If you don't already have a test for this case, could you add one?

src/main/haskell/language-kore/src/Data/Kore/Unification/UnifierImpl.hs
```diff
@@ -0,0 +1,386 @@
...
+                    )
+            }
+        , AndDistributionAndConstraintLifting
+            (applicationSymbolOrAlias ap)
+            subProofs
+        )
+
+-- finds ocurrences of x = t1 /\ x = t2 and transforms them
+-- into x = (t1 /\ t2)
+normalizeSubstitution1
```
virgil-serbanuta: why not "mergeConsecutiveAssigmenntsForSameVariable" or something similar?

src/main/haskell/language-kore/src/Data/Kore/Unification/UnifierImpl.hs
```diff
@@ -0,0 +1,386 @@
...
+    -- ^functions yielding metadata for pattern heads
+    -> CommonPurePattern level
+    -- ^left-hand-side of unification
+    -> CommonPurePattern level
+    -> Either
+        (UnificationError level)
+        (UnificationSubstitution level, UnificationProof level)
+unificationProcedure tools p1 p2
+    | p1Sort /= p2Sort
+    = Left (SortClash p1Sort p2Sort)
```
virgil-serbanuta: Either decrease the indentation by two spaces, or add 4. Or move the '=' on the previous line, I'm not sure.

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,481 @@
...
+                (extractPurePattern term1)
+                (extractPurePattern term2)
+            )
+        )
+
+unificationTests :: TestTree
+unificationTests =
+    testGroup
+        "Unification Tests"
+        [ unificationSuccess "Constant"
```
virgil-serbanuta: Wouldn't this be more readable as:
```
unificationSuccess "Constant"
    (UnificationTerm aA)
    (UnificationTerm aA)
    (UnificationResult aA)
   ...
```
?

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,481 @@
...
+unificationSuccess
+    :: (HasCallStack)
+    => String
+    -> CommonPurePatternStub Object
+    -> CommonPurePatternStub Object
+    -> CommonPurePatternStub Object
+    -> Substitution Object
+    -> UnificationProof Object
+    -> TestTree
+unificationSuccess message term1 term2 term subst proof =
```
virgil-serbanuta: Shouldn't `term` be called `result`?

src/main/haskell/language-kore/src/Data/Kore/Unification/UnifierImpl.hs
bmmoore: If we're trying to keep to 80-column lines we should not allow identifiers longer than 40 characters!

traiansf: :-)

virgil-serbanuta: We could do that, but why? There are some good arguments for having a limit for line length, and some weaker arguments for it being 80 (or 100) characters. However, I am not familiar with arguments for identifier length (except for one-letter identifiers being bad in many cases).

Anyway, do we have 40-char identifiers in our code? If we have 2-3 examples, we could check if long names help or hurt.

Returning to my comment, the mergeConsecutive... example above was not meant to be copy-pasted into code, I was hoping that Traian would find a better name (that's why I added "or something similar"). Still, I think it's better than normalizeSubstitution1, which, I think, is almost misleading.

src/main/haskell/language-kore/src/Data/Kore/ASTPrettyPrint.hs
```diff
@@ -72,6 +72,32 @@ writeOneFieldStruct
...
+writeTwoFieldStruct flags name contenta contentb =
+    writeOneFieldStructK
+        flags
+        name
+        (  prettyPrint NeedsParentheses contenta
+        >> write " "
+        >> prettyPrint NeedsParentheses contentb
+        )
+
+writeThreeFieldStruct
```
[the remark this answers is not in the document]
traiansf: I added a TODO to consider removing them when refactoring the proof objects stuff.

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,488 @@
...
+                    ]
+                    (var nonLinY)
+                    (extractPurePattern (applyS nonLinG [nonLinX]))
+                , Proposition_5_24_3
+                    [ FunctionalHead (symbolHead nonLinAS) ]
+                    (var nonLinX)
+                    (extractPurePattern nonLinA)
+                ]
+            )
+        , unificationSuccess
```
virgil-serbanuta: s/unificationSuccess/andSimplificationSuccess/?

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,488 @@
...
+                    ]
+                    (var expX)
+                    (extractPurePattern (applyS expBin [expY, expA]))
+                ]
+            )
+        , unificationProcedureSuccess
+            "times(times(a, y), x) = times(x, times(y, a))"
+            (applyS expBin [applyS expBin [expA, expY], expX])
+            (applyS expBin [expX, applyS expBin [expY, expA]])
+            [ ("a", expY)
```
virgil-serbanuta: Add a todo to solve somehow the a=y, y=a substitution?

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,488 @@
...
+        , unificationFailure "non-constructor symbolHead left"
+            a2A
+            aA
+            (NonConstructorHead (symbolHead a2))
+        , unificationFailure "nested failure"
+            fa
+            fa1
+            (ConstructorClash (symbolHead a) (symbolHead a1))
+        , unificationFailure "Unsupported constructs"
+            fa
```
virgil-serbanuta: I liked the inline definitions from above more.

src/main/haskell/language-kore/test/parser/Data/Kore/Unification/UnifierTest.hs
```diff
@@ -0,0 +1,488 @@
...
+            faImpliesXa1
+            UnsupportedPatterns
+             {- currently this cannot even be built because of builder checkd
+        , unificationFailure "Unmatching sorts"
+            aA
+            bA
+            UnificationError
+            -}
+        ]
```
virgil-serbanuta: How about a test with a unification of f(X, X) with f(a, g(a))?

PR #96, opened 2018-05-03
Imp defined as Kore definitions

hreada:
There are two main files related to this PR:
(1) imp.k contains the complete Kore definition of IMP
(2) imp2.k contains the Kore definition of IMP that doesn't use any AC matching and context. 

When @mattoxb finish defining built-in modules of INT, LIST, SET, and MAP, this PR will be ready to review.
UPDATE: Built-in modules are done in this PR https://github.com/kframework/kore/pull/101

@traiansf @virgil-serbanuta Do we have any coding style for Kore definitions? I was not following any coding style but I'd love to accommodate my code if there's one. If we don't have one yet, I think we may want to set up a simple one (for writing Kore definitions), as these definitions are getting bigger and bigger.


virgil-serbanuta:
@hreada I'll try to write a Kore style guide that encodes the current practice, maybe we can add as a task at the next iteration meeting. I took a quick look and your Kore code looks mostly readable, but you should never use tabs. There are many reasons for that, but, as an example, look at the `exists` indentation displayed by the github file diff for the following code extracted from your PR:

```
  /* intAsAExp is functional */

  axiom{R}
    \forall{R}(
      I:Int{},
      \exists{R}(
        E:AExp{},
	\equals{AExp{},R}(
	  intAsAExp{}(I:Int{}),
	  E:AExp{})))
  []
```

Besides this, don't forget to update the golden test files as described here:
https://github.com/kframework/kore/blob/master/src/main/haskell/language-kore/README.md
i.e.
`stack test --no-keep-going --ta --accept`

hreada:
@virgil-serbanuta Thanks for the feedback. I've converted all tabs to spaces and update golden test files.

hreada:
This PR is ready for review

hreada:
@virgil-serbanuta Sorry about that. I'll do a second examine on the code to make sure there's no tab.

hreada:
It'd be helpful if @traiansf could also take a look at it.

hreada:
I'll need to adjust the syntax for attributes and make it up to date with the resent commit of `attributes.kore` to the master branch.

hreada:
@virgil-serbanuta Is there a way to add new attributes syntax? I need `seqstrict`. I tried to modify `attributes.kore`, but that doesn't seem to work.

virgil-serbanuta:
@hreada you should change https://github.com/kframework/kore/blob/master/src/main/haskell/language-kore/src/Data/Kore/Implicit/Attributes.hs , which will generate the attributes.kore file (you have to run the tests with --accept twice since the attributes.kore file is actually a golden test file, and then you also need to update the attributes.kore.golden file).

src/main/haskell/language-kore/stack.yaml
```diff
@@ -0,0 +1,66 @@
+# This file was automatically generated by 'stack init'
```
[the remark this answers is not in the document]
hreada: I'll remove this auto-generated file and update the gitignore file accordingly.

src/main/k/kore.k
```diff
@@ -0,0 +1,61 @@
...
+                   | "\\iff"  "{" Sort "}" "(" Pattern "," Pattern ")"
+                   | "\\forall" "{" Sort "}" "(" Variable "," Pattern ")"
+                   | "\\exists" "{" Sort "}" "(" Variable "," Pattern ")"
+                   | "\\ceil" "{" Sort "," Sort "}" "(" Pattern ")"
+                   | "\\floor" "{" Sort "," Sort "}" "(" Pattern ")"
+                   | "\\equals"  "{" Sort "," Sort "}" "(" Pattern "," Pattern ")"
+                   | "\\in" "{" Sort "," Sort "}" "(" Pattern "," Pattern ")"
+                   | "\\top" "{" Sort "}" "(" ")"
+                   | "\\bottom" "{" Sort "}" "(" ")"
+                   | "\\dv" "{" Sort "}" "(" Pattern ")"
```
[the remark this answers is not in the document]
hreada: OK. I don't know why these stuffs show up. They belong to Kore-in-K and has nothing to do with this PR. Will remove them.

src/test/resources/imp.kore
```diff
@@ -69,45 +223,297 @@ module IMP
...
+
+  /* The injection function from any sort S to K. */
+  symbol asK{S}(S) : K{} []
+
+  /* Two injectivity axioms:
+   *   not (asK{S1}(X:S1) /\ asK{S2}(Y:S2))
+   *   asK{S}(X:S) /\ asK{S}(Y:S) -> asK{S}(X:S /\ Y:S)
+   */
+
+  axiom{S1,S2}
```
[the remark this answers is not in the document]
hreada: Nice point. This will need meta-level.

hreada:
@virgil-serbanuta I've got a problem with updating `Attribute.kore`. After I changed https://github.com/kframework/kore/blob/master/src/main/haskell/language-kore/src/Data/Kore/Implicit/Attributes.hs, I ran `stack test --no-keep-going --ta --accept` twice, but `Attributes.kore` is not modified. Could you please let me know if I did anything wrong?

src/test/resources/imp.kore
```diff
@@ -69,45 +235,311 @@ module IMP
...
+  /* Two injectivity axioms:
+   *   not (asK{S1}(X:S1) /\ asK{S2}(Y:S2)) if S1 and S2 are different
+   *   asK{S}(X:S) /\ asK{S}(Y:S) -> asK{S}(X:S /\ Y:S)
+   */
+
+  axiom{S1,S2,#R}
+    \implies{#R}(
+      \not{#R}(
+        \equals{#Sort{},#R}(
+          #S1:#Sort{},
```
virgil-serbanuta: You're relying here on S1 being lifted to the #S1 variable, which makes me feel a bit tensed because the connection is not obvious.

Maybe this is not a big issue that's worth fixing. However, a bigger problem is that there are wellformed patterns which are not liftable (i.e. their lifted form is not wellformed), e.g.
```
axiom{S, #R}
  \forall{#R}(
    #S:#Pattern{}
    \implies{K{}}(
      \and{K{}}(
        asK{S}(X:S),
        asK{S}(Y:S)),
      asK{S}(\and{S}(X:S, Y:S))))
```

virgil-serbanuta:
I forgot: please fix the tests.

src/test/resources/imp.kore
hreada: The fact that S1 is lifted to #S1:#Sort{}, is according to the Kore semantics (Section 9.2.3 of the Semantics of K document). The lift rule is the follows:
[[_]] : <object-sort> -> <meta-pattern>
[[s]] = '#' s ':' '#Sort{}'    if s is a sort variable.

hreada: @virgil-serbanuta Could you please inform me where your example comes from? I can recognize that part of it is the axiom in Line 367, but why there's a \forall{#R} in front of it?

hreada:
@virgil-serbanuta I'm trying to fix the failed tests issue. I ran `stack test --no-keep-going --ta --accept` to update the golden files, but it prompted this error msg:
```
Test suite failure for package matching-logic-0.0.1.0
    test:  exited with: ExitFailure 1
Full log available at .../.stack-work/logs/matching-logic-0.0.1.0-test.log
    Invalid option `--accept'
    Usage: test [-p|--pattern PATTERN] [-t|--timeout DURATION]
                [-j|--num-threads NUMBER] [--xml ARG] [-l|--list-tests]
                [-j|--num-threads NUMBER] [-q|--quiet] [--hide-successes]
                [--color never|always|auto]
```
It seems that the matching-logic-0.0.1.0 test suite doesn't accept the option `--accept`. Could you help on this issue?

src/test/resources/imp.kore
virgil-serbanuta: The origin of this example is not important. Yes, indeed, I took one of your axioms and changed it in order to provide an illustration for the issue I mentioned, but that's not the point. The main question is: do we want to allow kore to have wellformed patterns which cannot be lifted?

If yes, we should change our code to do a wellformedness check after lifting, and handle lifting errors (I think we currently assume that lifting just works). If not, we should either change the wellformedness requirements for kore or we should change how lifting is done.

virgil-serbanuta:
@hreada That's interesting. I have no clue how flag parsing for tests works, but it seems that Haskell ignores flags from libraries that are unused. The easiest fix is to do this:
```
cd language-kore
stack init
stack test --no-keep-going --ta --accept
cd ..
```

However, please note that the test that is currently failing on Jenkins is a Scala test, so it's unrelated to `stack`.

hreada:
@virgil-serbanuta Thank you. I noticed that (it took me a while to realize that I *do* have the access to see the details of tests failure). I am fixing that one now.

hreada:
It turns out that the Scala parse doesn't implement import and multiple modules yet. I'll fix that in this PR.

hreada:
I fixed the stack overflow issue. I also added hook-sort and hook-symbol to the scala parse, as they are in the same Trello card. Please review.

src/main/scala/org/kframework/kore/parser/KoreToText.scala
```diff
@@ -42,8 +42,25 @@ object KoreToText {
...
+    case HookSortDeclaration(params, sort, att) =>
+      "hook-sort" +
+        " " +
+        apply(sort) +
+        " " +
+        apply(att)
+
     case SymbolDeclaration(symbol, argSorts, returnSort, att) =>
       "symbol " +
+        apply(symbol) +
```
virgil-serbanuta: I'm not sure about the default Scala coding conventions, but here you're using a different indentation from below (i.e. for the string concatenation in HookSymbolDeclaration)

hreada: Thanks. Fixed.

Issue #94, opened 2018-04-29
Missing location information from error when unexpected sort


radumereuta:
Input:
```
[]
module TEST

sort Int{} []
sort String{} []
sort Exp{} []

symbol inj{SIn,SOut}(SIn): SOut []
symbol info{S}(String{}, S) : S []
symbol input{}(Int{}, Int{}) : String{} []

axiom{} // input: "-1"
 info{Exp{}}(input{}(\dv{Int{}}("1"),\dv{Int{}}("2")),
        inj{Int{},Exp{}}(
            info{Int{}}(\and{String{}}(input{}(\dv{Int{}}("1"),\dv{Int{}}("2")),
                            \equals{String{},String{}}(input{}(\dv{Int{}}("1"),\dv{Int{}}("2")), "-15")),
                        \dv{Int{}}("-1"))
    )) []
endmodule []
```
output:
```
c:\work\kore\src\main\haskell>stack exec -- kore-parser ..\..\..\..\test\test2.kore
"Reading the input fileTimeSpec {sec = 0, nsec = 635559}"
"Parsing the fileTimeSpec {sec = 0, nsec = 1488235}"
"Verifying the definitionTimeSpec {sec = 0, nsec = 1661928}"
kore-parser.EXE: Error in module 'TEST' -> axiom declaration -> symbol or alias 'info' -> symbol or alias 'inj' -> symbol or alias 'info' -> \and -> \equals -> <string>: Expecting object sort 'String{}' but got meta sort '#CharList{}'.
CallStack (from HasCallStack):
  error, called at app\parser\Main.hs:90:49 in main:Main
```
The problem: `"-15"` was supposed to be `\dv{Int{}}("-15")`, but I had to find out by removing parts of the term and trying to parse.
This example can be minimized further, but I chose to leave it like this so it becomes more clear why we need location information everywhere.

Version used: 0e9818a (latest master)

radumereuta:
Input:
```
[]
module INJ
  sort KInt{} []
  symbol input{}(KInt{}) : KInt{} []
  axiom{} input{}(\dv{KInt{}}("1"), \dv{KInt{}}("1")) []
endmodule []
```
Output:
```
c:\work\kore\src\main\haskell>stack exec -- kore-parser ..\..\..\..\test\test4.kore
"Reading the input fileTimeSpec {sec = 0, nsec = 691218}"
"Parsing the fileTimeSpec {sec = 0, nsec = 2134841}"
[...]

PR #100, opened 2018-05-08
Source code location information for IDs

virgil-serbanuta:


src/main/haskell/language-kore/src/Data/Kore/AST/Common.hs
```diff
@@ -39,6 +39,22 @@ import           Data.Proxy
 
 import           Data.Kore.AST.MetaOrObject
 
+data FileLocation = FileLocation
+    { fileName :: FilePath
+    , line     :: Int
+    , column   :: Int
+    }
+    deriving Show
```
radumereuta: If I may. In K, we have two classes that represent location information for terms: Location(<startLine>, <startColumn>, <endLine>, <endColumn>), Source(<string>).
When I'm trying to parse a term, sometimes I get ambiguities, and knowing where parsing ended can be really useful. The source of a term may be coming from other locations than files alone. Command line arguments, generated by front-end desugar steps... I've also been talking with @grosu lately and I think a more appropriate location format would be to have 6 fields: Location(<startLine>, <startColumn>, <endLine>, <endColumn>, <offset>, <length>). This would cover a wider range of situations and allow us to do more things with those terms. (one example that came up was to check if you have whitespace before or after a specific term)

virgil-serbanuta: Thanks, that's useful. I intended the FileLocation to be an actual position in a file, and the AstLocation to be the location for an AST term, and it has a much richer representation. However, I didn't include file ranges, which are likely to be useful.

I may or may not change its representation right now, but even if I don't, I added a comment explaining that it should be treated as an opaque token as much as possible, so it should be easy to change.

radumereuta:
The error message for the original example looks a lot better, but I found another one, where arity errors don't contain location information. Can you also look over that? See #94.

virgil-serbanuta:
@radumereuta I'll submit this as-is and let Traian implement locations for patterns.

PR #106, opened 2018-05-21
Fixing Haddock errors

traiansf:
Also added some compilation annotations, as `stylish-haskell` does not seem to be reading project setup info.

### DIFF
--- a/src/main/haskell/matching-logic/src/Kore/MatchingLogic/ProofSystem/Minimal.hs
+++ b/src/main/haskell/matching-logic/src/Kore/MatchingLogic/ProofSystem/Minimal.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-|
 Description: The minimal matching logic proof system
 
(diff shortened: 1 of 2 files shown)


traiansf:
Jenkins: test this please

traiansf:
I'd like to merge this, as the build currently fails for other pull requests because of haddock.  @virgil-serbanuta could you please take a look?

src/main/haskell/matching-logic/src/Kore/MatchingLogic/ProofSystem/Minimal.hs
```diff
@@ -175,7 +177,7 @@ instance (IsSignature sig, Eq (Sort sig), Eq (Label sig), Eq var) =>
                 phi1 == Just term1a, phi2 == Just term1b -> Right ()
             _ -> Left (Error [] "not proved")
 
-     -- ^ sigma(before ..,\phi1 \/ \phi2,.. after) <->
+     --   sigma(before ..,\phi1 \/ \phi2,.. after) <->
```
virgil-serbanuta: Since you removed the ^, could you move this before the thing that it is documenting? The same below.

PR #110, opened 2018-05-24
Minimal one plus one proof test

bmmoore:


src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne.hs
```diff
@@ -0,0 +1,446 @@
...
+ | Ax_out_ex Ctx Nat Formula
+ | Ax_out_or Ctx Formula Formula
+ | Ax_existence Nat
+ | Ax_singvar Ctx_path Ctx_path Nat Formula
+ | Ax_definedness
+ | Ax_proj1 Formula Formula
+ | Ax_proj2 Formula Formula
+ | Ax_and_intro Formula Formula
+ | Ax_or_elim Formula Formula Formula
+ | Ax_or_intro1 Formula Formula
```
virgil-serbanuta: A text description of these rules would be nice. I mean, I know some of them, e.g. Modus Ponens, but I think that this is the first time I see, say, Ax_or_intro1. That text description can be in the ProofSystem.hs file, but then it would be nice to have a pointer there. Also, why is there a naming difference between these and the ProofSystem stuff (besides the `Ax` prefix)?

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -0,0 +1,283 @@
...
+  checkDerivation conclusion rule = case rule & ruleTerms %~ Just of
+      Propositional1 a b ->
+          expect $ a --> b --> a
+      Propositional2 a b c ->
+          expect $ (a --> b --> c) --> (a --> b) --> (a --> c)
+      Propositional3 a b ->
+          expect $ (notP' a --> notP' b) --> (b --> a)
+      ModusPonens a (ImpliesP s a' b) | a == a' ->
+          expect $ Just b
+      ModusPonens _ _ -> Left (Error [] "hypotheses have wrong form")
```
virgil-serbanuta: Could you explain why this match produces an error?

bmmoore: This line is when the hypotheses don't have the form `A->B` and `A`.

bmmoore:
I just rebased onto master.

PR #112, opened 2018-05-29
Added definition for Kore written in K.

mattoxb:
This PR includes a Kore (syntax) definition in K. 
A bash script is provided to run all tests in `test/resources`, except string tests and exception tests.
All tests (except the above mentioned) passed.

### DIFF
--- a/docs/semantics-of-k.tex
+++ b/docs/semantics-of-k.tex
@@ -405,7 +405,8 @@
   \newcommand{\KgetSymbols}{\texttt{\sharpsymbol getSymbols}}
   \newcommand{\KsortDeclared}[1]{
   	\parametric{\texttt{\sharpsymbol sortDeclared}}{#1}}
-  \newcommand{\KsortsDeclared}{\texttt{\sharpsymbol sortsDeclared}}
+  \newcommand{\KsortsDeclared}[1]{
+        \parametric{\texttt{\sharpsymbol sortsDeclared}}{#1}}
   \newcommand{\KsymbolDeclared}[1]{
   	\parametric{\texttt{\sharpsymbol symbolDeclared}}{#1}}
   \newcommand{\KaxiomDeclared}{\texttt{\sharpsymbol axiomDeclared}}
@@ -1353,7 +1354,7 @@ $$
 \textrm{for any sorts $s,s'$} \\
 \end{array}
 $$
-And here are some \emph{pattern schemas}, each defining infinitely many patters:
+And here are some \emph{pattern schemas}, each defining infinitely many patterns:
 $$
 \begin{array}{rl}
 \parametric{\append}{s}(\parametric{\nil}{s},l'\cln\parametric{\List}{s})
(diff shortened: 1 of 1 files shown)


src/main/k/kore.k
```diff
@@ -2,7 +2,7 @@
 module KORE
   // imports STRING
 
-  syntax Module ::= "module" ModuleName Declarations "endmodule" Attribute
+  syntax Module ::= Attribute "module" ModuleName Declarations "endmodule" Attribute
```
hreada: See Semantics of K document, section 9.1.6 - 9.1.7, a module has its attributes in the end. A Kore definition consists of multiple modules, and the attribute for the entire Kore definition appear in the beginning.

mattoxb: Okay, that makse things clear.

virgil-serbanuta:
I've added Traian as a reviewer since I'm supposed to work on RV-Predict stuff this week.

src/main/k/tests/map.core
```diff
@@ -0,0 +1,128 @@
+module MAP
```
traiansf: `map.kore`?

src/test/resources/imp.kore
```diff
@@ -233,7 +233,7 @@ module MAP
 
   symbol merge{S1,S2}(Map{S1,S2}, Map{S1,S2}) : Map{S1,S2} []
 
-  symbol keys{S1,S2}(Map{S1,S2}) : Set{S1} []
+  symbol domains{S1,S2}(Map{S1,S2}) : Set{S1} []
```
traiansf: All files in `src/test/resources` are tested with the Kore parser;  therefore, if you changed this file, you also need to update the corresponding `.golden` file.  you can do that with `stack test --ta --accept`.

src/main/k/tests/bool.kore
```diff
@@ -0,0 +1,112 @@
+module BOOL
```
traiansf: It would be good if all self-contained definitions would also be tested through the KORE parser, to make sure things stay synchronized;
One suggestion would be that all files here would just be required to be parseable by the KORE parser.

src/main/k/tests/map.core
hreada: Fixed.

src/test/resources/imp.kore
hreada: Fixed.

src/main/k/tests/bool.kore
hreada: I think these tests should be moved to `test/resources`, and we should remove `k/tests` directory. I think it's a good practice to keep all tests in one place.
What do you think, @mattoxb ?

traiansf: Anyway, I suggest we merge it as it is now, and add a new task for reorganizing tests and updating the code appropriately.

traiansf:
Moved to a new branch and reopened as https://github.com/kframework/kore/pull/240

PR #116, opened 2018-06-04
minor edits for typos

njohnwalker:
Some things that stuck out as I was reading through the paper.
Maybe this doesn't deserve a pull request, but I'm just trying to get used to the workflow.


traiansf:
@njohnwalker you need to also ask somebody to review it; In general it should be preferably somebody managing/modifying often the code you altered

docs/semantics-of-k.tex
```diff
@@ -3884,7 +3884,6 @@ \subsection{Matching Logic Theories}
 \end{equation}
 checks whether a list of sorts are declared:
 \begin{align*}
-& \KsortsDeclared{\shs} \colon \KSortList \to \shs \\
```
njohnwalker: It seems like this is redundant but maybe I misunderstood it.

traiansf: It does seem redundant, indeed, so removing it seems like the right choice.
Since you're at it, though, maybe you could wrap the following usages into `\parametric`, to get their `{` `}` around the `\shs` argument.

PR #119, opened 2018-06-08
Generating approximate locations for patterns

traiansf:


virgil-serbanuta:
BTW, you should also link this to https://github.com/kframework/kore/issues/94 , unless this comment already does that.

PR #120, opened 2018-06-11
Command line arg lib

njohnwalker:
Adds command line argument parsing library to kore-parser.

Main.hs now uses [optparse-applicative](http://hackage.haskell.org/package/optparse-applicative)
This library supports adding tab completion in the shell, but this would require additional install scripts so I have held off for now.

I tried my best to emulate the flags that were already present but those are easy to change if need be.


src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -61,3 +98,38 @@ verifyMain willChkAttr definition =
                  Left err1 -> error (printError err1)
                  Right _   -> return ()
           }
+
+
+----------------------
+-- Helper Functions --
```
njohnwalker: The functions below only have a single dependency each so it makes sense to move them to a more general module(they will come in handy for every Main module) but I wasn't sure where to put them.

traiansf:
Looks reasonable to me, but I defer judgement to @virgil-serbanuta as he wrote the original code.  It is not clear to me how the default values (all flags are enabled by default) are set in the PR, although they seems to the set, as the tests pass.

src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -2,102 +2,135 @@ module Main where
...
+      ( fullDesc
+     <> progDesc "Parses Kore definition in FILE; optionally, \
+                 \Verifies well-formedness"
+     <> header "kore-parser - a parser for Kore definitions" )
+
+
+main :: IO ()
+main =
+    do {
+    ; KoreParserOptions
```
virgil-serbanuta: I have mixed feelings about this way of writing 'do' statements because it is fairly different from what we currently use, and consistency is nice. On the other hand, it looks nicer, because it's easier to see what is a top-level statement in a 'do'. @traiansf , what do you think?

njohnwalker:
I will add an indication of default values to the help text and documentation

src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -2,102 +2,135 @@ module Main where
...
-        | flag == "noprint" =
-            (parseCommandLineFlags commandLine)
-                { commandLineFlagsPrint = False }
-        | otherwise =
-            error ("Unknown flag: --" ++ flag)
-parseCommandLineFlags (flag : commandLine) =
-    (parseCommandLineFlags commandLine) { commandLineFlagsFileName = Just flag }
+    opts = info (commandLineParser <**> helper)
+      ( fullDesc
+     <> progDesc "Parses Kore definition in FILE; optionally, \
```
virgil-serbanuta: We usually indent this one more space and we add an extra space before 'fullDesc' to get the same alignment effect.

src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -2,102 +2,135 @@ module Main where
...
-                { commandLineFlagsPrint = True }
-        | flag == "noprint" =
-            (parseCommandLineFlags commandLine)
-                { commandLineFlagsPrint = False }
-        | otherwise =
-            error ("Unknown flag: --" ++ flag)
-parseCommandLineFlags (flag : commandLine) =
-    (parseCommandLineFlags commandLine) { commandLineFlagsFileName = Just flag }
+    opts = info (commandLineParser <**> helper)
+      ( fullDesc
```
virgil-serbanuta: Prefer indenting with multiples of 4 spaces.

src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -2,102 +2,135 @@ module Main where
...
-        , commandLineFlagsPrint = True
-        }
-parseCommandLineFlags (('-' : '-' : firstFlag) : commandLineReminder) =
-    addFlagAndContinue firstFlag commandLineReminder
+                    
+{-| Command Line Argument Parser -}
+commandLineParser :: Parser KoreParserOptions
+commandLineParser = KoreParserOptions
+      <$> argument str
+              ( metavar "FILE"
```
virgil-serbanuta: Prefer indenting with multiples of 4 spaces. Also, maybe one more space before 'metavar'.

src/main/haskell/language-kore/app/parser/Main.hs
njohnwalker: Maybe only in complicated 'do' statements? We don't need so much syntax when it's a straightforward 'do' block.  It just seems useful when we have to mix other nested language constructs. Like in 'main' definitions.

virgil-serbanuta: I allowed this usage in a restricted way (i.e. when it actually helps), see the style guide doc, Identation, Do. If you think that's not reasonable, please say so (in which case an example showing why the style guide is not reasonable would help).

src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -1,123 +1,129 @@
...
 main :: IO ()
 main =
     do {
     ; KoreParserOptions
-          { fileNameArgument  = fileName
-          , willPrintOption   = willPrint 
-          , willVerifyOption  = willVerify
-          , willChkAttrOption = willChkAttr
-          } <- commandLineParse
+      { fileName    = fileName
```
traiansf: @bmmoore would suggest to also use record puns https://downloads.haskell.org/~ghc/7.2.1/docs/html/users_guide/syntax-extns.html#record-puns 
in combination with DuplicateRecord fields, especially for a use-case like yours below.

PR #124, opened 2018-06-15
Simple stepper.

virgil-serbanuta:


src/main/haskell/language-kore/test/parser/Data/Kore/Step/StepTest.hs
```diff
@@ -0,0 +1,583 @@
...
+stepTests :: TestTree
+stepTests =
+    testGroup
+        "Step Tests"
+        [ singleStepTests
+        , multipleStepPickFirstTests
+        ]
+
+singleStepTests :: TestTree
+singleStepTests =
```
traiansf: could you please, as you did for the the step with axiom tests, also write a brief textual description of each test? i.e., what rules apply, on which term, and what is the result.

src/main/haskell/language-kore/test/parser/Data/Kore/Step/StepTest.hs
```diff
@@ -0,0 +1,583 @@
...
+    Proposition_5_24_3
+        functionalProof
+        (asMetaVariable variable)
+        (asPureMetaPattern patt)
+
+functionalVariable
+    :: MetaSort sort => MetaVariable sort -> FunctionalProof Meta Variable
+functionalVariable = FunctionalVariable . asMetaVariable
+
+runStep
```
traiansf: I'm wondering whether the next rule should be in the public module for the Stepper instead of the testing module, but I'm not sure.  What do you think about it?

virgil-serbanuta: I wasn't sure either. I was waiting for something that would help me decide whether I should expose an interface that takes an int and returns another one instead of returning an IntCounter

PR #125, opened 2018-06-18
`matching-logic` pedantic build

njohnwalker:
minor edits to `matching-logic` to clear the gauntlet of "-Wall -Werror"

mostly name shadowing and unused matches in cases of orphan instances, warnings were suppressed.  maybe refactors could fix these issues but it would be non-trivial.

A lot of warnings for shadowed variables in helper functions, maybe this is worth a snippet in the style guide.


src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ForallAST.hs
```diff
@@ -4,6 +4,8 @@
 {-# LANGUAGE TemplateHaskell            #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE ViewPatterns               #-}
+{-# OPTIONS_GHC -fno-warn-missing-signatures #-}
```
bmmoore: Why add `no-warn-missing-signatures` here? It shouldn't be hard to add the inferred signatures for whatever was missing.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -7,10 +7,10 @@ connectives, which were needed to complete the 1+1=2 proof.
  -}
 module Kore.MatchingLogic.ProofSystem.OnePlusOne.ProofSystem where
 
-import           Data.Text.Prettyprint.Doc       (Pretty (pretty))
+--import           Data.Text.Prettyprint.Doc       (Pretty (pretty))
```
bmmoore: Just delete this

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -155,17 +155,17 @@ instance (IsSignature sig, Eq (Sort sig), Eq (Label sig), Eq var) =>
                   Left (Error [] "right hand term does not match phi[y/x]")
               | otherwise -> Left (Error [] "conclusion does not agree with arguments")
             _ -> Left (Error [] "malformed conclusion")
-      ForallRule var term1 term2 ->
+      ForallRule _ term1 term2 ->
```
bmmoore: var shouldn't be unused here, it should be tested to be the same as `sVar` and `sVar1`. This file was probably copied before I fixed that bug in the other proof system file.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -179,9 +179,9 @@ instance (IsSignature sig, Eq (Sort sig), Eq (Label sig), Eq var) =>
               _ -> Left (Error [] "conclusion has wrong form")
       Framing _ _ _ ->
           Left (Error [] "hypothesis has wrong form")
-      PropagateOr label pos phi1 phi2 -> do
+      PropagateOr _ pos phi1 phi2 -> do
```
bmmoore: This `label` also needs to match `label1` instead of being ignored.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -196,9 +196,9 @@ instance (IsSignature sig, Eq (Sort sig), Eq (Label sig), Eq var) =>
 
      -- ^ sigma(before ..,\phi1 \/ \phi2,.. after) <->
      --     sigma(before ..,\phi1, .. after) \/ sigma(before ..,\phi2,.. after)
-      PropagateExists label pos var term ->
+      PropagateExists _ pos var term ->
```
bmmoore: This `label` should also be used.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
njohnwalker: Can I merge those changes?  If I add `var == sVar` to the guard, ghc complains about missing Eq constraints in the ToPattern class

bmmoore: Oh, wrong suggestion. The missing checks are `var == var1, var1 == var2`. You can compare with the guards in `Kore.MatchingLogic.ProofSystem.Minimal`.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne.hs
```diff
@@ -310,13 +310,13 @@ emit pat rule = ConvM (state (\(next,lines) -> (next,(next+1,(next,pat,rule):lin
 
 convert :: Simple_proof -> ConvM Int
 convert (Conclusion formula rule) = convertRule convert rule >>= \case
-    Right rule -> emit (convertFormula formula) rule
-    Left ix' -> return ix'
-
+    Right rule' -> emit (convertFormula formula) rule'
+    Left ix -> return ix
+{-
```
virgil-serbanuta: Is there a reason to keep this commented function? The same below.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne.hs
```diff
@@ -255,7 +255,7 @@ convertRule convertHyp r = case r of
     Ax_propositional1 f1 f2 -> Propositional1 (convertFormula f1) (convertFormula f2)
     Ax_propositional2 f1 f2 f3 -> Propositional2 (convertFormula f1) (convertFormula f2) (convertFormula f3)
     Ax_propositional3 f1 f2 -> Propositional3 (convertFormula f1) (convertFormula f2)
-    Ax_varSubst x y p1 p2 -> VariableSubstitution (SubstitutedVariable (nat2Int x))
+    Ax_varSubst x y p1 _ -> VariableSubstitution (SubstitutedVariable (nat2Int x))
```
virgil-serbanuta: It's a bit strange that p2 is not needed here, I think that it's the only rule where one of the pieces is ignored. On the other hand, I think that the ML variable substitution rule had only one pattern, so I guess it might be fine.

@bmmoore What does p2 mean here?

bmmoore: p2 here is the result of the substitution, but the proof checking for VariableSubstitution always computes the expected result with varSubst. The p2 parameter exists here because the Coq definition defined substitution as a relation, initially to reflect the matching logic paper allowing capture-avoiding substitution and also to more easily experiment with variants on side conditions prohibiting capture.

In the end the conditions we already put on the VariableSubstitution rule did not interfere too badly, and this translation should preserve the conclusion of the proof rule (the overall consistency of the translated proof is checked when the proof object is assembled).

bmmoore:
Changes look good. Does the build need to be rerun, or is that a real test failure?

njohnwalker:
The 1 + 1 = 2 proof failed, but it failed before I changed anything.

src/main/haskell/matching-logic/test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -226,7 +226,7 @@ instance (IsSignature sig, Eq (Sort sig), Eq (Label sig), Eq var) =>
                 AndP _ (VariableP sVar1 var1) term1' ->
                   case occ2 of
                     AndP _ (VariableP sVar2 var2) (NotP _ term2')
-                      | sVar1 == sVar2, var1 == var2, term1 == term2',
+                      | sVar1 == sVar2, var1 == var2, term1' == term2',
```
njohnwalker: Breaking the proof _was_ my fault, got too fancy with the variable renaming.

PR #126, opened 2018-06-22
Common module for Main modules

njohnwalker:
I added a `--version` flag to kore-parser and moved the common elements of the kore-parser main module to a new module `GlobalMain`.  GlobalMain exports a function `mainGlobal` that takes an arbitrary command line parser and combines it with it's predefined parser, handles its global flags and returns all of the parsed options.

I wasn't sure how to expand the project directory structure, but I'm guessing `GlobalMain.hs`  is not in the right place.

In order to print the version information in the `package.yaml` I would need to add some options to that file that would require setting a minimum version of stack (v1.7) so I have held off on that for now.


njohnwalker:
I forgot to merge this branch with upstream, should work now

PR #127, opened 2018-06-22
global main module and --version flag

njohnwalker:
I added a --version flag to kore-parser and moved the common elements of the kore-parser main module to a new module GlobalMain. GlobalMain exports a function mainGlobal that takes an arbitrary command line parser and combines it with it's predefined parser, handles its global flags and returns all of the parsed options.

I wasn't sure how to expand the project directory structure, but I'm guessing GlobalMain.hs is not in the right place.

In order to print the version information in the package.yaml I would need to add some options to that file that would require setting a minimum version of stack (v1.7) so I have held off on that for now.

### DIFF
--- a/src/main/haskell/language-kore/app/parser/GlobalMain.hs
+++ b/src/main/haskell/language-kore/app/parser/GlobalMain.hs
@@ -34,7 +34,7 @@ import           Options.Applicative                    ( Parser
                                                         , execParser )
 
 
-{- | Record Type containing common command-line arguments to each executable in 
+{- | Record Type containing common command-line arguments for each executable in 
 the project -}
 data GlobalOptions = GlobalOptions
     { willVersion    :: !Bool -- ^ Version flag [default=false]
--- a/src/main/haskell/language-kore/app/parser/GlobalMain.hs
+++ b/src/main/haskell/language-kore/app/parser/GlobalMain.hs
@@ -1,0 +1,161 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE CPP             #-}
+
+module GlobalMain
+    ( MainOptions(..)
+    , GlobalOptions(..)
+    , mainGlobal
+    , enableDisableFlag
+    , clockSomething
+    , clockSomethingIO
+    ) where
+
+import           Control.Monad                          ( when )
+import           Control.Exception                      ( evaluate )
+import           Data.Semigroup                         ( (<>) )
+import           Development.GitRev                     ( gitBranch
+                                                        , gitHash
+                                                        , gitCommitDate )
+import           System.Clock                           ( Clock (Monotonic)
+                                                        , diffTimeSpec
+                                                        , getTime )
+import           Options.Applicative                    ( Parser
+                                                        , InfoMod
+                                                        , (<|>)
+                                                        , (<**>)
+                                                        , flag
+                                                        , flag'
+                                                        , long
+                                                        , hidden
+                                                        , internal
+                                                        , help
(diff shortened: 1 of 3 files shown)


src/main/haskell/language-kore/app/parser/GlobalMain.hs
```diff
@@ -0,0 +1,161 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RecordWildCards #-}
```
virgil-serbanuta: Is using RecordWildCards really a good idea? Our current style guide does not say anything about it, but I may change it based on this discussion.

There are two problems that I see.

First, that it makes it harder to understand what happens there (i.e. that you have a projection which is implicitly applied to a record as opposed to a constant function), it requires a lookup for the projection name when reading the code instead of just glancing over the code.

Second, that it can overlap with a similarly named local thing, and it's not always trivial to say which is which.

Now, there is a good chance that --pedantic catches the second issue, but I didn't check.

About the first issue, the current style guide allows two ways of naming record fields: either with `RecordWildCards` and free-syle-ish names, but then one is not allowed to use them as getters/projections, or the fields should be prefixed with the record name.

Now, that almost solves the first problem: using projections is only allowed when they include the structure name, and the structure name is available locally, since one must match on it in order to have the implicit projection argument stuff.

Still, I can't help but think that code is more readable when using explicit arguments. What do you think?

src/main/haskell/language-kore/app/parser/GlobalMain.hs
```diff
@@ -0,0 +1,161 @@
...
+and returns the parsed options
+-}
+mainGlobal
+    :: Parser options                -- ^ local options parser
+    -> InfoMod (MainOptions options) -- ^ option parser information
+    -> IO      (MainOptions options)
+mainGlobal localOptionsParser modifiers = do
+  options@MainOptions
+            { globalOptions = GlobalOptions{..}
+            , localOptions = _
```
virgil-serbanuta: Is this match necessary?

src/main/haskell/language-kore/app/parser/GlobalMain.hs
njohnwalker: I was trying to think of the best way to introduce the options into the scope of the main function that would be the easiest to change, as I anticipate more flags and functionality being added.  But I will switch it back to the {NamedFieldPuns}  This would only mean adding another match to the main function when you add another option (compared to no change at all) but i agree it would be easier to read

src/main/haskell/language-kore/app/parser/GlobalMain.hs
```diff
@@ -0,0 +1,161 @@
...
+                                                        , internal
+                                                        , help
+                                                        , helper
+                                                        , info
+                                                        , execParser )
+
+
+{- | Record Type containing common command-line arguments to each executable in 
+the project -}
+data GlobalOptions = GlobalOptions
```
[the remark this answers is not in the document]
njohnwalker: I want to leave this for now, it comes out cleaner to use the record syntax  compared to a newtype over a tuple, it's easier to lift over the argument parser and the named projections seem like a more readable syntax

src/main/haskell/language-kore/app/parser/GlobalMain.hs
```diff
@@ -0,0 +1,160 @@
...
+                                                        , long
+                                                        , hidden
+                                                        , internal
+                                                        , help
+                                                        , helper
+                                                        , info
+                                                        , execParser )
+
+
+{- | Record Type containing common command-line arguments to each executable in
```
virgil-serbanuta: s/to/for/

PR #131, opened 2018-06-27
Function evaluation

virgil-serbanuta:


virgil-serbanuta:
And-ing patterns: I hope that Data.Kore.ASTUtils.SmartConstructors would make this easier after #137 gets submitted.

PR #132, opened 2018-06-27
K definition of IMP with no builtins

traiansf:
`.k` definition contains no builtins, but the generated `.kore` file bring in some from `DOMAINS`.


src/main/k/imp-no-domains/imp.k
```diff
@@ -0,0 +1,158 @@
...
+  syntax AInt ::=  ".AInt" | ABool AInt | "(" AInt ")" [bracket]
+endmodule
+
+module AINT
+  import AINT-SYNTAX
+  import ABOOL-SYNTAX
+
+  syntax AInt
+    ::= AInt "+AInt" AInt [function]
+      | addWithCarry(AInt, AInt, ABool) [function]
```
virgil-serbanuta: It seems that the function attribute was lost when translating to kore.

PR #133, opened 2018-06-28
Pattern loading and verification

virgil-serbanuta:


src/main/haskell/language-kore/app/parser/Main.hs
```diff
@@ -77,33 +95,73 @@ main = do
...
+            else return Map.empty
+        when willPrint $ putStrLn (prettyPrintToString parsedDefinition)
+
+        when (patternFileName /= "") $ do
+            parsedPattern <- mainPatternParse patternFileName
+            when willVerify $ do
+                indexedModule <-
+                    mainModule (ModuleName mainModuleName) indexedModules
+                mainPatternVerify indexedModule parsedPattern
+            when willPrint  $ putStrLn (prettyPrintToString parsedPattern)
```
traiansf: So, when willPrint, it will print both the definition and the pattern?  What if I only want to see one of them?

PR #134, opened 2018-06-28
Demo - DO NOT REVIEW

virgil-serbanuta:
stack build --pedantic
stack exec -- kore-parser imp.kore --verify --print --pattern imp-exp.pattern.kore --module IMP --evaluate


traiansf:
in the `demo` dir use `make %.run`  where `%` is the filename (without extension) of a `.demo` file.
Update the definition `demo.k` to change the code run by a demo, or to add other function definitions or main demo constants.

PR #136, opened 2018-07-05
Load test resources by absolute paths

ttuegel:
References to test resource files are resolved into absolute paths into the unpacked source tree used to build the tests so that tests requiring resources can run from `stack test` or `stack repl`. The resources must be listed in `extra-source-files`. The source tree must remain unpacked while the tests are run.

We discussed the possibility of doing this with the generated Paths module; that would work, but it would require that the test resources be listed as `data-files` that would be installed when the library is installed. To keep the test resources from leaving the source tree, a small test-specific Paths module is implemented using Template Haskell.

I also opted to use hpack for matching-logic; language-kore was already using it, so I assume it's allowed. Besides reducing redundancy in the package description, it helps avoid errors due to unlisted modules.


src/main/haskell/matching-logic/test/Paths.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Paths where
+
+import qualified Language.Haskell.TH     as TH
+import qualified Language.Haskell.TH.Lib as TH
+import           System.FilePath         (takeDirectory, (</>))
+
+
+-- | Resolve the relative file name of a test resource.
```
bmmoore: I think this should interpret the paths relative to `src/main/haskell/matching-logic`, which is where the existing paths have been relative to. `pwd </> ".." </> relative` should do it.

ttuegel: A more general way to do this would be traverse up the directory tree until we find `package.yaml`. That would eliminate the need to change this file if we move the module, too.

ttuegel:
@bmmoore I changed `dataFileName` so that test resource file names are resolved relative to the package root.

ttuegel:
@bmmoore Would you mind reviewing this again?

bmmoore:
Looks pretty good. I would only ask to have the `Paths.hs`  at `test/Paths.hs` in both packages.

Now that it's looking up the filesystem for `package.yaml` rather than just relative to the source file it might work to do the search at runtime too, but keeping TH is fine.

ttuegel:
> I would only ask to have the Paths.hs at test/Paths.hs in both packages.

I have done this now.

> Now that it's looking up the filesystem for package.yaml rather than just relative to the source file it might work to do the search at runtime too, but keeping TH is fine.

Even if we do the search at runtime, we still need Template Haskell to find the source file location at compile time.

PR #137, opened 2018-07-06
Smart constructors

u--:


src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,124 @@
...
+freeVars p = S.unions $ map freeVars $ p ^. partsOf allChildren
+
+alphaRename
+  :: MetaOrObject level 
+  => CommonPurePattern level 
+  -> CommonPurePattern level 
+alphaRename = (\case 
+ Forall_ s1 v p -> go Forall_ s1 v p 
+ Exists_ s1 v p -> go Exists_ s1 v p
+ _ -> error "Input must be an Exist or Forall at the top level"
```
ttuegel: Can we make this function total? For example, if its argument is not a binder, could it just return the argument unmodified? Or, can we encode this restriction in the type?

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,124 @@
+{-|
```
ttuegel: Is it necessary to implement substitution ourselves, or could we outsource this to a library like [bound](https://hackage.haskell.org/package/bound)?

u--: Hmm. Maybe. There was already a substitution function in Kore/Substitution, but I needed one that substitutes terms for terms and not just terms for variables.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,124 @@
...
+import           Data.Kore.ASTUtils.SmartConstructors
+
+
+subst
+  :: MetaOrObject level 
+  => CommonPurePattern level 
+  -> CommonPurePattern level 
+  -> CommonPurePattern level 
+  -> CommonPurePattern level
+subst a b (Forall_ s1 v p) = handleBinder a b Forall_ s1 v p
```
ttuegel: It would be better to write this as `subst a b = \case ...` since the first two variables are always bound the same.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,124 @@
...
+subst a b (Exists_ s1 v p) = handleBinder a b Exists_ s1 v p
+subst a b pat =
```
ttuegel: The intention here might be clearer using guards rather than `if`.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,124 @@
...
+  if pat == a then b else Fix $ fmap (subst a b) $ unFix pat
```
ttuegel: I think `subst` could actually be rewritten as a catamorphism using `embed`. Maybe that's a little tricky if we're doing whole-term substitution instead of only variables. Although, maybe not: I don't see that `pat` could contain `a` as a subterm and also be equal to `a`.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/SmartConstructors.hs
```diff
@@ -0,0 +1,328 @@
...
+  In_     s1 s2   a b -> In_       <$> i s1 <*> o s2           <*> c a <*> c b
+  Next_      s2   a   -> Next_     <$>          o s2           <*> c a
+  Not_       s2   a   -> Not_      <$>          o s2           <*> c a
+  Or_        s2   a b -> Or_       <$>          o s2           <*> c a <*> c b
+  Rewrites_  s2   a b -> Rewrites_ <$>          o s2           <*> c a <*> c b
+  Top_       s2       -> Top_      <$>          o s2
+  Var_          v     -> Var_      <$>                   var v
+  other -> error $ "Unknown constructor " ++ show other
+
+inputSort        f = patternLens f    pure pure pure
```
virgil-serbanuta: Could you add function signatures? Also, could you document all exported functions?

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/SmartConstructors.hs
```diff
@@ -0,0 +1,328 @@
...
+pattern Forall_       s2 v a   = Fix (ForallPattern (Forall s2 v a))
+pattern Iff_          s2   a b = Fix (IffPattern (Iff s2 a b))
+pattern Implies_      s2   a b = Fix (ImpliesPattern (Implies s2 a b))
+pattern In_        s1 s2   a b = Fix (InPattern (In s1 s2 a b))
+pattern Next_         s2   a   = Fix (NextPattern (Next s2 a)) 
+pattern Not_          s2   a   = Fix (NotPattern (Not s2 a))
+pattern Or_           s2   a b = Fix (OrPattern (Or s2 a b))
+pattern Rewrites_     s2   a b = Fix (RewritesPattern (Rewrites s2 a b))
+pattern Top_          s2       = Fix (TopPattern (Top s2))
+pattern Var_             v     = Fix (VariablePattern v)
```
virgil-serbanuta: Why didn't you add patterns for strings and chars?

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/SmartConstructors.hs
```diff
@@ -0,0 +1,328 @@
...
+  Iff_       s2   a b -> Iff_      <$>          o s2           <*> c a <*> c b
+  Implies_   s2   a b -> Implies_  <$>          o s2           <*> c a <*> c b
+  In_     s1 s2   a b -> In_       <$> i s1 <*> o s2           <*> c a <*> c b
+  Next_      s2   a   -> Next_     <$>          o s2           <*> c a
+  Not_       s2   a   -> Not_      <$>          o s2           <*> c a
+  Or_        s2   a b -> Or_       <$>          o s2           <*> c a <*> c b
+  Rewrites_  s2   a b -> Rewrites_ <$>          o s2           <*> c a <*> c b
+  Top_       s2       -> Top_      <$>          o s2
+  Var_          v     -> Var_      <$>                   var v
+  other -> error $ "Unknown constructor " ++ show other
```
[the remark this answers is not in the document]
u--: Pattern Synonyms + GADTS = GHC definitely can't tell if it's complete or not, so without the "other" it doesn't compile.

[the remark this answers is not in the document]
virgil-serbanuta: Oh, wow, I was waiting to use these synonyms in my condition-evaluation pull request, but now I'm not so sure anymore.

One of the reasons we chose Haskell is because all of the built-in safety, but it seems to me that Haskell programmers don't usually want it that much.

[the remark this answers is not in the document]
ttuegel: `PatternSynonyms` are a big problem for completeness checking. For our case, we really only want to be using the synonyms for construction, not matching. The benefit of using `Fix` is that we should be able to write algebras that only need to match on the underlying functior!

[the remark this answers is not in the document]
u--: Actually, in this case the COMPLETE pragma works. So writing a complete list of cases with 

Fix ( ApplicationPattern (Application ...
Fix (AndPattern (And ...

is equivalent to

App_ ...
And_ ...

and the compiler will know it's complete.

u--:
@ehildenb what do you mean w/ squashing? Like branch off the original commit that introduced the feature?

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/SmartConstructors.hs
u--: Added

src/main/haskell/language-kore/test/parser/Data/Kore/ASTUtils/ASTUtilsTest.hs
```diff
@@ -0,0 +1,171 @@
...
+import           Data.Kore.AST.PureML
+import           Data.Kore.ASTUtils.SmartConstructors
+import           Data.Kore.ASTUtils.Substitution
+
+astUtilsTests :: TestTree
+astUtilsTests = testGroup "astUtils" [ substitutions, sortAgreement ]
+
+substitutions :: TestTree
+substitutions = testGroup "substitutions" 
+  [ testCase "subTrivial" $ assertEqual "" subTrivial subTrivialSolution
```
virgil-serbanuta: This project uses 4 spaces for indentation almost everywhere, "where" and "=" (only sometimes and only when used in function definitions) are the only exceptions I can think of right now. The main reasons for using 4 spaces are uniformity and the fact that "where" and "=" use half of a normal indentation and it looks bad when they are indented by only one space. Note that function signatures are also indented by 4 spaces instead of 2.

BTW, I think that there are places above where you're using a one-space indentation.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,129 @@
+{-|
+Module      : Data.Kore.ASTUtils.Substitution
+Description : Substitute phi_1 for phi_2, avoiding capture
```
virgil-serbanuta: Could you add a short comment or example saying why this would be useful (as opposed to substituting variables)?

src/main/haskell/language-kore/src/Data/Kore/AST/Common.hs
```diff
@@ -101,14 +109,16 @@ data Id level = Id
     { getId      :: !String
     , idLocation :: !AstLocation
     }
-    deriving Show
+    deriving (Show, Generic)
```
virgil-serbanuta: Just curious, not asking for a fix: Where is this generic thing used (for any of the instances defined here, not necessarily this one)?

src/main/haskell/language-kore/test/parser/Data/Kore/ASTUtils/ASTUtilsTest.hs
```diff
@@ -0,0 +1,217 @@
+{-# LANGUAGE MultiParamTypeClasses  #-}
```
virgil-serbanuta: I didn't get an answer to one of my previous questions, so I'll try asking it this way: What's the test coverage for the new code?

src/main/haskell/language-kore/src/Data/Kore/AST/Common.hs
u--: Generic is needed to derive classes like Hashable.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,129 @@
...
+import           Data.List
+
+import           Data.Kore.AST.Common
+import           Data.Kore.AST.MetaOrObject
+import           Data.Kore.AST.PureML
+
+import           Data.Kore.ASTUtils.SmartConstructors
+
+
+subst
```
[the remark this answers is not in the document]
virgil-serbanuta: Old and new seems fine. Substituted and subtituting is also fine. Wrapping them in data types so that people calling your method do not mix the arguments is even better, but I can't usually convince others that it's worth doing. Anyway, the issue was that you had three patterns as arguments and it was not obvious which is which.

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,129 @@
...
+alphaRename = (\case 
+ Forall_ s1 v p -> go Forall_ s1 v p 
+ Exists_ s1 v p -> go Exists_ s1 v p
+ -- x -> x
+ _ -> error "Input must be an Exist or Forall at the top level"
+ ) where 
+    go binder s1 v p = 
+        binder s1 (replacementVar v p)
+        (subst (Var_ v) (Var_ $ replacementVar v p) p)
+    replacementVar v p = head $ alternatives v \\ (S.toList $ freeVars p)
```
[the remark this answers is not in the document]
virgil-serbanuta: Isn't it O(m*n) right now, where m is the first index that was not already used, and n is the free var count?

Even if my estimate is correct, I guess it may be reasonable if m is likely to be small (which, by default, it should be). I can't tell right now if it would increase when doing many similar substitutions as we execute code and produce proofs.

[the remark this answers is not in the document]
u--: Whoops, I thought \\ assumes sorted. Yeah, it is slower in this case.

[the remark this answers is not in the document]
u--: But I just noticed: `freeVars p` is being called over and over in the "filter" case. That's another accident that can't happen when you use operations over whole datastructures like union/intersect/difference.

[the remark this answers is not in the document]
virgil-serbanuta: Thanks. As I said, I was just curious. We're not supposed to optimize things right now, and we should prefer code that is safer/easier to read/whatever.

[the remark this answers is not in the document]
u--: Yeah I'm just explaining why I think it's a good default. For now I filtered with S.member, in the future maybe substitution could be outsourced to kmett's bound or something.

src/main/haskell/language-kore/test/parser/Data/Kore/ASTUtils/ASTUtilsTest.hs
u--: Do you mean `hpc report` ? About 70-80%

src/main/haskell/language-kore/src/Data/Kore/ASTUtils/Substitution.hs
```diff
@@ -0,0 +1,127 @@
...
+subst old new = \case 
+    Forall_ s1 v p -> handleBinder old new Forall_ s1 v p 
+    Exists_ s1 v p -> handleBinder old new Exists_ s1 v p 
+    pat 
+     | pat == old -> new
+     | otherwise  -> Fix $ fmap (subst old new) $ unFix pat 
+
+-- subst2 a b = cata (\pat -> if Fix pat == a then b else Fix pat)
+
+-- handleBinder
```
virgil-serbanuta: You can leave it if it makes sense, but I'm not a fan of code that's commented out. If you allow time to pass, the code will be more and more out of date, but people will avoid deleting it since they assume it has a purpose.

src/main/haskell/language-kore/test/parser/Data/Kore/ASTUtils/ASTUtilsTest.hs
virgil-serbanuta: We usually use `stack test --coverage` and look at the html report, I guess that may use `hpc` behind the scenes.

That number, by itself, does not mean much. We are aiming for 100% coverage. Until now we allowed the coverage to be lower for things like automatic instance derivations and struct getters, but we usually want close to 100% for everything else. So, if the missing 20-30% are things like that, it should be fine.

PR #145, opened 2018-07-13
fix strong-next notation. \circ to \bullet

hreada:
Fix in Section 4.4 Rewiting and Reachability the notation for "strong next" symbol. Change it from \circ to \medbullet. In order to do that, I used the package `txfonts` which has an affect in the appearance of the fonts of the entire document.


traiansf:
Jenkins: test this please

PR #146, opened 2018-07-16
Generalize smart constructors to use PureMLPattern instead of CommonP…

virgil-serbanuta:
…urePattern.


u--:
My bad, am using stylish-haskell now.

PR #147, opened 2018-07-16
User meta sort

njohnwalker:
Tentative change to allow user-defined meta sorts.

Running the parser on the file `user-meta-nat.kore` generates a Definition that passes the well-formedness checking.

This change breaks the tests that have to do with parsing meta-sorts:
- parsing meta sorts is total now, (no failure on typos like `#Chart`)
- it looks like some golden files may need to be regenerated, but I want to hold off for now.

~For a less extensive change, possible attribute implementation forthcoming~

I had to mess around with the scala tests in order to pass the Jenkins check, but this hides the fact that now the two test suites both pass but with different targets since the scala and haskell implementations have diverged with this change.


src/test/resources/expected/test-pattern-16.kore.golden
```diff
@@ -0,0 +1,24 @@
+Definition
+    { definitionAttributes = Attributes []
+    , definitionModules =
+        [ Module
+            { moduleName = ModuleName "TEST-EXCEPTION-9"
```
virgil-serbanuta: You should change the module name (here and in the other similar tests).

src/test/scala/org/kframework/TextToKoreTest.scala
```diff
@@ -231,13 +231,13 @@ class TextToKoreTest {
     parseFromFileExpectParseException("test-exception-8.kore")
   }
   @Test def test_exception_9(): Unit = {
-    parseFromFileExpectParseException("test-exception-9.kore")
+    parseFromFileExpectParseException("test-pattern-16.kore")
```
virgil-serbanuta: The Scala parser is supposed to be in sync with the Kore one, so these files should not generate exceptions.

src/main/haskell/language-kore/src/Data/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -162,6 +163,11 @@ verifyMetaSentence
     (SentenceAxiomSentence axiomSentence)
   =
     verifyAxiomSentence axiomSentence indexedModule attributesVerification
+verifyMetaSentence
+    _indexedModule
+    _attributesVerification
+    (SentenceSortSentence _sortSentence)
+  = error "Meta sort declaration well-formedness not yet defined"
```
virgil-serbanuta: When do you intend to fix this?

src/main/haskell/language-kore/src/Data/Kore/Parser/ParserImpl.hs
```diff
@@ -899,6 +898,15 @@ sentenceConstructorRemainderParser sentenceType
                     (symbolParser Object)
                     (Rotate31 <....> SentenceSymbol)
 
+sentenceSortRemainderParser :: Parser KoreSentence
```
virgil-serbanuta: The Haskell code should stay in sync with the doc, so if the meta sorts are not already in the Sematics of K doc, could you add a Trello task for Xiaohong to add them?

src/main/haskell/language-kore/src/Data/Kore/ASTVerifier/SentenceVerifier.hs
njohnwalker: Should been in the next commit, was waiting for feedback, but I just did it

src/test/scala/org/kframework/TextToKoreTest.scala
virgil-serbanuta: I see that you mentioned that in the PR description, but I still think that they should stay in sync.

njohnwalker: What do you think is the best course of action?

virgil-serbanuta: I see three options, I think 1 is the most reasonable:

1. Fix the Scala parser to accept these (i.e. to parse meta-sorts) and call parseFromFile here. It should be fairly simple, but I didn't look at that particular part of the code. That's the approach we used until now.

2. If 1 seems too hard (e.g. you don't know Scala and you think you can't figure it out enough to update the parser), you could try to convince Xiaohong to update it, since he wrote the initial code. Or someone else from the team.

3. Convince the other people that we should give up on the Scala parser for now (e.g. by deleting the code) and that we should rewrite it latter when we actually need it. You might some weak support from me, but you don't have that many chances since people are convinced that we'll need it, and updating it every now and then is less effort than reimplementing/updating it later.

virgil-serbanuta: BTW, that someone else I mentioned before can be me, of course.

njohnwalker: It looks like the scala parser was out of sync already, https://github.com/kframework/kore/commit/7ba1247896aeb5409019944c14f1c4f9fe915c33 introduces new syntax and tests but they weren't added to scala.  I can fix this too but I'm afraid that this PR is getting stale...

virgil-serbanuta: Ok, how about this: I'll approve this pull request in order to minimize merge pain, and you'll update the Scala parser to handle meta-sorts in a follow-up PR. Optionally you can either fix the 'next' and 'rewrites' issues, and change the way tests are being run, or you can make a Trello task for that.

PR #148, opened 2018-07-18
Produce human-readable output from the unparser

ttuegel:
The unparser produces human-readable output (as far as possible) by reducing the
use of extra whitespace.

### Questions

The `Unparse` class is reduced to an alias for the `Pretty` class provided by `prettyprinter`. Is it worth keeping around?

The golden tests for the implicit Kore definitions are reformatted using `kore-format`. Do we really want golden tests here, or would it be more useful to test the round-trip through the parser and unparser?


src/main/haskell/language-kore/app/format/Main.hs
```diff
@@ -0,0 +1,56 @@
...
+    <> progDesc "Parse a Kore definition and render it in standard format"
+    <> header "kore-format - parse and render Kore definitions"
+
+main :: IO ()
+main =
+  do
+    options <- mainGlobal commandLine infoMod
+    case localOptions options of
+        Nothing -> return () -- global options parsed, but local failed; exit gracefully
+        Just KoreFormatOptions {..} ->
```
virgil-serbanuta: A few days ago I changed the style guide to discourage this since one can't resolve the `fileName` and `width` symbols locally, i.e. without looking at the struct definition. This is also used in other places, but some of those look less bad since the field name is prefixed with the struct name.

src/main/haskell/language-kore/app/format/Main.hs
```diff
@@ -0,0 +1,56 @@
...
+          do
+            defn <- readFile fileName >>= either error return . fromKore fileName
```
virgil-serbanuta: Do you think it makes sense to extract this to a `fromKoreOrDie` function?

src/main/haskell/language-kore/app/format/Main.hs
```diff
@@ -0,0 +1,56 @@
...
+            let
+              layoutOptions
+                | width > 0 =
+                    defaultLayoutOptions
+                    { layoutPageWidth = AvailablePerLine width 1.0 }
```
virgil-serbanuta: Indent one more level?

src/main/haskell/language-kore/test/parser/Data/Kore/Unparser/UnparseTest.hs
```diff
@@ -136,18 +128,15 @@ unparseUnitTests =
                     ]
                 }::KoreDefinition
             )
-            (  "[]\n\n"
-            ++ "    module i\n    endmodule\n    []\n"
-            ++ "    module k\n    endmodule\n    []\n"
-            )
+            "[]\nmodule i\nendmodule\n[]\nmodule k\nendmodule\n[]"
```
virgil-serbanuta: Could you split this after "\n", as it was before? That way it's much easier to understand what should happen in this test.

src/main/kore/kore.kore
```diff
@@ -1,3518 +1,166 @@
...
-                    ),
-                    \equals{
-                        #Sort{},
-                        #sp
-                    }(
-                        #getSort{}(
-                            #psi:#Pattern{}
-                        ),
-                        #s:#Sort{}
-                    )
```
virgil-serbanuta: I guess that these long lines are caused by using `layoutPrettyUnbounded` in the definition for `unparse`.

This module had two purposes:
1. To catch regressions related to the implicit Kore definitions.
2. To document what is implicitly defined in Kore.

Having log lines means that it becomes somewhat harder to identify the effect of a regression and the kore.kore file becomes less useful for people interested in implicit definitions.

Could you fix this? Preferably for all regression tests, not just this one.

src/main/haskell/language-kore/app/format/Main.hs
ttuegel: As far as I can see, the style guide does not indicate additional indentation here.

ttuegel:
@virgil-serbanuta I made the changes you suggested, except where I commented otherwise.

src/main/haskell/language-kore/app/format/Main.hs
virgil-serbanuta: You're right. When I wrote "Indent everything with multiples of 4 spaces, unless specified otherwise." I wanted to write "Indent everything that is a subcomponent ..." or something like that, but it turns out that I didn't.

After discussing this with @traiansf , this is reasonable so you can leave it as-is. I dislike it a bit since 4 spaces of indentation and a curly brace is a better visual cue than only a curly brace, but that's not the only criteria that matters and I don't have to like everything.

However, there is another interesting issue related to the rule above. In this PR you're using half-indentation for `do` and `let` variables. This may actually be interesting (more interesting for `do`, less interesting for `let` IMO, but others may think differently). However, in the PR you sent to Traian you're using the standard 4 spaces indentation for `do` and `let` variables, so we wondered why they are inconsistent.

So, if you think we should use this half-indentation in some cases (I think you wanted to also use this for `case` statements at the top level of functions), you can submit this as-is, but we should discuss and maybe include these in the style guide doc.

ttuegel: > However, there is another interesting issue related to the rule above. In this PR you're using half-indentation for do and let variables. This may actually be interesting (more interesting for do, less interesting for let IMO, but others may think differently).

Sorry, I didn't intend to do that, but old habits die hard. In my personal projects, I have (for years) indented all layout keywords two spaces; when you've done something one way for a long time, your eyes just sail past it without noticing. :smiley:  We can discuss that for the style guide, but for now I will fix this to match what the draft style guide says.

> However, in the PR you sent to Traian you're using the standard 4 spaces indentation for do and let variables, so we wondered why they are inconsistent.

I don't think that PR doesn't actually contains any new code, which probably explains the discrepancy.

PR #149, opened 2018-07-18
Automatic test discovery

ttuegel:
Automatic test discovery removes some boilerplate code by using `tasty-discover` to infer the structure of the test tree automatically. Only a few hundred lines of boilerplate are removed, but discovery will prevent adding more in the future.

While I implemented this, I also changed the test module names to conform to my suggestion for the style guide.


ttuegel:
> However it's pretty hard to parse; I assume you did nothing but renamed files and function names for the most part.

I did these things:
1. Renamed existing files, per discussion on the style guide.
2. Where a module had only one top-level test group, the call to `testGroup` was removed and it became a `[TestTree]`.
3. The test cases are renamed to have the prefix `test_` which is the convention `tasty-discover` uses to select tests.
4. Modules that only import tests to construct a `testGroup` are removed because they are unnecessary.

> If the new files you created, you might consider keeping the line length below 80, if possible.

I didn't actually create any new files, but Git becomes confused if you change the indentation of many lines (such as when deleting a `testGroup` call).

ttuegel:
> The 80 chars comment was about

Oh, sorry, that was from a previous pull request. I didn't realize there was a violation, but I'll fix it up.

traiansf:
I'm also sorry. I only now realized it was form a previous pull request.  Not being able to follow all changes, I took a look at the repo as-is, and this was one of the few which seemed to be connected with the PR

PR #150, opened 2018-07-19
Proof System + New Unification

u--:


virgil-serbanuta:
I'm not sure if this is ready for review or not (nobody's assigned, there's no 'DO NOT REVIEW YET' warning), but I tried to take a look and it shows a lot of diffs related to Thomas' test refactoring, so you may want to remove them.

u--:
Sorry about that, I merged instead of rebased early on when I needed to try pretty-printing proofs... Is there a way to remove them?

ttuegel:
If you can revert the merge commit from the old `prettyprinter` branch, you should be able to rebase your changes onto the new master or merge the new master into your branch.

virgil-serbanuta:
I'm not that good with git. I think that Thomas merged those PRs, so rebasing everything on master might work. Make sure you can undo everything (e.g. by making a backup branch) before trying that.

Merging the master branch might also work, but then you'd leave all that messy history inside our tree. I'm not sure it matters, though, but others may care about that.

Another option would be to do something like rebase -i HEAD~<number> and marking those commits as 'd'. I never tried that before, but it might work.

u--:
This is really dumb, but I may have confused things to the point that it's easier just to copy the files I wrote back in. I didn't modify any existing files besides adding some Hashable instances here and there.

PR #153, opened 2018-07-20
Full KORE syntax support in prover REPL

dfilaretti:
###### Reviewer checklist

- [x] Style conformance

---


ttuegel:
@dfilaretti This is still marked "DO NOT REVIEW", but the other day it sounded like it was ready. Are we still waiting to review it?

src/main/haskell/kore/src/Kore/Parser/ParserImpl.hs
```diff
@@ -498,6 +498,14 @@ variableOrTermPatternParser childParser x = do
         then VariablePattern <$> variableRemainderParser x identifier
         else symbolOrAliasPatternRemainderParser childParser x identifier
 
+symbolOrAliasParser
+    :: MetaOrObject level
+    => level  -- ^ Distinguishes between the meta and non-meta elements.
+    -> Parser (SymbolOrAlias level)
+symbolOrAliasParser x = do
+    identifier <- idParser x
+    (SymbolOrAlias identifier <$> inCurlyBracesListParser (sortParser x))
```
virgil-serbanuta: You could reduce code duplication by extracting this line and naming it "symbolOrAliasReminderParser", then you could also call it in the symbolOrAliasPatternReminderParser function above.

njohnwalker: I'm reluctant to make this change as I'm not sure how long this `symbolOrAliasParser` will be around and the new definition would only be used twice.  It's easier to read each definition when there is no "jump" to another definition.

PR #156, opened 2018-07-24
Merge language-kore with matching-logic

ttuegel:
This pull request merges our two packages because the interface between them is so poorly defined.

At the same time, I made a few changes related to the module hierarchy. Modules under `Kore.MatchingLogic` now fall under `Logic.Matching` since they generally don't depend on Kore and the `Logic` top-level category is common to other libraries. That frees up the `Kore` category for the modules previously under `Data.Kore`. The `Data` top-level category is disfavored outside Haskell boot libraries because it's just noise: most libraries define data types, so having that `Data` prefix on the module names doesn't give you any additional information. A handful of modules were moved outside either category because they are general definitions used in the implementation that have nothing to do with matching logic or Kore.

I also made a set of unrelated changes left over from my `stylish-haskell` pull request, on the grounds that I'm already making tree-wide changes. This PR would add a `stylish-haskell` configuration file with a conservative configuration for automatically formatting import statements.

I have removed redundancy in a few places. The `FromString` class is already in `base`, but it is named `IsString`. Before merging, the packages used different implementations of `Fix` (from `data-fix` and `recursion-schemes`). Now only `recursion-schemes` is used; that will be important later for syntax tree annotations.

###### Reviewer checklist

- [x] Style conformance

---


ttuegel:
@bmmoore This is fully ready for review. I have updated to latest master and fixed up the Jenkins configuration.

ttuegel:
@bmmoore This is up-to-date with master again.

src/main/haskell/kore/test/prover/Test/Kore/MatchingLogic/ProofSystem/OnePlusOne/ProofSystem.hs
```diff
@@ -99,7 +98,8 @@ transformRule _sort label var term hypothesis rule = case rule of
     OrIntroR t1 t2 -> OrIntroR <$> term t1 <*> term t2
 
 -- | Lens focusing on the terms within a Rule.
-ruleTerms :: (Applicative f)
+ruleTerms :: forall (f :: * -> *) (sort :: *) termA termB hyp label var.
```
bmmoore: Why was an explicit forall added here?

ttuegel: I needed to add an explicit kind signature because of enabling `PolyKinds` by default.

PR #159, opened 2018-07-24
Scala parser update

njohnwalker:
Updated the scala parser to accept meta sort declarations and the `\next` pattern production

###### Reviewer checklist

- [ ] Style conformance

---

### DIFF
--- a/src/main/scala/org/kframework/kore/Interface.scala
+++ b/src/main/scala/org/kframework/kore/Interface.scala
@@ -35,7 +35,7 @@ object Import {
 }
 
 trait SortDeclaration extends Declaration {
-  def params: Seq[SortVariable]
+  def params: Seq[Sort]
 
   def sort: Sort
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/kframework/kore/Interface.scala
```diff
@@ -35,15 +35,15 @@ object Import {
 }
 
 trait SortDeclaration extends Declaration {
-  def params: Seq[SortVariable]
+  def params: Seq[Sort]
```
virgil-serbanuta: Why this change? I think that the semantics-of-k doc specifies a variable here.

njohnwalker:
As discussed, I am closing this pull request because we wish to freeze development of the Scala parser for the time being, with it's implementation in sync with a specific revision of the Semantics of K document. (This update would be a branch from the semantics of k document)

PR #161, opened 2018-07-25
Attributes

traiansf:
This PR is a reimplementation of Attributes in the following sense:
- implicit validation of attributes is reduced to basic checks such as attribute is a (not necessarily well-formed) term pattern consisting of the application of an object symbol identifier to a (possibly empty) list of attribute arguments.
- validation and the ImplicitModule structure are parameterized to allow users of it to implement their own validation and post-processing of attributes, by providing a type for Processed attributes implementing the `ParsedAttributes` class
- two such instances are provided: `ImplicitAttributes` an empty structure, doing no postprocessing, and `StepperAttributes` postprocessing attributes useful for unification/evaluation/stepping.

In addition to that, an inconsistency with the definition of `DomainValues` which made them useless in the `Pure` world was discovered and fixed.

###### Reviewer checklist

- [ ] Style conformance

---


src/test/resources/expected/imp.kore.golden
```diff
@@ -3382,27 +3338,11 @@ Definition
...
-                                        }))))
-                                    , Fix (UnifiedPattern (UnifiedObject (Rotate31 (ApplicationPattern Application
-                                        { applicationSymbolOrAlias =
-                                            SymbolOrAlias
-                                                { symbolOrAliasConstructor = (Id "firstArgument" AstLocationNone) :: Id Object
-                                                , symbolOrAliasParams = []
-                                                }
-                                        , applicationChildren = []
-                                        }))))
+                                    [ Fix (UnifiedPattern (UnifiedMeta (Rotate31 (StringLiteralPattern (StringLiteral "divAExp,1")))))
```
virgil-serbanuta: I don't think that this is how we want this to work.

src/test/resources/imp2.kore
```diff
@@ -554,11 +552,11 @@ module IMP
           HOLE:AExp{},
           divAExp{}(HOLE:AExp{}, Y:AExp{})),
         X:AExp{}))
-  [strict{}(\dv{Strict{}}("divAExp"), firstArgument{}())]
+  [ strictAxiom{}("divAExp,1") ]
```
virgil-serbanuta: I think that you should use divaexp as as symbol here and, even if you're not doing that, 1 should be in a different string.

src/test/resources/expected/imp.kore.golden
traiansf: ?

src/test/resources/imp2.kore
traiansf: ok...

src/main/haskell/language-kore/src/Data/Kore/AST/MLPatterns.hs
```diff
@@ -229,7 +223,7 @@ applyPatternLeveledFunction function (BottomPattern a) =
 applyPatternLeveledFunction function (CeilPattern a) =
     patternLeveledFunctionML function a
 applyPatternLeveledFunction function (DomainValuePattern a) =
-    patternLeveledFunctionML function a
+    domainValueLeveledFunction function a
```
virgil-serbanuta: This is not tested.

src/main/haskell/language-kore/src/Data/Kore/ASTVerifier/PatternVerifier.hs
```diff
@@ -413,6 +419,23 @@ verifyVariableUsage variable _ verifyHelpers _ _ = do
...
 
+-- TODO: properly verify child pattern if it's not a stringLiteral
+verifyDomainValue
+    :: (MetaOrObject level)
+    => DomainValue Object
+    -> VerifyHelpers level
+    -> Set.Set UnifiedSortVariable
+    -> Either (Error VerifyError) (Sort Object)
+verifyDomainValue dv verifyHelpers declaredSortVariables =
+    case isMetaOrObject verifyHelpers of
```
virgil-serbanuta: This function has some holes in testing, could you test what is worth testing (maybe everything)?

src/main/haskell/language-kore/src/Data/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -307,40 +348,43 @@ internalIndexModuleIfNeeded
...
         )
   where
+    moduleAtts = moduleAttributes koreModule
     indexedModulesAndStartingIndexedModule =
         ( indexedModules
         , (indexedModuleWithDefaultImports
                 (moduleName koreModule) implicitModule)
             { indexedModuleAttributes =
-                moduleAttributes koreModule
+                (parseAttributes moduleAtts, moduleAtts)
```
virgil-serbanuta: Not tested

src/main/haskell/language-kore/src/Data/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -355,7 +399,9 @@ indexModuleMetaSentence
             { indexedModuleMetaAliasSentences =
                 Map.insert
                     (aliasConstructor (sentenceAliasAlias sentence))
-                    sentence
+                    ( parseAttributes (sentenceAliasAttributes sentence)
```
virgil-serbanuta: Test?

src/main/haskell/language-kore/src/Data/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -386,8 +434,10 @@ indexModuleMetaSentence
   =
     return
         ( indexedModules
-        , indexedModule { indexedModuleAxioms = sentence : sentences }
+        , indexedModule { indexedModuleAxioms = (atts, sentence) : sentences }
         )
+  where
+    atts = parseAttributes (sentenceAxiomAttributes sentence)
```
virgil-serbanuta: Test?

src/main/haskell/language-kore/src/Data/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -412,7 +462,8 @@ indexModuleMetaSentence
         ( newIndexedModules
         , indexedModule
             { indexedModuleImports =
-                (attributes, importedModule) : indexedImports
+                (parseAttributes attributes, attributes, importedModule)
```
virgil-serbanuta: Test? And you have the same issue in other similar places below.

src/main/haskell/language-kore/src/Data/Kore/Step/StepperAttributes.hs
```diff
@@ -0,0 +1,70 @@
+{-|
```
virgil-serbanuta: This file seems to be missing some things from testing, could you fix?

traiansf: Thanks for pointing it out. I forgot to add the test file to Git :-)

src/main/haskell/language-kore/src/Data/Kore/IndexedModule/IndexedModule.hs
traiansf: Since there are no tests at all for IndexedModule, I will add a separate task for that.

traiansf:
@ttuegel could you please help me fix `Common.hs` ? I can't manage to do it by myself.  I know there might be more errors later (introduced by the merge) but I got stuck here and I can't seem to be able to make any progress.

On another note, as Ralph was remarking earlier, if the level of the code gets too high, it will limit the people which can contribute to it.  

I won't claim I'm a proficient Haskell programmer, but if I, knowing the codebase, get so frustrated about a piece of code which used to work fine before, I can only imaging how "easy" it would be for people outside the company to contribute to this project.

ttuegel:
> On another note, as Ralph was remarking earlier, if the level of the code gets too high, it will limit the people which can contribute to it.
> 
> I won't claim I'm a proficient Haskell programmer, but if I, knowing the codebase, get so frustrated about a piece of code which used to work fine before, I can only imagine how "easy" it would be for people outside the company to contribute to this project.

I agree with you: Using a fixpoint functor to define our abstract syntax is limiting:
1. it limits who can contribute to the code,
1. it limits our ability to leverage external libraries, and
1. it limits our ability to use compiler features (mainly deriving).

It isn't clear to me what we gain by sacrificing these things to use a fixpoint type instead of an ordinary recursive data type, but that decision was made before I joined the project.

Let me recap our earlier conversation, to remind everyone else how we arrived at the current situation.
When I merged `language-kore` and `matching-logic` into `kore`, I observed that each package used a different implementation of `Fix`, from the libraries [`data-fix`](https://hackage.haskell.org/package/data-fix) and [`recursion-schemes`](https://hackage.haskell.org/package/recursion-schemes) respectively. The implementations in each library are isomorphic but not interoperable. One of the next changes I need to make is to add annotations to the syntax tree; the usual way to do that is with the [`Cofree`](https://hackage.haskell.org/package/free-5.1/docs/Control-Comonad-Cofree.html#t:Cofree) comonad. `Cofree` is another fixpoint functor supported by the `recursion-schemes` library. (`data-fix` only supports `Fix`, as the name implies.) I decided that it didn't make sense for the combined library to depend on two different implementations of the same type, especially because the next thing I needed to do would force me to get rid of the `data-fix` dependency anyway. Therefore, I changed everything that depended on `data-fix` to depend on `recursion-schemes` instead. `recursion-schemes` places slightly different requirements on base functors; for example, these instances:
```.hs
-- data-fix: This is not a proper instance.
-- FlexibleContexts can cause problems with type inference.
instance Eq (f (Fix f)) => Eq (Fix f) where ...

-- recursion-schemes: This is a proper instance,
-- but providing an Eq1 instance is more complicated.
instance Eq1 f => Eq (Fix f) where ...
```
Most of the `Eq1`, `Ord1`, etc. instances can be derived, but a few cannot because of the way we use higher-kinded types to place restrictions on the structure of the syntax tree. We now face the difficulty of manually writing these instances; the root cause is using a fixpoint type instead of ordinary recursive data types, for which the required instances could be derived. (Using fixpoint types also cuts us off from standard libraries like [`uniplate`](https://hackage.haskell.org/package/uniplate) for deriving traversals and [`unbound`](https://hackage.haskell.org/package/unbound) or [`bound`](https://hackage.haskell.org/package/bound) for deriving operations on free variables and binders).

I guess the summary of this comment is: I don't know what to do about the increasing-complexity problem because we seem to have gotten here by a sequence of steps that are individually reasonable and necessary.

src/main/haskell/kore/src/Kore/AST/Common.hs
```diff
@@ -964,7 +966,7 @@ data Pattern level variable child where
     CeilPattern
         :: !(Ceil level child) -> Pattern level variable child
     DomainValuePattern
-        :: !(DomainValue Object child) -> Pattern Object variable child
+        :: !(DomainValue Object (Fix (Pattern Meta Variable))) -> Pattern Object variable child
```
ttuegel: With this constructor, I don't think `Pattern` is going to do what we want. Traversals (`cata` or even just `project`) are not going to be able to descend into this child tree. The caller will have to write a second traversal inside the first just to deal with this constructor, defeating the purpose of using `Pattern` as a base functor.

traiansf: @ttuegel I know this patttern will not be able to be descended automatically, but I think it is acceptable to pay this price to be able to have a proper type for the child of a domain value.

The problem I was trying to solve here is that having it be a child like for any other pattern would make it impossible to represent object-only patterns containing domain values, because the child of a domainvalue must be a meta-pattern of sort  `#String`.

If you can think of an alternative solution, I would be glad to hear it.

virgil-serbanuta: Also, I think that we do not usually want to descend into a domain value's child, except in a few narrow cases (e.g. serialization), or after it was lifted to meta, when it's not a DomainValuePattern anymore.

virgil-serbanuta:
"I guess the summary of this comment is: I don't know what to do about the increasing-complexity problem because we seem to have gotten here by a sequence of steps that are individually reasonable and necessary." 

One way to solve the tragedy of commons (which is what this seems to be) is to restrict somehow the resource's use (in our case, complexity). Restricting and enforcing the restricts can be done by a person that watches over this, or by our consensus, or whatever.

Consider this two options, none of which is optimal, and maybe none of us would actually suggest them, but which may guide the discussion to a more productive direction. We could decide between
* not increasing the complexity of the constructs we are using period
* using whatever language construct/extension we wish without any constraint.

I think that the second one would give us short-term gains, but has a high chance of killing the project in the long term or, in the best case, making development really slow because some of the existing coders would struggle with other's code (and maybe also with their own code and with the unexpected interactions between the various extensions/coding techniques/whatever), they might leave out of frustration, and it would become really hard to get and/or train new coders. It would provide some of us with really nice job security, though.

I think that the first option would give us a steady development pace that would locally be slower than what's reasonable, and sometimes it would be globally slower. (Fix may be an example of global slowness, recursion schemes may be another one, lenses may be another one, I don't want to give an opinion right now, since it would distract from the main discussion).

To me it seems obvious that, between the two artificial options, the second is unacceptable, and the first is suboptimal, but can be followed. Unless we find a better option, we might just do that.

So, going back to the prisoner's dilemma, weighing the issues involved in increasing the complexity could be a manager's job (Ralph's? Brandon's? I don't actually know who is my manager) or, perhaps, a tech lead's job (Traian's). Do we want that? How would we want to decide when to use a new extension, or a new library that adds new concepts (as opposed to just providing some functionality)?

ttuegel:
> One way to solve the tragedy of commons (which is what this seems to be) is to restrict somehow the resource's use (in our case, complexity). Restricting and enforcing the restricts can be done by a person that watches over this, or by our consensus, or whatever.

I think that everything you said is part of an important discussion, but alongside that discussion I need to make sure one thing is clear: The problem here is that trying to restrict the type of child terms is not compatible with defining the AST as the fixpoint of a functor. To be a fixpoint, each child node in the AST must have the same type as its parent. The type system is telling us that Kore patterns (with type-level `Object` and `Meta` annotations) aren't the fixpoint of any functor. My adding `TemplateHaskell` to `Common` makes it very complicated to work around what the compiler is saying, but it didn't create the underlying complexity. We can remove this use of `TemplateHaskell` and we can remove this use of `recursion-schemes`, but we are still going to have to live with the underlying complexity of having an abstraction that doesn't match what is being modeled.

src/main/haskell/kore/src/Kore/AST/Common.hs
```diff
@@ -622,6 +630,14 @@ instance (Eq (Sort level), Eq (v level)) => Eq1 (Exists level v) where
         && (existsVariable a == existsVariable b)
         && liftedEq (existsChild a) (existsChild b)
 
+instance (Ord (Sort level), Ord (v level)) => Ord1 (Exists level v) where
+    liftCompare liftedCompare a b =
+        case compare (existsSort a) (existsSort b) of
```
ttuegel: For this `Ord1` instance and the one below, you can use the `Monoid` instance for `Ordering`, which takes the first non-`EQ` result:
```.hs
   compare (existsSort a) (existsSort b)
<> compare (existsVariable a) (existsVariable b)
<> liftedCompare (existsChild a) (existsChild b)
```

PR #162, opened 2018-07-26
Add a design doc for attributes as patterns.

virgil-serbanuta:
###### Reviewer checklist

- [ ] Style conformance

---


design-decisions/2018-07-25 Attributes as patterns.md
```diff
@@ -0,0 +1,120 @@
...
+
+Attributes are identified by a name and might have (or not) parameters.
+
+When present, attribute parameters are usually very simple, e.g. the `strict` attribute
+can have a list of argument indexes in which it is strict.
+
+Sometimes we may
+want to refer to user-defined symbols, e.g. we may use the `unit` attribute
+to specify the neutral element for an operation, and we may want to use this
+attribute both for the operation symbol declaration and for the axioms that
```
grosu: In K and ML, symbols are not necessarily "operation" symbols.  In FOL, "operation symbols" are interpreted as total functions in models, while "predicate symbols" as relations.  So stick to just "symbols".

design-decisions/2018-07-25 Attributes as patterns.md
```diff
@@ -0,0 +1,120 @@
...
+attribute with an operation symbol as argument to, say,
+link an associativity axiom with an operation.
+
+However, we can't tell what's the general structure of an attribute.
+
+Attribute syntax
+----------------
+
+We could encode each attribute as a string (name/key) or as a pair of
+strings (key-value).
```
grosu: I think that would also be more inefficient than needed.  Looking up a string key in a map is slower than looking up a constant key from a finite and likely very small set of such constants, which is what you'd get if the key is just an ordinary symbol.

design-decisions/2018-07-25 Attributes as patterns.md
```diff
@@ -0,0 +1,120 @@
...
+However, that does not seem a natural way of encoding
+the structure that they have, and each backend would need its own
+string-parsing library.
+
+*Decision*: We will encode each attribute as a Kore pattern that uses only
+application and string literals, with the top symbol being an application
```
grosu: Why not just say "term pattern"?  That's what term patterns are.  @hreada can you please put  a reference to term patterns here?  Also, I don't see why we have to break the general rule and write `1` instead of `\dv("1","Int")`.  It just feels to me that we are breaking a uniform rule to represent things as patterns, that we know works, for no clear benefits.  "More human readable" is probably your argument.  But why should we bother with that in KORE?  We are going to have layers on top of KORE that add syntactic sugar, as part of implementing K in K.  Frontends can choose to generate those if they want to, and then use a general procedure to "desugar" them, instead of just picking some adhoc encoding now and them being stuck with it.

design-decisions/2018-07-25 Attributes as patterns.md
```diff
@@ -0,0 +1,120 @@
...
+of an object identifier giving the name of the attribute to the list of its
+parameters.
+Numbers will be encoded as string literals. Therefore the following may be valid
+attributes:
+```
+constructor{}()
+strict{}("1", "2")
+unit{}(epsilon{}())
+operation{}(plus{}())
+```
```
grosu: There were more examples discussed in the Slack channel.  For example, how do you say that a binary operation has two units?

virgil-serbanuta: `unit{}(epsilon{}(), zero{}())`, I added an example.

design-decisions/2018-07-25 Attributes as patterns.md
hreada: Term patterns in matching logic are patterns built with only functional symbols. See Definition 5.21 at page 28 of the paper _Matching Logic_ at this link http://fsl.cs.illinois.edu/FSL/papers/2017/rosu-2017-lmcs/rosu-2017-lmcs-public.pdf.
If in @traiansf's definition, the "top symbol" is always a functional symbol, then what the definition defines a subset of term patterns. (Notice that string literals, i.e. cons lists of characters, are term patterns by definition).

design-decisions/2018-07-25 Attributes as patterns.md
virgil-serbanuta: Solved.

design-decisions/2018-07-25 Attributes as patterns.md
virgil-serbanuta: I don't think that this is a strong argument, but I added it.

design-decisions/2018-07-25 Attributes as patterns.md
virgil-serbanuta: According to @hreada, this is a bit more generic. I'm not sure we need that much precision, so I guess I could just use 'term'.

design-decisions/2018-07-25 Attributes as patterns.md
```diff
@@ -0,0 +1,146 @@
...
+We will *not* check that this attributes are well-formed in the usual sense,
+e.g. we will not check whether the `plus` symbol in the example above has
+two arguments or that the `strict` symbol arguments are object patterns or have
+the `#Pattern` sort.
+We *will* check that the top-level element is an application pattern since we
+must be able to find its name.
+We *may* check that the pattern uses only application and string literals.
+We *will* require a backend to provide a validation function for each attribute
+it can understand.
+
```
hreada: It looks to me that the only mendatory checking happening is to check that attributes are symbol application patterns. IMO we could even drop that checking, too; but I'm fine with leaving it here.

hreada: Also the last requirement seems a bit redundant, since the backend may provide a idle validation function that checks nothing. Having this requirement doesn't enforce anything to happen.

PR #164, opened 2018-07-27
LTS Haskell 12.2

ttuegel:
The Stack resolver is updated to LTS Haskell 12.2 using GHC 8.4.3. In addition to package updates, the newer GHC version has some fixes for `COMPLETE` pragmas that I would like to use. I believe @bmmoore was having trouble with some derived instances and unreachable code, which is also addressed in this release.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Data/Function/Compose.hs
+++ b/src/main/haskell/kore/src/Data/Function/Compose.hs
@@ -2,7 +2,7 @@
 Module      : Data.Function.Compose
 Description : Short-hand notation for function composition with many arguments
 Copyright   : (c) Runtime Verification, 2018
-License     : UIUC/NCSA
+License     : NCSA
 Maintainer  : traian.serbanuta@runtimeverification.com
 Stability   : experimental
 Portability : portable
--- a/src/main/haskell/kore/src/Kore/Parser/LexemeImpl.hs
+++ b/src/main/haskell/kore/src/Kore/Parser/LexemeImpl.hs
@@ -3,7 +3,7 @@ Module      : Kore.Parser.LexemeImpl
 Description : Lexical unit definitions for Kore and simple ways of composing
               parsers. Meant for internal use only.
 Copyright   : (c) Runtime Verification, 2018
-License     : UIUC/NCSA
+License     : NCSA
 Maintainer  : virgil.serbanuta@runtimeverification.com
 Stability   : experimental
 Portability : POSIX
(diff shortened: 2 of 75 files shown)


ttuegel:
The build is failing while it is preparing to build the documentation, but I'm not sure why.

src/main/haskell/kore/src/Data/Function/Compose.hs
```diff
@@ -2,7 +2,7 @@
 Module      : Data.Function.Compose
 Description : Short-hand notation for function composition with many arguments
 Copyright   : (c) Runtime Verification, 2018
-License     : UIUC/NCSA
+License     : NCSA
```
traiansf: Why the license change?

src/main/haskell/kore/src/Kore/Parser/LexemeImpl.hs
```diff
@@ -121,9 +124,10 @@ skipWhitespace = L.space Parser.space1 (L.skipLineComment "//") blockComment
     blockComment = Parser.string "/*" >> void (manyTill commentBody (Parser.string "*/"))
     commentBody = Parser.anyChar <|> (eof >> fail "Unfinished comment.")
 
-koreKeywordsSet :: Trie.Trie ()
-koreKeywordsSet = Trie.fromList $ map (\s -> (Char8.pack s, ()))
-    ["module", "endmodule", "sort", "symbol", "alias", "axiom"]
+koreKeywordsSet :: HashSet Char8.ByteString
```
traiansf: Why replacing Trie by set?

u--: I think Trie is unnecessarily high-powered for short lists of short strings.

ttuegel: The package that provides `Trie` (`bytestring-trie`) was removed from LTS Haskell because it is not maintained. `HashMap` and `HashSet` both have O(1) lookup; the only advantage of `Trie` is possible sharing between the keys. With so few keys, and such small keys, we don't get any sharing anyway.

src/main/haskell/kore/src/Data/Function/Compose.hs
ttuegel: The license itself does not change, but `NCSA` is the proper [SPDX license identifier](https://spdx.org/licenses/) for the license we're using.

ttuegel:
> Could we perhaps prebuild haddocks for the dependencies on the server, to avoid running out of memory during the build?

I'm not sure that pre-building would help; I think it would still run out of memory. We don't need it to build our own Haddocks, but some links will be missing. Are we building Haddocks on Jenkins as a test of the documentation, or for our own reference?

ttuegel:
The build is timing out on Jenkins. I don't see any reason to pursue this further.

Issue #163, opened 2018-07-27
PolyKinds shouldn't be on by default


bmmoore:
Apparently I was using the wrong extension, if KindSignatures suffices.
We have enough types with unused type parameters that we shouldn't turn on PolyKinds and silently change their meaning.

ttuegel:
Actually, the GHC documentation is unclear about this. It seems that `PolyKinds` is required for polymorphism itself and not just inference. Unfortunately, there seems to be no way to have polymorphism and disable inference, so I'm not sure what we want to do.

PR #165, opened 2018-07-27
Disable PolyKinds by default

ttuegel:
PolyKinds makes many of our constructors much more polymorphic than
intended. The extension is only actually required in
Data.Functor.Impredicative.

Fixes: #163

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/AST/Builders.hs
+++ b/src/main/haskell/kore/src/Kore/AST/Builders.hs
@@ -71,7 +71,7 @@ It can also be used to transform a symbol or alias sentence to something that
 can be applied to patterns.
 -}
 applyPS
-    :: SentenceSymbolOrAlias s
+    :: (SentenceSymbolOrAlias s, Show level)
     => s level (Pattern level) Variable
     -> [Sort level]
     -> [CommonPurePatternStub level]
@@ -101,7 +101,7 @@ It can also be used to transform a symbol or alias sentence to something that
 can be applied to patterns.
 -}
 applyS
-    :: SentenceSymbolOrAlias s
+    :: (SentenceSymbolOrAlias s, Show level)
     => s level (Pattern level) Variable
     -> [CommonPurePatternStub level]
     -> CommonPurePatternStub level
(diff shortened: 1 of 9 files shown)


src/main/haskell/kore/src/Kore/AST/Common.hs
```diff
@@ -80,18 +75,15 @@ prettyPrintAstLocation :: AstLocation -> String
...
 prettyPrintAstLocation AstLocationTest = "<test data>"
-prettyPrintAstLocation
-    (AstLocationFile FileLocation
-        { fileName = name
-        , line = line'
-        , column = column'
-        }
-    )
-    = name ++ " " ++ show line' ++ ":" ++ show column'
+prettyPrintAstLocation (AstLocationFile FileLocation { fileName = name
```
virgil-serbanuta: Please do not reformat code, especially not in a such bad way. Please follow this style guide:
https://docs.google.com/document/d/1h_rNcOzZQi0pCKMCdtmUm-MDKklY33LujpMzCEuQPCQ/edit?ts=5ada4619#

This comment is valid for this entire pull request.

src/main/haskell/kore/src/Kore/AST/Builders.hs
```diff
@@ -71,7 +71,7 @@ It can also be used to transform a symbol or alias sentence to something that
 can be applied to patterns.
 -}
 applyPS
-    :: SentenceSymbolOrAlias s
+    :: (SentenceSymbolOrAlias s, Show level)
```
virgil-serbanuta: Why is this "Show level" needed?

ttuegel: `deriveShow1` and the like functions require class constraints on all types of kind `*`, even phantom types. With `PolyKinds` on, `level` couldn't be inferred to kind `*` because it's phantom, but not it defaults to `*`.

bmmoore: That will go away if we put Object and Meta in their own kind rather than just being dummy types in `*`

ttuegel: @bmmoore Right, I have a branch for that like we discussed, but it is waiting on this and on #164.

PR #168, opened 2018-07-31
Document plan for Object-level/Mixed-level proofs

bmmoore:
Note this only touches documentation, code should not be affected.

### DIFF
--- a/docs/proof_object.md
+++ b/docs/proof_object.md
@@ -275,3 +275,143 @@ proof checker extremely simple.
 * Check that `P11` is syntactically equal to `Q1`
 * Check that `P12` is syntactically equal to `P21`
 * Check that `P22` is syntactically equal to `Q2`
+
+# Mixed-level Proof Object Syntax
+
+Language definitions are expected to use mostly
+object-level or mixed-level patterns, which is why
+we have defined the Kore syntax to allow such patterns,
+rather than simply requiring everything to be encoded
+as meta-level patterns.
+For the same reason, we wish to define a syntax of proofs
+that allows writing mixed-level patterns, and that allows
+object-level reasoning steps to be written in the
+form `by <ProofRule>` rather than by explicitly
+constructing meta-level proofs of patterns involving `#provable`
+using the hypotheses described in section 7.7 of
+The Semantics of K.
+
+To support this we allow as patterns any Kore pattern,
+which follows the syntax of `<pattern>` in 9.1.4.
+```
+  <ProofObject>  ::= <AssumedClaim>* <DerivedClaim>*
+
+  <AssumedClaim> ::= <Id> : <Pattern>
+
+  <DerivedClaim> ::= <Id> : <Pattern> by <ProofRule>
+
+  <Id> ::= /* Quite flexible syntax. */
+```
+
+Proofs will be defined so that a valid mixed-level proof can
+be transformed into a valid simple proof that assumes or
+proves (resp.) the lifting of each assumed or derived claim
+of the mixed-level proof.
+
+The patterns in claims will be use a special "claim lifting" that
+transforms a pattern `phi` whose topmost element is an meta-level
+construct into the pure meta pattern `[[phi]]`, and lifts
+a pattern which has an object-level construct as the topmost element
+and has no metavariables into the pure meta pattern `#provable([[phi]])`.
+
+The claim lifting of a pattern that has metavariables needs to be defined
+in detail, but it will generally add implications before `#provable([[phi]])`
+asserting that the metavariables are instantiated with well-formed
+sorts or object patterns.
+
+Following the conventions of lifting we the meta-level proof rules
+will have a '#' prefix added to the name of the object-level proof rules.
+All the same proof rules that we have above for the meta-level should
+have object-level versions as well, but expanding the object-level
+proof to a pure matching logic proof will involve instantiating and
+using hypotheses about `#provable` rather than using matching logic
+proof rules directly.
+Compared to the previous section, the meta-level proof rules are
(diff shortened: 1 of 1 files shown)


docs/proof_object.md
```diff
@@ -275,3 +275,91 @@ proof checker extremely simple.
 * Check that `P11` is syntactically equal to `Q1`
 * Check that `P12` is syntactically equal to `P21`
 * Check that `P22` is syntactically equal to `Q2`
+
+# Mixed-level Proof Object Syntax
+
+Language definitions are expected to use mostly
+object-level or mixed-level formulas, which is why
```
hreada: formulas => patterns

docs/proof_object.md
```diff
@@ -275,3 +275,91 @@ proof checker extremely simple.
...
+
+Proofs will be defined so that a valid mixed-level proof can
+be transformed into a valid simple proof that assumes or
+proves (resp.) the lifting of each assumed or derived claim
+of the mixed-level proof.
+
+#### Note
+We may wish to disambiguate the object-level proof rule
+
+A pattern is considered equivalent to its lifting.
```
hreada: Maybe it is worthy to point out that there's a difference between the lifting of a pattern and the one of a pattern as a claim; that is, the lifting process defined in _Section 9.2.8 Lift Patterns_ in _Semantics of K_ lifts a pattern phi to its meta-representation (AST) [[phi]]. The lifting of a claim, say `id : phi`, is `id : #provable([[phi]])`.

docs/proof_object.md
```diff
@@ -275,3 +275,91 @@ proof checker extremely simple.
...
+that `#wellFormed` holds of the lifting of P whenever
+P is well-formed.
+
+Similarly, hypotheses of the form `#occursFree{s}(x,P)`
+or `\not{s}(#occursFree{s}(x,P))` are also automatically
+discharged, and new proof rules `check-free(x,P)`
+and `check-not-free(x,P)`
+
+```
+  <ProofRule>    ::=
```
hreada: Maybe make it clear and say `<MixedLevelProofRule> ::= <ObjectLevelProofRule> | <MetaLevelProofRule>` and explicitly use `#propositional1`, `#propositional2`, etc., for the meta-level rules.

docs/proof_object.md
```diff
@@ -275,3 +275,91 @@ proof checker extremely simple.
...
+All the proof rules above should be available at the object level as well,
+but now the actual reasoning will be in the form of properly
+instantiating and applying a corresponding meta-level hypothesis.
+
+The "object-level" form of the proof rules are equivalent to proving
+`#provable` of the lifting of the formula.
+To exactly resemble the meta-level proof rules, the `#wellFormed`
+hypotheses are automatically discharged.
+
+The new proof rule `check-well-formed(P)` proves
```
hreada: Are `check-well-formed` and `check-free` added to the syntax below?

docs/proof_object.md
```diff
@@ -275,3 +275,91 @@ proof checker extremely simple.
...
+forms, prefixing a '#' to unambiguously refer to the meta-level form.
+
+All the proof rules above should be available at the object level as well,
+but now the actual reasoning will be in the form of properly
+instantiating and applying a corresponding meta-level hypothesis.
+
+The "object-level" form of the proof rules are equivalent to proving
+`#provable` of the lifting of the formula.
+To exactly resemble the meta-level proof rules, the `#wellFormed`
+hypotheses are automatically discharged.
```
hreada: Why are they automatically discharged?

docs/proof_object.md
bmmoore: Do you think we should not overload the rule names? If not, consider that it would be the prefixed names in a mixed proof that desugar directly to the (unprefixed) meta-level proof rules.

hreada: I think we shouldn't overload rule names. In addition, I prefer we adopt a consistent naming convention here, that is `#...` for meta-level proof rules and non-`#...` for object-level proof rules.

bmmoore: Do you think we should also add the `#` prefix to pure matching logic proofs, or just to meta-level rules in mixed proofs?

docs/proof_object.md
bmmoore: As it says, "To exactly resemble the meta-level proof rules".
A rule like the matching logic "singvar" doesn't have an explicit hypothesis about the formulas being well-formed.

docs/proof_object.md
```diff
@@ -275,3 +275,143 @@ proof checker extremely simple.
...
+Proofs will be defined so that a valid mixed-level proof can
+be transformed into a valid simple proof that assumes or
+proves (resp.) the lifting of each assumed or derived claim
+of the mixed-level proof.
+
+The patterns in claims will be use a special "claim lifting" that
+transforms a pattern `phi` whose topmost element is an meta-level
+construct into the pure meta pattern `[[phi]]`, and lifts
+a pattern which has an object-level construct as the topmost element
+and has no metavariables into the pure meta pattern `#pattern([[phi]])`.
```
hreada: #pattern => #provable?

bmmoore:
@u-- The only free variable things being proven automatically are `#occursFree` claims talking about object-level patterns.

u--:
Yes I see that. The question is what happens to substitution operations at the meta-level.

u--:
@bmmoore I still do not see how this is resolved. Grigori has said that checking meta-level proofs should be 100% string comparison, with "no parsing required," though maybe that's a bit of poetic license. If there are proof rules at the meta level that do substitution, we do not have this property.

PR #170, opened 2018-08-03
Parser benchmarks

ttuegel:
@njohnwalker warned me that the parser may be fragile with respect to performance, so before I make changes to extract location information, I am adding benchmarks to ensure I do not create any performance regressions.

I modified the library to add `NFData` instances for all our abstract syntax types; no additional dependencies are introduced because the library already depends on `deepseq`. `NFData` is used by the benchmark to evaluate all data structures to normal form, ensuring that we are not bypassing the benchmark with laziness.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/bench/parser/Main.hs
+++ b/src/main/haskell/kore/bench/parser/Main.hs
@@ -1,0 +1,28 @@
+module Main where
+
+import Criterion.Main
+
+import Kore.Parser.Parser
+       ( fromKore )
+
+import qualified Paths
+
+main :: IO ()
+main =
+    defaultMain
+    [ parse "Parse implicit Kore definition" (Paths.dataFileName "../../kore/kore.kore")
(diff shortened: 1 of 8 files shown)


src/main/haskell/kore/src/Control/DeepSeq/Orphans.hs
```diff
@@ -0,0 +1,11 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
virgil-serbanuta: File comment?

src/main/haskell/kore/bench/parser/Main.hs
```diff
@@ -0,0 +1,35 @@
+module Main where
```
virgil-serbanuta: How does one run this? Should it be added to package.yaml?

PR #174, opened 2018-08-08
Kore simplification.

virgil-serbanuta:
###### Reviewer checklist

---


u--:
The function `evaluate` is duplicated in every file, can that be factored out?

src/main/haskell/kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -0,0 +1,238 @@
...
+        , Show (variable level)
+        , Ord (variable level)
+        )
+    => PureMLPattern level variable
+    -> PureMLPattern level variable
+    -> PureMLPattern level variable
+makeTermAnd b@(Bottom_ _) _ = b
+makeTermAnd (Top_ _) term = term
+-- TODO: (partial) unification / other simplifications
+makeTermAnd first second = makeTermAndSecond first second
```
traiansf: Why not in the same function?

virgil-serbanuta:
@ttuegel @bmmoore : I'll merge this PR now, since it stops other people from advancing, but, if you have comments, I'll address them in follow-up PRs.

Also, I promise to write follow-up PRs for tests and for solving the TODOs I left in the code (at least the important ones).

src/main/haskell/kore/src/Kore/Step/Simplification/And.hs
ttuegel: > Because the Haskell pattern matching stuff is o(n^2) and it will emit a warning of it taking over 2M iterations to check a function that uses the underscore matchers (e.g. Bottom_) on at least two arguments.

The exhaustiveness checker does not work well for recursive matching in general, but the situation does get much worse with pattern synonyms. We can write the `case`s ourselves, which is still reasonably simple:
```.hs
makeTermAnd first second =
    case first of
        Bottom_ _ -> first
        Top_ _ -> second
        _ -> case second of
            Bottom_ _ -> second
            Top_ _ -> first
            _ -> mkAnd first second
```

(I think pattern matching on the LHS of `=` was a huge design mistake for Haskell, and recursive matching in general, but that's a long rant.) Pattern synonyms are supported better in GHC 8.4, but we don't have that on Jenkins at the moment.

src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
```diff
@@ -0,0 +1,383 @@
...
+@
+
+-}
+crossProductGeneric
+    :: (child1 -> child2 -> child3)
+    -> MultiOr child1
+    -> MultiOr child2
+    -> MultiOr child3
+crossProductGeneric joiner (MultiOr first) (MultiOr second) =
+    MultiOr $ map (uncurry joiner) $ List.crossProduct first second
```
ttuegel: This is often written one of these ways:
```.hs
map (uncurry joiner) $ List.crossProduct first second
===
-- Sometimes you see `liftA2` instead of `liftM2`; they do the same thing, 
-- but `liftA2` has a relaxed class constraint.
liftM2 joiner first second
===
-- This is probably the most common variant.
joiner <$> first <*> second
```
The second and third variants do less allocation at runtime.

src/main/haskell/kore/src/Kore/Step/Simplification/Bottom.hs
```diff
@@ -0,0 +1,31 @@
...
+import           Kore.AST.Common
+                 ( Bottom (..) )
+import           Kore.Step.OrOfExpandedPattern
+                 ( OrOfExpandedPattern )
+import qualified Kore.Step.OrOfExpandedPattern as OrOfExpandedPattern
+                 ( make )
+import           Kore.Step.Simplification.Data
+                 ( SimplificationProof (..) )
+
+{-| simplifies a Bottom pattern, which means returning an always-false or.
```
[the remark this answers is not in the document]
ttuegel: Because of the global instance coherence constraint, creating a class works best if there is only one valid instance you can make for a type, or at least one canonical instance. I don't think we can say there is "one valid simplifier" you can define for a type. I'm not even sure there is a clear "canonical simplifier" for each type. I can imagine that someone might want to substitute their own simplifier in the future, which would be ~~impossible~~ fairly difficult if it was a class. But feel free to tell me, "No, that's silly, nobody would ever want to define their own simplifiers!" :grinning:

PR #176, opened 2018-08-08
Include axiom "requires" conditions

ttuegel:
- Added a field `axiomPatternRequires` to `AxiomPattern`, carried through `stepWithAxiom`.
- Added commentary to `stepWithAxiom`.
- Incidentally, generalized `Kore.ASTUtils.SmartPatterns` using GADT-style pattern synonym constraints (a different kind of pre- and post-condition).

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/AxiomPatterns.hs
```diff
@@ -27,6 +28,7 @@ it can only represent axioms that look like left-pattern => right-pattern.
 data AxiomPattern level = AxiomPattern
```
hreada: Update the comments accordingly; probably `left-pattern => right-pattern` => `left-pattern => right-pattern requires requires-pattern`?

src/main/haskell/kore/src/Kore/ASTUtils/SmartPatterns.hs
```diff
@@ -70,9 +70,10 @@ pattern Ceil_
     -> PureMLPattern level var
 
 pattern DV_
-  :: Sort Object
+  :: () => (level ~ Object) =>
```
virgil-serbanuta: Why the `() =>`?

src/main/haskell/kore/src/Kore/Predicate/Predicate.hs
```diff
@@ -279,3 +283,13 @@ makeFalsePredicate
     => Predicate level var
 makeFalsePredicate =
     GenericPredicate mkBottom
+
+{- | Replace all variables in a @Predicate@ using the provided mapping.
+-}
+mapVariables :: (from level -> to level) -> Predicate level from -> Predicate level to
+mapVariables f = wrapPredicate . mapPatternVariables f . unwrapPredicate
```
virgil-serbanuta: Is `fmap (mapPatternVariables variableMapper) predicate` better?

Regardless of whether it's better, could you replace former uses of the fmap stuff  with your function? There should be at leas one of these in ExpandedPattern

src/main/haskell/kore/src/Kore/Predicate/Predicate.hs
```diff
@@ -279,3 +283,13 @@ makeFalsePredicate
...
+
+{- | Extract the set of all (free and bound) variables from a @Predicate@.
+-}
+allVariables :: Ord (var level) => Predicate level var -> Set (var level)
+allVariables = pureAllVariables . unwrapPredicate
```
virgil-serbanuta: Could you remove `variableSetFromPredicate`?

src/main/haskell/kore/test/Test/Kore/Step/BaseStep.hs
```diff
@@ -75,6 +75,7 @@ test_baseStep =
                 AxiomPattern
                     { axiomPatternLeft = asPureMetaPattern (x1 PatternSort)
                     , axiomPatternRight = asPureMetaPattern (x1 PatternSort)
+                    , axiomPatternRequires = makeTruePredicate
```
virgil-serbanuta: Could you add a test that uses a different predicate? See `testCase "Preserving initial condition."` for an example.

You could do the same for the Integration test, but that's less important.

Could you also add one for the UserDefined test? There is a task for changing the axiom handling code, and having a test for a non-true condition would be nice.

src/main/haskell/kore/src/Kore/ASTUtils/SmartPatterns.hs
ttuegel: Pattern synonyms take *two* contexts: the first is the required context, the second is the provided context. (Or, precondition and postcondition if you prefer.) This says: nothing is required to attempt matching this pattern, and if it does match then you may assume `level ~ Object`. The prior version of `DV_` had `level` replaced by `Object` explicitly; that is equivalent to having the constraints `(level ~ Object) => (level ~ Object) => ...`. In other words, the old pattern synonym required you to know the pattern was an object-level pattern before attempting to match `DV_`.  The [GHC manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#typing-of-pattern-synonyms) describes all the details of typing pattern synonyms if you want to know more.

src/main/haskell/kore/test/Test/Kore/Step/BaseStep.hs
```diff
@@ -887,6 +887,52 @@ test_baseStep =
                         }
                 )
             )
+    -- x => x requires g(x)=f(x)
```
ttuegel: @virgil-serbanuta I added this test in `BaseStep` for axiom conditions.

src/main/haskell/kore/test/Test/Kore/Step/Function/UserDefined.hs
```diff
@@ -92,6 +92,28 @@ test_userDefinedFunction =
                 (asApplication (metaF (x PatternSort)))
             )
         )
+    , testCase "Cannot apply step with unsat axiom pre-condition"
```
ttuegel: @virgil-serbanuta Is this what you had in mind for an axiom condition test in `UserDefined`?

src/main/haskell/kore/src/Kore/ASTUtils/SmartPatterns.hs
virgil-serbanuta: Thanks for the explanation! I removed the `=> ()` part and it seemed to compile, so I thought I'd ask.

PR #177, opened 2018-08-08
Derive Eq1/Ord1/Show1, improve Hashable, centralize orphans

bmmoore:
This uses the various make* functions from deriving-compat
to replace handwritten instances in cases where the derive*
functions didn't work.

The handwritten Hashable for Pattern is updated to follow
the instructions for writing an instance from the Data.Hashable
documentation.

Several instances explicitly given on type synonyms expanding
to (Fix ...) were removed by adding more orphan instances for Fix.
As all the orphans so far were for Fix I reorganized the orphan
instance modules to a single module grouped by type rather than
many modules grouped by class.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/AST/Common.hs
```diff
@@ -1085,161 +1085,17 @@ data Pattern level variable child where
...
-            (StringLiteralPattern _, _) -> LT
-            (_, StringLiteralPattern _) -> GT
-            (CharLiteralPattern a', CharLiteralPattern b') -> compare a' b'
-            (CharLiteralPattern _, _) -> LT
-            (_, CharLiteralPattern _) -> GT
-            (TopPattern a', TopPattern b') -> liftCompare liftedCompare a' b'
-            (TopPattern _, _) -> LT
-            (_, TopPattern _) -> GT
-            (VariablePattern a', VariablePattern b') -> compare a' b'
+    liftCompare liftedCompare a b = $(makeLiftCompare ''Pattern) liftedCompare a b
```
virgil-serbanuta: Just curious: You removed (say) `pat` from the argument list of `liftShowsPrec` but you didn't remove `a` and `b` here. Is there a reason for this difference?

PR #182, opened 2018-08-13
Merge and normalize

njohnwalker:
I solved a cyclical dependency issue that kept normalization from being incorporated in the `mergeSubstitutions` definition.  But once I started integrating  normalization in the function, it became clear that normalization requires a lot more assumptions than a simple merge, so I created a new definition that did both merging and normalization (`mergeAndNormalizeSubstitutions`) and replaced that in BaseStep.

I can modify the original function, but I was reluctant to thread all the extra dependencies through the additional usages of `mergeSubsttutions`

~Two tests fail now~
edit: fixed the way errors are handled, tests pass

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
+++ b/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
@@ -256,9 +256,9 @@ stepWithAxiom
             $ unificationOrSubstitutionToStepError
             $ ctorSubstitutionCycleToBottom
             ( return ( PredicateSubstitution
-                       { predicate = makeFalsePredicate
-                       , substitution = []
-                       }
+                           { predicate = makeFalsePredicate
+                           , substitution = []
+                           }
                      , EmptyUnificationProof
                      )
             )
@@ -266,9 +266,9 @@ stepWithAxiom
 
     return $ do
         ( PredicateSubstitution
-          { predicate = normalizedCondition
-          , substitution = normalizedSubstitution
-          }
+              { predicate = normalizedCondition
+              , substitution = normalizedSubstitution
+              }
           , _  -- TODO: Use this proof
           ) <- normalizedSubstitutionWithCounter
 
--- a/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
+++ b/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
@@ -24,7 +24,7 @@ import           Data.Maybe
 import           Data.Monoid
                  ( (<>) )
 import           Data.Reflection
-                 ( give )
+                 ( Given, give )
 import qualified Data.Set as Set
 
 import           Kore.AST.Common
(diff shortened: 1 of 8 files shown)


src/main/haskell/kore/src/Kore/Step/Substitution.hs
```diff
@@ -92,22 +70,68 @@ mergeSubstitutions tools first second = do
...
+    return $ do --IntCounter Monad
+      { PredicateSubstitution
+        { predicate = _condition
+        , substitution = substitutionList'
+        } <- predSubstitution
+      -- ; condition' <- fst $ mergeConditionsWithAnd --TODO: merge conditions
+      --                 [ makeTruePredicate
+      --                 , condition ]
+      ; return ( PredicateSubstitution
+                 { predicate = makeTruePredicate
```
virgil-serbanuta: You should not be dropping the _condition, the merged substitution is not valid without it. I.e. if you merge `[x=f(a)]` with `[x=g(b)]` you should get `[x=f(a)]` and `f(a)=g(b)` (this particular example is not implemented yet, but there's a task for it).

njohnwalker: Oh you are right, I fixed that, but the test still doesn't pass.  I think I'm not catching an error in `BaseStep` that should be caught (and converted to a `Bottom`)

src/main/haskell/kore/src/Kore/Step/Substitution.hs
```diff
@@ -92,22 +70,66 @@ mergeSubstitutions tools first second = do
...
+    -> UnificationSubstitution level variable
+    -> UnificationSubstitution level variable
+    -> Either
+        ( UnificationOrSubstitutionError     level variable )
+        ( IntCounter
+          ( PredicateSubstitution level variable
+          , UnificationProof level variable
+          )
+        )
+mergeAndNormalizeSubstitutions tools first second = do
```
ttuegel: The nested `Monad` makes this code difficult to parse, but a much simpler implementation of this function is
```.hs
mergeAndNormalizeSubstitutions tools first second = do
    (substitutionList, proof) <-
        normalizeSubstitutionDuplication' (first ++ second)
    predSubstitution <- normalizeSubstitution' substitutionList
    -- TODO(virgil): Return the actual condition here. and proofs
    return $ (,) <$> predSubstitution <*> pure proof  -- Inside IntCounter
```

njohnwalker: Oh wow, I didn't look at rewriting this after I removed the mergeConditions section. you are right, much simpler...

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -256,45 +250,39 @@ stepWithAxiom
...
-
     normalizedSubstitutionWithCounter <-
-        normalizeSubstitutionError
-            (return PredicateSubstitution
-                { predicate = makeFalsePredicate
-                , substitution = []
-                }
+        stepperVariableToVariableForError
+            existingVars
+            $ unificationOrSubstitutionToStepError
```
virgil-serbanuta: Note that having two different functions (`unificationOrSubstitutionToStepError` and `ctorSubstitutionCycleToBottom`) for error conversion solves the naming problem that the previous functions had (i.e. `substitutionToStepError` was misleading), but allows us to call one function without the other. I think that we never need to call the former without the latter, but doing so would produce hard to detect bugs, so I'd have preferred to either have this as a single function or to have an `unificationOrSubstitutionToStepError` which crashes when encountering a constructor cycle error.

PR #187, opened 2018-08-17
Always parse matching logic constructors the same way

ttuegel:
`Kore.Parser.ParserImpl.mlConstructorRemainderParser` is modified so that it
does not take a continuation. Instead, it parses all the matching logic
constructors itself. It takes a level parameter which is used to determine if a
particular constructor should be rejected. Dropping the continuation ensures
that matching logic constructors are always parsed by the same code path, which
fixes the parsing of domain values in aliases.
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Parser/ParserImpl.hs
+++ b/src/main/haskell/kore/src/Kore/Parser/ParserImpl.hs
@@ -601,26 +601,10 @@ koreMLConstructorParser = do
         c <- ParserUtils.peekChar'
         if c == '#'
             then asKorePattern <$>
-                mlConstructorRemainderParser korePatternParser
-                    Meta patternType (unsupportedPatternType Meta)
+                mlConstructorRemainderParser korePatternParser Meta patternType
             else asKorePattern <$>
-                mlConstructorRemainderParser korePatternParser
-                    Object patternType objectMlConstructorRemainderParser
-    objectMlConstructorRemainderParser patternType =
-        case patternType of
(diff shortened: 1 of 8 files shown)


ttuegel:
So that I can avoid making the same mistake in the future: What should my unit tests have checked for?

virgil-serbanuta:
Any test that would have failed before, but worked after your change, would have been a reasonable start. You added an integration test checking similar stuff, but an integration test tests more stuff than an unit-test, so it will also fail for unrelated reasons, while an unit-test would ideally point at the exact problem.

Now, we don't have an ideal separation between our components, so a unit-test usually tests more than just one thing (we might try to write mockable stuff for Haskell, but I can't tell yet if it would look reasonable or not; though I should probably try it at least once). Before the change that automatically includes all tests, they were topologically-sorted, so the first test that failed almost always identified the root cause of the problem (IIRC the main exceptions are two or three test files where I mistakenly thought it would be a good idea to test all possible combinations of stuff without testing first the individual pieces of functionality and some of the serialization tests which look more like integration tests, but should have also had unit tests). After that change, it's still true that, almost always, one of the failing tests identifies the problem, though one needs to sort according to dependencies to see which one.

So, with that in mind, there were a few options:
1. The functionality change in purePatternParser.
2. The functionality change in koreSentenceParser.
For both, tests for dv, next and rewrites would have been nice.

ttuegel:
I was going to push the unit tests to this branch, but on second thought, I will include them in my pull request for the domain value syntax, instead, as some of the new tests require that.

PR #193, opened 2018-08-27
SMT

u--:
###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/SMT/SMT.hs
```diff
@@ -0,0 +1,171 @@
...
+          return (tx1 `op` tx2)
+        goBoolean (And_     _   x1 x2) = goBinaryOp (&&&) goBoolean x1 x2
+        goBoolean (Or_      _   x1 x2) = goBinaryOp (|||) goBoolean x1 x2
+        goBoolean (Implies_ _   x1 x2) = goBinaryOp (==>) goBoolean x1 x2
+        goBoolean (Iff_     _   x1 x2) = goBinaryOp (<=>) goBoolean x1 x2
+        goBoolean (Not_     _   x1)    = goUnaryOp (bnot) goBoolean x1
+        goBoolean (Equals_  s _ x1 x2) = c x1 x2
+          where
+            c = case getHookString $ getSortHook s of
+                "BOOL.Bool" -> goBinaryOp (.==) goInteger
```
traiansf: Aren't `INT.Int` and `BOOL.Bool` switched?

u--: Evidently do need more tests..

traiansf:
@u-- if you can merge with master, I'll take another look.

src/main/haskell/kore/src/Kore/IndexedModule/MetadataTools.hs
```diff
@@ -24,7 +24,8 @@ import Kore.IndexedModule.Resolvers
 -- |'MetadataTools' defines a dictionary of functions which can be used to
 -- access the metadata needed during the unification process.
 data MetadataTools level attributes = MetadataTools
-    { attributes :: SymbolOrAlias level -> attributes
+    { symAttributes :: SymbolOrAlias level -> attributes
```
traiansf: `symAttributes` seems a little vague;  one could also read it as symbolic attributes.
Maybe rename it to `symbolAttributes` or `headAttributes` ?

u--: I think we can agree that sym = kore symbol in this codebase, idk of anything called "Symbolic" outside of SBV's API.

src/main/haskell/kore/test/Test/Kore/SMT/SMT.hs
```diff
@@ -0,0 +1,173 @@
...
+prop_true = run propTrue
+
+prop_false :: Property
+prop_false = run propFalse
+
+run :: CommonPurePattern Object -> Property
+run prop = (give tools $ provePattern prop) === Just True
+
+prop1 :: CommonPurePattern Object
+prop1 = dummyEnvironment $ mkNot $
```
virgil-serbanuta: Note that the prop_1 test is somewhat less readable than when inlining the definition for this.

src/main/haskell/kore/test/Test/Kore/SMT/SMT.hs
```diff
@@ -0,0 +1,173 @@
...
+prop_demorgan :: Property
+prop_demorgan = run propDeMorgan
+
+prop_true :: Property
+prop_true = run propTrue
+
+prop_false :: Property
+prop_false = run propFalse
+
+run :: CommonPurePattern Object -> Property
```
virgil-serbanuta: expectTrue might have been a better name.

src/main/haskell/kore/src/Kore/SMT/SMT.hs
```diff
@@ -0,0 +1,181 @@
...
+-- | Returns `Just True` if the sentence is satisfied in all models,
+-- `Just False` if it has a counterexample,
+-- Nothing if timeout/undecidable.
+provePattern
+    :: Given (MetadataTools Object SMTAttributes)
+    => CommonPurePattern Object
+    -> Maybe Bool
+provePattern p =
+    case res of
+      Unsatisfiable _   -> Just True
```
virgil-serbanuta: You're using a lot of strange indentations in this file and its tests: 2 spaces instead of 4, 1 space instead of 2, 0 spaces instead of 4 (e.g. in `prop4`). Indenting a `case` at the top level in a function two spaces and then the case branches another two spaces (as done in `getSortAttributes`) is not that bad (although it is not explicitly allowed by the coding style doc), but the others are kind of bad.

If you think it's useful I'll point out all the issues, but it would be easier if you'd prune them first this function has two (the `case` branch indentation and the `where` one).

PR #197, opened 2018-08-28
Tests for the function registry

traiansf:
###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Kore/Step/Function/Registry.hs
```diff
@@ -0,0 +1,274 @@
...
+            , updateAttributes
+                (Attributes [functionAttribute, constructorAttribute])
+                (simpleObjectSymbolSentence (SymbolName "g") (SortName "S"))
+            , asSentence
+                (SentenceAxiom
+                    { sentenceAxiomParameters = [asUnified sortVar]
+                    , sentenceAxiomAttributes = Attributes []
+                    , sentenceAxiomPattern =
+                        (patternPureToKore
+                            ( Implies_ sortVarS
```
virgil-serbanuta: You can leave it as-is, but, FWIW, the mk* constructors will automatically infer sorts and IIRC you can force the sort at the top to be sortVarS with some function I don't remember.

src/main/haskell/kore/test/Test/Kore/Step/Function/Registry.hs
```diff
@@ -0,0 +1,274 @@
...
+            testDef
+    in
+        case verifyResult of
+            Left err1            -> error (printError err1)
+            Right indexedModules ->
+                fromMaybe
+                    (error
+                        (  "The main module, '"
+                        ++ "test"
+                        ++ "', was not found. Check the --module flag."
```
virgil-serbanuta: The error message, specifically the '--module flag' part, was probably copy-pasted and should be changed.

src/main/haskell/kore/test/Test/Kore/Step/Function/Registry.hs
```diff
@@ -0,0 +1,274 @@
...
+testMetadataTools :: MetadataTools Object StepperAttributes
+testMetadataTools = extractMetadataTools testIndexedModule
+
+test_functionRegistry :: [TestTree]
+test_functionRegistry =
+    [ testCase "Checking that two axioms are found for f"
+        (assertEqual ""
+            2
+            (length $ fromMaybe
+                (error "Should find precisely two axiom for f")
```
virgil-serbanuta: s/axiom/axioms/

src/main/haskell/kore/test/Test/Kore/Step/Function/Registry.hs
```diff
@@ -0,0 +1,274 @@
...
+        (assertEqual ""
+            2
+            (Map.size testEvaluators)
+        )
+    , testCase "Checking that the indexed module contains a rewrite axiom"
+        (assertEqual ""
+            (1::Int)
+            (length (koreIndexedModuleToAxiomPatterns Object testIndexedModule))
+        )
+    , testCase "Checking that evaluator simplifies correctly"
```
virgil-serbanuta: Not sure this is the right place to put this test, but it's better to have it than not to have it, I guess.

traiansf: It's mostly for coverage;  If I didn't put it, then the evaluator part in a lookup would show as unevaluated.

src/main/haskell/kore/test/Test/Kore/Step/Function/Registry.hs
traiansf: Well, I thought about it, but for `mk*` constructors I need to provide some  `SortTools` which might be more cumbersome, as they would be for the definition I'm trying to build.
Of course, I could create another definition just with the syntax, or mock them, but it seems more work than just spelling the sorts.

PR #199, opened 2018-08-29
Inj unification

traiansf:
##### Reviewer checklist

---

Should be reviewed after #195 
Implements most cases of unification modulo injections.  Notable exception:  
`inj{s1,s3}(a{}())` vs. `inj{s2,s3}(x)`

This should succeed if `s1 < s2 < s3` but fail otherwise.


dlucanu:
A more general situation is
`inj{s1,s3}(y)` vs. `inj{s2,s3}(x)`
when there is `s0` such that `s0 < s1 < s3` and `s0 < s2 < s3`.
I guess that for such cases the retracts should be used as well.

ttuegel:
I was able to get this working (now that I *really* have the latest K). I did have to do `mv src/main/k/in-progress/lib src/main/k/`.

src/main/haskell/kore/src/Kore/Unification/UnifierImpl.hs
```diff
@@ -323,36 +398,47 @@ matchConstructor
...
-                                (applicationChildren ap1)
-                                (applicationChildren ap2)
-                    }
-            else Left $ Left
-                    (PatternClash
-                        (HeadClash head1)
-                        (HeadClash head2)
-                    )
-    | otherwise = Left $ Left $ NonConstructorHead head1
+matchInjTriangle tools head1 ap1 head2 ap2
```
virgil-serbanuta: Could you move this inj merging stuff in Simplification.Application?

src/main/haskell/kore/src/Kore/Step/PatternAttributes.hs
```diff
@@ -109,16 +112,22 @@ isFunctionalPattern tools = cata  reduceM
         proofs <- concat <$> sequence patt
         return (proof : proofs)
 
+isPreconstructedPattern
+    :: err
+    -> Pattern level variable pat
+    -> Either err (FunctionalProof level variable)
+isPreconstructedPattern _ (DomainValuePattern dv) =
+    return (FunctionalDomainValue dv)
```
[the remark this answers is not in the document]
traiansf: Although it does not return `Bool` it does return a proof when it succeds.

src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -352,16 +352,22 @@ kItemSort = groundObjectSort "SortKItem"
 -- The function below works around several limitations of
 -- the current tool by tricking the tool into believing that
 -- functions are constructors (so that function patterns can match)
--- and that @inj@ is both functional and constructor.
+-- and that @kseq@ and $dotk@ are both functional and constructor.
```
ttuegel: `$dotk@` => `@dotk@`

src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -352,16 +352,22 @@ kItemSort = groundObjectSort "SortKItem"
...
+        , isFunctional = isFunctional atts || isCons h || isInj h
+        , isInjective =
+            isInjective atts || isFunction atts || isCons h || isInj h
+        , isSortInjection = isSortInjection atts || isInj h
         }
     }
   where
     isInj :: SymbolOrAlias Object -> Bool
-    isInj h = getId (symbolOrAliasConstructor h) `elem` ["inj","kseq","dotk"]
+    isInj h =
```
ttuegel: Could we change the front-end to annotate `inj`, `kseq`, and `dotk` to be annotated with appropriate attributes, instead of matching the symbol name? Not _now_ necessarily, but eventually? Having certain names be special is not very nice.

src/main/haskell/kore/src/Kore/Implicit/ImplicitVarsInternal.hs
```diff
@@ -48,5 +51,5 @@ v2 = implicitUnparameterizedVariable "#v2"
 vx = implicitUnparameterizedVariable "#x"
 vx' = implicitUnparameterizedVariable "#x'"
 
-pS = sortParameter Meta "#sp" AstLocationImplicit
+pS = sortParameter (Proxy::Proxy Meta) "#sp" AstLocationImplicit
```
ttuegel: `Proxy` is `Applicative`, so that
```.hs
Proxy :: Proxy Meta
===
pure Meta
```
I'm not saying you need to change it here, I'm just mentioning it. It is especially convenient when a function is parametric in the level, so you write something like,
```.hs
foo :: MetaOrObject level => level -> Proxy level
foo level = pure level
```
instead of using `ScopedTypeVariables` (which requires writing an explicit `forall` and generally gets messy).

General question: Is there a reason you are using a proxy (`Proxy level`) instead of a witness (`IsMetaOrObject level`)?

src/main/haskell/kore/app/exec/Main.hs
traiansf: I totally agree with that. This entire `constructorFunctions` needs to disappear. They are here only temporary, until the relevant changes are effected in the frontend.

src/main/haskell/kore/src/Kore/Implicit/ImplicitVarsInternal.hs
traiansf: Why Proxy?  I needed a function of the form
```
sortParam :: String -> SortVariable level
sortParam name = sortParameter Proxy name AstLocationTest
```
Here, the parameter `level` is on the result sort, so I thought `Proxy` would perform the job better.
Would you think of a better solution in this context?

src/main/haskell/kore/src/Kore/Step/PatternAttributes.hs
```diff
@@ -127,7 +136,20 @@ checkFunctionalHead tools (ApplicationPattern ap) =
...
+-}
+isConstructorTop
+    :: MetadataTools level StepperAttributes
+    -> Pattern level variable pat
+    -> Bool
+isConstructorTop tools (ApplicationPattern ap) =
+    isConstructor (MetadataTools.symAttributes tools patternHead)
+  where
+    patternHead = applicationSymbolOrAlias ap
+isConstructorTop _ p = isRight (isPreconstructedPattern undefined p)
```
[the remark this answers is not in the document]
virgil-serbanuta: I still wish to have a better name for this, maybe `isConstructorLike`...

PR #204, opened 2018-09-06
stack.yaml: Include Z3 in nix.packages

ttuegel:
The SMT tests require Z3 to be installed. By adding it to nix.packages, Stack
will ensure it is provided automatically for Nix users.

###### Reviewer checklist

---

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -18,7 +18,9 @@
 resolver: lts-11.3
 
 nix:
-  packages : [ zlib ]
+  packages:
+    - z3
+    - zlib
 
 # User packages to be built.
 # Various formats can be used as shown in the example below.


stack.yaml
```diff
@@ -18,7 +18,9 @@
 resolver: lts-11.3
 
 nix:
-  packages : [ zlib ]
+  packages:
+    - z3
```
virgil-serbanuta: Will this install at least the 4.7.0 version or whatever is needed for our project?

ttuegel: It will use exactly the version of Z3 is in the user's Nixpkgs expression. Nix packages are pure functions from dependencies to build artifacts; all package versions are _exact_ to ensure reproducibility. If you would prefer, I can provide a Nix expression in the Kore repository which would provide an appropriate version.

PR #205, opened 2018-09-06
Simplifier cannot return error

ttuegel:
Any error that the simplifier would catch has already been detected during the
well-formedness check. Therefore, it does not make sense to have complicated
error handling.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/Builtin.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/Builtin.hs
@@ -41,8 +41,6 @@ module Kore.Builtin.Builtin
 
 import           Control.Monad
                  ( zipWithM_ )
-import           Control.Monad.Except
-                 ( MonadError )
 import qualified Control.Monad.Except as Except
 import qualified Data.Functor.Foldable as Functor.Foldable
 import           Data.HashMap.Strict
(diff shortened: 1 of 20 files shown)


src/main/haskell/kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -546,11 +544,25 @@ functionEvaluator ctx impl =
...
+ -}
+wrongArity :: HasCallStack => String -> a
+wrongArity ctx = error (ctx ++ ": Wrong number of arguments")
+
+{- | Evaluation failure due to an unparseable domain value.
+
+  Such a failure indicates a bug in the well-formedness checker.
+
+ -}
+liftParser :: HasCallStack => String -> Either (Error e) a -> a
```
virgil-serbanuta: You can leave it as-is, but consider naming this `extractParsedValue` or something similar.

src/main/haskell/kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -546,11 +544,25 @@ functionEvaluator ctx impl =
...
+
+{- | Evaluation failure due to a builtin call with the wrong arity.
+
+  Such a failure indicates a bug in the well-formedness checker.
+
+ -}
+wrongArity :: HasCallStack => String -> a
+wrongArity ctx = error (ctx ++ ": Wrong number of arguments")
+
+{- | Evaluation failure due to an unparseable domain value.
```
virgil-serbanuta: This comment does not describe this function, it only describes its error case.

PR #212, opened 2018-09-12
Make substitution/unification work for a=a and x=x substitutions

vladciobanu:
###### Reviewer checklist

---


vladciobanu:
A brief overview of the changes in this PR:

`Makefile` - I added this because I use it. I could easily add it to my local ignore list if it bothers people; I just find it easy to work with (especially starting `ghcid` and `test-ghci` are useful).

`ASTPrettyPrint.hs` - added a couple of instances for `Predicate` and `ExpandedPattern` which helped me with debug/tracing while figuring out what's going on. I did not add any tests for these instances.

`SubstitutionNormalization.hs` - main area with changes is here, inside `normalizeSubstitution`, where the base idea is I'm filtering out nodes that match `(a, [a])`. I forgot a `Show` constraint in the definition of `notReflexive` from when I was using `trace` here; I'll remove it.

`test/SubstitutionNormalization.hs` - edited the assertion for the test that described the previous behavior.

`test/BaseStep.hs` is where I am a bit confused. Three tests here failed (the ones that I switched from using `runStep` to `runStepTwice`). This was the only modification needed to make all tests pass. I would like to understand why these tests worked under the assumption that `a=a` would error during `normalizeSubstitution` and why running it twice works.

After some additional digging, I now think it might be because of how `normalizeSubstitutionAfterMerge` works, specifically since it normalizes the substitution list? This doesn't mean a lot to me right now (I need to get more familiar with the codebase), but it might help the reviewer.

src/main/haskell/kore/src/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -82,8 +82,12 @@ normalizeSubstitution tools substitution = do
     interestingVariables = extractVariables substitution
     variableToPattern = Map.fromList substitution
     dependencies = buildDependencies substitution interestingVariables
+    depsWithoutReflexivity = Map.filterWithKey (notReflexive) dependencies
+    notReflexive :: (Eq a, Show a) => a -> [a] -> Bool
+    notReflexive a [x]= a /= x
```
virgil-serbanuta: This is a quick reply from my mobile phone, so I may be wrong.

This does not seem right.

The following must happen when calling normalizeSubstitution (in Kore.Substitution.SubstitutionNormalization):

`[a=a]` --> `[]`

`[a=f(a)]` --> `error`

`[a=a,a=b]` --> `[a=b]`

`[a=b, b=f(a)]`>`error`

I don't think that the second and the third are working now. BTW, you should probably add these as tests for normalizeSubstitution.

src/main/haskell/kore/src/Kore/Variables/Free.hs
```diff
@@ -43,6 +44,19 @@ freeVariables = patternBottomUpVisitor freeVarsVisitor
         Set.delete (asUnified (forallVariable f)) (forallChild f)
     freeVarsVisitor p = fold p  -- default rule
 
+-- | 'isVariablePattern' returns true iff the pattern is just a VariablePattern.
+isVariablePattern
```
virgil-serbanuta: If you still need this, 'Common.hs' would probably be a better place.

src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
```diff
@@ -0,0 +1,152 @@
...
+                    )
+                ]
+                [   ( Mock.x
+                    , Mock.constr10 (Mock.f Mock.a)
+                    )
+                ]
+            )
+        )
+    , testCase "Constructor circular dependency?"
+        -- [a=b] + [b=constructor(a)]  === bottom
```
virgil-serbanuta: Huh... There are a few things broken here (not by you).

First, could you delete `substitutionToStepError`? It is not used anywhere.

Second, could you call `ctorSubstitutionCycleToBottom` in `normalizeSubstitutionAfterMerge`? It should fix this test. I guess then you could delete its usage from `BaseStep.hs`.

Actually, even better, I guess you could fully delete the `CtorCircularVariableDependency` error case and you could make `normalizeSubstitution` return bottom in this case.

src/main/haskell/kore/src/Kore/ASTPrettyPrint.hs
```diff
@@ -776,3 +780,15 @@ instance (MetaOrObject level, PrettyPrint (variable level))
         writeOneFieldStruct flags "FunctionalStringLiteral" l
     prettyPrint flags (FunctionalCharLiteral l) =
         writeOneFieldStruct flags "FunctionalCharLiteral" l
+
+instance (MetaOrObject level, PrettyPrint (variable level))
+    => PrettyPrint (Predicate level variable)
+  where
+    prettyPrint flags pat =
+        prettyPrint flags (unsafeCoerce pat :: PureMLPattern level variable)
```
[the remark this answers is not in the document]
vladciobanu: You are absolutely right, I just failed to notice `unwrapPredicate` exists. The reason I used `unsafeCoerce` is because it's just a newtype wrapper so it should always be safe (Coercible/`coerce` would not work because the constructor is hidden).

I'll change to `unwrapPredicate`

src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
vladciobanu: How would you feel about renaming `CtorCircularVariableDependency vars` to a unit ctor `Bottom`, and mapping it to `Right bottom` in `normalizeSubstitutionAfterMerge`?

virgil-serbanuta: It would work, I guess, but it would be less clean (i.e. it would be obvious that you're handling a non-error case as an error, and you couldn't even pretend that it would be useful somewhere as an actual error) why would that be better than handling this in `normalizeSubstitution`

vladciobanu:
I removed `CtorCircularVariableDependency` entirely (which is why you'll see a few extra files that needed to be updated). I also removed `substitutionToStepError` and `ctorSubstitutionCycleToBottom` since they don't seem to be needed anymore. This is now happening inside `normalizeSubstitution`.

I also had to update a couple of tests for `normalizeSubstitution` that used to assert constructor circular dependency and now return `Right []` instead. I feel these are okay; perhaps we should remove one of the two tests?

There is only one remaining test that has a `TODO`, the one that errors instead of returning bottom. Is that something that I need to tackle in this PR?

vladciobanu:
Updated again as per the Slack conversation. I also added type signatures everywhere, let me know if you think it makes it easier. I know that it would have been a lot easier for me to parse what this function does in its current state, but maybe that's just me.

I also added a few tests (whose names end with _"with variable"_) as per the previous comment. They result in slightly different errors than the ones without variables (I think it's `NonFunctionalPattern` instead of `PatternClash`, but still an `UnificationError`).

src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
```diff
@@ -0,0 +1,213 @@
...
+                ]
+                [   ( Mock.x
+                    , Mock.constr10 (Mock.f Mock.a)
+                    )
+                ]
+            )
+        )
+
+    , testCase "Constructor and constructor of function errors with variables"
+        -- [x=constructor(y)] + [x=constructor(f(y))]  === error
```
virgil-serbanuta: Could you also add a test with a functional symbol instead of f? Also, is the TODO below needed?

src/main/haskell/kore/test/Test/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -119,21 +119,65 @@ test_substitutionNormalization =
       in
         testCase "Simplest cycle"
             (assertEqual ""
-                (Left (CtorCircularVariableDependency [var1]))
+                (Right
+                    [    ( asVariable (v1 PatternSort)
```
virgil-serbanuta: If you really feel like it, you could make sure that normalization removes x=x assignments, but leaving a TODO in the normalization code is also fine.

vladciobanu:
Added the test for functional symbols (see https://github.com/kframework/kore/pull/212/files#diff-d0e47c7b43d4e1bf02352c1c8091d7a1R156 )

Also attempted fixing x=x assignments; see just this commit: https://github.com/kframework/kore/pull/212/commits/420de390dcf68d7a6abfac9b3547339538201cfc

traiansf:
Jenkins: add to whitelist

PR #215, opened 2018-09-13
Execution strategy

ttuegel:
This PR implements execution strategies. @bmmoore pointed out to me that trying to implement heating/cooling strategies is actually *wrong* if we encode heating and cooling as separate `\rewrites` rules without side conditions. I am not going to pursue strategies further at this time because that is how we currently encode strictness rules. Therefore, this PR only re-implements the existing stepper in terms of a simple strategy. Although it doesn't actually add any useful feature, I want to merge it before it falls behind because we will eventually want strategies.

There are two changes in this PR not directly related to strategies:

First, I added a class in `Control.Monad.Counter` which generalizes and replaces the `IntCounter` class. An earlier implementation of strategies needed to generate fresh variables in a more general monadic context, so it made sense to decouple variables from the monad that generates them. I still think that makes sense, and the generalization performs better anyway.

Second, I changed the representation of `StepProof` to ensure that it is always flat. Joining two `StepProof`s is now `O(log(n))` instead of `O(n)`. Joining `StepProof`s no longer associates to the right; because it did not freely associate, `Kore.Step.Step.pickFirstStepper` was not tail-recursive. It is now, or at least until I removed it.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Strategy.hs
```diff
@@ -0,0 +1,323 @@
...
+  @Strategy prim@ represents a strategy for execution by applying rewrite axioms
+  of type @prim@.
+
+ -}
+data Strategy prim where
+
+    -- The recursive arguments of these constructors are /intentionally/ lazy to
+    -- allow strategies to loop.
+
+    And :: Strategy prim -> Strategy prim -> Strategy prim
```
virgil-serbanuta: Could you consider describing these cases? Since you didn't export them, the style guide does not require it, but it may be useful, at least for some of them.

And and Or may be obvious-ish, but Step is less so, and Apply is fairly confusing at first sight (e.g. why does it take a 'prim' and a 'Strategy' to produce a 'Strategy'? Why isn't called 'Sequence' or 'ApplySequence' or something similar? Why aren't 'Apply' and 'Sequence' independent concepts?).

ttuegel: @virgil-serbanuta I added more documentation about `Strategy` which I believe addresses your questions here. However, I was not entirely sure if your questions were intended to direct me in writing documentation, or if you actually wanted to discuss the design here. Either way, I think the documentation addresses your questions, but I am happy to discuss more.

PR #221, opened 2018-09-18
design-decisions: Builtin Map

ttuegel:
@daejunpark 
This document summarizes my understanding of our discussion from Mon 17 Sep 2018. Could you have a look and tell me if anything is missing or incorrect? I would also appreciate if you and @traiansf could address the questions in the document regarding certain builtin operations.

###### Reviewer checklist

---


design-decisions/2018-09-17-Builtin-Map.md
```diff
@@ -0,0 +1,71 @@
...
+    symbol lookup{}(Map{}, K{}) : K{} []
+
+    // inKeys(map, key): Is key in map?
+    symbol inKeys{}(Map{}, K{}) : Bool{} []
+endmodule
+```
+
+Questions:
+
+- What does `update` do if the key is missing from the map?
```
daejunpark: @ttuegel @traiansf I think the `update(m,k,v)` should add the new entry `k |-> v` when `k` doesn't exist in `m`. In that sense, we may want to change the name to `insert` to capture this behavior.

What do you think?

design-decisions/2018-09-17-Builtin-Map.md
```diff
@@ -0,0 +1,71 @@
...
+- What does `concat` do if the key is present in both maps?
```
daejunpark: @ttuegel @traiansf I think `concat(m1,m2)` shouldn't be defined when there is a key in both `m1` and `m2`, like the separating conjunction `*` operator.

daejunpark:
Looks good to me. I put my answer for your question, to initiate the further discussion here.

design-decisions/2018-09-17-Builtin-Map.md
ttuegel: I agree that it makes sense for `update` to insert a missing key, but I don't want to change the name. If it had been named `insert` instead, I would have asked, what happens if the key is already in the map? :smile:

design-decisions/2018-09-17-Builtin-Map.md
```diff
@@ -0,0 +1,71 @@
...
+
+Eventually we will support both concrete and symbolic map operations.
+Once we support concrete operations, we can add support for symbolic operations
+by keeping symbolic constructions unevaluated, i.e. a symbolic map will be a
+pattern consisting of an unevaluated symbolic operation upon a symbolic or
+concrete map.
+Evaluating lookups from a symbolic map will require traversing the symbolic map
+pattern and attempting unification against each key.
+Looking up a symbolic key will traverse the entire map and will potentially be
+very expensive, so we will want to fall back to concrete lookup whenever
```
virgil-serbanuta: Insertions will be even more fun (shudder) (just a comment, not a request for changes).

ttuegel: I think we need to leave symbolic insertions unevaluated. @daejunpark and I discussed different ways to handle symbolic operations. The main tradeoff is between evaluating insertions or evaluating lookups; should we generate a large disjunction of terms when we do an insertion or when we do a lookup? We settled on evaluating lookups because we can throw out all the disjunction branches where the key does not unify, so there may be fewer branches to explore. On the other hand, the result of evaluating insertions is always the largest possible disjunction. I didn't put this much detail into the design document because this is all preliminary and tentative, but now I think I should update it with a note for whomever eventually implements symbolic operations.

PR #223, opened 2018-09-19
Design decision template

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/design-decisions/0000-00-00-Template.md
+++ b/design-decisions/0000-00-00-Template.md
@@ -1,0 +1,89 @@
+Decision Template
+=================
+
+Background
+----------
+
+The background section should contain the background information needed for
+understanding what is being decided.
+This is not supposed to be a burden to write or read, 1-2 paragraphs should be
+enough.
+
+If the decision changes the way we do something, please mention this here and
+describe the current way either here or in the "Other options" section.
+
+Example:
+
+    We are writing decision documents for each decision that we want to record.
+    These live in the `desing-decisions` directory, using file names with
+    the `yyyy-mm-dd-<dash-separated-title>.md` format, with the date
+    being roughly the date when the decision was made.
+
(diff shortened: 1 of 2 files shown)


bmmoore:
The format looks nice, but I think there should also be a template that's not already filled out with example text.

Your suggestion for obsolete files is unclear - is adding an `OBSOLETE` prefix supposed to override the earlier decision that filenames should start with the date? If so, that should be mentioned, if not then the word "prefixed" is not an adequate description of how the filename should be modified.

design-decisions/0000-00-00-Template.md
```diff
@@ -0,0 +1,100 @@
...
+This is not supposed to be a burden to write or read, 1-2 paragraphs should be
+enough.
+
+If the decision changes the way we do something, please mention this here and
+describe the current way either here or in the "Other options" section.
+
+Example:
+
+    We are writing decision documents for each decision that we want to record.
+    These live in the `desing-decisions` directory, using file names with
```
traiansf: `desing-decisions` => `design-decisions`

virgil-serbanuta:
@bmmoore I split this in two parts, the template itself, and the decision to have a template. I still found it useful to leave some text inside the template. Does that seem reasonable?

design-decisions/2018-09-20-Decision-Template.md
```diff
@@ -0,0 +1,64 @@
...
+
+Background
+----------
+
+We are writing decision documents for each decision that we want to record.
+These live in the `design-decisions` directory, using file names with
+the `yyyy-mm-dd-<dash-separated-title>.md` format, with the date
+being roughly the date when the decision was made.
+
+Currently (2018-09-19) the document format is unrestricted, but loosely
```
ttuegel: Now that this file has a date in the name, we probably don't need to specify the date in the text.

PR #226, opened 2018-09-21
Free form pattern matching - DO NOT REVIEW YET

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
+++ b/src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
@@ -1,0 +1,431 @@
+{-|
+Module      : Kore.Step.Function.Matcher
+Description : Matches free-form patterns for Equals rules.
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : virgil.serbanuta@runtimeverification.com
+Stability   : experimental
+Portability : portable
+-}
+module Kore.Step.Function.Matcher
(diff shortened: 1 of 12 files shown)


src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -0,0 +1,431 @@
...
+                (Ceil_ _ _ secondChild) ->
+                    match tools quantifiedVariables firstChild secondChild
+                _ -> Nothing
+        (CharLiteral_ _) ->
+            case second of
+                (CharLiteral_ _) ->
+                    if first == second
+                    then Just $ return topPredicateSubstitution
+                    else Nothing
+                _ -> Nothing
```
u--: You can just do if first == second then Just $ return topPredicateSubstitution else Nothing without the case second of

src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -0,0 +1,431 @@
...
+        (DV_ _ _) ->
+            case second of
+                (DV_ _ _) ->
+                    if first == second
+                    then Just $ return topPredicateSubstitution
+                    else Nothing
+                _ -> Nothing
```
u--: In fact, that comment applies to each of these. `first == second` can only be true if they have the same head constructor, so the `case second of` is always redundant.

PR #232, opened 2018-09-24
Smt pruning

u--:
###### Reviewer checklist

---


u--:
Re: whitespace, my bad, will fix.

Regarding the variable parameter, we already have proliferation of type parameters, I was sort of hoping that Simplification would not require `variable`. The alternative would be to extend `SortedVariable`, like you said, with the name of the variable as well. But these efforts to simulate row polymorphism in haskell never work out that well, and we are already at a point where many of type signatures are longer than the implementation. If Vlad is implementing `And` as unification, the `UnificationVariable`s should be restricted to that module, right? As I understand it, the only two special `Variable` types are `StepperVariable` and `UnificationVariable`, so they basically only exist within certain algorithms, and aren't passed around outside of them.

virgil-serbanuta:
In BaseStep.stepWithAxiom, when it calls unificationProcedure, axiomLeft and startPattern are based on StepperVariable, and unificationProcedure will pass that directly to the simplification code. I think that the simplification code directly involves only the code in AndTerms.hs (though I did not do a thorough check), but that code also calls other code (I think it calls code from Ceil.hs, maybe other code also). So, a few things:

1. It can be done, but there is a risk that it will create code that's messier and harder to change (e.g. we'd have to move things back to Variable if we add new dependencies to the unification code). If you think that's not the case, you might be right, but I'd like to know why.
2. If you still want to do the variable->Variable change, I'd prefer to have it in a separate PR, but I won't insist on it (though other reviewers might).
3. PureMLPattern level Variable is the same as CommonPurePattern level. ExpandedPattern level Variable is the same as CommonExpandedPattern level. The same for other types. We should probably use the Common version.

virgil-serbanuta:
For future reference, we're going to call the entire simplification code with step variables to solve special unification stuff (associative, commutative, list, map set).

PR #234, opened 2018-09-25
Add LIST builtins

ttuegel:
This pull request will add the LIST builtins needed for EVM. Documentation is also added for INT builtins.

###### Reviewer checklist

---

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -168,3 +168,59 @@ the result is `\dv{Bool{}}("false")`.
     hooked-symbol inKeys{}(Map{}, Key{}) : Bool{}
         [hook{}("MAP.in_keys")]
 ~~~
+
+## LIST
+
+Depends on `INT`.
+
+The sort of list elements is arbitrary, but must be consistent between symbol
+declarations. The sort consistency of hooked symbols is *not* checked. The
+element sort is referred to as `Elem{}` below.
+
+### LIST.List
+
+The builtin sort of associative lists.
+
+~~~
+    hooked-sort List{}
+        [hook{}("LIST.List")]
+~~~
+
+### LIST.unit
+
+The empty list.
+
+~~~
+    hooked-symbol unit{}() : List{}
+        [hook{}("LIST.unit")]
+~~~
+
+### LIST.element
+
+The singleton list.
+
+~~~
+    hooked-symbol element{}(Elem{}) : List{}
+        [hook{}("LIST.element")]
+~~~
+
+### LIST.concat
+
+Concatenate both arguments.
+
+~~~
+    hooked-symbol concat{}(List{}, List{}) : List{}
+        [hook{}("LIST.concat")]
+~~~
+
+### LIST.get
+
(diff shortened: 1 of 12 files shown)


src/main/haskell/kore/test/Test/Kore/Builtin/List.hs
```diff
@@ -0,0 +1,275 @@
...
+            , ExpandedPattern.top === evaluate predicate
+            ]
+
+{- |
+    @
+        get{}(concat{}(element{}(e), ...), 0) === e
+    @
+ -}
+prop_getFirstElement :: [Integer] -> Property
+prop_getFirstElement (Seq.fromList -> values) =
```
virgil-serbanuta: What does `Seq.fromList -> values` construct mean?

src/main/haskell/kore/test/Test/Kore/Builtin/List.hs
```diff
@@ -0,0 +1,275 @@
...
+    let patGet = App_ symbolGet [ patList , Test.Int.asPattern 0 ]
+        patList = asPattern (Test.Int.asPattern <$> values)
+        value =
+            case values of
+                Empty -> Nothing
+                v :<| _ -> Just v
```
virgil-serbanuta: What does the "expressionless face" emoticon (`:<|`) do? hoogle does not find it.

ttuegel: It is [`cons` on `Data.Sequence.Seq`](https://www.stackage.org/haddock/lts-12.10/containers-0.5.11.0/Data-Sequence.html#t:Seq). I should just import it qualified instead.

I was going to recommend using Stackage search instead of Hoogle, but both seem to miss pattern synonyms (probably a bug?).

src/main/haskell/kore/test/Test/Kore/Builtin/List.hs
ttuegel: That is a [view pattern](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#view-patterns); it applies `Seq.fromList` before binding `values`, so that `values :: Seq Integer` even though the first argument to `prop_getFirstElement` has type `[Integer]`. I should change the function signature.

docs/hooks.md
```diff
@@ -168,3 +298,59 @@ the result is `\dv{Bool{}}("false")`.
...
+~~~
+    hooked-symbol concat{}(List{}, List{}) : List{}
+        [hook{}("LIST.concat")]
+~~~
+
+### LIST.get
+
+Get an element from the list by index. Positive indices count from the beginning
+of the list and negative indices count from the end. The first element is
+`\dv{Int{}}("0")` and the last element is `\dv{Int{}}("1")`. The result is
```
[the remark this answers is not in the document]
ttuegel: It should be "last" and `\dv{Int{}}("-1")`!

PR #240, opened 2018-09-27
Kore in k

traiansf:
I've restarted the `kore-in-k` PR.  I would like to get this merged, but it seems that the latest changes to K make tests fail for the definition.  @hreada could you please take a look at it?

###### Reviewer checklist

---


nishantjr:
@traiansf @hreada There were a couple of problems: one is that the regular expressions for `MetaVariable` isn't working. I'm not sure what the issue was. The other is that the tests had updated to the new syntax for alias. Simply uncommenting some code fixed that. I've cleaned up the history and force pushed for now.

*We also need to get these tests integrated with the CI system.* There's probably something in the `stack.yaml` that needs changing for that, but I'm not sure how.

It'd be awesome if you could rebase instead of merging when trying to bring in the latest code on master. The history becomes much cleaner  to read and debug (tools like `git-bisect` are great for finding bugs). Besides that, when you resolve a conflict in a merge commit, the conflict resolution gets hidden away in the merge commit and gets overlooked easily.

PR #248, opened 2018-10-02
Outstanding issues for unification as simplification

vladciobanu:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
+++ b/src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
@@ -12,21 +12,21 @@ import Control.Monad.Counter
 import Data.Reflection
        ( give )
 
-import           Kore.AST.Common
-                 ( Variable )
-import           Kore.AST.MetaOrObject
-                 ( Object )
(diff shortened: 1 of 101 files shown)


src/main/haskell/kore/test/Test/Kore/Step/Substitution.hs
```diff
@@ -12,21 +12,21 @@ import Control.Monad.Counter
...
-import Kore.Predicate.Predicate
-       ( makeFalsePredicate, makeTruePredicate )
-import Kore.Step.ExpandedPattern
-       ( PredicateSubstitution (..) )
-import Kore.Step.Substitution
-       ( mergeAndNormalizeSubstitutions )
-import Kore.Unification.Error
-import Kore.Unification.Unifier
-       ( UnificationSubstitution )
+import           Kore.AST.Common
```
virgil-serbanuta: I think that, since you don't have a `qualified` in here, stylish-haskell, which we all use, will remove the spaces between `import` and the package name, adding more noise to code reviews. You should probably set your editor to follow what stylish-haskell does.

PR #249, opened 2018-10-02
Concrete collection keys

ttuegel:
The keys of builtin collections (Map and Set) are fixed to be concrete patterns, in the sense that they do not contain variables. The restriction to concrete patterns is made at the type level, which makes some more instances derivable for `DomainValue` and `BuiltinDomain`.

During builtin function evaluation, keys are also checked for normalization. Keys should consist only of constructors, `\dv` patterns, and sort injections, so that two keys are equal under `(==)` if and only if they are actually equivalent patterns.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/AST/Common.hs
+++ b/src/main/haskell/kore/src/Kore/AST/Common.hs
@@ -547,7 +547,11 @@ data DomainValue level child = DomainValue
     { domainValueSort  :: !(Sort level)
     , domainValueChild :: !child
     }
-    deriving (Eq, Generic, Ord, Show)
+    deriving (Eq, Foldable, Functor, Generic, Ord, Show, Traversable)
+
+deriveEq1 ''DomainValue
+deriveOrd1 ''DomainValue
+deriveShow1 ''DomainValue
 
 instance Hashable child => Hashable (DomainValue level child)
 
@@ -927,7 +931,7 @@ data Pattern level variable child where
     CeilPattern
         :: !(Ceil level child) -> Pattern level variable child
     DomainValuePattern
-        :: !(DomainValue Object (BuiltinDomain (Fix (Pattern Meta Variable))))
+        :: !(DomainValue Object (BuiltinDomain child))
         -> Pattern Object variable child
     EqualsPattern
         :: !(Equals level child) -> Pattern level variable child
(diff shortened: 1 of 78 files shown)


src/main/haskell/kore/src/Kore/AST/Common.hs
```diff
@@ -910,7 +931,7 @@ data Pattern level variable child where
     CeilPattern
         :: !(Ceil level child) -> Pattern level variable child
     DomainValuePattern
-        :: !(DomainValue Object (BuiltinDomain (Fix (Pattern Meta Variable))))
+        :: !(DomainValue Object (BuiltinDomain child))
```
virgil-serbanuta: It seems that, when I sent the previous reply, I somehow managed to lose the code comment I made.

I think that using `child` here is a step in the right direction, but it made me realize that using ConcretePurePattern for keys is kind of strange. When we do transformations/traversals on a pattern, we implicitly assume that they will go through all the patterns, so they should also go through the map keys / set entries. This might be safe with the current stuff we have, since many of those are variable-based, but I'm not sure of that.

Speaking of which, we should probably include these children when simplifying/substituting, not sure if that's included in the change below.

ttuegel: > I think that using `child` here is a step in the right direction, but it made me realize that using ConcretePurePattern for keys is kind of strange. When we do transformations/traversals on a pattern, we implicitly assume that they will go through all the patterns, so they should also go through the map keys / set entries.

Yes, I thought it was strange at first, but I think it's right. `ConcretePurePattern` only ensures that the keys are concrete, which is necessary, but not sufficient. The builtin function evaluators also check that the keys are _normalized_, which is the other condition required: `(==)` in Haskell must be `\equals` in matching logic for the internal representation to be valid. If the keys are concrete and normalized, I don't think any of our transformations would do anything to them. If that is not enough, then we may need to introduce a type for the keys that represents a single fully-normalized value.

If we do want to transform the keys, they will have to have their own type parameter in `Pattern`. The keys cannot depend on the `child` parameter: then they would have to be mapped by the `Functor` instance, which is impossible because of the `Ord` constraint. To have `child` in the keys, we would have to throw out the recursion-schemes approach entirely.

> Speaking of which, we should probably include these children when simplifying/substituting, not sure if that's included in the change below.

The children are automatically included in simplification because the `Functor` and `Traversable` instances include them! I did not implement a test for this explicitly, but the concrete map unification tests (in another branch) would not work without it. I will add tests specifically for simplification and substitution to this branch.

src/main/haskell/kore/src/Kore/Step/Simplification/DomainValue.hs
```diff
@@ -48,3 +74,110 @@ simplify dv =
...
+                Map.foldlWithKey' simplifyBuiltinDomainMapElement
+                    (pure (Map.empty, Predicate.makeTruePredicate, mempty))
+                    _map
+            let term =
+                    (asPurePattern . DomainValuePattern)
+                    DomainValue
+                        { domainValueSort
+                        , domainValueChild = BuiltinDomainMap _map
+                        }
+            OrOfExpandedPattern.make
```
virgil-serbanuta: The standard for the simplify functions which are similar to the ones in this file (Application and And) is that they `mergePredicatesAndSubstitutions` which also does some normalization and simplification. Arguably, we may want to change that at some point, and maybe only normalize them when that's actually needed, but, for now, it's likely that there are edge cases where simplification will not work properly without it.

src/main/haskell/kore/src/Kore/Step/Simplification/DomainValue.hs
```diff
@@ -48,3 +74,110 @@ simplify dv =
         ]
     , SimplificationProof
     )
+-}
```
virgil-serbanuta: Are there tests for the changes in this file (and not only this one)?

ttuegel:
@virgil-serbanuta Could you please re-review this?

PR #250, opened 2018-10-02
Smt pruning test

u--:
Forgot to include this file before.


src/main/haskell/kore/test/Test/Kore/Step/Condition/Evaluator.hs
```diff
@@ -0,0 +1,85 @@
...
+import           Kore.Step.ExpandedPattern
+import qualified Kore.Step.Condition.Evaluator as Eval
+import           Kore.Step.Simplification.Data
+
+import           Kore.Predicate.Predicate
+
+import qualified Kore.Builtin as Builtin
+
+import           Test.Kore.Builtin.Bool 
+  (boolDefinition, boolModuleName, boolSort)
```
ttuegel: This import list should be indented to the start of the module name above it.

src/main/haskell/kore/test/Test/Kore/Step/Condition/Evaluator.hs
```diff
@@ -0,0 +1,85 @@
...
+                    )
+                )
+                ( PredicateSubstitution 
+                    { predicate = makeFalsePredicate :: CommonPredicate Object
+                    , substitution = []
+                    }
+                )
+            )
+        ]
+    where a :: Given (SymbolOrAliasSorts Object) => CommonPurePattern Object
```
virgil-serbanuta: You're contributing enough code that your should actually take the time to learn our coding style.
https://github.com/kframework/kore/wiki/Coding-Standards
(see the half-indentation section for this specific example).

PR #251, opened 2018-10-02
Add Data.Result

ttuegel:
FunctionResult is factored out of Kore.Step.Simplification.AndTerms and
generalized in Data.Result to encompass success, failure, and
indeterminacy. Tests are added for the various instances.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Data/Result.hs
+++ b/src/main/haskell/kore/src/Data/Result.hs
@@ -1,0 +1,66 @@
+{-|
+Module      : Data.Result
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : thomas.tuegel@runtimeverification.com
+-}
+module Data.Result
+    ( Result (..)
+    , definite
+    , fromResult
+    ) where
+
+import           Control.Applicative
+                 ( Alternative (..) )
+import qualified Control.Monad as Monad
+
+{- | @Result a@ encompasses success, failure, and indeterminacy.
+
+    We may view @Failure@ as indicating a failed attempt, whereas @Unknown@
+    indicates that no attempt was made.
+ -}
+data Result a
+    = Success a  -- ^ successful result producing an @a@
+    | Failure  -- ^ unsuccessful result
+    | Unknown  -- ^ indeterminate result
+    deriving (Eq, Ord, Foldable, Functor, Read, Show, Traversable)
+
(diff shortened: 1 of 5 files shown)


src/main/haskell/kore/src/Data/Result.hs
```diff
@@ -0,0 +1,66 @@
...
+
+-- | Return the result, or the default value if not successful.
+fromResult
+    :: a  -- ^ default value
+    -> Result a  -- ^ result
+    -> a
+fromResult _a =
+    \case
+        Success _a -> _a
+        _ -> _a
```
15joeybloom: What do you think about
```
fromResult _ (Success _a) = _a
fromResult default _ = default
```
instead? This avoids the `_a` inside the first arm of the lambda case shadowing the first argument of the function.

src/main/haskell/kore/test/Test/Data/Result.hs
```diff
@@ -0,0 +1,74 @@
...
+
+prop_biasFailure :: Result Integer -> Property
+prop_biasFailure m = do
+    let r = Failure in (r <|> m) === r
+
+prop_leftUnitMonad :: Integer -> Fun Integer (Result Integer) -> Property
+prop_leftUnitMonad a (applyFun -> k) = (return a >>= k) === (k a)
+
+prop_rightUnitMonad :: Integer -> Fun Integer (Result Integer) -> Property
+prop_rightUnitMonad a (applyFun -> k) = (return a >>= k) === (k a)
```
15joeybloom: This is the same as the left unit property above. Maybe you meant to say
```
prop_rightUnitMonad :: Result Integer
prop_rightUnitMonad m = (m >>= return) === m
```

src/main/haskell/kore/test/Test/Data/Result.hs
```diff
@@ -0,0 +1,74 @@
...
+            Failure -> variant (1 :: Int)
+            Unknown -> variant (2 :: Int)
+
+prop_biasSuccess :: Integer -> Result Integer -> Property
+prop_biasSuccess a m =
+    let r = Success a in (r <|> m) === r
+
+prop_biasFailure :: Result Integer -> Property
+prop_biasFailure m = do
+    let r = Failure in (r <|> m) === r
```
virgil-serbanuta: How about testing `Unknown <|> m`?

PR #253, opened 2018-10-03
Make ExpandedPattern Applicative

u--:
###### Reviewer checklist

---

This has the potential to simplify a lot of code in `Simplification`. `ExpandedPattern` is now a pattern synonym, and all code works, so there's no rush to refactor.

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
+++ b/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
@@ -74,6 +74,9 @@ data Predicated level variable child = Predicated
 instance Functor (Predicated level variable) where
     fmap f (Predicated a p s) = Predicated (f a) p s
 
+-- `<*>` does not do normalization for now. 
+-- Don't use it until we figure that out.
+
 instance
   ( MetaOrObject level
   , SortedVariable variable
--- a/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
+++ b/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
@@ -11,9 +11,9 @@ Portability : portable
 -}
 module Kore.Step.ExpandedPattern
     ( CommonExpandedPattern
-    , CommonPredicateSubstitution  -- TODO(virgil): Stop exporting this.
-    , ExpandedPattern (..)
-    , PredicateSubstitution (..)  -- TODO(virgil): Stop exporting this.
+    , PredicateSubstitution.CommonPredicateSubstitution  -- TODO(virgil): Stop exporting this.
+    , ExpandedPattern
+    , PredicateSubstitution.PredicateSubstitution (..)  -- TODO(virgil): Stop exporting this.
     , allVariables
     , bottom
     , isBottom
@@ -23,6 +23,7 @@ module Kore.Step.ExpandedPattern
     , toMLPattern
     , top
     , fromPurePattern
+    , Predicated(..)
     ) where
 
 import           Data.List
@@ -49,7 +50,7 @@ import           Kore.Predicate.Predicate
                  makeAndPredicate, makeEqualsPredicate, makeFalsePredicate,
                  makeTruePredicate, unwrapPredicate )
 import qualified Kore.Predicate.Predicate as Predicate
-import           Kore.Step.PredicateSubstitution
+import qualified Kore.Step.PredicateSubstitution as PredicateSubstitution
                  ( CommonPredicateSubstitution, PredicateSubstitution (..) )
 import           Kore.Unification.Data
 import           Kore.Variables.Free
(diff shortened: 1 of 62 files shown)


src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
```diff
@@ -59,16 +60,40 @@ import           Kore.Variables.Free
...
+  , SortedVariable variable
+  , Show (variable level)
+  , Eq (variable level)
+  , Given (SymbolOrAliasSorts level)
+  )
+  => Applicative (Predicated level variable) where
+    pure a = Predicated a makeTruePredicate []
+    a <*> b = Predicated
+        { term = f x 
+        , predicate = fst (predicate1 `makeAndPredicate` predicate2)
```
virgil-serbanuta: @traiansf @u-- Doing this kind of joining instead of `mergePredicatesAndSubstitutions` is likely to be the right long-term solution, but, right now, we don't have the infrastructure to do the normalization part when it is actually needed. Our current code relies on things being normalized (not sure how much, exactly).

I don't think that `mergePredicatesAndSubstitutions` fits nicely in this instance. Should we try to add the on-demand normalization thing to this PR?

u--: I saw `mergePredicatesAndSubstitutions` and wasn't sure how to proceed from there. We can certainly add it using a `Given` constraint (which again will be completely nuked once we get proper AST annotations).

Or: can't we write a function `normalize`, such that `normalize (PredicateSubst p1 s1 `mappend` PredicateSubst p2 s2) = mergePredicatesAndSubstitutions [p1, p2] [s1, s2]`? Here `mappend` is the obvious monoid instance that combines the predicates with `mkAnd` and the substitutions with `++`.

u--: At any rate, since `ExpandedPattern` is now a type synonym and everything works as before, we can make these refactorings at our leisure.

u--: However, if we don't merge this now then the merge conflicts will keep piling up.

virgil-serbanuta: Ok, could you add a code comment saying that people should not use this until we have the infrastructure to handle on-demand normalization?

This is not necessarily related to this PR, but we may want to split the Predicated type in two - NormalizedPredicated and NotNormalizedPredicated or something like that.

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -347,8 +345,8 @@ addToolsArg
 addToolsArg = pure
 
 toExpanded
-    :: MetaOrObject level
-    =>  (  MetadataTools level StepperAttributes
+    :: (MetaOrObject level, SortedVariable variable, Show (variable level), Eq (variable level))
```
virgil-serbanuta: Why the one-liner? It goes over the 90 character limit, which seems to be the latest agreed hard limit (though, from the Slack discussions, it is not clear to me yet if we still want to fit things in 80 chars and only go to 90 in exceptional cases, or 90 is the only limit that matters).

src/main/haskell/kore/src/Kore/Step/Simplification/Application.hs
```diff
@@ -207,13 +205,13 @@ makeExpandedApplication tools symbol children
         , _proof) <-
             mergePredicatesAndSubstitutions
                 tools
-                (map ExpandedPattern.predicate children)
-                (map ExpandedPattern.substitution children)
+                (map predicate children)
```
virgil-serbanuta: You shouldn't use Predicate's getters unqualified (see the coding standards, naming, records).

The previous kind of imports, which split the ExpandedPattern in two, an unqualified import for the type and the constructor, and a qualified import for everything in the ExpandedPattern type, made sure that this unqualified access can't happen by mistake.

Of course, you don't have to import it like that, but you need to qualify these getters somehow.

Please fix this and the other getters in this pull request.

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
virgil-serbanuta: Sorry, it seems that the style guide does not contain anything about 90 chars, it's still 80 chars, expandable to 100 in rare cases where it's needed.

src/main/haskell/kore/src/Kore/Step/Simplification/Application.hs
u--: That's actually a really weird issue. The fact that the name `type ExpandedPattern` coincides with the module name makes it confused.

virgil-serbanuta: True. That's why I recommend (but do not enforce since the style guide does not require it) the splitted imports I mentioned above.

u--: I can still prefix it with `ExpandedPattern` instead of `Predicated`, which is misleading but w/e. I suppose in a future PR the module will be renamed `Predicated`.

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -609,9 +612,9 @@ equalInjectiveHeadsAndEquals
                 (map (ExpandedPattern.predicate . fst) children)
                 (map (ExpandedPattern.substitution . fst) children)
         return
-            ( ExpandedPattern
+            ( Predicated
                 { term = give (MetadataTools.symbolOrAliasSorts tools) $
-                    mkApp firstHead (map (ExpandedPattern.term . fst) children)
+                    mkApp firstHead (map (term . fst) children)
```
virgil-serbanuta: Not sure if you intended to fix everything with the previous commit, but it didn't include all unqualified usages.

u--:
So, Tom showed me how to squash and `push -f` to collapse long commit chains on a PR, but that command didn't seem to work for the merge commits. Before when this happened I would just manually copy and paste into a new branch, but this time I was trying to be a little more principled...

virgil-serbanuta:
For the future, you can usually rebase on master instead of merging (if you didn't do that before, I recommend making a backup of the branch the first few times).

PR #256, opened 2018-10-05
Added K as a submodule and created Makefiles

traiansf:
###### Reviewer checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,47 @@
+# Settings
+# --------
+
+BUILD_DIR:=$(CURDIR)/.build
+K_SUBMODULE:=$(BUILD_DIR)/k
+export K_BIN=$(K_SUBMODULE)/k-distribution/target/release/k/bin/
+
+.PHONY: all clean deps k-deps distclean clean-submodules kore-exec
+
+all: kore-exec
+
+test: kore-test k-test
+
+kore-test:
+	stack test --pedantic
+
+k-test: kore-exec k-deps
+	$(MAKE) -C src/main/k/working k-test
+
+kore-exec:
+	stack build kore:exe:kore-exec
+
+deps: k-deps
+
+distclean: clean
+	cd $(K_SUBMODULE) \
+		&& mvn clean -q
+	git submodule deinit --force -- ./
+
+clean: clean-submodules
+	stack clean
+	$(MAKE) -C src/main/k/working clean
+	rm -rf $(BUILD_DIR)/bin
+
+clean-submodules:
+	rm -rf $(K_SUBMODULE)/make.timestamp
+
+
+k-deps: $(K_SUBMODULE)/make.timestamp
+
+$(K_SUBMODULE)/make.timestamp:
+	@echo "== submodule: $@"
+	git submodule update --init -- $(K_SUBMODULE)
+	cd $(K_SUBMODULE) \
+		&& mvn package -q -DskipTests -U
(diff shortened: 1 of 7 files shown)


traiansf:
@ttuegel I've added you as a reviewer, since this would need to be called from Jenkins;  Therefore I wanted to get your opinion about it.

The `make` target to run the integration tests would be `k-tests`

Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+kore-tests:
+	stack test --pedantic
+
+k-tests: kore-exec k-deps
+	$(MAKE) -C src/main/k/working test
+
+kore-exec: $(BUILD_DIR)/bin/kore-exec
+
+$(BUILD_DIR)/bin/kore-exec:
```
ttuegel: Instead of installing `kore-exec` into a custom location, could we get the path from `stack path --local-install-root` instead? That would allow reusing any previous build and `stack clean` would have the desired effect in the `clean` target below.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	$(MAKE) -C src/main/k/working clean
+	rm -rf $(BUILD_DIR)/bin
+
+clean-submodules:
+	rm -rf $(K_SUBMODULE)/make.timestamp
+
+
+k-deps: $(K_SUBMODULE)/make.timestamp
+
+$(K_SUBMODULE)/make.timestamp:
```
ttuegel: What does this timestamp do?

src/main/k/working/Makefile
```diff
@@ -0,0 +1,13 @@
+TOPTARGETS := all clean test
+
+export KOMPILE:=$(K_BIN)/kompile
+export KORE_EXEC:=$(BUILD_DIR)/bin/kore-exec
```
ttuegel: Why not run `kore-exec` through Stack?

Makefile
traiansf: it marks that K was built.  I was inspired from the identical mechanism used by the `evm-semantics`

traiansf:
@ttuegel please re-review.  I've also changed the tests so that only one program is run for each definition, as running time seems to have increased of late.

ttuegel:
@traiansf Could you set up the Makefiles so that we can run individual tests? Right now, it only works if one runs `make k-tests` from the top level.

traiansf:
@dwightguth I think that when we decide to package the Haskell backend as part of K we should indeed make the Haskell backend a submodule of K and move these integration tests in the K repository.  I think that would be an easy change to make.  However, at the moment these tests break quite often, and most of the time it is because of changes in the haskell backend; therefore, I thought that, for the moment, they could stay here.  I'm willing to supervise moving them over to the K repo whenever the haskell backend becomes stable enough for distribution with K.

traiansf:
@ttuegel I thought I did, and I was working fine for the individual directories under `working`.  But indeed, if you ran it from the `working` directory itself, it would break.  Now it should work fine, with the observation that if it is run from the individual directories, it assumes by default that K is installed and on the path.  If K is not in the path, then one would need to set `K_BIN` either in the environment when calling make.  Does this seem reasonable?

ttuegel:
> Does this seem reasonable?

It does. I think this is the reason I had trouble getting the integration tests to work at first: I had the latest K from `master` installed system-wide, and it was using that instead of the submodule.

traiansf:
Jenkins: test this please

PR #257, opened 2018-10-05
Fix integration tests

vladciobanu:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/SMT/SMT.hs
+++ b/src/main/haskell/kore/src/Kore/SMT/SMT.hs
@@ -139,7 +139,7 @@ unsafeTryRefutePattern
     => PureMLPattern Object variable
     -> Maybe Bool
 unsafeTryRefutePattern p = unsafePerformIO $ do
-  let smtPredicate = setTimeOut 20 >> patternToSMT True p -- 20ms
+  let smtPredicate = setTimeOut 200 >> patternToSMT True p -- 20ms
         >>= (\case {
                Right p' -> return $ bnot p' ;
                Left _ -> sBool "TranslationFailed"
--- a/src/main/haskell/kore/src/Kore/SMT/SMT.hs
+++ b/src/main/haskell/kore/src/Kore/SMT/SMT.hs
@@ -19,9 +19,7 @@ module Kore.SMT.SMT
 where
 
 import           Control.Error.Util
-                 ( hush )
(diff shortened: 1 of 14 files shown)


src/main/haskell/kore/src/Kore/SMT/SMT.hs
```diff
@@ -225,9 +230,11 @@ patternToSMT sloppy p =
             => CommonPurePattern Object
             -> String
             -> Translating (SBV a)
-        goLiteral (DV_ s (BuiltinDomainPattern (StringLiteral_ i))) hookName
+        goLiteral p@(DV_ s (BuiltinDomainPattern (StringLiteral_ i))) hookName
          | (getHookString $ getSortHook s) == hookName
-             = return $ literal $ read i
+             = liftEither $ note
+                 (MalformedDVLiteral p)
+                 (literal <$> readMaybe i)
```
ttuegel: The domain value literals should be parsed using the appropriate parser in `Kore.Builtin.Int` or `Kore.Builtin.Bool`. I think @u-- is taking care of this.

In the long-term, all these builtins will be decoded into an `Integer` or `Bool` before they ever get to SMT.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -291,6 +291,18 @@ stepWithAxiom
...
         $ unificationProofStepVariablesToCommon
             existingVars variableMapping1 rawSubstitutionProof
+
+    let
+        variablesInLeftAxiom = pureAllVariables axiomLeftRaw
+        toVariable (AxiomVariable v) = v
+        toVariable (ConfigurationVariable v) = v
+        substitutions =
+            Set.fromList . fmap (toVariable . fst) $ normalizedSubstitution
+    if Predicate.isFalse condition
```
ttuegel: The first part of this condition, `Predicate.isFalse condition`, does not seem to be related to the error message below; is that right?

src/main/haskell/kore/src/Kore/SMT/SMT.hs
vladciobanu: The only reason for this change is to make the integration tests work again, for @traiansf 's PR. I would rather not do too big of a change, especially since @u-- has more changes coming up. This change was suggested by him, actually.

PR #267, opened 2018-10-11
Apply unit laws to builtin functions

ttuegel:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/List.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/List.hs
@@ -178,6 +178,7 @@ evalGet :: Builtin.Function
 evalGet =
     Builtin.functionEvaluator evalGet0
   where
+    ctx = "LIST.get"
     evalGet0
         :: MetadataTools Object StepperAttributes
         -> PureMLPatternSimplifier Object variable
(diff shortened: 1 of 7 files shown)


virgil-serbanuta:
Oh, and also: You might think about adding unit laws for all the builtins (e.g. ints), though that's less important.

@traiansf Alternately, we might add actual axioms for these simplifications, and use those instead of builtin specific stuff.

traiansf:
I agree with @virgil-serbanuta that eventually we should simplify these using axioms rather than handle them case by case.  However, I also agree with this PR, as it helps moving things along.  @ttuegel could you please add a TODO in the code and a Trello task for the general simplification problem.

ttuegel:
> This will solve the current problem, but it looks like a half-concrete case for lists might work

That's a good idea! Actually, a (different) semi-concrete representation could work for Map and Set, too.

> Alternately, we might add actual axioms for these simplifications, and use those instead of builtin specific stuff.

I was confused about this when we discussed axioms for builtins. It should fall on the frontend to provide these axioms, right? K gives `unit` and `element` attributes for the `concat` symbols, but it doesn't provide the corresponding axioms. Anyway, I agree that providing axioms would be a more principled solution.

src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -196,16 +210,41 @@ evalConcat :: Builtin.Function
...
             let (_list1, _list2) =
                     case arguments of
                         [_list1, _list2] -> (_list1, _list2)
-                        _ -> Builtin.wrongArity "LIST.concat"
-            _list1 <- expectBuiltinDomainList "LIST.concat" _list1
-            _list2 <- expectBuiltinDomainList "LIST.concat" _list2
-            returnList resultSort (_list1 <> _list2)
+                        _ -> Builtin.wrongArity ctx
+                leftIdentity = do
+                    _list1 <- expectBuiltinDomainList ctx _list1
```
vladciobanu: In this and `Set`, `Map`, etc.: is it a problem that we run `expectBuiltinDomainList` up to twice more than we need to in case it goes to the `bothConcrete` case?

ttuegel: No, `expectBuiltinDomainList` is just a wrapper to check for certain types of error, it doesn't actually do anything. It might be a little inefficient, but not much.

virgil-serbanuta:
> > Alternately, we might add actual axioms for these simplifications, and use those instead of builtin specific stuff.
> 
> I was confused about this when we discussed axioms for builtins. It should fall on the frontend to provide these axioms, right? K gives `unit` and `element` attributes for the `concat` symbols, but it doesn't provide the corresponding axioms. Anyway, I agree that providing axioms would be a more principled solution.

Yes, if we're going to use axioms in this case, then the frontend should provide them. With my comment I wasn't asking for an implementation, I was just wondering what would be the best thing to do here.

PR #268, opened 2018-10-12
Decision concerning configuration splits

traiansf:
I tried to sketch something regarding splitting configuration when encountering new constraints on variables from configuration.

Please demolish it, either here, or on slack, and please suggest improvements.

I would like to get approval on this from at least one of @grosu @hreada @bmmoore before I merge it.
That is, of course, after we converge to a solution we don't disagree upon.

###### Reviewer checklist

---


design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+### Example (from side conditions)
+
+Consider rewriting term `3 / A:Int` using 
+
+`rule X:Int / Y:Int => X /Int Y requires Y =/=Int 0`
+which would be written as 
+`Y =/= 0 = true /\ X:Int / Y:Int => X /Int Y`
+
+Then `3 / A:Int /\ (Y =/=Int 0 = true /\ X:Int / Y:Int)` will simplify to
+`(X:Int = 3 /\ Y:Int = A:Int) /\ (Y =/=Int 0) = true`
```
virgil-serbanuta: Not fully true, you forgot to add a `3 /Int A:Int` term in the simplification. And, since you're simplifying, the final result should actually be `3 /Int A:Int /\ A:Int =/= 0`.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+which is satisfiable, but not `T`.
+
+### Example (from unification)
+
+Consider rewriting term `if A:Int <Int 0 then -1 else 1` using
+
+`rule if true then E1:Exp else E2:Exp => E1`
+
+Then `if A:Int <Int 0 then -1 else 1 /\ if true then E1:Exp else E2:Exp` will
+simplify to `(E1:Exp = -1 /\ E2:Exp = 1) /\ A:Int <Int 0 = true`
```
virgil-serbanuta: Same comment as above.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+which is satisfiable, but not `T`.
+
+Problem
+-------
+
+Simply continuing the rewriting process when a constraint is introduced
```
virgil-serbanuta: s/constraint/constraint on a symbolic variable/, I guess.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+is not sound when the constraint does not simplify to top.  The reason is
+that, going further, the coverage is reduced for the inputs which satisfy
+the constraint.
+
+Questions
+---------
+
+* When/how we decide that a new constraint is satisfiable but not top?
```
virgil-serbanuta: s/we/do we/

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+prevents undesirable unrolling of functions.
+
+Decision: check whether new constraints are implied by existing ones
+--------------------------------------------------------------------
+
+To exhibit the problem described above, a new constraint must not be implied
+by the existing constraint, and their conjunction should be satisfiable but
+not top.  That is, supposing `CC` is the existing configuration constraint
+and `NC` is the new constraint, it must be that 
+* `~(CC -> NC)`, which is `CC /\ ~ NC`, is not unsatisfiable
```
virgil-serbanuta: not unsatisfiable means satisfiable, right?

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,87 @@
...
+* `CT /\ CC /\ (NC \/ \not(NC))`
+* `(CT /\ CC /\ NC) \/ (CT /\ CC /\ \not(NC))`
+
+We now have two configuration patterns. 
+
+* For the first configuration, retrying the application of the axiom should
+now succeed, as now `NC` is implied by `CC /\ NC`.
+* The second will carry the information that the inputs satisfying `NC`
+  are handled by a separate branch so they don't need to be considered
+  on this branch further.
```
virgil-serbanuta: I'd add a few words about special-casing this for evaluating functions with multiple branches.

design-decisions/2018-10-12-Condition-Split.md
traiansf: no. it means either satisfiable or unknown

virgil-serbanuta: Well, your explanation below contradicts that:
`'CC /\ NC' is not unsatisfiable (meaning 'CC /\ NC' has solutions)`
so maybe use different terms.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,96 @@
...
+```
+which is satisfiable, but not `T`.
+
+Problem
+-------
+
+Simply continuing the rewriting process when a new constraint is introduced
+(on some symbolic variable) is not sound when the constraint does not simplify
+to top.  The reason is that, going further, the coverage is reduced to the
+inputs which satisfy the constraint.
```
bmmoore: I don't think this is a good description of the problem, this whole thing only comes up as an issue if we are not handling evaluation in a sound way! If the rule above was the only applicable rule, then it *is true* that `((3 / A:int) /\ \next(E2:Exp)) <->((A:Int =/= 0)  = true /\ (E2 = 3t /Int A:Int)`.

The ideas in this proposal are ok as a means of trying to keep the configuration in an easily-understood form, but if soundness is ever at risk that's a warning that we are not following the principle of expressing everything in logically sound ways.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,96 @@
...
+
+* When/how do we decide that a new constraint is satisfiable but not top?
+
+* How about such constraints coming from function application?
+
+Decision: do not apply function axioms if they introduce new constraints
+------------------------------------------------------------------------
+
+We might revert this decision later, but for now this is a safe bet, as it
+prevents undesirable unrolling of functions.
```
bmmoore: Here, what does it even mean to "apply a function" if it introduces new constraints? Soundness shouldn't be at risk here. If it's just a policy for whether or not we try to unfold function applications it might be a reasonable policy. Are you making a distinction between "new constraints" added as a side condition and the constraints arising from unifying a non-trivial LHS with the configuration?
I think you probably mean to allow function applications if unification reports a successful pattern matching. But, if that's the intent you should write that explicitly - from a logical point of view I don't see any fundamental distinction between a pattern and a side condition, and regard unification success as just a special case of a rule stated as "only apply function axioms if new constraints are implied by existing ones".

Regarding soundness, given a current configuration `C[... f(X) ...] /\ P` applying a rule for `f` would be witnessed by a proof of something like `C[... f(X) ...]  /\ P = C[... rhs of rule ...] /\ P` saying the new expression for the configuration after unrolling the function is equal to the old. If the rule for `f` was instead `side-condition(X) /\ (f(X) = rhs of rule)` then we could only prove the bare equation `C[... f(X) ...] /\ P = C[... rhs of rule ...] /\ P` if we can prove that `P` implies `side-condition(X)`, otherwise we could only prove an equation like `C[... f(X) ...] /\ P = ((C[... rhs of rule ...] /\ P /\ side-condition(X)) \/ (C[... f(X) ...] /\ P /\ ~side-condition(X))`

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,96 @@
...
+
+Decision: check whether new constraints are implied by existing ones
+--------------------------------------------------------------------
+
+To exhibit the problem described above, a new constraint must not be implied
+by the existing constraint, and their conjunction should be satisfiable but
```
bmmoore: Isn't talking about the conjunction being top totally redundant? `CC /\ NC` can only be top if `CC -> NC` was true and `CC` was already `top, right?

design-decisions/2018-10-12-Condition-Split.md
bmmoore: I don't understand either. multiple overlapping step rules rules should probably lead to a case split, but I don't think we want to make a case split when a function application isn't ready to reduce.

design-decisions/2018-10-12-Condition-Split.md
hreada: Even if the above rule isn't the only applicable one, we can still claim a sound one-path step as follows.
* Grab the unification result `(X:Int = 3 /\ Y:Int = A:Int) /\ (A:Int =/=Int 0) = true`.
* Existentially quantify free variables in the rule, i.e., `X` and `Y`.
* Simplify. We get `(A:Int =/=Int 0) = true`.
* Construct the term after one-step rewriting `3 /Int A /\ (A:Int =/=Int 0) = true`.

And, it is indeed that
```
|- (exists X Y . (X / Y /\ Y =/=Int 0)) /\ (3 / A => 3 /Int A /\ (A:Int =/=Int 0) = true)
```

hreada: I think the problem here is the fact that in the above claim, we have `(exists X Y . (X / Y /\ Y =/=Int 0))` as part of the lhs of the one-step reachability claim. Intuitively, it means the set of all patterns where the rule
`rule X:Int / Y:Int => X /Int Y requires Y =/=Int 0`
is applicable. 

I guess what @traiansf wanted to point out is that, if these "constraints of applicability", denoted `NC`, are not equivalent to `\top`, then continuing the process will end up with the following claim:
```
|- NC1 /\ NC2 /\ ... /\ NCk /\ (CT /\ CC) => CT'
```
where the accumulative constraints, `NC1 /\ NC2 /\ ... /\ NCk`, may only cover a very small portion of the input space.

hreada: I'd say as long as `NC1 /\ NC2 /\ ... /\ NCk` is not unsatisfiable, then we're in a good shape, if one-path reachability is considered. The search/branch should be cut off if `NC1 /\ NC2 /\ ... /\ NCk` is unsatisfiable.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+
+By unrolling ◇  we obtain:
+```
+∀ X . φ(X) ->  ∃ Y . ψ(X, Y) ∨ ○ ◇ ∃ Y . ψ(X, Y)
+```
+
+We can now move `∃ Y . ψ(X, Y)` to the right of the implication,
+and (assuming law of excluded middle) we obtain the equivalent:
+```
+(∀ X . (φ(X) ∧ ¬∃ Y . ψ(X, Y)) -> ○ ◇ ∃ Y . ψ(X, Y))
```
hreada: typo? `(∀ X . ((φ(X) ∧ ¬∃ Y . ψ(X, Y)) -> ○ ◇ ∃ Y . ψ(X, Y)))`

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+```
+for some `φ'` determined by the algorithm.
+
+It is then sound to replace the first conjunct in the goal above with
+`∀ X . φ'(X) -> ψ(X)`.  That is indeed so, because
+
+`∀ X . φ'(X) -> ψ(X)` implies `∀ X . ○ φ'(X) -> ○ ψ(X)` which,
+by using `(Step)` and implication transitivity, implies that
+```
+φ(X) ∧ (∃ Y . α(Y)) -> ○ \psi(X)
```
hreada: `ψ(X)`

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+where `X` and `Y` are sets of variables.
+
+### Reduction to proving a ○ implication
+
+By unrolling ◇ we obtain:
+```
+∀ X . φ(X) ->  ∃ Y . ψ(X, Y) ∨ ○ ◇ ∃ Y . ψ(X, Y)
+```
+
+We can now move `∃ Y . ψ(X, Y)` to the right of the implication,
```
bmmoore: to the *left* of the implication

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+
+Then `if A:Int <Int 0 then -1 else 1 ∧ if true then E1:Exp else E2:Exp` will
+simplify to
+```
+if A:Int <Int 0 then -1 else 1 ∧ (E1:Exp = -1 ∧ E2:Exp = 1)
+                               ∧ A:Int <Int 0 = true
+```
+which is satisfiable, but not `T`.
+
+Problem
```
bmmoore: What is the Problem even? This goes on to describe the reachability, then one step execution, then propose an algorithm, then examples, and I still don't see any statement of what is actually the problem this is trying to solve.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+### Proving `∀ X . φ(X) -> ○ ψ(X)`
+
+We use just `ψ(X)` for simplicity.  However, note that `ψ(X)` could be
+of the form `◇ ∃ Y . ψ(X, Y)` encountered above.
+
+Now, say the definition contains an axiom of the form
+`∀ Y . α(Y) -> ○ β(Y)`.
+
+We can now split the original goal onto `∃ Y. α(Y)` to obtain
+the equivalent goal:
```
bmmoore: "split onto"?

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+It is then sound to replace the first conjunct in the goal above with
+`∀ X . φ'(X) -> ψ(X)`.  That is indeed so, because
+
+`∀ X . φ'(X) -> ψ(X)` implies `∀ X . ○ φ'(X) -> ○ ψ(X)` which,
+by using `(Step)` and implication transitivity, implies that
+```
+φ(X) ∧ (∃ Y . α(Y)) -> ○ ψ(X)
+```
+ 
+Decision: Reduce premise when unrolling `◇ `
```
bmmoore: What premise is being reduced?

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+--------------------------------------------
+That is, replace the goal
+```
+∀ X . φ(X) -> ◇ ∃ Y . ψ(X, Y)
+```
+by the goal
```
bmmoore: When do we make this replacement? Any time we encounter a ◇? Without a clear problem statement, I don't see what question this is even a decision about. Are there alternatives? Grounds for making this decision?

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+(∀ X . (φ(X) ∧ ¬∃ Y . ψ(X, Y)) -> ○ ◇ ∃ Y . ψ(X, Y))
+```
+
+Decision: Split on matching when using rule to advance one-path proof
+---------------------------------------------------------------------
+That is, replace the goal
+```
+∀ X . φ(X) -> ○ ψ(X)
+```
+by the goal(s)
```
bmmoore: This section looks a lot like a definition of what it even means to try to "use a rule to advance a proof", rather than a decision about any clear question. I think the interesting question is more how we decide what rules to try to use.

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+-------
+
+Consider the following "one-path" reachability problem
+
+```
+∀ X . φ(X) -> ◇ ∃ Y . ψ(X, Y)
+```
+where `X` and `Y` are sets of variables.
+
+### Reduction to proving a ○ implication
```
bmmoore: Why is this stuff here, and with a bigger heading than the one-step stuff below? I think it's the one-step problem which is more general, and more exactly the problem being addressed. This looks like it's still background explaining why we care about the

design-decisions/2018-10-12-Condition-Split.md
```diff
@@ -0,0 +1,159 @@
...
+and (assuming law of excluded middle) we obtain the equivalent:
+```
+(∀ X . (φ(X) ∧ ¬∃ Y . ψ(X, Y)) -> ○ ◇ ∃ Y . ψ(X, Y))
+```
+
+#### Basic Symbolic Execution
+
+**Input:** `φ(X)`, `∀ Y .α(Y) -> ○ β(Y)`
+
+**Output:** a pattern `φ'(X)` and satisfying the same assumptions we put on
```
bmmoore: What assumption on φ? No assumptions have been given about the form of any formulas so far?

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+```
+∀ X . φ(X) → ◆ ∃ Y . ψ(X, Y)
+```
+where `X` and `Y` are sets of variables and ◆ denotes "strong-eventually"
+
+Consider also the axiom
+
+`∀ Z . α(Z) → ⚪ β(Z)`.
+
+where ⚪  denotes "strong-next".
```
virgil-serbanuta: The `applying-axioms.md` document uses a full circle for strong-next, why not use a consistent notation?

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+We can now split the original goal onto `∃ Y. α(Y)` to obtain
+the equivalent goal:
+
+```
+(∀ X . Φ(X) ∧ (∃ Z. α(Z)) → ⚪ Ψ(X))
+∧
+(∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)) → ⚪ Ψ(X))
+```
+
+Now we know, from the basic symbolic execution algorithm that
```
virgil-serbanuta: Could you clarify which algorithm? The one in the `Algorithm for applying one rule to a pattern` document? Another one?

Also, by default, I would expect to apply the symbolic execution algorithm to φ, not to Φ. In particular, even if φ is functional, it's not obvious that Φ is functional. Could you clarify that?

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+```
+Φ(X) ∧ (∃ Y . α(Y)) → ⚪ Ψ(X)
+```
+
+
+Therefore, it is sound to replace the original goal with 
+```
+(∀ X . Φ'(X) → ◆ ∃ Y . ψ(X, Y))
+∧
+(∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)) → ⚪ ◆ ∃ Y . ψ(X, Y))
```
virgil-serbanuta: You could move ψ out of Φ to get something that resembles the original goal more, since this part is supposed to be unchanged except for the `¬ ∃ Z. α(Z)` part. If that's also possible for the first conjunct, you should also do that.

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+using axiom
+```
+forall X:Int, Y:Int .  (Y =/=Int 0 = true) \and X / Y -> \next X /Int Y
+```
+
+Then the goal will simplify to
+```
+(\forall A:Int . 3 /Int A \and (A =/=Int 0 = true)-> \diamond \exists B:Int . B)
```
virgil-serbanuta: I think it would be nice to have a form where Φ is visible, which then simplifies to the form that you have here, which uses φ.

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+### Example (from side conditions)
+
+Consider proving that
+```
+forall A:Int . 3 / A -> \diamond \exists B:Int . B
+```
+
+using axiom
+```
+forall X:Int, Y:Int .  (Y =/=Int 0 = true) \and X / Y -> \next X /Int Y
```
virgil-serbanuta: Above you're using unicode for math symbols. For consistency, you could also use it here, it should be a simple search-and-replace.

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+The first conjunct can be discharged immediately, leaving the second
+one as a goal.
+```
+(\forall A:Int . 3 / A \and \not (A =/=Int 0 = true) -> \next \diamond \exists B:Int . B
+```
+
+
+### Example (from unification)
+
+Consider proving goal
```
virgil-serbanuta: s/goal/the goal/

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+\forall E1:Exp, E2:Exp . if true then E1 else E2 -> \next E1
+```
+and
+```
+\forall E1:Exp, E2:Exp . if false then E1 else E2 -> \next E2
+```
+
+Then, by using the first axiom, the goal will simplify to
+```
+(\forall A:Int . -1 \and A <Int 0 = true -> \diamond (-1 \or 1))
```
virgil-serbanuta: You're not using Φ here, and it's not obvious to me that ψ can be simplified.

docs/applying-axioms.md
```diff
@@ -0,0 +1,48 @@
+Algorithm for applying one rule to a pattern
+============================================
+
+**Input:** `φ(X)`, `∀ Y .α(Y) -> ◍ β(Y)`
+
+Note: ◍ is strong next in this document
```
bmmoore: Why that character? I have never seen a circle full of vertical stripes for that. Unicode has white and black circles in a few sizes: `⚪⚫○●`. Those are `MEDIUM? (BLACK|WHITE) CIRCLE`. (Maybe you looked only for character names starting with CIRCLE).

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+given axiom.
+
+Unrolling "strong-eventually"
+------------------------------
+
+Let `Δ = ν f . ⚪ f` denote the formula for diverging computation.
+Then,  for any formula ψ:
+
+- `◆ ψ = ◇ ψ ∨ Δ` by the definition of "strong eventually"
+- `    = ψ ∨ ⚪ ◇ ψ ∨ Δ` by unrolling `◆`
```
bmmoore: The previous equation didn't have a `◆`. Maybe this unrolled `◇`?

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+- `    = ψ ∨ ⚪ ◇ ψ ∨ ⚪ Δ` by unrolling `Δ`
+- `    = ψ ∨ ⚪ (◇ ψ ∨ Δ)`
+  because `⚪` commutes with `∨` 
+- `    = ψ ∨ ⚪ ◆ ψ`
```
bmmoore: I would prefer an argument that avoids the non-constructive case split `◆ ψ = ◇ ψ ∨ Δ` if you have one.

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+Therefore proving
+```
+∀ X . φ(X) → ◆ ∃ Y . ψ(X, Y)
+```
+is equivalent to proving
+```
+∀ X . φ(X) →  ∃ Y . ψ(X, Y) ∨ ⚪ ◆ ∃ Y . ψ(X, Y)`
+```
+
+We can now move `∃ Y . ψ(X, Y)` to the right of the implication,
```
bmmoore: Isn't that moving to the left side?

docs/one-path-reachability-proofs.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+Applying the axiom `∀ Z . α(Z) → ⚪ β(Z)`
+------------------------------------------
+
+In the following let `Φ(X) = φ(X) ∧ ¬∃ Y . ψ(X, Y)`
+and `Ψ(X) = ◆ ∃ Y . ψ(X, Y)`
+ 
+
+We can now split the original goal onto `∃ Y. α(Y)` to obtain
```
bmmoore: This should be "split on", or perhaps "split over", not "split onto".

traiansf:
I think it got better now (also including Virgil's design doc on And-simplification).  @bmmoore @virgil-serbanuta please re-review.

docs/2018-11-08-One-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,220 @@
...
+- `    = ψ ∨ •◆ ψ`
+
+
+Therefore proving
+```
+∀ X . φ(X) → ◆ ∃ Y . ψ(X, Y)
+```
+is equivalent to proving
+```
+∀ X . φ(X) →  ∃ Y . ψ(X, Y) ∨ •◆ ∃ Y . ψ(X, Y)`
```
virgil-serbanuta: Do you need the backtick at the end of the line?

PR #269, opened 2018-10-12
Unify builtin lists

u--:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/List.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/List.hs
@@ -27,6 +27,9 @@ module Kore.Builtin.List
     , lookupSymbolElement
     , lookupSymbolConcat
     , lookupSymbolGet
+    , isSymbolConcat
+    , isSymbolElement
+    , unify
     ) where
 
 import           Control.Monad.Except
@@ -37,31 +40,40 @@ import qualified Data.HashMap.Strict as HashMap
 import           Data.Map.Strict
                  ( Map )
 import qualified Data.Map.Strict as Map
+import           Data.Reflection
+                 ( give )
+import           Data.Result
 import           Data.Sequence
-                 ( Seq )
+                 ( Seq, (><) )
 import qualified Data.Sequence as Seq
 
+import           Kore.AST.Common
 import qualified Kore.AST.Common as Kore
 import           Kore.AST.MetaOrObject
-                 ( Object )
+import           Kore.ASTUtils.SmartPatterns
 import qualified Kore.ASTUtils.SmartPatterns as Kore
 import qualified Kore.Builtin.Builtin as Builtin
+import           Kore.Builtin.Hook
+                 ( Hook )
 import qualified Kore.Builtin.Int as Int
 import qualified Kore.Error as Kore
 import           Kore.IndexedModule.IndexedModule
                  ( KoreIndexedModule )
 import           Kore.IndexedModule.MetadataTools
-                 ( MetadataTools )
+                 ( MetadataTools (..) )
+import           Kore.Predicate.Predicate
 import           Kore.Step.ExpandedPattern
-                 ( ExpandedPattern )
 import qualified Kore.Step.ExpandedPattern as ExpandedPattern
 import           Kore.Step.Function.Data
-                 ( AttemptedFunction (..) )
 import           Kore.Step.Simplification.Data
-                 ( PureMLPatternSimplifier, Simplifier )
 import           Kore.Step.StepperAttributes
                  ( StepperAttributes )
-
+import qualified Kore.Step.StepperAttributes as StepperAttributes
+import           Kore.Step.Substitution
+                 ( normalize )
+import           Kore.Substitution.Class
+                 ( Hashable )
+import           Kore.Variables.Fresh
 
 {- | Builtin variable name of the @List@ sort.
  -}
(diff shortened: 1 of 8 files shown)


src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -288,3 +301,93 @@ lookupSymbolGet
...
+      (h1 :<| t1, h2 :<| t2) -> do
+        h12Unified <- simplifyChild h1 h2
+        let list1Tail = DV_ dvSort $ BuiltinDomainList t1
+        let list2Tail = DV_ dvSort $ BuiltinDomainList t2
+        t12Unified <- unify tools simplifyChild list1Tail list2Tail
+        return $ do
+            (h12Unified', _) <- h12Unified
+            (t12Unified', _) <- t12Unified
+            res <- 
+              normalize tools $
```
virgil-serbanuta: Could you make your indentation a multiple of 4 throughout this entire PR, except in the cases allowed by the style guide?

src/main/haskell/kore/src/Kore/Step/Substitution.hs
```diff
@@ -52,6 +54,31 @@ import           Kore.Unification.UnifierImpl
...
+     , OrdMetaOrObject variable
+     , ShowMetaOrObject variable
+     ) 
+  => MetadataTools level StepperAttributes 
+  -> ExpandedPattern level variable 
+  -> m (ExpandedPattern level variable)
+normalize tools r =
+    runExceptT (normalizePredicatedSubstitution tools r)
+        >>= \case
+            Left _ -> return ExpandedPattern.bottom
```
virgil-serbanuta: Bottom means unification is impossible, Left means we have no clue how to unify, but it might work. So, if you don't know how to unify your lists/list elements (i.e. the error case) you should also return an error (the other option is making a predicate out of that unification).

Please delete `normalize` and use `normalizePredicatedSubstitution` directly.

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -599,6 +599,45 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
                     (Mock.builtinMap [(Mock.aConcrete, fOfA), (Mock.bConcrete, fOfB)])
                 )
         )
+    , testCase "builtin List domain"
```
virgil-serbanuta: Could you add test cases for:
* Unification returning a non-empty substitution
* Unification that returns an error (unifying `Mock.plain10(Mock.x)` with `Mock.a` should fail)
* Unification that merges two substitutions for the same variable, discovering a new substitution for another variable (`Mock.x=Mock.functionalconstr10(Mock.a)` with `Mock.x=Mock.functionalconstr10(Mock.y)`)

src/main/haskell/kore/src/Kore/Step/Substitution.hs
u--: `normalize` was already used in Tom's Map PR, I just moved it to a different spot. So far as I know, my use of it in `Builtin.List.unify` is as correct as the use in `Builtin.Map.unify`

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
u--: On it

u--: Well, I'm not sure if those fit the definition of a unit test, which is the smallest possible test of the new functionality. What I'm trying to do right now is add a unit test for list builtins that use a hooked "concat" symbol.

src/main/haskell/kore/src/Kore/Step/Substitution.hs
virgil-serbanuta: Ok, I'll let @ttuegel decide what should be done with this 'normalize' issue.

u--: I'm not sure I understand what the issue is. `normalizePredicatedSubstitution` returns `Left` if it doesn't know how to normalize a list of substitutions. That list of substitutions is itself correct and unique for a given unification (sub-) problem.

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
virgil-serbanuta: Ok, fine. Then, the following seem part of the list unification functionality, please unit-test all of them as you see fit.
* List unification can return a non-empty substitution.
* List unification returns an error because child unification fails.
* List unification merges child substitutions.
* List unification normalizes the merged child substitution.

u--: Oh, you're saying basically to test that all possible results of unification "pass through" the list unification. This is precisely why we try to use combinators:

- List unification returns an error because child unification fails.

Should be automatic from use of the Result monad.

- List unification merges child substitutions.
- List unification normalizes the merged child substitution.

Should be automatic from the use of ExpandedPattern's `Applicative` instance, though that's not 100% fixed yet.

u--: Well, there already is one test that tries to unify `[a, a]` with `[a, b]` and expects an error. Technically that isn't even the smallest "unit" to test, `[a] /\ [b]` is...

src/main/haskell/kore/src/Kore/Step/Substitution.hs
virgil-serbanuta: Let me try to explain it this way:

Let phi and psi be two patterns that we can't unify (e.g. `Mock.plain10(Mock.a)` and `Mock.b`).

If someone calls `normalize [x=phi, x=psi]`, and normalize returns bottom, the caller concludes that there is a contradiction in the substitution, so it should give up on whatever it is doing and return bottom (or, if that happened in a `not`, it will return `top`; or something like that). Note that returning bottom is a perfectly valid thing and it should happen for, say, `[x=constr1(a), x=constr2(a)]`.

Now, if `normalize` returns an error, then the caller can postpone normalizing the substitution (by transforming it into a predicate) for a later time, when more information is available. Note that we actually want to do this when, say, postponing a map lookup when the key is a variable, but we probably want to do this in other cases, also.

This difference is also relevant when evaluating functions defined on multiple branches (e.g. signum). For the moment, most likely, we will not evaluate functions on symbolic variables except in some very narrow cases, but, in the future, we may want to know that one of the cases cannot apply to our variables (i.e. bottom), that it might apply (error), or that it can be applied when a certain predicate is satisfiable (normal return value).

virgil-serbanuta: So, if `normalize` returns bottom instead of Left, or if the error is not propagated above the List/Map unification, we can't make that difference anymore, and we can't postpone evaluating functions.

u--: Hmm. I honestly am not sure what the correct behavior is then. I'll have to ask Tom on monday whether his use of `normalize` is correct.

ttuegel:
Jenkins is failing because some files are not up-to-date with respect to the style guide:
```
./scripts/check-stylish-haskell.sh \
	src/main/haskell/kore/app \
	src/main/haskell/kore/src \
	src/main/haskell/kore/test
Found dirty files:
 M src/main/haskell/kore/src/Kore/Builtin/List.hs
 M src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
 M src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
 M src/main/haskell/kore/src/Kore/Step/Substitution.hs
 M src/main/haskell/kore/test/Test/Kore/Step/MockSymbols.hs
 M src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
ttuegel: I don't think list unification should propagate the `\bottom` from a failed child to the top. The simplifier already does that, and I don't think we should implement the same functionality twice. (I'm adding tests for it in #274.)

src/main/haskell/kore/src/Kore/Step/Substitution.hs
ttuegel: I don't understand what `Left` from `normalize` means; is it an error or not? By "error", I mean "unrecoverable", as opposed to "exception".

PR #274, opened 2018-10-15
Simplification.DomainValue: Propagate \bottom through domain values

ttuegel:
This actually just adds tests because `MultiOr` already propagates `\bottom`. However, I have started to implement this functionality twice before I realized that, so I am making a test and leaving notes for myself in case I start trying to do that again.

###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Kore/Step/Simplification/DomainValue.hs
```diff
@@ -60,7 +64,33 @@ test_domainValueSimplification =
...
+            (OrOfExpandedPattern.make [])
+            (evaluate
+                mockMetadataTools
+                (DomainValue
+                    testSort
+                    (BuiltinDomainMap (Map.fromList [(Mock.aConcrete, bottom)]))
+                )
+            )
+        )
+    , testCase "\\bottom propagates through builtin List"
```
virgil-serbanuta: Why not a test for sets?

ttuegel: Set keys are concrete values, so they cannot contain `\bottom`.

PR #278, opened 2018-10-16
Kore.Variables.Free: Replace `pureFreeVariables` with `freePureVariables`

ttuegel:
The new function `freePureVariables` is a simple traversal down the tree,
tracing the bound variables and recording those that occur free.

Note: The implementation is slightly different than I outlined in the associated Trello card, but it is morally the same.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Unification/SubstitutionNormalization.hs
+++ b/src/main/haskell/kore/src/Kore/Unification/SubstitutionNormalization.hs
@@ -22,9 +22,11 @@ import           Data.Foldable
                  ( traverse_ )
 import           Data.Functor.Foldable
                  ( Fix, cata )
-import qualified Data.Map as Map
-import           Data.Maybe
-                 ( mapMaybe )
+import           Data.Map.Strict
+                 ( Map )
+import qualified Data.Map.Strict as Map
+import           Data.Set
+                 ( Set )
 import qualified Data.Set as Set
 
 import           Data.Graph.TopologicalSort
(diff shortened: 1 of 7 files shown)


src/main/haskell/kore/src/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -208,42 +213,22 @@ normalizeSortedSubstitution
...
-    deps =
-        mapMaybe
-            (`Map.lookup` interestingVariables)
-            (Set.toList (freeVariables patt))
-    isSameVar = case patt of
-        (Var_ v) -> v == var
-        _        -> False
-    deps' = if isSameVar then [] else deps
+    => Set (variable level)
+    -> variable level -> PureMLPattern level variable
```
vladciobanu: Is there a reason for the two arguments on the same line?

src/main/haskell/kore/src/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -208,42 +213,22 @@ normalizeSortedSubstitution
...
+    -> Set (variable level)
+getDependencies interesting var =
+    \case
+        Var_ v | v == var -> Set.empty
+        patt -> Set.intersection interesting (freePureVariables patt)
```
vladciobanu: Wow, nice refactor here. Big :+1:

src/main/haskell/kore/src/Kore/Variables/Free.hs
```diff
@@ -11,20 +11,58 @@ Portability : portable
...
                  ( Fix, cata )
+import qualified Data.Functor.Foldable as Functor.Foldable
+import           Data.Set
+                 ( Set )
 import qualified Data.Set as Set
 
 import Kore.AST.Common
 import Kore.AST.MetaOrObject
 
+freePureVariables
```
vladciobanu: I was going to comment on using different algorithms for very similar things, but I assume https://trello.com/c/H3VxRw12/294-investigate-a-different-approach-for-freevariables would fix it? In that case, I would only mention that this function needs some docs.

ttuegel: The card you linked is actually what this pull request does; the card for the other is [here](https://trello.com/c/x2OwZsls).

PR #280, opened 2018-10-17
Ensure that branch is rebased onto master

ttuegel:
When I updated the Jenkins builder a few days ago, I also wanted to check that pull request history is linear on top of master, i.e. rebased and not merged. At that time the check was failing, but I did not understand why!

This pull request reinstates the linear-history check. This pull request will fail to build (at least the first time) until I change the Jenkins configuration. That change is backwards-compatible.

###### Reviewer checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,12 +44,19 @@ clean: clean-submodules
 	$(MAKE) -C src/main/k/working clean
 
 check:
-	if ! ./scripts/git-assert-clean.sh; then \
+	if ! ./scripts/git-assert-clean.sh; \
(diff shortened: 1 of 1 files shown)


ttuegel:
> However, one question that I had was what happens if after the PR passes tests and gets accepted, another PR gets merged before it.

If Jenkins was asked to re-test the PR, the check would fail. However, GitHub already would not allow merging the PR because master was updated, so the PR author would need to update their PR; at that point, Jenkins would re-test the updated PR anyway.

PR #284, opened 2018-10-18
Search in kore-exec

ttuegel:
The search function in `krun` is implemented.

@traiansf This would update the K submodule to your `haskell-search` branch, which is not merged into K. Therefore, this pull request is not technically ready to merge until that is ready.

###### Reviewer checklist

---


ttuegel:
@ehildenb I would also appreciate your feedback on this pull request, since we were discussing the search feature not long ago.

src/main/k/working/search/tests/initial.star.initial.output.golden
```diff
@@ -0,0 +1,3 @@
+VarDotVar0 ==K <generatedCounter>
```
traiansf: I was confused by this at first, but I think this is due to the fact that your `pattern` gets completed to a full configuration to allow matching on it.  I wonder whether it would be a good idea to eliminate `DotVar`s from the substitution before we convert it back to a pattern.   Not necessarily in this PR, though :-)

src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -334,17 +358,75 @@ main = do
...
+                            searchFileName
+                    clockSomething "Searching"
+                        $ evalSimplifierWithTimeout smtTimeOut
+                        $ do
+                            axiomsAndSimplifiers <-
+                                makeAxiomsAndSimplifiers indexedModule metadataTools
+                            let
+                                (rewriteAxioms, simplifier, substitutionSimplifier) =
+                                    axiomsAndSimplifiers
+                                runStrategy' pat =
```
traiansf: Really nice job refactoring!  But, while you're at it, the definition of `runstrategy'` here seems identical to that of `computeStrategy` above.  Do you think it would make sense to extract it?

src/main/haskell/kore/src/Kore/Step/Strategy.hs
```diff
@@ -63,117 +66,97 @@ data Strategy prim where
...
+    Continue :: Strategy prim
 
-     -}
-    Step :: Strategy prim -> Strategy prim
+    deriving (Eq, Show)
+
+{- | Apply two strategies in sequence.
 
-    {- | Terminate execution; the end of all strategies.
+The first strategy is applied, then the second is applied all its children.
```
virgil-serbanuta: s/all/to all/?

PR #295, opened 2018-10-24
Pure-equality function axioms, recurse for axioms, and IMP with Map

ttuegel:
This pull request allows Kore to use `\equals` axioms without pre- or post-conditions as function evaluators. While implementing this, I discovered that we only check the top-most module for function axioms; I correct that in this pull request. Finally, I added an integration test for IMP using Map instead of State. It does not actually do unification on Map yet, but it does require function axioms for `append` and that depends on the first two changes.

@dfilaretti The contents of this pull request should help you get past the issue you were having with Set unification.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -756,3 +762,29 @@ internalIndexedModuleSubsorts imod = let
                  | (_,SentenceAxiom { sentenceAxiomAttributes })
                      <- indexedModuleAxioms imod]
     in concat <$> mapM parseAttributesM attributes
+
+{- | Determine all indexed modules in scope from the given module.
+
+@indexedModulesInScope@ resolves all imported modules recursively so that
+traversing the 'Map' is equivalent to traversing all the (transitively) improted
```
traiansf: improted => imported

src/main/haskell/kore/src/Kore/Step/Function/Registry.hs
```diff
@@ -30,40 +29,34 @@ import Kore.AST.PureML
...
-                    (snd <$> indexedModuleAxioms indexedModule))
-            )
+    extractFunctionAxioms0 imod =
+        mapM_ (mapM_ extractAxiomPattern) axiomsInScope
+      where
+        axiomsInScope = indexedModuleAxioms <$> indexedModulesInScope imod
+    insertAxiom (name, patt) =
+        Monad.State.modify' $ Map.alter (Just . (patt :) . fromMaybe []) name
+    extractAxiomPattern (_, axiomSentence) =
+        mapM_ insertAxiom (axiomToIdAxiomPatternPair level axiomSentence)
```
traiansf: While I can agree that the refactoring above makes it more compact (and perhaps more efficient, due to using the `State` monad), I'm not sure about it being more readable.  
Could you please at least add signatures for the functions (esp. for the ones involving `mapM_`) so one would follow more easily what is `mapM_` mapping on?  I found the one for pairs a little strange, but I guess I could get used to it :-)

src/main/haskell/kore/src/Kore/Step/Function/Registry.hs
```diff
@@ -77,8 +70,21 @@ axiomToIdAxiomPatternPair
...
             case fromPurePattern (axiomPatternLeft axiomPat) of
-                ApplicationPattern Application {applicationSymbolOrAlias = s} ->
-                    return (symbolOrAliasConstructor s, axiomPat)
+                ApplicationPattern Application { applicationSymbolOrAlias } ->
+                    case symbolOrAliasParams of
+                        [] -> return (symbolOrAliasConstructor, axiomPat)
+                        _ ->
+                            -- TODO (thomas.tuegel): Handle matching for
+                            -- parameterized symbols, then enable extraction of
+                            -- their axioms.
```
traiansf: Thanks for adding this!

src/main/haskell/kore/src/Kore/Step/Function/Registry.hs
```diff
@@ -87,9 +93,34 @@ axiomToIdAxiomPatternPair
 axiomPatternsToEvaluators
     :: Map.Map (Id level) [AxiomPattern level]
     -> Map.Map (Id level) [ApplicationFunctionEvaluator level]
-axiomPatternsToEvaluators = fmap (map axiomPatternToEvaluator)
+axiomPatternsToEvaluators axiomPatterns =
+    -- remove the key if there are no associated function evaluators
+    Map.filter (not . null)
+        (mapMaybe axiomPatternEvaluator <$> axiomPatterns)
```
traiansf: So, if I understand correctly, currently none of the generated axioms will be added as evaluators due to the filtering below, as only assoc, comm, and unit were of the new form we accept (the others like functional or subsort have an existential quantifier so they don't qualify).  So we'll only have the hand-written axioms in the prelude for now. Is that right?

ttuegel: The generated axioms for `append` are added. Axioms marked `assoc`, `comm`, and `unit` are filtered out below because we don't implement matching for general equalities. The `inj` axioms are also filtered out because `inj` is a parametric symbol.

PR #296, opened 2018-10-24
Unification tests (not passing yet)

dfilaretti:
I'm trying to add some simple tests for unification of `set`, `list` and `map`. 
Although there is a different test for each, the logic is the same. 
They don't work yet, and they all seem to fail for the same reason. 

```
kore-exec: Unexpected non-false predicate GenericPredicate (Fix (AndPattern (And {andSort = SortActualSort (SortActual ...
```

(needs more investigation)

###### Reviewer checklist

---


src/main/k/in-progress/tests/list-unify-1/tests/2.output.golden
```diff
@@ -0,0 +1,15 @@
+<generatedTop>
```
traiansf: I think the output here would include b, as (unlike for set) you're only removing from one end.

src/main/k/in-progress/tests/list-unify-1/tests/3.output.golden
```diff
@@ -0,0 +1,15 @@
+<generatedTop>
```
traiansf: Same comment as above

dfilaretti:
I also realized I forgot to test the simple case of two concrete sets/lists. E.g. unifying a set with itself. I'm going to add those as well.

src/main/k/in-progress/tests/list-unify-1/tests/2.output.golden
dfilaretti: you are right, I got confused between lists and sets. fixed now.

PR #298, opened 2018-10-24
Symbolic execution fix

vladciobanu:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/IndexedModule/MetadataTools.hs
+++ b/src/main/haskell/kore/src/Kore/IndexedModule/MetadataTools.hs
@@ -16,7 +16,12 @@ module Kore.IndexedModule.MetadataTools
     ) where
 
 import           Data.Graph
+import           Data.Map.Strict
+                 ( Map )
 import qualified Data.Map.Strict as Map
+import           Data.Set
+                 ( Set )
+import qualified Data.Set as Set
 
 import Kore.AST.Common
 import Kore.AST.MetaOrObject
@@ -38,6 +43,8 @@ data MetadataTools level attributes = MetadataTools
     , isSubsortOf :: Sort level -> Sort level -> Bool
     {- ^ @isSubsortOf a b@ is true if sort @a@ is a subsort of sort @b@,
        including when @a@ equals @b@. -}
+    , subsorts :: Sort level -> Set (Sort Object)
+    -- ^ get the subsorts for a sort
     }
   deriving Functor
 
@@ -49,7 +56,8 @@ type SymbolOrAliasSorts level = SymbolOrAlias level -> ApplicationSorts level
 -- its argument and result sorts.
 --
 extractMetadataTools
-    ::forall level atts . MetaOrObject level
+    :: forall level atts
+    .  MetaOrObject level
     => KoreIndexedModule atts
     -> MetadataTools level atts
 extractMetadataTools m =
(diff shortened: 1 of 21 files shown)


src/main/haskell/kore/test/Test/Kore/IndexedModule/MockMetadataTools.hs
```diff
@@ -39,6 +40,7 @@ makeMetadataTools symbolOrAliasSorts attr isSubsortOf =
         , sortAttributes = const functionAttributes
         , symbolOrAliasSorts = symbolOrAliasSorts
         , isSubsortOf = \first second -> (first, second) `elem` isSubsortOf
+        , subsorts = const Set.empty
```
virgil-serbanuta: makeMetadataTools shoud get a table/graph/somehting on top of which it should compute `isSubsortOf` and `subsorts`, otherwise our code would get contradictory information in tests. Alternately, `isSubsortOf` could work as it currently does, but `makeMetadataTools` should also get a lookup table or something for `subsorts` and we would rely on the caller to keep them consistent.

src/main/haskell/kore/src/Kore/IndexedModule/MetadataTools.hs
```diff
@@ -58,21 +66,39 @@ extractMetadataTools m =
     , sortAttributes = getSortAttributes m
     , symbolOrAliasSorts  = getHeadApplicationSorts m
     , isSubsortOf = checkSubsort
+    , subsorts = Set.fromList . fmap getSortFromId . getSubsorts
     }
   where
+    subsortTable :: Map (Sort Object) [Sort Object]
```
virgil-serbanuta: You should add unit-tests for this.

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -732,12 +733,19 @@ sortInjectionAndEqualsAssumesDifferentHeads
         else if isConstructorLikeTop tools (project firstChild)
              || isConstructorLikeTop tools (project secondChild)
             then return (return (ExpandedPattern.bottom, SimplificationProof))
-            else
-                empty
+            else if Set.null sortIntersection
```
virgil-serbanuta: You should add unit-tests for this.

src/main/haskell/kore/src/Kore/IndexedModule/MetadataTools.hs
vladciobanu: Done; see https://github.com/kframework/kore/pull/298/commits/2ffd4b051b9fa1c141e1ba7d52e9555f049fcfbc#diff-c1602aa6b3c4f86ae4b16a7770174399R193 (and below)

vladciobanu:
@virgil-serbanuta I did a few changes. First off, I realised that `isSubsortOf` is reflexive while `subsorts` did not include the requested sort in the resulting set, so I changed it to do so. This does not change the algorithm because the set intersection used in `AndTerms` still applies identically.

I added tests for both areas you mentioned -- see the replies to your review for exact links. Please let me know if you would rather i rebase before you re-check this. I think rebasing mid-review makes it harder to do subsequent reviews so I will not do it unless you specifically request it. The downside is, Jenkins will not run.

PR #299, opened 2018-10-24
Configuration splitting simplification

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/design-decisions/2018-10-24-And-Not-Exists-Simplification.md
+++ b/design-decisions/2018-10-24-And-Not-Exists-Simplification.md
@@ -1,0 +1,65 @@
+And-Not-Exists Simplification
+======================
+
+Background
+----------
+
+Terms containing `not` show up naturally in various contexts. Examples include
+`owise` clauses and configuration left-overs when applying rewriting rules in
+proofs.
+
+Owise clauses, described in 2018-09-03-Owise-encoding.md,
+are solved naturally by applying the substitution resulting from matching and
+simplifying.
+
+The configuration left-overs, described in
+?TODO: add reference after Traian merges his PR.?, are not handled currently.
+Their format looks something like
+```
+Φ(X) ∧ (¬ ∃ Z. α(Z))
+```
+where `Φ(X)` is the term that we attempted to rewrite and `α(Z)` is the
+LHS term of a rewrite rule we managed to apply.
+
+Other terms containing `not` are also not handled currently.
+
+Problem/Questions
+-----------------
+
+We could implement generic handling for simplifying `not` terms based on
+constructors, but a `not` would be expanded to an `or` of a very large list
+of terms, which would need to be simplified. On the other hand, this should
+solve the `not` evaluation once and for all.
+
+Example:
+
+If `C1(a, b)`, `C2(a)` and `C3(a, b, c)` are the constructors of a certain sort,
+then
+```
+¬C1(phi, psi) =
+    C2(T)
+    ∨ C3(T, T, T)
+    ∨ C1(phi, ¬psi)
+    ∨ C1(¬phi, T)
+```
+
+We could also implement specific handling for the configuration left-overs,
+leaving the generic solution for later. This would be much simpler and faster,
+but would not solve the general problem.
+
+
+Decision: Implement what's needed for configuration left-overs
+--------------------------------------------------------------
+
+We will implement particular handling for now, as described in
+`2018-10-24-Configuration-Splitting-Simplification.md`.
+
+Reasoning
+---------
+
+For performance reasons, we would need to implement specific handling anyway.
+The generic handling part may or may not be needed in the future. It seems
+that the only reason for implementing generic handling first is to reduce the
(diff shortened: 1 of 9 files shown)


design-decisions/2018-10-24-And-Not-Exists-Simplification.md
```diff
@@ -0,0 +1,65 @@
...
+Example:
+
+If `C1(a, b)`, `C2(a)` and `C3(a, b, c)` are the constructors of a certain sort,
+then
+```
+¬C1(phi, psi) =
+    (∃ a . C2(a))
+    ∨ (∃ a b c . C3(a, b, c))
+    ∨ C1(phi, ¬psi)
+    ∨ (∃ b . C1(¬phi, b))
```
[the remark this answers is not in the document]
bmmoore: As you say, not really. but `C1(¬phi,⊤)` is ok here.

In general I think we should prefer `⊤` to quantifiers when we really just want to say any term is allowed in that position. It seems especially messy to put the quantifiers outside the symbol application so a variable meant just for one position is bound in the rest. `C1(¬phi,∃ b . b)` should also be equivalent to this expression.

docs/2018-10-24-Configuration-Splitting-Simplification.md
```diff
@@ -0,0 +1,88 @@
...
+
+Problem
+-------
+
+We want to rewrite `φ(X) ∧ (¬ ∃ Z. α(Z))` part of the pattern above
+to something more manageable, preferably something that does not use `not`
+and `exists`, except in cases where it can be handled by an SMT solver.
+
+We will assume that
+1. `φ(X) = t(X) ∧ p(X)` where `p(X)` is a predicate, and `t(X)` is a
```
bmmoore: Confusing because `=` is also a matching logic operator. Use "`φ(X)` is `t(X) ∧ p(X)`" (preferred), or perhaps another operator like `≡`.

bmmoore: Wait, are you assuming `φ(X)` is given in that form, or just that it's provably equal to something in that form? (the later doesn't sound good for an algorithm, unless you require `t` and `p` as inputs).

docs/2018-10-24-Configuration-Splitting-Simplification.md
```diff
@@ -0,0 +1,88 @@
...
+Solution
+--------
+
+```
+φ(X) ∧ (¬ ∃ Z. α(Z)) = t(X) ∧ p(X) ∧ (¬ ∃ Z. α(Z))
+```
+We will examine only the `t(X) ∧ (¬ ∃ Z. α(Z))` term next.
+```
+t(X) ∧ (¬ ∃ Z. α(Z))
+    = t(X) ∧ ⌈t(X) ∧ (¬ ∃ Z. α(Z))⌉
```
bmmoore: a little explanation of the steps would be nice. This first line is justified because `t(X)` is (or should be) assumed functional above, right?

docs/2018-10-24-Configuration-Splitting-Simplification.md
```diff
@@ -0,0 +1,88 @@
...
+Problem
+-------
+
+We want to rewrite `φ(X) ∧ (¬ ∃ Z. α(Z))` part of the pattern above
+to something more manageable, preferably something that does not use `not`
+and `exists`, except in cases where it can be handled by an SMT solver.
+
+We will assume that
+1. `φ(X) = t(X) ∧ p(X)` where `p(X)` is a predicate, and `t(X)` is a
+   function-like term. It may or may not unify with `α(Z)`.
```
bmmoore: What is "function-like"? Is that stronger than "functional"?

docs/2018-10-24-Configuration-Splitting-Simplification.md
```diff
@@ -0,0 +1,88 @@
...
+`∀ Z . α(Z) → ● β(Z)`, where `●` is the strong-next symbol (aka 'next'),
+to the following reachability goal
+```
+∀ X . φ(X) → ◆ ∃ Y . ψ(X, Y)
+```
+the result will look something like
+```
+￼(∀ X . Φ'(X) → ◆ ∃ Y . ψ(X, Y))
+￼∧
+￼(∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)) → ● ◆ ∃ Y . ψ(X, Y))
```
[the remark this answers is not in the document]
virgil-serbanuta: I agree that it's odd, I made similar comments about Traian's PR (both in writing and in-person). I think he managed to prove this, but it's still odd. Since it is copied from his doc, I'll add a TODO to update this when he merges.

docs/2018-10-24-Configuration-Splitting-Simplification.md
virgil-serbanuta: It's weaker, I added some definitions at the top.

virgil-serbanuta:
@bmmoore I think I have fixed all the comments, I hope it looks better now.

PR #300, opened 2018-10-24
A test for unification of concrete lists

dfilaretti:
The interesting part here is that it seems we can decide whether two lists are the same or not, but this only works when they have the same length. 
If we have two lists of different length instead of getting "not equal" as a result, we get a crash/error. 

(adding @u-- since I understand he is going to investigate list issues, so maybe this can be useful)

###### Reviewer checklist

---


u--:
`  | Seq.length list1 /= Seq.length list2 = Failure`

This line ought to be executing in those test cases with lists of different lengths. The `Failure` should then tell the unification algorithm to halt and return a negative answer. Currently trying to figure out why this does not happen.

dfilaretti:
Jenkins: test this please

PR #301, opened 2018-10-24
More list unification tests

dfilaretti:
Some more list unification tests. 
Should address some of the missing cases we discussed. 
Will probably add more, as well as similar/related cases for sets and maps. 

###### Reviewer checklist

---


src/main/k/in-progress/tests/list-unify-3/list-unify-3.k
```diff
@@ -0,0 +1,19 @@
...
+  configuration
+    <T>
+      <k> $PGM:Val </k>
+      <l> ListItem(a) ListItem(b) ListItem(c) </l>
+      <v> .K </v>
+    </T>
+
+  rule 
+    <k> V:Val </k>
+    <l> ... ListItem(V) ... </l>
```
traiansf: Actually I don't think we're aiming on handling this for now.  Maybe you could make two tests, one with matching at one end, one with matching at the other.

dfilaretti: I added two more tests very similar to this one, to try matching on the left and right.

PR #304, opened 2018-10-25
Split the configuration in the stepper.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
+++ b/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
@@ -10,6 +10,7 @@ Portability : portable
 module Kore.Step.BaseStep
     ( AxiomPattern (..)
     , StepperConfiguration (..)
+    , StepResult (..)
     , StepperVariable (..)
     , StepProof (..)
     , StepProofAtom (..)
@@ -27,6 +28,8 @@ import           Control.Monad.Except
 import qualified Data.Map as Map
 import           Data.Maybe
                  ( fromMaybe, mapMaybe )
+import           Data.Reflection
+                 ( give )
 import           Data.Semigroup
                  ( Semigroup (..) )
 import           Data.Sequence
@@ -41,18 +44,24 @@ import           Kore.AST.PureML
 import           Kore.ASTUtils.SmartConstructors
                  ( mkBottom )
 import           Kore.IndexedModule.MetadataTools
+                 ( MetadataTools, SymbolOrAliasSorts )
+import qualified Kore.IndexedModule.MetadataTools as MetadataTools
                  ( MetadataTools (..) )
 import           Kore.Predicate.Predicate
-                 ( Predicate )
+                 ( Predicate, makeAndPredicate, makeNotPredicate )
 import qualified Kore.Predicate.Predicate as Predicate
 import           Kore.Proof.Functional
                  ( FunctionalProof (..) )
 import           Kore.Step.AxiomPatterns
 import           Kore.Step.Error
 import           Kore.Step.ExpandedPattern
-                 ( PredicateSubstitution, Predicated (..) )
+                 ( ExpandedPattern, PredicateSubstitution, Predicated (..) )
 import qualified Kore.Step.ExpandedPattern as ExpandedPattern
+import qualified Kore.Step.ExpandedPattern as PredicateSubstitution
+                 ( toPredicate )
 import           Kore.Step.Pattern
+import           Kore.Step.RecursiveAttributes
+                 ( isFunctionPattern )
 import           Kore.Step.Simplification.Data
                  ( PredicateSubstitutionSimplifier (..),
                  SimplificationProof (..), Simplifier )
@@ -61,6 +70,7 @@ import           Kore.Step.StepperAttributes
 import           Kore.Step.Substitution
                  ( mergePredicatesAndSubstitutionsExcept )
 import           Kore.Substitution.Class
+                 ( Hashable (..), substitute )
 import qualified Kore.Substitution.List as ListSubstitution
 import           Kore.Unification.Data
                  ( UnificationProof (..), UnificationSubstitution,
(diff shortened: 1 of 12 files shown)


src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
```diff
@@ -216,9 +217,20 @@ isBottom
   = True
 isBottom _ = False
 
+{-| If an ExpandedPattern is bottom, it is converted to the canonical
+representation of bottom.
+-}
+loBottomize
```
u--: I'm not sure why we need to convert things to the canonical bottom, considering we can always call `isBottom`. 

In general, when you have an equivalence class of things, there are two options: either try to convert things to a canonical representation when we produce them, or just test for membership in the class when we consume them. Both work fine but it seems like we've already settled on the latter, so I think we should be consistent and stick to it.

virgil-serbanuta: That's true, I think... The reason for converting to bottom is to prevent the code that does not use isBottom yet from computing useless stuff. As an example, `stepWithAxiom` only tests the predicate returned by `stepWithAxiom`, and will recursively call itself if only the term is bottom.

I agree that `axiomFunctionEvaluator` shoud use `isBottom`. I agree that it's useless to convert the remainder to bottom. However, the stepWithAxiom function was already converting the result to bottom, and changing that should be in a separate PR, so I will either revert to the old stepWithAxiom code and leave a TODO.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -170,6 +179,18 @@ instance
     freshVariableWith (ConfigurationVariable a) n =
         ConfigurationVariable $ freshVariableWith a n
 
+{-! The result of applying an axiom to a pattern. Contains the rewritten
```
ttuegel:
```suggestion
{-| The result of applying an axiom to a pattern. Contains the rewritten
```

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -170,6 +179,18 @@ instance
...
+pattern (if any) and the unrewritten part of the original pattern.
+-}
+data StepResult level variable =
```
ttuegel: Is this not `OrOfExpandedPattern` by another name, and with only two disjuncts?

src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
ttuegel: Are we now requiring to normalized `\bottom` `ExpandedPattern`s? If so, should we combine this with other normalizers for `ExpandedPattern`? If not, why is it required?

I also think the joke is going to be lost on people.

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -118,7 +120,12 @@ transitionRule tools substitutionSimplifier simplifier =
...
         case result of
             Left _ -> pure []
-            Right (config', proof') ->
+            Right
+                ( StepResult
+                    { rewrittenPattern = config'
+                    -- TODO(virgil): Also use the remainder
+                    }
+                , proof') ->
                 if ExpandedPattern.isBottom config'
```
ttuegel: Since we are using `ExpandedPattern.isBottom` here, did we need to normalize `\bottom` patterns above?

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -118,7 +120,12 @@ transitionRule tools substitutionSimplifier simplifier =
             $ stepWithAxiom tools substitutionSimplifier config a
         case result of
             Left _ -> pure []
-            Right (config', proof') ->
+            Right
+                ( StepResult
+                    { rewrittenPattern = config'
+                    -- TODO(virgil): Also use the remainder
```
ttuegel: If `StepResult` is equivalent to `OrOfExpandedPattern`, we would only need to extract and return the list below.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
virgil-serbanuta: Well, maybe, but it's tricky. First, @traiansf uses this formula in https://github.com/kframework/kore/pull/268 :
```
(∀ X . Φ'(X) → ◆ ∃ Y . ψ(X, Y))
∧
(∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)) → ⚪ ◆ ∃ Y . ψ(X, Y))
```
I believe that "next" can be removed from the last term, i.e. we have something like
```
(∀ X . Φ'(X) → ◆ ∃ Y . ψ(X, Y))
∧
(∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)) →  ◆ ∃ Y . ψ(X, Y))
```
in which case this becomes
```
(
    (∀ X . Φ'(X))
    ∨ 
    (∀ X . Φ(X) ∧ (¬ ∃ Z. α(Z)))
) → ◆ ∃ Y . ψ(X, Y))
```
in which case, indeed, it becomes an `or`. I'll wait to see what @traiansf does with his PR before agreeing that this is, indeed, the case. We may have a small issue because the `Φ` formula above is not what we are actually rewriting, but whatever.

Anyway, we have a second issue: the remainder might be part of an `or`, but we might want to treat it as an `or` only when we do proofs, not when we execute. To be more precise, I guess that when we execute a step we always want to rewrite the initial pattern, and we may want to ignore the reminders (or, at most, we may to rewrite the initial pattern in all possible ways, then we may want to generate a reminder for all applied rules at once). Again, @traiansf to confirm this.

src/main/haskell/kore/src/Kore/Step/Step.hs
virgil-serbanuta: As mentioned above, even if it is so, we might want to do this only when doing proofs.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
virgil-serbanuta: Things have become a bit more clear about the first issue mentioned in the above message. Here's how I understand this problem:

This rewriting stuff is useful when we want to do single-path execution and we want to make sure we reach a certain target. Roughly, we want to first remove the target from our current pattern, then we want to rewrite the result entirely. 

There are two ways of doing this. The first one is described in Traian's PR mentioned above, and there we want to treat the remainder separately, rewriting it, and rewriting the new remainder, and so on, until we get a bottom remainder. Failing to do so means failing to reach the target.

In the second one, we would represent the pattern as an "or" and we would take random components (`rc`) of that or, removing the target (`t`) from them (`rt := rc and not (exists x . t)`). If `rt` is null, then we just remove `rc` from the or and we restart the algorithm. If `rt` is not null we would try to an axiom `lhs->o rhs` axiom with which to rewrite `rt`. If we can't find one, we fail the algorithm. If we find one, then we remove `rc` from the or and we add the rewrite result together with the `rc` from which we removed the axiom `lhs` (`rc and not (exists x . lhs)`) back to the or. We repeat everything until the or becomes empty.

My guess is that @traiansf prefers the former, but not the latter.

The second issue from above still remains unsolved.

PR #305, opened 2018-10-25
Remove Data.Result

ttuegel:
The `Result` type functioned as `Maybe` with an extra constructor `Failure`,
which was only used to simulate the `Left` return of the unification
procedure (which runs in the `ExceptT (UnificationOrSubstitutionError ...)`
monad). Unification is much simpler if the unification cases run in `MaybeT m`
where `m` is `Simplifier` for `\and`-simplification and `ExceptT ...` for
unification.

While implementing this refactoring, a bug was discovered in List and Set
unification which could cause invalid substitutions (of non-function patterns)
to be generated.

I also cleaned up the documentation for `Kore.Step.Simplification.AndTerms` a bit, and removed some stale comments.

@u-- Could you check in particular the changes I made to the List unification code?

@dfilaretti Could you check in particular the changes I made to the Set unification code?

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -91,20 +108,15 @@ termEquals
...
-    termEquals0 = do  -- Result monad
-        result <- termEqualsAnd tools substitutionSimplifier first second
-        return $ do  -- Counter monad
-            (Predicated {predicate, substitution}, _pred) <- result
-            return
-                ( PredicateSubstitution
-                    {predicate = predicate, substitution = substitution}
-                , SimplificationProof
-                )
+    erasePredicatedTerm (Predicated { predicate, substitution }, proof) =
```
virgil-serbanuta: I think this is generic enough that it should live in ExpandedPattern.hs .

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -627,52 +633,65 @@ equalInjectiveHeadsAndEquals
...
+            predicates
+            substitutions
+    let PredicateSubstitution
+            { predicate = mergedPredicate
+            , substitution = mergedSubstitution
+            }
+          =
+            merged
+    return
+        ( Predicated
```
virgil-serbanuta: This is not related to this PR, but could you add a TODO to somehow solve the case when one of the children terms is bottom? I think we don't want to create an application pattern. Most likely this will not break our code since subsequent simplifications will make everything bottom, but it will be confusing, at least. This may also happen in other places.

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -506,168 +509,175 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
...
+        ,
+            let term3 = Mock.builtinList [Mock.a, Mock.a]
+                term4 = Mock.builtinList [Mock.a, Mock.b]
+                unified34 =
+                    let unifiedElements = [pure Mock.a, ExpandedPattern.bottom]
+                    in Mock.builtinList <$> sequenceA unifiedElements
+            in
+                testCase "[same head, different head]"
+                $ assertEqualWithExplanation ""
+                    (Just unified34)
```
virgil-serbanuta: I think that the result should be just bottom, not a list containing bottom. They are equivalent, but the former is much more useful since it will probably transform the entire rewrite term to bottom. Why the change?

ttuegel: The result is `\and(/* term: a list containing bottom */, /* predicate: */ \bottom)`, so the entire rewrite will be transformed into bottom. This is not immediately evident from the way the test is written, so I should change the way the test result is written, but is that result acceptable? I believe this is also what Map unification does. The change is because the List unification code is not doing its own ad-hoc unification now.

src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -357,61 +355,84 @@ unify
...
+    unify0
+        (DV_ dvSort (BuiltinDomainList list1))
+        (DV_ _      (BuiltinDomainList list2))
+      =
+        Monad.Trans.lift $ unifyConcrete dvSort list1 list2
+
+    unify0
+        (App_ symConcat [DV_ _ (BuiltinDomainList list1), v@(Var_ _)])
+        (DV_ dvSort (BuiltinDomainList list2))
+        | isSymbolConcat hookTools symConcat =
```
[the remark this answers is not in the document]
ttuegel: I [updated the style guide](https://github.com/kframework/kore/wiki/Coding-Standards#guards) to match what you said, please let me know if you do not agree.

src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
ttuegel: That will also require that it is a symbol, not alias, application, right? I don't know what it would mean to have an injective alias, but it's possible to give that attribute to an alias.

src/main/haskell/kore/src/Kore/Builtin/List.hs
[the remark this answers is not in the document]
virgil-serbanuta: I agree. Could you update the new code to match it?

src/main/haskell/kore/src/Kore/Builtin/Map.hs
```diff
@@ -550,33 +524,36 @@ unify
...
-                            -- remainder of map2.
-                            ExpandedPattern.bottom
-                        | otherwise =
-                            asBuiltinMap <$> propagatePredicates _intersect
-                normalized <-
-                    normalize tools substitutionSimplifier result
-                return (normalized, SimplificationProof)
-        return unified
+            predicatedIntersect :: Predicated level variable (Map k p)
+            predicatedIntersect = (propagatePredicates . discardProofs) intersect
```
virgil-serbanuta: Line length.

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -506,168 +509,175 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
...
+                , predicate = makeTruePredicate
+                , substitution =
+                    [ (Mock.m, Mock.builtinMap [(Mock.bConcrete, fOfB)])
+                    , (Mock.x, fOfA)
+                    ]
+                }
+            )
+            (unify
+                mockMetadataTools
+                (Mock.builtinMap [(Mock.aConcrete, fOfA), (Mock.bConcrete, fOfB)])
```
virgil-serbanuta: Line length.

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -506,168 +509,175 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
...
+                    (mkVar Mock.m)
                 )
-                (unify
-                    mockMetadataTools
-                    (Mock.concatMap
-                        (mkVar Mock.m)
-                        (Mock.builtinMap [(Mock.aConcrete, mkVar Mock.x)])
-                    )
-                    (Mock.builtinMap [(Mock.aConcrete, fOfA), (Mock.bConcrete, fOfB)])
+                (Mock.builtinMap [(Mock.aConcrete, fOfA), (Mock.bConcrete, fOfB)])
```
virgil-serbanuta: Line length

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -506,168 +509,175 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
...
+            )
+        , testCase "framed Map with concrete Map" $ assertEqualWithExplanation ""
```
virgil-serbanuta: Line length

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -506,168 +509,175 @@ test_andTermsSimplification = give mockSymbolOrAliasSorts
...
+            (unify
+                mockMetadataTools
+                (Mock.concatMap
+                    (mkVar Mock.m)
+                    (Mock.builtinMap [(Mock.aConcrete, mkVar Mock.x)])
                 )
-        )
-    , testCase "builtin List domain"
-        (do
+                (Mock.builtinMap [(Mock.aConcrete, fOfA), (Mock.bConcrete, fOfB)])
```
virgil-serbanuta: Line length

src/main/haskell/kore/test/Test/Kore/Step/Simplification/AndTerms.hs
virgil-serbanuta: Ah, `sequenceA` will `and` the predicate from ExpandedPattern.bottom with the one from Mock.a, producing a bottom predicate... Yes, this is acceptable. Yes, making this clear would be nice.

dfilaretti:
I took a look at `Set.hs` and the refactoring looks OK to me.

ttuegel:
> I think that `sigma(..., ⊥, ...) = ⊥` is a property of all symbols, not only the injective ones

@virgil-serbanuta Yes, I just meant that the case you commented on only handles unification of injective symbols. I think we should have a step in the simplifier that turns such symbol applications to ⊥ before they reach the unifier.

> Also, I think that the map, set and list constructor-like symbols (`concat` and `unit` or whatever they are called) behave as symbols, even if they are defined as aliases (@traiansf do we want to allow that?), so they should become `⊥` if one of their arguments is `⊥`.

That is carried out during simplification: the ⊥-children are turned into an empty `OrOfExpandedPattern` that eliminates the entire term as it is propagated through builtin domain value.

PR #307, opened 2018-10-26
Add predicate to search

vladciobanu:
I'm not sure what kind of tests I should add for this. Could I get some suggestions @traiansf ?

###### Reviewer checklist

---


vladciobanu:
Jenkins: test this please

src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -428,8 +430,8 @@ mainWithOptions
                             (orPredicate, _proof) =
                                 give symbolOrAliasSorts
                                 $ makeMultipleOrPredicate
-                                $ map
-                                    ExpandedPattern.substitutionToPredicate
+                                $ fmap
```
virgil-serbanuta: Why the map->fmap change?

src/main/haskell/kore/src/Kore/Step/Search.hs
```diff
@@ -128,32 +127,45 @@ matchWith
...
     -> PureMLPatternSimplifier level variable
     -> ExpandedPattern level variable
     -> ExpandedPattern level variable
-    -> MaybeT Simplifier (UnificationSubstitution level variable)
+    -> MaybeT Simplifier (PredicateSubstitution level variable)
 matchWith tools substitutionSimplifier simplifier e1 e2 = do
     (unifier, _proof) <- hushT (unificationProcedure tools substitutionSimplifier t1 t2)
-    let (predicate, _proof) = give (MetadataTools.symbolOrAliasSorts tools)
-            $ makeMultipleAndPredicate
+    (predicate, _proof) <-
```
virgil-serbanuta: I think that `predicate` also contains a substitution, please rename if so.

src/main/haskell/kore/app/exec/Main.hs
vladciobanu: I apologize for the extra noise, but I consider `map` a legacy (pre-`Functor`) and I thought the general consensus is to always use `fmap`. I can revert if people disagree.

src/main/haskell/kore/src/Kore/Step/Search.hs
vladciobanu: I think you are right, I have a note that I forgot about. I'll double check and rename. Thanks!

PR #315, opened 2018-10-30
Kore.Step.OrOfExpandedPattern.merge: Use idempotency property

ttuegel:
`merge` takes advantage of the idempotency property of disjunction to eliminate
identical patterns from disjunctions. Only patterns that are syntactically
identical are de-duplicated. (Simplification should eventually rewrite
semantically-equivalent patterns into syntactically identical ones.)

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
+++ b/src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
@@ -32,10 +32,11 @@ module Kore.Step.OrOfExpandedPattern
     , traverseFlattenWithPairsGeneric
     ) where
 
-import Data.List
-       ( foldl' )
-import Data.Reflection
-       ( Given )
+import           Data.List
+                 ( foldl' )
+import           Data.Reflection
+                 ( Given )
+import qualified Data.Set as Set
 
 import           Kore.AST.Common
                  ( PureMLPattern, SortedVariable, Variable )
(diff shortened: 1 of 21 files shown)


src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
```diff
@@ -337,14 +336,23 @@ crossProductGenericF
...
-    :: OrOfExpandedPattern level variable
+    :: (Ord level, Ord (variable level))
+    => OrOfExpandedPattern level variable
     -> OrOfExpandedPattern level variable
     -> OrOfExpandedPattern level variable
-merge patts1 patts2 =
-    filterOr (mergeGeneric patts1 patts2)
+merge (MultiOr patts1) (MultiOr patts2) =
+    (filterOr . MultiOr . Set.toList)
+        (Set.fromList patts1 <> Set.fromList patts2)
```
virgil-serbanuta: 1. Could you add a note that this does not work as expected when, e.g., the patterns contain `\exists` and `\forall` statements.

2. It seems to me that this idempotency thing should also apply to, say, `make` and `traverseWithPairs` (in general: to all or most functions that call `filterOr`).

3. Could you add tests? At least one or more unit-tests for for OrOfEXpandedPattern, but maybe also a test for the original problem (`(a or a)=a` simplification or whatever).

src/main/haskell/kore/test/Test/Kore.hs
```diff
@@ -294,8 +296,24 @@ patternGen childGen x =
         , VariablePattern <$> variableGen x
         ]
 
-purePatternGen :: MetaOrObject level => level -> Gen (CommonPurePattern level)
-purePatternGen level = embed <$> patternGen (purePatternGen level) level
+purePatternGen :: forall level. MetaOrObject level => level -> Gen (CommonPurePattern level)
```
virgil-serbanuta: There are a lot of lines over 80 chars in this PR, but this one goes over 90.

PR #317, opened 2018-10-31
Add KEQUAL.ite and STRING.lt builtins

vladciobanu:
I only added `STRING.lt` since this seems to be the only required builtin by `evm-semantics`, at least for now.

It is odd that strings are not between `"`. Should add a card to investigate whether this is a problem in the frontend or here.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
@@ -19,6 +19,7 @@ module Kore.Builtin.KEqual
     , builtinFunctions
     ) where
 
+import qualified Data.Functor.Foldable as Functor.Foldable
 import qualified Data.HashMap.Strict as HashMap
 import           Data.Map
                  ( Map )
@@ -27,10 +28,14 @@ import           Data.Text
                  ( Text )
 
 import           Kore.AST.Common
-                 ( Application (..), PureMLPattern, SortedVariable )
+                 ( Application (..), BuiltinDomain (..), DomainValue (..),
+                 Pattern (..), PureMLPattern, SortedVariable )
 import           Kore.AST.MetaOrObject
+import           Kore.AST.Sentence
+                 ( SentenceSymbol (..) )
 import qualified Kore.Builtin.Bool as Bool
 import qualified Kore.Builtin.Builtin as Builtin
+import qualified Kore.Error
 import qualified Kore.IndexedModule.MetadataTools as MetadataTools
 import qualified Kore.Step.ExpandedPattern as ExpandedPattern
 import           Kore.Step.Function.Data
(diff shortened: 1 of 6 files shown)


dwightguth:
It is deliberate that the representation of the string "foo" in kore is \dv{String{}}("foo") rather than \dv{String{}}("\"foo\""). I believe there is a bug in the k frontend that might not correctly add back the quotes when unparsing though.

src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -113,3 +154,44 @@ evalKEq true false tools substitutionSimplifier _ pat =
...
+    -> PureMLPatternSimplifier Object variable
+    -> Application Object (PureMLPattern Object variable)
+    -> Simplifier
+        ( AttemptedFunction Object variable
+        , SimplificationProof Object
+        )
+evalKIte _ _ _ pat =
+    case pat of
+        Application
+            {  applicationChildren = [expr, t1, t2] } ->
```
ttuegel: This must throw `Builtin.wrongArity` if called with the wrong arity. Although the definition and pattern are verified, it is possible that the pattern was rewritten in the meantime. If the arity is wrong, we must not return `notApplicableFunctionEvaluator`.

src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -113,3 +154,44 @@ evalKEq true false tools substitutionSimplifier _ pat =
...
+        _ -> notApplicableFunctionEvaluator
+  where
+    evaluate :: Functor.Foldable.Fix (Pattern Object variable) -> Maybe Bool
+    evaluate expr = case Functor.Foldable.project <$> [expr] of
+        [DomainValuePattern a] -> Just (get a)
+        _                      -> Nothing
+
+    get :: DomainValue Object (BuiltinDomain child) -> Bool
+    get =
+        Builtin.runParser (Text.unpack "ite")
```
ttuegel: The call to `Text.unpack` should not be necessary. Please use the full hook name of the builtin; some builtin domains have overlapping function names, and I think this is one.
```suggestion
        Builtin.runParser "KEQUAL.ite"
```

src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -61,20 +67,55 @@ symbolVerifiers =
...
+            Kore.Error.koreFailWhen
+                (secondSort /= thirdSort)
+                "Expected continuations to match"
+            Kore.Error.koreFailWhen
+                (secondSort /= result)
+                "Expected continuations to match"
+            return ()
+      where
+        arity = length sorts
+        firstSort = head sorts
```
ttuegel: You can avoid the (admittedly, guarded) partial functions using a pattern match:
```.hs
case sorts of
    [firstSort, secondSort, thirdSort] -> _
    _ -> Kore.Error.koreFail _
```

vladciobanu: Whoops, I was going to make this prettier, but I forgot after all the debugging. Thanks!

src/main/haskell/kore/test/Test/Kore/Builtin/String.hs
```diff
@@ -0,0 +1,176 @@
...
+
+-- | Specialize 'String.asPattern' to the builtin sort 'stringSort'.
+asPattern :: String -> CommonPurePattern Object
+asPattern = String.asPattern stringSort
+
+-- | Specialize 'String.asConcretePattern' to the builtin sort 'stringSort'.
+asConcretePattern :: String -> ConcretePurePattern Object
+asConcretePattern = String.asConcretePattern stringSort
+
+-- | Specialize 'Int.asExpandedPattern' to the builtin sort 'intSort'.
```
15joeybloom: Looks like you forgot to change this line from `Int` to `String`

15joeybloom:
`docs/hooks.md` should be updated to mention `<String` and `ite`. Also, it looks like the existing `KEQUAL.eq` and `KEQUAL.neq` are missing from `docs/hooks.md` as well.

src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -61,20 +67,58 @@ symbolVerifiers =
...
+            { sentenceSymbolSorts = sorts
+            , sentenceSymbolResultSort = result
+            }
+      =
+        Kore.Error.withContext "In argument sorts" $
+            case sorts of
+                [firstSort, secondSort, thirdSort] -> do
+                    Kore.Error.koreFailWhen
+                        (arity /= 3)
+                        ("Expected 3 arguments, found " ++ show arity)
```
15joeybloom: You don't need this arity check anymore, right? If `arity /= 3` we would be in the other arm of the `case`.

ttuegel:
I added a [card](https://trello.com/c/wKedMqvE) for the missing hooks documentation.

src/main/haskell/kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -113,3 +157,44 @@ evalKEq true false tools substitutionSimplifier _ pat =
...
+    get =
+        Builtin.runParser "KEQUAL.ite"
+        . Builtin.parseDomainValue Bool.parse
+
+    evalIte expr t1 t2 =
+        case evaluate expr of
+            Just result ->
+                if result
+                    then purePatternFunctionEvaluator t1
+                    else purePatternFunctionEvaluator t2
```
15joeybloom: I'm nitpicking, but I think this is more readable:
```
            Just True -> purePatternFunctionEvaluator t1
            Just False -> purePatternFunctionEvaluator t2
```

ttuegel: It's generally not a good idea to have backtracking `case` statements like that; this would avoid backtracking:
```.hs
Just result
    | result -> purePatternFunctionEvaluator t1
    | otherwise -> purePatternFunctionEvaluator t2
```

15joeybloom: Oh, ok. Is this an efficiency concern? Or a readability concern?

PR #318, opened 2018-10-31
Unification for IMP

ttuegel:
This pull request fixes unification and collections enough to run the tutorial version of IMP.

The tutorial version is still quite slow: `sum.imp` takes about 20 s on my machine, compared to about 2 s in `imp-map-unify-strict`.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Builtin/Set.hs
```diff
@@ -519,3 +551,34 @@ unify
...
+                    (key, _) <- unifyChildren key1 key2
+                    let result = App_ element' <$> propagatePredicates [key]
+                    return (result, SimplificationProof)
+            _ ->
+                -- Cannot unify a non-element Set with an element Set.
+                return (ExpandedPattern.bottom, SimplificationProof)
+
+    -- | Return an unsolved predicate
+    unsolved
+        :: (level ~ Object, k ~ Kore.ConcretePurePattern Object)
```
vladciobanu: `k` should also not be needed here

src/main/haskell/kore/test/Test/Kore/Builtin/Set.hs
```diff
@@ -264,13 +272,110 @@ prop_unifyFramingVariable set n =
...
 -- | Specialize 'Set.asPattern' to the builtin sort 'setSort'.
 asPattern :: Set Integer -> CommonPurePattern Object
 Right asPattern = (. Set.map Test.Int.asConcretePattern) <$> Set.asPattern indexedModule setSort
 
 -- | Specialize 'Set.asPattern' to the builtin sort 'setSort'.
-asExpandedPattern :: Set.Builtin -> CommonExpandedPattern Object
-Right asExpandedPattern = Set.asExpandedPattern indexedModule setSort
+asExpandedPattern :: Set Integer -> CommonExpandedPattern Object
+Right asExpandedPattern =
+    (<$>)
```
vladciobanu: I'd prefer `fmap` here as mentioned above. I'm not a huge fan of excessive pointfree usage, and this is close to the limit. I'd probably not use pointfree here, but I don't mind that you did. Do we have some sort of clear style guide about this? @virgil-serbanuta ?

src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -418,27 +444,70 @@ unify
...
+        -> m (expanded, proof)
+    unifyFramedRight resultSort list1 symConcat prefix2 frame2
+      | Seq.length prefix2 > Seq.length list1 =
+        return (ExpandedPattern.bottom, SimplificationProof)
+      | otherwise =
+        do
+            (prefixUnified, _) <- unifyConcrete resultSort prefix1 prefix2
+            (suffixUnified, _) <- simplifyChild frame2 listSuffix1
+            let result =
+                    (<$>)
```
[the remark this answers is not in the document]
ttuegel: > Why not fmap if you're using it as prefix?

Because I hate that name. :laughing: As a practical matter, though, @u-- is working on [this](https://trello.com/c/BMUSsyAK) which will remove the application pattern entirely.

src/main/haskell/kore/test/Test/Kore/Builtin/Set.hs
u--: `<$>` is a synonym for `fmap`, they are both pointfree. FWIW I agree with tom that the name `fmap` is ugly and generally prefer `<$>`. Maybe because operators are visually distinct from alphanumeric functions, and it helps signal that there's a functor involved.

src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -418,27 +453,70 @@ unify
...
+        => PureMLPattern level variable
+        -> PureMLPattern level variable
+        -> m (expanded, proof)
+    unsolved a b =
+        let
+            unified = give symbolOrAliasSorts mkAnd a b
+            predicate = give symbolOrAliasSorts makeCeilPredicate unified
+            expanded = (give symbolOrAliasSorts pure unified) { predicate }
+        in
+            return (expanded, SimplificationProof)
```
u--: Maybe define
```
unsolved p = pure p { predicate = makeCeilPredicate p }
```
and call ``giveSymbolOrAliasSorts $ unsolved (a `mkAnd` b)``

src/main/haskell/kore/src/Kore/Builtin/Set.hs
```diff
@@ -519,3 +557,34 @@ unify
...
+        => Kore.PureMLPattern level variable
+        -> Kore.PureMLPattern level variable
+        -> m (expanded, proof)
+    unsolved a b =
+        let
+            unified = give symbolOrAliasSorts mkAnd a b
+            predicate = give symbolOrAliasSorts makeCeilPredicate unified
+            expanded = (give symbolOrAliasSorts pure unified) { predicate }
+        in
+            return (expanded, SimplificationProof)
```
u--: Duplicate?

src/main/haskell/kore/src/Kore/Builtin/List.hs
ttuegel: Then `unsolved` would just be a `\ceil` predicate, and not an "unsolved" unification problem.

src/main/haskell/kore/src/Kore/Builtin/List.hs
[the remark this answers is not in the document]
bmmoore: `App_ symConcat <$>` would look just fine here.

src/main/haskell/kore/src/Kore/Builtin/Map.hs
```diff
@@ -508,28 +559,51 @@ unify
...
+                        [ DV_ _ (BuiltinDomainMap map2), x@(Var_ _) ] ->
+                            unifyFramed1 resultSort map1 symbol2 map2 x
+                        [ x@(Var_ _), DV_ _ (BuiltinDomainMap map2) ] ->
+                            unifyFramed1 resultSort map1 symbol2 map2 x
+                        [ _, _ ] ->
+                            give symbolOrAliasSorts
+                                (Builtin.unifyUnsolved dv1 app)
+                        _ ->
+                            Builtin.wrongArity "MAP.concat"
+                | isSymbolElement hookTools symbol2 ->
```
virgil-serbanuta: Hi, it's late in the night and I'm not fully trusting my ability to review code.

My current guess is that this handles the case when we have a map matching without a framing variable, i.e. `<store> X -> Y </store>`, and that this will match a map with a single element.

I also guess that the same case, but with framing variables, i.e. `<store> X -> Y ... </store>`, is handled by the `[ _, _ ]` case branch above.

Is that correct?

PR #319, opened 2018-11-01
Unify return concrete

u--:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/List.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/List.hs
@@ -414,9 +414,9 @@ unify
               | isSymbolConcat hookTools symbol2 ->
                 Monad.Trans.lift $ case args2 of
                     [ DV_ _ (BuiltinDomainList list2), x@(Var_ _) ] ->
-                        unifyFramedRight resultSort list1 symbol2 list2 x
+                        unifyFramedRight resultSort dv1 list2 x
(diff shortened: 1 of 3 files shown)


src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -457,17 +457,17 @@ unify
         -> Seq p
         -> p
         -> m (expanded, proof)
-    unifyFramedRight resultSort list1 symConcat prefix2 frame2
```
ttuegel: Could you remove the unused argument in all the framing variable unification cases?

src/main/haskell/kore/src/Kore/Builtin/List.hs
```diff
@@ -457,17 +457,17 @@ unify
...
       | otherwise =
         do
             (prefixUnified, _) <- unifyConcrete resultSort prefix1 prefix2
             (suffixUnified, _) <- simplifyChild frame2 listSuffix1
-            let result =
-                    (<$>)
-                        (App_ symConcat)
-                        (propagatePredicates [prefixUnified, suffixUnified])
+            let result = give symbolOrAliasSorts $
+                    pure (DV_ resultSort (BuiltinDomainList list1))
```
ttuegel: Could we pass in the original `DV_` pattern so that it is shared instead of reconstructing it?

PR #321, opened 2018-11-05
Translate predicates for SMT, redux

ttuegel:
This pull request fixes our predicate translation for SMT. There is also a significant refactoring of our predicate evaluator test suite because we were sending a number of things to SMT which were not predicates. Shockingly, this almost worked because we conflated `false` with `\bottom{}()`.

I added a design decision document describing the predicate translation; reviewers should start there.

The `PredicateProof` type was in the way of some of the work I had to do, so I removed it. It was being discarded everywhere it was used, so it did not represent progress toward our proof-objects goal.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Simplification/Implies.hs
```diff
@@ -177,13 +177,13 @@ makeEvaluateImpliesNonBool
             { term = t
             , predicate =
                 -- TODO: Remove fst
```
u--:
```suggestion
```

src/main/haskell/kore/src/Kore/Step/Simplification/Implies.hs
```diff
@@ -177,13 +177,13 @@ makeEvaluateImpliesNonBool
...
-                fst $ makeImpliesPredicate
+                makeImpliesPredicate
                     -- TODO: Remove fst
```
u--:
```suggestion
```

src/main/haskell/kore/src/Kore/Step/Simplification/Implies.hs
```diff
@@ -177,13 +177,13 @@ makeEvaluateImpliesNonBool
...
-                    (fst $ makeAndPredicate
+                    (makeAndPredicate
                         firstPredicate
                         (substitutionToPredicate firstSubstitution))
                     -- TODO: Remove fst
```
u--:
```suggestion
```

src/main/haskell/kore/src/Kore/Step/Simplification/Not.hs
```diff
@@ -122,8 +122,8 @@ makeEvaluate
             -- TODO: Remove fst.
```
u--:
```suggestion
```

ttuegel:
I am going to switch our SMT solver to use `sat` instead of `prove`; that should mitigate the problem described above. The SBV library is too opinionated for our purposes: Kore uses scoped quantifiers, SMT-LIB supports scoped quantifiers, but SBV forces everything to be given in prenex normal form. In the future, we should use another library for SMT.

design-decisions/2018-11-05-Translating-predicates.md
```diff
@@ -0,0 +1,122 @@
...
+translated as variables, i.e. left uninterpreted by the solver. `\in{S, I}` is
+also translated as a variable. Identical patterns will be translated to the
+same variable, so that the solver can reason over the predicates even if it
+cannot reason under them.
+
+As indicated above, `\equals{Bool, I}` and `\equals{Int, I}` are interpreted as
+the corresponding builtin equalities. This is possible because the predicate in
+`I` is translated to `Bool` in the solver. Equality on predicates (`\equals{I,
+I}`) is translated as `\iff{I}`. Equality on all other sorts is translated to a
+variable, i.e. left uninterpreted as `\ceil{S, I}`.
```
traiansf: I wonder whether, for all other sorts, we could actually interpret equality as equality, and have the operands be variables. This might let the SMT do things like transitivity and substitution.

ttuegel: I'm almost certain we can't do that for arbitrary sorts because the patterns might not be functional.

src/main/haskell/kore/src/Kore/Predicate/Predicate.hs
```diff
@@ -187,20 +176,15 @@ makeAndPredicate
...
-makeAndPredicate (GenericPredicate first) (GenericPredicate second)
-  | first == second =
-    ( GenericPredicate first
-    , PredicateProof
-    )
-makeAndPredicate (GenericPredicate first) (GenericPredicate second) =
-    ( GenericPredicate $ mkAnd first second
-    , PredicateProof
-    )
+    -> Predicate level variable
```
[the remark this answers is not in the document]
traiansf: @virgil-serbanuta I think that, for now, we stopped adding them and are just dropping them whenever we feel they're in the way.  I have no strong feeling about this; I tend to agree that, whenever we decide to reincorporate proofs we'll have to do much more work than simply adding back an extra result.  Moreover, we should probably have somebody sit and thing about an architecture for modularly (and elegantly, and uniformly) adding proofs.

PR #325, opened 2018-11-06
Refactor PredicateSubstitution

ttuegel:
`PredicateSubstitution` is a special case of `Predicated`.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
+++ b/src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
@@ -10,10 +10,11 @@ Stability   : experimental
 Portability : portable
 -}
 module Kore.Step.ExpandedPattern
-    ( CommonExpandedPattern
-    , CommonPredicateSubstitution  -- TODO(virgil): Stop exporting this.
+    ( Predicated (..)
     , ExpandedPattern
-    , PredicateSubstitution (..)  -- TODO(virgil): Stop exporting this.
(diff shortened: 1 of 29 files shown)


traiansf:
What is the rationale for this?  I thought keeping `PredicateSubstitution` as a separate type was useful in that we don't accidentally mix types.  Can't we, at least, have a `newtype PredicateSubstitution` wrapping the `Predicated`?

src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
```diff
@@ -49,58 +59,72 @@ import           Kore.Predicate.Predicate
...
+{- | @Predicated@ represents a value conditioned on a predicate.
 
-type ExpandedPattern level variable =
-    Predicated level variable (PureMLPattern level variable)
+@Predicated level variable child@ represents a @child@ conditioned on a
+@predicate@ and @substitution@ (which is a specialized form of predicate).
+
+The 'Applicative' instance conjoins the predicates and substitutions so that the
+result is conditioned on the combined predicates of the inputs. The combined
+predicate and substitution are /not/ normalized.
```
virgil-serbanuta: IMO it's very easy to use the applicative stuff and to forget to normalize at the end, and I'm not confident that using non-normalized substitutions would always work out properly (though it might since substitutions are equalities). It would be nice to have a type-safe way of combining this stuff.

PR #326, opened 2018-11-09
Refactor attribute parser

ttuegel:
In preparation to add more attributes (`smtlib` and `smt-lemma` next), and in response to remarks by other contributors that the attribute parser is difficult to understand, I simplified the parser and added many test cases.

Besides simplifying the attribute parser, I also changed `StepperAttributes` to distinguish clearly between attributes that are declared in the Kore definition and attributes that are inferred from the declared attributes.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -720,11 +722,12 @@ metaNameForObjectSort name = "#`" ++ name
 
 See also: 'parseAttributes'
```
traiansf: Qualify `parseAttributes` in the comment?

vladciobanu:
I'm not a big fan of `Default`, but I realize it was already there for `Attributes`. If we all agree, we should create a trello card to remove it.

ttuegel:
@traiansf The lenses are necessary on some level to make the attribute parsers [composable](https://github.com/kframework/kore/pull/326/files#diff-670ed0a1188c6486add14ac9bd0951e2R77), but I can change it so that the lenses do not need to be the "default" way to access `StepperAttributes`.

ttuegel:
> If we all agree, we should create a trello card to remove it.

@vladciobanu I would like to remove it too, but I'm not sure what it should be replaced with. Every attribute must have a default value in case it is omitted.

PR #329, opened 2018-11-14
Compare mod alpha

u--:
May not be what we need for merging branches, but still a very basic feature which should probably exist for other purposes...

Also, extended the generator in Test.Kore to use previously bound variables.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
+++ b/src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
@@ -1,6 +1,6 @@
 {-|
 Module      : Kore.ASTUtils.AlphaCompare
-Description : Compare 
+Description : Compare
 Copyright   : (c) Runtime Verification, 2018
 License     : NCSA
 Maintainer  : phillip.harris@runtimeverification.com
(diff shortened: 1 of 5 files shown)


src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+module Kore.ASTUtils.AlphaCompare
+    ( alphaCompare
+    ) where
+
+import Data.List
+
+import Kore.AST.Common
+import Kore.ASTUtils.SmartPatterns
+
+alphaCompare
```
ttuegel: Could we name this `alphaEq` instead? The name is slightly misleading because `compare` exists, and this does not do what that does.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+    :: (Eq (var level), Eq (var level))
+    => PureMLPattern level var
+    -> PureMLPattern level var
+    -> Bool
+alphaCompare e1 e2 = go [] [] e1 e2
+
+go
```
ttuegel: Please move `go` into a `where` clause of `alphaCompare`:
1. The profiling report will be much easier to read.
1. The `Eq` constraint will not need to be passed over and over at runtime.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+    ( alphaCompare
+    ) where
+
+import Data.List
+
+import Kore.AST.Common
+import Kore.ASTUtils.SmartPatterns
+
+alphaCompare
+    :: (Eq (var level), Eq (var level))
```
ttuegel: The `Eq` constraint is duplicated.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+        && go ctx1 ctx2 a1 a2
+        && go ctx1 ctx2 b1 b2
+    (Top_ s1, Top_ s2) ->
+        s1 == s2
+    (StringLiteral_ s1, StringLiteral_ s2) ->
+        s1 == s2
+    (CharLiteral_ c1, CharLiteral_ c2) ->
+        c1 == c2
+    (App_ h1 c1, App_ h2 c2) ->
+        h1 == h2
```
ttuegel: The application heads can contain sort variables, but I'm not sure what we should do about that.

Actually, everywhere occurrence of a sort in a pattern could be a sort variable.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+        && and (zipWith (go ctx1 ctx2) c1 c2)
+    (Var_ v1, Var_ v2) ->
+        elemIndex v1 ctx1 == elemIndex v2 ctx2
```
ttuegel: If I understand correctly (and I am not sure I do): comparison is implemented by a de Bruijn indexing scheme for bound variables, is that right?

What happens if `v1` and `v2` are free?

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,101 @@
...
+alphaCompare e1 e2 = go [] [] e1 e2
+
+go
+    :: (Eq (var level), Eq (var level))
+    => [var level]
+    -> [var level]
+    -> PureMLPattern level var
+    -> PureMLPattern level var
+    -> Bool
+go ctx1 ctx2 e1 e2 = case (e1, e2) of
```
ttuegel: Please document how this function works. It would particularly help:
1. to know what the arguments are, and
1. not to use single-letter variable names everywhere.

src/main/haskell/kore/test/Test/Kore.hs
```diff
@@ -125,108 +125,141 @@ moduleNameGen :: Gen ModuleName
...
-variableGen :: MetaOrObject level => level -> Gen (Variable level)
-variableGen x = pure Variable
-    <*> scale (`div` 2) (idGen x)
-    <*> scale (`div` 2) (sortGen x)
+variableGen
+    :: MetaOrObject level
+    => [Variable level]
+    -> level
+    -> Gen (Variable level)
+variableGen vars x = oneof (newVar : if null vars then [] else [elements vars])
```
ttuegel: Why not this?
```suggestion
variableGen vars x = oneof (newVar : map elements vars)
```

src/main/haskell/kore/test/Test/Kore.hs
```diff
@@ -237,78 +270,125 @@ domainValueGen _ x =
...
-        , ForallPattern <$> forallGen childGen x
-        , IffPattern <$> iffGen childGen x
-        , ImpliesPattern <$> impliesGen childGen x
-        , InPattern <$> inGen childGen x
-        , NotPattern <$> notGen childGen x
-        , OrPattern <$> orGen childGen x
-        , TopPattern <$> topGen x
-        , VariablePattern <$> variableGen x
+patternGen vars childGen x =
+    frequency
```
ttuegel: What is the reason for using `frequency` here?

u--: VariablePattern is too rare to obtain nontrivial test cases otherwise.

u--: At least, that was what I thought when I was inspecting it manually. Maybe it's silly.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
u--: I think structural equality for the sorts is fine because you can't bind new sort variables.

src/main/haskell/kore/test/Test/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,60 @@
...
+    "alphaCompare x x == True" $
+    forAll (purePatternGen Object) $
+    (\(x :: CommonPurePattern Object) -> alphaEq x x)
+
+alphaCompareNegatives :: TestTree
+alphaCompareNegatives =
+    QC.testProperty
+    "alphaCompare x x == True" $
+    forAll pairs $
+    (\(x, y) -> not (alphaEq x y) || x == y)
```
ttuegel:
```suggestion
    (\(x, y) -> (x /= y) ==> not (alphaEq x y))
```

src/main/haskell/kore/test/Test/Kore.hs
ttuegel: Yeah, I think I meant `map pure vars`. With the original, there is a 50% chance to introduce a new variable, instead of a uniform probability to introduce a new variable or re-use a bound variable. I'm not sure which distribution is better for these tests.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,108 @@
...
+module Kore.ASTUtils.AlphaCompare
+    ( alphaEq
+    ) where
+
+import Data.List
+
+import Kore.AST.Common
+import Kore.ASTUtils.SmartPatterns
+
+-- alphaEq compares terms modulo renaming of bound variables
```
virgil-serbanuta: Does this work as documentation? Shouldn't it start with -- |?

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,108 @@
...
+        (App_ h1 c1, App_ h2 c2) ->
+            h1 == h2
+            && and (zipWith (go ctx1 ctx2) c1 c2)
+        (Var_ v1, Var_ v2) ->
+            case elemIndex v1 ctx1 of
+                Just i -> (Just i) == elemIndex v2 ctx2
+                Nothing -> v1 == v2
+        (DV_ s1 dv1, DV_ s2 dv2) ->
+            s1 == s2
+            && dv1 == dv2
```
virgil-serbanuta: Hm, map and list DVs can contain variables, shouldn't you also test alpha equivalence instead of equality for those?

If you change this, you should also change domainValueGen.

src/main/haskell/kore/test/Test/Kore/ASTUtils/AlphaCompare.hs
```diff
@@ -0,0 +1,60 @@
+{-|
```
virgil-serbanuta: I suppose that this is random-ish, but, when I checked test coverage, ctx2 was not evaluated in almost all functions, while next, rewrites, charliteral and stringliteral were not covered at all. Could you add tests that cover the entire code?

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
u--: It's implementation details, which I don't think belong in Haddock docs.

src/main/haskell/kore/test/Test/Kore/ASTUtils/AlphaCompare.hs
u--: That's just quickcheck not generating enough random terms.

`ctx1` and `ctx2` are only evaluated when the function has to compare a variable, and `ctx2` is not evaluated if the first variable being compared is free.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
u--: I can try.. I decided to just ignore DV for now, after all alpha equivalence is still just a subset of the possible equivalence relations we might care about.

src/main/haskell/kore/src/Kore/ASTUtils/AlphaCompare.hs
virgil-serbanuta: Fine, but this function is public, so it needs its own haddock documentation. Please add it.

PR #334, opened 2018-11-19
Equals for domain values.

virgil-serbanuta:
###### Reviewer checklist

---


virgil-serbanuta:
Jenkins: test this please

src/main/haskell/kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -190,9 +193,71 @@ makeEvaluateTerm
...
+        , SortedVariable variable
+        , Eq (variable level)
+        , Show (variable level)
+        )
+    => MetadataTools level StepperAttributes
+    -> BuiltinDomain (PureMLPattern level variable)
+    -> (Predicate level variable, SimplificationProof level)
+makeEvaluateBuiltinDomain
+    _tools
+    (BuiltinDomainPattern (StringLiteral_ _))
```
ttuegel: You might add a note that this is true if-and-only-if the pattern has passed the verifier. Of course, that should be true at this point. Eventually, I don't think we should even have string-literal domain values at this point in the stepper.

PR #336, opened 2018-11-20
Debugging helpers

virgil-serbanuta:
###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Debug.hs
```diff
@@ -0,0 +1,167 @@
...
+Module      : Kore.Debug
+Description : Debugging helpers.
+Copyright   : (c) Runtime Verification, 2018
+License     : UIUC/NCSA
+Maintainer  : virgil.serbanuta@runtimeverification.com
+Stability   : experimental
+Portability : portable
+
+Usage: When you have a function whose result is in a monad which is handled by
+the functions below, i.e.
```
traiansf: use e.g. instead of i.e. since you don't list all monads handled

src/main/haskell/kore/src/Kore/Debug.hs
```diff
@@ -0,0 +1,167 @@
...
+        Left err ->
+            endThing name ("error: " ++ show err)
+            $ Left err
+        Right r ->
+            endThing name ("result: " ++ show r)
+            $ Right r
+
+{-|Wraps a generic monad action for printing debug messages, similar to 'trace'.
+
+The monad must not have an error case.
```
traiansf: Maybe say why (the end case will be lost?)

src/main/python/debugFilter.py
```diff
@@ -0,0 +1,179 @@
+import sys
```
traiansf: Would it make sense to document this script, too?

virgil-serbanuta: I was seeing this as a throw-away script, but, you're right, If I'm adding this to source control, who knows what will happen, so I should document it.

PR #214, opened 2018-09-13
Refactor AST to have `domain` parameter.

u--:
###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`
- [ ] Style conformance: `stylish-haskell`

---

PR #337, opened 2018-11-21
Parameterize patterns on domains

ttuegel:
Parameterizing patterns on domains will be helpful when introducing pattern
annotations because it prevents awkward circular dependencies.

`Kore.Step.Pattern` introduces new patterns synonyms for pattern types specific
to execution phase, particularly `StepPattern` that serves as a replacement for
`PureMLPattern` (which has a new type parameter that is fixed throughout
execution).

The modules `Data.Functor.Impredicative` and `Data.Function.Compose` are no
longer used and have been removed.

The class `PatternSubstitutionClass` has been removed; it could be implemented
entirely from its `SubstitutionClass` superclass constraint.

Two domains are implemented in `Kore.Domain.Builtin` and
`Kore.Domain.External`. The latter refers to the external syntax of Kore domain
values, i.e. string literals. At the moment, everything uses the `Builtin`
domain, but eventually the parser and unparser should use the `External` domain
and the verifier should transform from `External` to `Builtin` representations
as part of pattern verification.

Replaces: #214 

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/AST/Kore.hs
```diff
@@ -31,171 +34,182 @@ module Kore.AST.Kore
...
 
-pattern UnifiedMetaPattern :: Pattern Meta var child -> UnifiedPattern var child
-pattern UnifiedMetaPattern pat = UnifiedPattern (UnifiedMeta (Rotate31 pat))
+    UnifiedObjectPattern
+        :: !(Pattern Object domain variable child)
+        -> UnifiedPattern domain variable child
 
-pattern UnifiedObjectPattern :: Pattern Object var child -> UnifiedPattern var child
-pattern UnifiedObjectPattern pat = UnifiedPattern (UnifiedObject (Rotate31 pat))
+$(return [])  -- Begin a new definition group where UnifiedPattern is in scope.
```
virgil-serbanuta: It may be obvious for the others in our team, but, for my education, could you explain what's happening here? I see the comment, but I don't get it - wouldn't UnifiedPattern be in scope below even without doing anything special?

ttuegel: @virgil-serbanuta `UnifiedPattern` is in scope for expressions, but not for quoting, so that quoting the type name below (`''UnifiedPattern`) would fail. Template Haskell limits the ways in which definitions can be mutually recursive; a quoted name must be defined in a previous _definition group_ before it can be used. A top-level Template Haskell splice always starts a new definition group, even if the splice is empty, as is the case here.

Thank you for asking about this. I doubt it is clear to everyone on the team; this is an unusual use of Template Haskell. I certainly have never seen this done outside this project!

virgil-serbanuta: Thank you!

PR #339, opened 2018-11-23
One-path step execution

virgil-serbanuta:
###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Kore/Step/Step.hs
```diff
@@ -260,6 +277,218 @@ test_simpleStrategy =
...
+            metadataTools
+            (Limit 0)
+            (ExpandedPattern.fromPurePattern Mock.a)
+            (simpleAxiom Mock.a mkBottom)
+            [simpleAxiom Mock.a Mock.b]
+            [simpleAxiom Mock.a Mock.c]
+        assertEqualWithExplanation ""
+            (ExpandedPattern.fromPurePattern Mock.a)
+            actual
+    , testCase "Axiom priority, first step" $ do
```
traiansf: I would expect here to get `bottom`, as it indicates progress.

src/main/haskell/kore/test/Test/Kore/Step/Step.hs
```diff
@@ -260,6 +277,218 @@ test_simpleStrategy =
...
+            metadataTools
+            (Limit 1)
+            (ExpandedPattern.fromPurePattern Mock.a)
+            (simpleAxiom Mock.d mkBottom)
+            [simpleAxiom Mock.a Mock.b]
+            [simpleAxiom Mock.a Mock.c]
+        assertEqualWithExplanation ""
+            (ExpandedPattern.fromPurePattern Mock.c)
+            _actual
+    , testCase "Axiom priority, second step" $ do
```
traiansf: Same here: expected `bottom`

src/main/haskell/kore/test/Test/Kore/Step/Step.hs
```diff
@@ -260,6 +277,218 @@ test_simpleStrategy =
...
+        -- Removal axiom: b => bottom
+        -- Coinductive axiom: b => c
+        -- Normal axiom: b => d
+        -- Normal axiom: a => b
+        -- Start pattern: a
+        -- Expected: b, since a->b->bottom and bottom is ignored
+        [(_actual, _proof)] <- runOnePathSteps
```
traiansf: I would expect bottom here, as well, as after `a -> b` you can apply `b -> bottom`

src/main/haskell/kore/test/Test/Kore/Step/Step.hs
```diff
@@ -260,6 +277,218 @@ test_simpleStrategy =
...
+            (ExpandedPattern.fromPurePattern Mock.a)
+            (simpleAxiom Mock.e mkBottom)
+            [simpleAxiom Mock.b Mock.c]
+            [ simpleAxiom Mock.b Mock.d
+            , simpleAxiom Mock.a Mock.b
+            ]
+        assertEqualWithExplanation ""
+            (ExpandedPattern.fromPurePattern Mock.c)
+            _actual
+        -- Removal axiom: e => bottom
```
traiansf: ? The comment here doesn't look right

src/main/haskell/kore/test/Test/Kore/Step/Step.hs
```diff
@@ -260,6 +277,218 @@ test_simpleStrategy =
...
+        -- Normal axiom: constr11(b) => g(b)
+        -- Normal axiom: constr11(c) => f(c)
+        -- Normal axiom: constr11(x) => h(x)
+        -- Normal axiom: constr10(x) => constr11(x)
+        -- Start pattern: constr10(x)
+        -- Expected:
+        --   f(a) and x=a
+        --   or (f(b) and x=b)
+        --   or (f(c) and x=c)
+        --   or (h(x) and x!=a and x!=b and x!=c )
```
traiansf: This might be ok for one-path, since here we want to constrain terms as much as possible, but for all-path, for example, here `h(x)` should be without constraints.

virgil-serbanuta: This is mandatory for one-path.

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -178,3 +305,71 @@ heatingCooling axiomStrategy axioms =
...
+    -> [axiom]
+    -- ^ normal axioms
+    -> Strategy (Prim axiom)
+onePathFirstStep destinationRemovalAxiom axioms =
+    Strategy.sequence
+        (  [axiomWithRemainderStep destinationRemovalAxiom]
+        ++ map
+            axiomOnRemainderStep
+            axioms
+        ++ [Strategy.apply nonEmptyRemainderFail]
```
ttuegel: Why should the step fail on a non-empty remainder? I guess this question might answer itself if I understood what the remainder was about.

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -178,3 +305,71 @@ heatingCooling axiomStrategy axioms =
...
+onePathFirstStep
+    :: axiom
+    -- ^ axiom with which to remove the destination we`re trying to reach
+    -> [axiom]
+    -- ^ normal axioms
+    -> Strategy (Prim axiom)
+onePathFirstStep destinationRemovalAxiom axioms =
+    Strategy.sequence
+        (  [axiomWithRemainderStep destinationRemovalAxiom]
+        ++ map
```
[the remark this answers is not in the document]
virgil-serbanuta: Because each is applied on the previous axiom's remainder.

src/main/haskell/kore/src/Kore/Step/Step.hs
virgil-serbanuta: I added a comment, but https://github.com/kframework/kore/blob/master/docs/2018-11-08-One-Path-Reachability-Proofs.md may help more, especially the "applying the axiom" section.

virgil-serbanuta:
@ttuegel I tried to improve the documentation, did I manage to make things more clear?

src/main/haskell/kore/src/Kore/Step/Step.hs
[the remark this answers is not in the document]
ttuegel: That somewhat abuses what `Strategy.sequence` means, which is: that each of the configurations is a child of the configuration reached by applying the preceding axioms. The remainder pattern is not a child configuration, but exists in parallel (at the same execution step; I understand that `Strategy.parallel` is not what you need here).

src/main/haskell/kore/src/Kore/Step/Step.hs
ttuegel: I was and am reading that document, but some connections are still unclear. For example, the terminology is different; I'm still not sure what corresponds to the "remainder" in that document. Another thing that is not clear is that everything in that document is conjoined at the top level, but in the code it is disjoined.

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -65,16 +69,96 @@ import qualified Kore.Step.Strategy as Strategy
 
 {- | A strategy primitive: a rewrite rule or builtin simplification step.
  -}
-data Prim rewrite = Simplify | Rewrite !rewrite
+data Prim rewrite =
```
ttuegel: When I wrote the strategy module, I imagined that we would use different "primitive rule" types for different modes of execution, and I think we could really benefit from that here. This could be much simpler if the one-path proof strategy used its own type different from the one used for concrete execution:
~~~.hs
data Prim rewrite = Simplify | Split ![rewrite]
~~~
The corresponding transition rule (`applySplit` vs `applyRewrite`) would apply each argument of `Split` to the remainders of the previous patterns. The `NonEmptyRemainderFail`, `CancelBottom` and `CancelRemainder` cases would not need to exist to preserve the concrete execution behavior because nothing about concrete execution would change. There wouldn't need to be an unusual use of `Strategy.sequence` (see my other comment). We would not have to introduce `StrategyPattern` or do extra post-processing of the execution tree.

src/main/haskell/kore/src/Kore/Step/Step.hs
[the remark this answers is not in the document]
virgil-serbanuta: I see. Here's how I was interpreting it at the strategy level, which does not know anything about axioms: we have a primitive strategy that produces two results (the rewrite and the remainder), and we follow that with a strategy that works only on the remainder. By definition, this is a sequence. Of course, it remains true that the remainder is not a true child, but we could change the way we think about strategies.

There are two other options. First, we could not use strategies. I'm not that happy with this, but it's a possibility.

Second, we could change Strategy.hs so that the patterns returned by a strategy are tagged as either as `Result` or as `Remainder` and we could have something like `daisyChain` as an alternative to `sequence`. Would we gain enough to do this, or is it better to keep strategies simple?

src/main/haskell/kore/src/Kore/Step/Step.hs
virgil-serbanuta: Added explanations for both.

PR #341, opened 2018-11-26
Expand a comment about template Haskell with Tom's explanations.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/AST/Kore.hs
+++ b/src/main/haskell/kore/src/Kore/AST/Kore.hs
@@ -62,6 +62,16 @@ data UnifiedPattern domain variable child where
         -> UnifiedPattern domain variable child
 
 $(return [])  -- Begin a new definition group where UnifiedPattern is in scope.
+              -- Without it, UnifiedPattern would be in scope for expressions,
+              -- but not for quoting, so using ''UnifiedPattern below would
+              -- fail.
+              --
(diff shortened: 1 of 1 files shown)


src/main/haskell/kore/src/Kore/AST/Kore.hs
```diff
@@ -62,6 +62,16 @@ data UnifiedPattern domain variable child where
         -> UnifiedPattern domain variable child
 
 $(return [])  -- Begin a new definition group where UnifiedPattern is in scope.
```
ttuegel: This technique is actually used many other places, for example in `Kore.AST.Common`. Would it be better to define somewhere:
```.hs
import Language.Haskell.TH

-- | Begin a new definition group and so on...
newDefinitionGroup :: Q [Dec]
newDefinitionGroup = return []
```

I think it would be overly repetitive to copy your comment to each place this is used, but I also think it would be good to have the technique documented if it is used so extensively.

vladciobanu: I'm not very familiar with TH, but I know of its reputation regarding the increase in compilation times. Would implementing this suggestion affect us in an observable way?

ttuegel: No, because we already use Template Haskell extensively.

ttuegel:
Merged in #374.

PR #346, opened 2018-11-28
One-path verification

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/test/Test/Kore/OnePath/Verification.hs
+++ b/src/main/haskell/kore/test/Test/Kore/OnePath/Verification.hs
@@ -1,0 +1,349 @@
+module Test.Kore.OnePath.Verification
+    ( test_onePathVerification
+    ) where
+
+import Test.Tasty
+       ( TestTree )
+import Test.Tasty.HUnit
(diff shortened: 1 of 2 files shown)


src/main/haskell/kore/test/Test/Kore/OnePath/Verification.hs
```diff
@@ -0,0 +1,349 @@
...
+        -- Expected: error a
+        actual <- runVerification
+            metadataTools
+            (Limit 0)
+            [simpleAxiom Mock.a Mock.b]
+            [simpleClaim Mock.a Mock.b]
+        assertEqualWithExplanation ""
+            (Left $ ExpandedPattern.fromPurePattern Mock.a)
+            actual
+    , testCase "Runs one step" $ do
```
traiansf: I was confused here because I thought it should verify it in one step. Maybe add a comment about it

traiansf: I was expecting it to succeed in one step. Maybe add a comment why.

PR #353, opened 2018-12-05
Added conditional debug functions

traiansf:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Debug.hs
+++ b/src/main/haskell/kore/src/Kore/Debug.hs
@@ -58,11 +58,18 @@ module Kore.Debug
     , traceExceptT
     , traceMaybeT
     , traceNonErrorMonad
+    , applyWhen
     ) where
 
 import Control.Monad.Trans.Except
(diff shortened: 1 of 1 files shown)


src/main/haskell/kore/src/Kore/Debug.hs
```diff
@@ -90,6 +92,28 @@ traceExceptT name startValues action = ExceptT $ do
...
 
+{-|Wraps an 'ExceptT' action for printing debug messages, similar to 'trace'.
+
+In addition to 'traceExceptT' it takes as the first argument a condition
+controlling when the action should be traced
+
+If the condition is true, it behaves as 'traceExceptT' otherwise it reverts
+to the action.
+-}
+traceWhenExceptT
```
virgil-serbanuta: I wonder if it's worth making a `traceWhen` function that would be used something like this:
`traceWhen condition traceExceptT "str1" "str2" action` and
`traceWhen condition traceMaybeT "str1" "str2" action`

PR #362, opened 2018-12-07
Revert "And idempotence (#352)"

traiansf:
This reverts commit 9b20c5aac7f3a069e6fb895f3d67c9857869b45f.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Simplification/And.hs
+++ b/src/main/haskell/kore/src/Kore/Step/Simplification/And.hs
@@ -13,22 +13,11 @@ module Kore.Step.Simplification.And
     , simplifyEvaluated
     ) where
 
-import Data.List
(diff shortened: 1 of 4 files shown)


virgil-serbanuta:
We assumed that it introduced a large performance issue, the build time seems to have increased ~3x after it. We expected the original commit to have a performance penalty, but not that big, so this was the most likely cause for the issue. However, this build's time is similarly bad, so we're not so sure anymore.

ttuegel:
> We assumed that it introduced a large performance issue, the build time seems to have increased ~3x after it.

Do you mean a performance issue in `kore-exec` itself? Or a performance issue compiling `kore` with GHC?

traiansf:
I'm closing this since it does not seem to affect performance.

PR #379, opened 2018-12-12
implement STRING.chr builtin

dfilaretti:
Implements the `STRING.chr` builtin. 

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Builtin/String.hs
+++ b/src/main/haskell/kore/src/Kore/Builtin/String.hs
@@ -37,6 +37,8 @@ import           Control.Error
                  ( MaybeT )
 import           Control.Monad
                  ( void )
+import           Data.Char
+                 ( chr )
 import qualified Data.HashMap.Strict as HashMap
 import           Data.List
                  ( findIndex, isPrefixOf, tails )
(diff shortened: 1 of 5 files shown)


src/main/haskell/kore/src/Kore/Builtin/String.hs
```diff
@@ -380,6 +385,27 @@ evalString2Int = Builtin.functionEvaluator evalString2Int0
...
+    evalChr0 _ _ resultSort arguments =
+        Builtin.getAttemptedFunction $ do
+            let _n =
+                    case arguments of
+                        [_n] -> _n
+                        _    -> Builtin.wrongArity lengthKey
+            _n <- Int.expectBuiltinInt chrKey _n
+            Builtin.appliedFunction
+                . asExpandedPattern resultSort
+                $ [ toEnum (fromIntegral _n) ]
```
vladciobanu: Minor nitpick: I would find it easier to read this line if it used `chr` (from `Data.Char`) instead of `toEnum`, but I don't feel too strongly about it.

dfilaretti: I agree, `chr` is definitely easier to read then `toEnum`. I am going to fix this now.

PR #381, opened 2018-12-13
Added Set.set2list builtin

vladciobanu:
###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Builtin/Set.hs
```diff
@@ -294,16 +299,43 @@ evalDifference =
...
+        -> [StepPattern Object variable]
+        -> Simplifier (AttemptedFunction Object variable)
+    evalToList0 tools _ resultSort arguments =
+        Builtin.getAttemptedFunction $ do
+            let _set =
+                        case arguments of
+                            [_set] -> _set
+                            _      -> Builtin.wrongArity toListKey
+            _set <- expectBuiltinSet toListKey tools _set
+            Builtin.appliedFunction
```
ttuegel: There is a function [`returnList`](https://github.com/kframework/kore/blob/master/src/main/haskell/kore/src/Kore/Builtin/List.hs#L145) which you could export from `Kore.Builtin.List`:
```.hs
returnList . Seq.fromList . Set.toList $ _set
```

src/main/haskell/kore/src/Kore/Builtin/Set.hs
```diff
@@ -370,61 +402,91 @@ asExpandedPattern symbols resultSort =
     asExpandedPattern0 = \asPattern0 builtin ->
         ExpandedPattern.fromPurePattern $ asPattern0 builtin
 
+concatKey :: String
+concatKey = "SET.concat"
```
virgil-serbanuta: Could you use these constants in other places where their value is used, e.g. Test.Kore.Builtin.Definition?

src/main/haskell/kore/src/Kore/Builtin/Set.hs
vladciobanu: I didn't think I can do that because I needed to call `fromConcretePurePattern`, but I can do this instead:

```Haskell
List.returnList resultSort
    . fmap Kore.fromConcretePurePattern
    . Seq.fromList
    . Set.toList
    $ _set
```

src/main/haskell/kore/src/Kore/Builtin/Set.hs
vladciobanu: I'll add the `Set` ones now and create a different PR to update all builtins in the same way.

vladciobanu:
Jenkins: test this please.

PR #385, opened 2018-12-17
Add an 'OrOfPredicateSubstitution'

virgil-serbanuta:
###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/ExpandedPattern.hs
```diff
@@ -257,19 +277,20 @@ top =
 
 {-| 'isTop' checks whether an ExpandedPattern is equivalent to a top Pattern.
 -}
-isTop :: ExpandedPattern level variable -> Bool
```
traiansf: Why not renamining dropping `Term` from `TopBottomTerm` and its methods, and making `Predicated` instance of `TopBottom`?  `instance TopBottom term => instance TopBottom (Predicated level variable term) where ...` ?

PR #389, opened 2018-12-19
Trusted claims

traiansf:
###### Reviewer checklist

---

You should probably review the commits separately, as the `Refactoring IndexedModule to include axiom attrs` touches a lot of code (and does not do much).

Also, it will probably be untestable until https://github.com/kframework/k/pull/225 is merged

### DIFF
--- a/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
+++ b/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
@@ -11,8 +11,12 @@ module Kore.Attribute.Trusted
     , trustedId, trustedSymbol, trustedAttribute
     ) where
 
+import           Control.DeepSeq
+                 ( NFData )
 import qualified Control.Monad as Monad
 import           Data.Default
+import           GHC.Generics
+                 ( Generic )
 
 import           Kore.AST.Kore
 import           Kore.Attribute.Parser
@@ -22,7 +26,9 @@ import qualified Kore.Attribute.Parser as Parser
 {- | @Trusted@ represents the @trusted@ attribute for claim sentences.
  -}
 newtype Trusted = Trusted { isTrusted :: Bool }
-    deriving (Eq, Ord, Show)
+    deriving (Eq, Ord, Show, Generic)
+
+instance NFData Trusted
 
 instance Default Trusted where
     def = Trusted False
--- a/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
+++ b/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
@@ -1,9 +1,13 @@
 {-|
 Module      : Kore.Attribute.Trusted
-Description : Trusted claim sentence attribute
+Description : Attribute identifying trusted claim sentences
 Copyright   : (c) Runtime Verification, 2018
 License     : NCSA
-Maintainer  : thomas.tuegel@runtimeverification.com
+Maintainer  : traian.serbanuta@runtimeverification.com
+
+
+A trusted claim is a reachability logic verification claim
+which can be used as a circularity without needing to be proven.
 
 -}
 module Kore.Attribute.Trusted
--- a/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
+++ b/src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
@@ -1,0 +1,60 @@
+{-|
+Module      : Kore.Attribute.Trusted
+Description : Trusted claim sentence attribute
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : thomas.tuegel@runtimeverification.com
+
+-}
+module Kore.Attribute.Trusted
+    ( Trusted (..)
+    , trustedId, trustedSymbol, trustedAttribute
+    ) where
+
+import qualified Control.Monad as Monad
+import           Data.Default
+
+import           Kore.AST.Kore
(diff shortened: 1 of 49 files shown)


src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
```diff
@@ -0,0 +1,60 @@
+{-|
+Module      : Kore.Attribute.Trusted
+Description : Trusted claim sentence attribute
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : thomas.tuegel@runtimeverification.com
```
virgil-serbanuta: Why aren't you the maintainer?

src/main/haskell/kore/src/Kore/Attribute/Trusted.hs
```diff
@@ -0,0 +1,60 @@
+{-|
+Module      : Kore.Attribute.Trusted
+Description : Trusted claim sentence attribute
```
virgil-serbanuta: How about "Attribute identifying trusted claim sentences"?

Regardless of whether you like my suggestion or not, could explain in a few words what is a "trusted claim"?

src/main/haskell/kore/src/Kore/Exec.hs
```diff
@@ -348,24 +349,26 @@ prove limit definitionModule specModule = do
         (rewriteAxioms, simplifier, substitutionSimplifier) =
             axiomsAndSimplifiers
     specAxioms <-
-        simplifyRewriteAxioms tools
+        mapM (sequence . fmap (simplifyRewriteRule tools))
```
virgil-serbanuta: Could you extract the "sequence . fmap (simplifyRewriteRule tools)" or something similar to a separate function that has a signature?

src/main/haskell/kore/src/Kore/Exec.hs
```diff
@@ -348,24 +349,26 @@ prove limit definitionModule specModule = do
...
 
     return $
         Bifunctor.first
             (give symbolOrAlias toMLPattern)
             result
 
   where
-    makeClaim claim = (claim, limit)
+    makeClaim (attributes, rule) = Claim { rule , attributes }
+    extractUntrustedClaims =
```
virgil-serbanuta: Could you add a signature for this function?

src/main/haskell/kore/src/Kore/IndexedModule/IndexedModule.hs
```diff
@@ -123,14 +126,60 @@ data IndexedModule param pat atts =
         -- ^ map from builtin domain (symbol and sort) identifiers to the hooked
         -- identifiers
     }
-    deriving (Functor, Generic, Show)
+    deriving (Generic, Show)
+
+instance Bifunctor (IndexedModule param pat) where
```
virgil-serbanuta: Is this really a good idea? I mean, when someone is reading the code, why would it make sense that "first" and "second" for an indexed module would refer to a rather obscure part of it (i.e. one kind of attributes)?

I know that those are the last two arguments, and Bifunctor works that way, but does this actually make sense, or does it look more like a hack that saves a bit of work?

Assuming that my questions make some sense, then why not make it a Trifunctor? That way you're not making it a functor over the obscure things only, you're making it a functor over (almost) everything.

My questions are genuine, not rethorical.

src/main/haskell/kore/test/Test/Kore/OnePath/Verification.hs
```diff
@@ -262,6 +265,42 @@ test_onePathVerification = give symbolOrAliasSorts
         assertEqualWithExplanation ""
             (Left $ ExpandedPattern.fromPurePattern Mock.b)
             actual
+    , testCase "trusted second proves first" $ do
```
virgil-serbanuta: Could you also add a test that trusted claims are not used in the first step? E.g. from
trusted claim: a=>b
claim: a=>b
you get a failure?

src/main/haskell/kore/src/Kore/IndexedModule/IndexedModule.hs
traiansf: Well, I'm not sure.  I was just continuing existing work which made it a `Functor`.  On the other hand, the only usage is to replace the attributes by `Null`.  I guess it made more sense when it was just a `Functor` and the instance was automatically derived.

vladciobanu: I also found that bit weird even from before (that `IndexedModule` is a `Functor` for its attributes. Perhaps lenses would be a better choice here?

traiansf: I just went on an removed teh Bifunctor instance. If we ever need it for more than one function, we can revisit this issue.

traiansf:
@virgil-serbanuta are you satisfied with the changes?

dwightguth:
Do we really want to replicate the trusted attribute of the java backend? wouldn't it be better to simply tell the prover when you prove which claims you want to prove and which claims you want to trust?

traiansf:
@dwightguth I don't have anything against using a different mechanism, but that would have to be designed and work from the frontend towards the backend.  We could separate them in different modules and send two such modules to the backend, but I'm not sure it would be such a great improvement. 

Also, languages such as Coq sometimes use something like `Admitted` to accept a result without proof.

traiansf:
Jenkins: test this please

PR #392, opened 2018-12-19
Remove SymbolOrAliasSorts

ttuegel:
Remove `SymbolOrAliasSorts` from `MetadataTools`.

Removes modules `Kore.ASTUtils.SmartConstructors` and `Kore.ASTUtils.SmartPatterns`. New constructors and pattern synonyms for verified patterns reside in `Kore.AST.Valid` with helpers defined in `Kore.AST.Lens`.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Predicate/Predicate.hs
```diff
@@ -120,14 +117,29 @@ We should consider deleting this and implementing the functionality otherwise.
...
+outside of that.  We should consider deleting this and implementing the
+functionality otherwise.
+
+ -}
 unwrapPredicate :: Predicate level variable -> StepPattern level variable
 unwrapPredicate (GenericPredicate p) = p
 
+{- | Return the 'StepPattern' corresponding to the given 'Predicate'.
+
+All predicates are flexibly-sorted; the sort argument is used to force the
```
virgil-serbanuta: This is true in in practice, but it's not necessarily true, in theory we could have symbols that only evaluate to top and bottom (@traiansf unless we previously agreed to wrap those in ceil, but I don't remember that; if we did, we should have a design decision doc).

How about prefixing this with "In practice"?

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -176,6 +172,15 @@ instance
...
         ConfigurationVariable $ freshVariableWith a n
 
+instance
+    Unparse (variable level) =>
+    Unparse (StepperVariable variable level)
+  where
+    unparse =
+        \case
+            AxiomVariable var -> unparse var
+            ConfigurationVariable var -> unparse var
```
virgil-serbanuta: Unparsing these the same way makes me somewhat nervous. It will probably work in most cases (almost all?), but in the rare case when we have two distinct variables unparsing to exactly the same thing, reparsing the resulting string will produce a pattern with a completely different meaning; also, debugging output will be confusing.

Prefixing the variables somehow would make the problem less likely, and would probably make the debug output more readable.

src/main/haskell/kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -118,39 +115,34 @@ makeEvaluate tools child
 makeEvaluateNonBoolCeil
     ::  ( MetaOrObject level
         , SortedVariable variable
-        , Show (variable level)
         , Ord (variable level)
+        , Show (variable level)
```
virgil-serbanuta: Could we group these two constraints into a single one, DebugVariable or something similar (not necessarily in this PR)? I guess that's their purpose, so we should be explicit about what happens.

src/main/haskell/kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -161,50 +153,37 @@ makeEvaluateTerm
...
+    case projected of
+        ApplicationPattern app
+          | StepperAttributes.isTotal headAttributes ->
+            let
+                (ceils, _proofs) = unzip (map (makeEvaluateTerm tools) children)
+                result = makeMultipleAndPredicate ceils
+            in
+                (result, SimplificationProof)
+          where
+            Application { applicationSymbolOrAlias = patternHead } = app
```
virgil-serbanuta: I'm starting to like this way of extracting field values.

virgil-serbanuta:
Oh, one more thing: if I remember correctly, you said that you expected some performance improvement. It would be nice if that information would be present in the PR description and if it would preserved in the commit history after merging.

vladciobanu:
It would be even nicer if we could somehow use Jenkins to time at least the integration suite and export the data somewhere, to generate a report. Now that I think about it, would be interesting to also track coverage and docs.

src/main/haskell/kore/src/Kore/AST/Lens.hs
```diff
@@ -0,0 +1,304 @@
+{-|
+Module      : Kore.AST.Lens
+Description : Lenses into Kore patterns
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : phillip.harris@runtimeverification.com
```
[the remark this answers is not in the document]
ttuegel: These were originally part of `Kore.ASTUtils.SmartConstructors` (which Phillip wrote). I moved them into their own module because I was removing the rest of that module and I thought it would be better to delete it than leave a module called "SmartConstructors" which contained no smart constructors.

src/main/haskell/kore/src/Kore/AST/Lens.hs
```diff
@@ -0,0 +1,304 @@
...
+            { applicationSymbolOrAlias
+            , applicationChildren
+            }
+      =
+        Application applicationSymbolOrAlias
+            <$> traverse lensChild applicationChildren
+
+-- | The sort returned by a top level constructor.
+-- NOTE ABOUT NOTATION:
+-- In the this haskell code, this is always `s2`.
```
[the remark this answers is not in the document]
ttuegel: I'm going to remove this note because it is obsolete: the notation it documents was used in `Kore.ASTUtils.SmartConstructors`, but not anymore in this module.

src/main/haskell/kore/src/Kore/Predicate/Predicate.hs
ttuegel: I will add the prefix, but note that `Application` patterns can only be turned into a `Predicate` by `wrapPredicate`, which has a note about being unsafe.

src/main/haskell/kore/src/Kore/Parser/ParserImpl.hs
```diff
@@ -459,13 +459,13 @@ BNF definitions:
 @
 ⟨variable⟩ ::= ⟨object-variable⟩ | ⟨meta-variable⟩
 @
-unifiedVariableParser :: Parser (UnifiedVariable Variable)
+-}
+unifiedVariableParser :: Parser (Unified Variable)
```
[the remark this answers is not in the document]
traiansf: I think that we originally though we might need it, and it somehow seemed to fit into the overall parsing picture;  if it's not used, we should drop  it; it's not hard to write it again if we decide it's useful later.

ttuegel:
@virgil-serbanuta This is ready for re-review now. I will have performance numbers tomorrow, but I would expect < 5% change from this. `SymbolOrAliasSorts` was the cheapest lookup in `MetadataTools`. The biggest gains will come from attribute annotations, but I started with sort annotations because I think that is easier.

src/main/haskell/kore/src/Kore/Parser/ParserImpl.hs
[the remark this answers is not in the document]
ttuegel: It's used in the `Unparse` tests to check that we can correctly do the parsing-unparsing round-trip.

src/main/haskell/kore/src/Kore/Step/Simplification/Ceil.hs
ttuegel: Yes, I will make a separate pull request for that. I think it should be a class for any syntactic object, not just variables; the `Unparse (variable level)` is just getting pulled in to satisfy an `Unparse (Pattern level domain variable child)` instance somewhere. But at this location, the constraint could just be `DebugSyntax (variable level)`.

PR #393, opened 2018-12-20
Remove the K submodule from Kore

ehildenb:
###### Reviewer checklist

---

This is in preparation to have the K repository have Kore as a submodule.
I'm in the process of making sure KEVM works as before with this setup, probably wait for that to merge.


ehildenb:
Jenkins: test this please

ehildenb:
This is ready for review, I've confirmed that it works with KEVM.

Once this is in, I'll fix the submodule issues on https://github.com/kframework/k/pull/228 and then we can get the corresponding branch in KEVM in.

ehildenb:
Alright, looks like I'm putting the coverage/documentation reports in the correct places now. This is ready for review if https://github.com/kframework/k/pull/228 passes.

Makefile
```diff
@@ -1,70 +0,0 @@
-include include.mk
```
ttuegel: Could we leave the `Makefile`? Some of the targets are useful for developers, such as `stylish` and `docs`. We could remove all the Jenkins-specific stuff, though.

scripts/jenkins
```diff
@@ -0,0 +1,58 @@
...
+# ---------------
+./scripts/git-assert-clean.sh \
+    || fatal "Please commit your changes!"
+
+./scripts/git-rebased-on.sh "$(git rev-parse $MASTER_BRANCH)" --linear \
+    || fatal "Please rebase your branch on 'master'!"
+
+notif "Stylish ..."
+# -----------------
+for hs_file in $(find ${HS_SOURCE_DIRS[@]} -name '*.hs' -o -name '*.hs-boot'); do
```
ttuegel: I think `find -exec` returns a non-zero exit code if `stylish-haskell` fails on any file; this should do the same.

scripts/jenkins
```diff
@@ -0,0 +1,58 @@
...
+    stylish-haskell -i "$hs_file"
+done
+./scripts/git-assert-clean.sh \
+    || fatal "Please commit fixed stylizations!"
```
ttuegel: This is hard(er) to do without a `make stylish` target!

scripts/jenkins
ehildenb: I have `set -e` at the top, which means that if at any point in the process a process returns a non-zero exit code the whole script will exit with that code.

traiansf:
@ehildenb I don't like the fact that we changed the way we call Jenkins in a PR which was not just that.

Now all others PRs are blocked until this gets merged.  Also, I might be missing something, but our old tests included some integration tests which are found neither here nor in the K PR.

I also subscribe to @ttuegel 's comment that the `Makefile` was also used by us and not just Jenkins.

Makefile
ehildenb: I'm slightly opposed to using a `Makefile` when the targets aren't directly files, because then you end up over-building or under-building some of the phony targets. I can put it back, but was wondering if you'd be ok with changing `./script/jenkins => ./scripts/build` and making it have subcommands like `./scripts/build [clean|check|stylish|build|test|docs]`, then the Jenkins script will just do:

```
#!/usr/bin/env bash
set -e

./scripts/build clean
./scripts/build check
./scripts/build stylish
./scripts/build build
./scripts/build test
./scripts/build docs
```

Can also put it at `./build` instead of `./scripts/build`.

Let me know which you prefer, I'll get it in ASAP.

ehildenb: One argument against multiple scripts is that we'll have to make sure any environment setup they have is synchronized. We could do this by having a `./build/env` which is `source`d by the other scripts.

If we do it as one script, I would use this style https://github.com/kframework/evm-semantics/blob/82c71231f74ed96d042494b680105e1a996235ed/kevm#L199. We lose tab-completion, but it is fairly straightforward/lightweight.

ehildenb:
Regarding running the K integration tests, would you all prefer:

(1) Clone and build a copy of `master` K to a specified location (not using submodules to avoid recursive submodules), to ensure that each PR works on current `master` K, or
(2) Rely on K being on PATH.

I prefer (1) because I think it will work better on the CI machines.

dwightguth:
I mean one possible solution for (2) would be to grab the artifact built by the latest k-master job and extract it into the workspace for this job.

Makefile
ttuegel: > We could do this by having a `./build/env` which is `source`d by the other scripts.

I think this is fine. It also has the advantage that a developer could source that file into their own shell to do debugging.

ehildenb:
I guess I prefer solutions which don't rely on Jenkins-isms because it increases the amount of external code we rely on for the build process (ie. what if the copy-artifact plugin stops being maintained, or we migrate away from Jenkins).

We could also download the latest nightly release tarball directly, which would avoid having to call `mvn package -DskipTests`, and will always be behind latest K master.

ttuegel:
> I mean one possible solution for (2) would be to grab the artifact built by the latest k-master job and extract it into the workspace for this job.

I think this is a good idea for the Jenkins job. I would prefer (2) because when I run the tests locally, I probably have the version of K on `PATH` that I want to use.

ehildenb:
Alright, I'm pulling changes to the build system out into another PR, so that we can get the dependency reversed with this build system then focus on how we want the build system to change later as @traiansf pointed out. I'll work on the "where do we get K from" issue now, it sounds like we have 2 votes for Jenkins copy-artifact plugin, and 1 for `git clone && cd && mvn package -DskipTests`, so I'll setup the Jenkins copy-artifact plugin.

Makefile
ehildenb: Going to move this discussion to a future PR.

dwightguth:
@ehildenb that's fine if you can get it to work but I don't know if github has a URL to download the latest pre-release.

ehildenb:
@ttuegel this is ready for re-review, overview of what this PR does:

-   Parameterized the `Makefile` in variables `NPROCS` and `UPSTREAM_BRANCH` for overriding how many cores to run integration tests on and what branch to make sure we're rebased on, respectively.
-   Remove the K submodule, change to `k-frontend` target which unpacks and builds the latest k-nightly release.
-   The Makefile based integration testing now does not call the build of K before (which was resulting in over-building), but assumes that you have already built K (9d128312).
-   Many files have `dos2unix` ran on them, and many whitespace errors are taken care of (this is to quiet the K checkstyle in the branch where the haskell backend is added as a submodule to the K frontend).
-   Testing is sped up a bit by calculating the number of cores and running the integration tests on all cores (from 17min -> 11min).

The changes regarding switching from a `Makefile` to a `build` script have been pulled out and are in my development branch for after we get the dependency K -> Kore flipped.

Changes important for developers:

-   You must run `make k-frontend` prior to trying to do any of the integration tests. `all` target will also build `k-frontend`.
-   Manually set `NPROCS` lower if needed by doing `make NPROCS=N TARGET`.

ehildenb:
@dwightguth good point. In the PR where I add the Haskell backend as a submodule, I'll just ignore checkstyle for this directory as well.

I figure since the tests pass here, it's OK to leave the changes which fix the `dos2unix` and whitespace errors.

Makefile
```diff
@@ -3,12 +3,20 @@ include include.mk
...
         test test-kore test-k
 
 kore:
 	stack build $(STACK_BUILD_OPTS)
 
+k-frontend:
+	mkdir -p $(BUILD_DIR)
+	rm -rf $(K_DIR) $(K_NIGHTLY)
+	wget --output-document $(K_NIGHTLY) \
+	    $$(curl 'https://api.github.com/repos/kframework/k/releases' | jq --raw-output '.[0].assets[0].browser_download_url')
```
ttuegel: Please add a note to `README.md` which documents the dependency on `curl` and `jq`.

Makefile
```diff
@@ -3,12 +3,20 @@ include include.mk
...
+	mkdir --parents $(K_DIR)
+	tar --extract --verbose --file $(K_NIGHTLY) --strip-components 1 --directory $(K_DIR)
```
ttuegel: I don't think `tar` needs to be `--verbose`.

Makefile
```diff
@@ -28,28 +36,23 @@ test-kore:
 	fi
 
 test-k:
-	$(MAKE) -C src/main/k/working test-k
-
-jenkins: distclean check all test docs
+	$(MAKE) --directory src/main/k/working test-k -j$(NPROCS)
```
ttuegel: Using multiple jobs here creates a race condition; right now my machine is running 16 processes of `stack build`. This could probably be mitigated by making the `test-k` target build `kore-exec` first; Stack does rebuild-avoidance but it does not do locking (intentionally, AFAICT).

ehildenb: I'll make the `test-k` target depend on the `kore` target. I'm still not a fan of this stacking of phony targets, as it results in things being rebuilt too often. But that can wait for another PR.

Makefile
```diff
@@ -3,12 +3,20 @@ include include.mk
...
+.PHONY: all clean docs haddock jenkins kore k-frontend stylish \
         test test-kore test-k
 
 kore:
 	stack build $(STACK_BUILD_OPTS)
 
+k-frontend:
+	mkdir -p $(BUILD_DIR)
+	rm -rf $(K_DIR) $(K_NIGHTLY)
+	wget --output-document $(K_NIGHTLY) \
```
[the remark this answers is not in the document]
ehildenb: Done.

Makefile
ehildenb: Done.

Makefile
ehildenb: Removed.

README.md
```diff
@@ -33,3 +33,15 @@ The `/src` directory contains a parser for the Kore language implemented
 in scala.
 
 The `/src/test` directory contains a collection of Kore definition examples.
+
+## Build Dependencies
+
+-   [curl](https://curl.haxx.se/) and [jq](https://stedolan.github.io/jq/) for downloading and building latest nightly K.
+-   [git](https://git-scm.com/) for checkstyle on the repository history and files.
+-   [stack](https://www.haskellstack.org/) for building the Kore project.
```
ttuegel: Stack is only required to build the project using `make`:
```suggestion
-   [stack](https://www.haskellstack.org/) for building the Kore project using `make`.
```
Any Haskell build tool would suffice to build the project.

ttuegel:
The changes to `--coverage` and the integration test dependency on `$(KORE_EXEC)` look acceptable to me. You should probably remove the `$(KORE_EXEC)` target from `include.mk`.

Unfortunately, the integration tests will now behave badly if we forget to run `stack build` before `make test-k`. A different way to remove the integration test race condition--that would fail more nicely--would be to leave the dependency on `$(KORE_EXEC)` but remove the target. `make` would then complain that it can't find `kore-exec` and doesn't know how to build it, which would serve as an appropriate reminder to run `stack build`. Otherwise, `make` will attempt the test and `krun` will complain about a missing `kore-exec`.

ehildenb:
I like that! Will make the change.

ttuegel:
@ehildenb Would you mind to rebase this?

ttuegel:
I cannot reproduce the (apparently transient) build failure, so I'm going to merge this.

PR #395, opened 2019-01-07
Added builtins and added a hack for a simplification bug

vladciobanu:
Builtins added: Set.size, String.ord, String.chr, and Krypto.keccak256

I will create a card about the technical debt created in the simplification code. I think I'll need @virgil-serbanuta 's help in fixing it.

However, please note that this actually fixes a lot of EVM programs that were not working before.

###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Kore/Builtin/String.hs
```diff
@@ -106,6 +106,34 @@ test_length =
         (Test.Int.asExpandedPattern 0)
     ]
 
+test_chr :: [TestTree]
+test_chr =
+    [ testString
+        "48 == '0'"
```
traiansf: you could consider writing a more informative string here and below

src/main/haskell/kore/src/Kore/Step/Simplification/PredicateSubstitution.hs
```diff
@@ -110,7 +110,7 @@ simplify
         traverse
             (`substitute` unifiedSubstitution)
             predicate
-    if substitutedPredicate == predicate && not first
+    if substitutedPredicate == predicate && times > 1
```
virgil-serbanuta: I don't think this is the right fix. If a bool is not enough here, i.e. if the formula is not simplified in the first iteration, then there is no guarantee that one more iteration would be enough.

IIRC the main issue was caused by simplification not descending into builtins (e.g. maps), which is needed after substitutions.

If you insist on keeping this, please change the RP description with
`s/fixed a simplification bug/made an ugly hack to work around builtin simplification/`
and please provide an argument for why you prefer this hack instead of the proper fix.

src/main/haskell/kore/src/Kore/Builtin/Krypto.hs
```diff
@@ -0,0 +1,101 @@
+{- |
+Module      : Kore.Builtin.Krypto
+Description : Built-in cryptographic functions.
+Copyright   : (c) Runtime Verification, 2018
```
virgil-serbanuta: 2019?

PR #399, opened 2019-01-09
STRING.ord should return bottom if the input's length is not 1

vladciobanu:
I will add integration tests with my next PR ( https://trello.com/c/A1yfIgnB/409-correctly-sanitize-variables-in-exists-and-forall )

###### Reviewer checklist

---


traiansf:
maybe add a unit test with empty and more that one characters?

PR #402, opened 2019-01-11
Basic functionality for declaring smt-lemmas to SMT

u--:
This had to be walked back a bit from the original goal. Rather than declare every symbol in Kore as an SMT symbol, we expect the user to declare them themselves in the SMT prelude. This allows for EVM's `chop` lemma but not arbitrary reasoning about Kore terms. 

In order to extend SMT's capabilities in Kore we need to figure out how to deal with collections, which are monomorphic in K but polymorphic in Kore and SMT.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/TranslateSMT.hs
```diff
@@ -0,0 +1,261 @@
+{-|
+Module      : Kore.Step.TranslateSMT
+Description : Evaluates conditions.
+Copyright   : (c) Runtime Verification, 2018
```
virgil-serbanuta: 2019?

src/main/haskell/kore/src/Kore/Step/TranslateSMT.hs
```diff
@@ -0,0 +1,261 @@
...
+License     : NCSA
+Maintainer  : thomas.tuegel@runtimeverification.com
+Stability   : experimental
+Portability : portable
+-}
+
+module Kore.Step.TranslateSMT
+    ( translatePredicate
+    , Translator
```
virgil-serbanuta: This looks like a refactoring, so you can ignore this comment, but all exported stuff should be documented.

src/main/haskell/kore/src/SMT.hs
```diff
@@ -191,11 +203,27 @@ runSMT config SMT { getSMT } = do
...
+-- to escape special chars
+escapeId :: Text -> Text
+escapeId name = "|" <> name <> "|"
+
 declare :: MonadSMT m => Text -> SExpr -> m SExpr
 declare name typ =
     liftSMT $ withSolver $ \solver ->
         SimpleSMT.declare solver name typ
 
+declareFun :: MonadSMT m => Text -> [SExpr] -> SExpr -> m SExpr
```
virgil-serbanuta: Please document exported symbols (here and below). I'm not sure if above you intended that comment to work as documentation (probably not), but you should also make sure that `escapeId` has documentation.

src/main/haskell/kore/src/SimpleSMT.hs
```diff
@@ -459,6 +459,11 @@ declareFun proc f as r =
   do ackCommand proc $ fun "declare-fun" [ Atom f, List as, r ]
      return (const f)
 
+declareSort :: Solver -> Text -> Int -> IO SExpr
```
virgil-serbanuta: Please document.

src/main/k/working/search/search.k
```diff
@@ -8,6 +8,8 @@ module SEARCH
       | "final1" | "final2"
       | "unreachable"
 
+  rule 123 => 456 [smt-lemma]
```
virgil-serbanuta: I guess that this tests that the code does not crash and (maybe) that the SMT accepts the syntax you are sending. Could you add some meaningful tests for the new code (preferably both unit tests and integration ones)?

ttuegel:
@u-- Can this pull request be tested properly now that the `concrete` attribute is implemented?

src/main/haskell/kore/src/Kore/Step/SmtLemma.hs
```diff
@@ -0,0 +1,133 @@
...
+
+normalizeRule
+    :: ( Ord (var Object)
+       , Unparse (var Object)
+       )
+    => StepPattern Object var
+    -> StepPattern Object var
+normalizeRule pat =
+    case pat of
+        Rewrites_ s a b -> mkEquals s a b -- `s` is arbitrary/dummy value
```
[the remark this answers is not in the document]
u--: This is how they get translated to SMT. A rule like `1 + 1 => 3 [smt-lemma]` becomes an equality axiom to SMT.

u--:
@ttuegel Probably.

src/main/haskell/kore/src/Kore/Sort.hs
```diff
@@ -104,6 +105,10 @@ substituteSortVariables substitution sort =
                     substituteSortVariables substitution <$> sortActualSorts
                 }
 
+isConcreteSort :: Sort level -> Bool
```
ttuegel: Should this be `isActualSort`, to keep the terminology consistent?

u--:
Actually they were addressed, github just doesn't update it sometimes.

PR #403, opened 2019-01-11
AxiomPatterns: Fix the form of rewrite rules

ttuegel:
See also: kframework/k#241
Trello: https://trello.com/c/GmxON75i

###### Reviewer checklist

---


ttuegel:
I am alarmed that those tests passed locally. :confused:

traiansf:
@ehildenb: as discussed in the #kore channel, we will disallow the usage of ensures for rules defining a language (which have an existential flavor), and will only be using ensures for properties to be proven (and maybe some trusted lemmas) for which we actually want to ensure that the condition holds.

PR #405, opened 2019-01-11
Sanity checks for function evaluation.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
+++ b/src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
@@ -14,6 +14,7 @@ module Kore.Step.Function.Matcher
 
 import           Control.Applicative
                  ( (<|>) )
+import qualified Control.Comonad.Trans.Cofree as Cofree
 import           Control.Error.Util
                  ( just, nothing )
 import           Control.Monad.Counter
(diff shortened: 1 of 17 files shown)


src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -454,6 +464,13 @@ matchVariableFunction
                     ]
 matchVariableFunction _ _ _ _ = nothing
 
+data ArbitrarySymbols a = ArbitrarySymbols
```
ttuegel: Does "arbitrary" here mean "not constructor-like"?

src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -478,14 +495,25 @@ matchNonVarToPattern
...
                 tools substitutionSimplifier first second
         (return . return)
             (fmap secondVariablesSubstitutionToPredicate result)
+    let
+        firstVariables = freePureVariables first
+        resultVariables =
+            foldMap
+                (freeVariablesUnderArbitrarySymbols tools)
+                (OrOfExpandedPattern.extractPatterns finalResult)
+    if null (Set.intersection firstVariables resultVariables)
```
ttuegel: Could you add a comment here explaining what the logic is? I'm struggling to understand the big picture here.

virgil-serbanuta: Thanks, I should have done that before sending the code review. Is it easy enough to understand now?

src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
virgil-serbanuta: Well... Yes. Or almost, it means not constructor-modulo-like. I renamed it.

traiansf:
I'm OK with it in general, and when @ttuegel approves it you can merge it, but I was thinking this might deserve a design decision doc, just in case it stays in the codebase long enough.

src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
ttuegel: :+1

PR #411, opened 2019-01-15
Exists variable renaming

vladciobanu:
###### Reviewer checklist

---


src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -343,6 +343,19 @@ externalizeFreshVars pat = Recursive.fold renameFreshLocal pat
       where
         name :: Text
         name = getId variableName
+    renameFreshLocal
```
virgil-serbanuta: You might as well add `forall`, even if that's not needed right now.

vladciobanu: Good point, I've added. Will merge once the tests pass.

PR #412, opened 2019-01-15
Attributes doc

tomtomjhj:
###### Reviewer checklist

---


docs/attributes.md
```diff
@@ -0,0 +1,188 @@
+# Attributes
+
+As pointed out in [Attributes as Patterns](../design-decisions/2018-07-25
```
ttuegel: This link should be on one line, or it will not be parsed correctly.

docs/attributes.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Assoc
+
+Associativity axiom
+
+```
+assoc{}()
+```
+
+Axioms with this attribute should not be used for evaluating functions.
```
ttuegel: The `assoc`, `comm`, `idem`, and `unit` axiom attributes identify certain axioms as associativity, commutativity, idempotency, or unit axioms. The Haskell backend does not currently do anything with those axioms, but it eventually will, so this statement is not accurate.

docs/attributes.md
```diff
@@ -0,0 +1,188 @@
...
+
+## misc.
+
+### Subsort
+
+```
+subsort{Sub,Super}()
+```
+
+Used to define subsorting relations, (not necessarily) attached to subsort
```
ttuegel: I think this is always attached to the subsort axiom. At least, it is ignored otherwise, if I understand correctly.

tomtomjhj: The current implementation seems to just collect this attribute from all of the axioms in the indexed module.
https://github.com/kframework/kore/blob/654e5c58b9b54b94a4df936a42bc342336d3394f/src/main/haskell/kore/src/Kore/IndexedModule/IndexedModule.hs#L814-L837

docs/attributes.md
```diff
@@ -0,0 +1,210 @@
...
+A symbol is functional if it is given the `functional` attribute or the
+`sortInjection` attribute.
+
+### Hook
+
+```
+hook{}("BUILTIN.name")
+```
+
+The string `"BUILTIN.name"` is used to lookup for known builtins.
```
virgil-serbanuta: I'd add a reference to https://github.com/kframework/kore/blob/master/docs/hooks.md

docs/attributes.md
ttuegel: Aha! That would be a bug. It should never be attached to anything but an `axiom`.

PR #418, opened 2019-01-18
use Text for StringLiterals

tomtomjhj:
functions related to error reporting and parser still use `String`s

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Builtin/Map.hs
```diff
@@ -411,6 +424,41 @@ asExpandedPattern symbols resultSort =
     asExpandedPattern0 = \asPattern0 builtin ->
         ExpandedPattern.fromPurePattern $ asPattern0 builtin
 
+concatKey :: String
```
ttuegel: We can reduce duplication by giving `concatKey` the following type:
~~~.hs
concatKey :: IsString s => s
~~~
(with `IsString` imported from `Data.String`) and removing `concatKeyT`--and doing the same for the others.

tomtomjhj:
I guess this can be merged now.

PR #420, opened 2019-01-21
Updated lts, fixed tests (megaparsec)

vladciobanu:
Only breaking changes in `lts-13.x` are found in `megaparsec`. Trivially, `anyChar` is now `anySingle`. The error type was updated to also have positions, so I changed `parseErrorPretty` with `errorBundlePretty`. This is where most of the diff comes from, because failing test cases (golden and otherwise) for parser have changed. I do think it is worth the trouble, since the output is much nicer now. For example, before:

```
Parse error: Error in module 'BOOL' -> alias 'falseBool' declaration (../../../test/resources//test-bool-4.kore 4:11) -> \dv (../../../test/resources//test-bool-4.kore 4:61) -> Verifying builtin sort 'BOOL.Bool' -> While parsing domain value: <string literal>:1:1:
unexpected "untru"
expecting "false" or "true"
```

And now:
```
Parse error: Error in module 'BOOL' -> alias 'falseBool' declaration (../../../test/resources//test-bool-4.kore 4:11) -> \dv (../../../test/resources//test-bool-4.kore 4:61) -> Verifying builtin sort 'BOOL.Bool' -> While parsing domain value: <string literal>:1:1:
  |
1 | untrue
  | ^^^^^
unexpected "untru"
expecting "false" or "true"
```

We can probably create a card to simplify our code and use that position in the error instead of trying to calculate our own.

###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Kore/Parser/Lexeme.hs
```diff
@@ -71,14 +71,20 @@ idParserTests =
         , Failure FailureTest
             { failureInput = "["
             , failureExpected =
-                "<test-string>:1:1:\n"
-                ++ "genericIdRawParser: Invalid first character '['.\n"
+                "<test-string>:1:1:\n\
```
ttuegel: :+1: for string gaps!

stack.yaml
```diff
@@ -45,7 +45,8 @@ packages:
 
 # Dependency packages to be pulled from upstream that are not in the resolver
 # (e.g., acme-missiles-0.3)
-# extra-deps:
+extra-deps:
+  - tasty-hedgehog-0.2.0.0@sha256:83a8b777fa472040979e44dba43c32441f55d5ddb9641a4d53deee4b0e09fa34
```
ttuegel: I'm unhappy about this dependency being dropped from Stackage. That shouldn't block this update, but given this and all the other problems we have had, should we consider using something other than `tasty`?

vladciobanu: Not sure what the alternatives are. I'll create a card to look into this.

vladciobanu:
Thanks for the review!

I Pushed the requested change, and added https://trello.com/c/1dh5JRs9/430-improve-source-location-reporting-from-parsing and https://trello.com/c/0j6Qgv92/431-label-parsers-properly

PR #429, opened 2019-01-25
Permanent debugger instrumentation

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/app/exec/Main.hs
+++ b/src/main/haskell/kore/app/exec/Main.hs
@@ -2,6 +2,8 @@ module Main (main) where
 
 import           Control.Applicative
                  ( Alternative (..), optional, (<$) )
+import           Control.Concurrent
+                 ( threadDelay )
 import qualified Control.Lens as Lens
 import           Data.Function
                  ( (&) )
@@ -376,6 +378,7 @@ externalizeFreshVars pat = Recursive.fold renameFreshLocal pat
 main :: IO ()
 main = do
     options <- mainGlobal parseKoreExecOptions parserInfoModifiers
+    threadDelay 10000000
     case localOptions options of
         Nothing ->
             -- global options parsed, but local failed; exit gracefully
--- a/src/main/haskell/kore/src/Kore/Debug.hs
+++ b/src/main/haskell/kore/src/Kore/Debug.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -Wno-unused-top-binds    #-}
 {-|
 Module      : Kore.Debug
 Description : Debugging helpers.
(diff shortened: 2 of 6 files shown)


ehildenb:
Jenkins: add to whitelist

ehildenb:
Jenkins: test this please

ttuegel:
@vladciobanu This looks like it intersects with logging, so I think you should have a look at it.

vladciobanu:
Ths indeed looks like something that will likely be refactored out with the logging PR, but if it's needed now I don't see a problem. The logging PR is still some time away (I postponed it until @ttuegel is able to clean up / remove `CounterT`).

virgil-serbanuta:
I find myself adding this instrumentation to our code every other day. If nobody minds this, I'll just merge.

src/main/haskell/kore/app/exec/Main.hs
```diff
@@ -376,6 +378,7 @@ externalizeFreshVars pat = Recursive.fold renameFreshLocal pat
 main :: IO ()
 main = do
     options <- mainGlobal parseKoreExecOptions parserInfoModifiers
+    threadDelay 10000000
```
ttuegel: Why do we need this long delay on startup?

src/main/haskell/kore/src/Kore/Debug.hs
```diff
@@ -60,42 +61,117 @@ module Kore.Debug
...
+        _ -> case Map.lookup place enabledPlaces of
+            Nothing -> id
+            Just debugResult -> logger debugResult
+
 {-|Wraps an 'ExceptT' action for printing debug messages, similar to 'trace'.
 
 It prints the name and the start values before the action, and the action
 result after.
 -}
 traceExceptT
```
ttuegel: We should make sure the `trace*` functions have the `HasCallStack` constraint so that the call stack of the child is passed through properly.

src/main/haskell/kore/app/exec/Main.hs
virgil-serbanuta: Ooops, this was some debugging instrumentation that I failed to remove, will send a PR.

src/main/haskell/kore/src/Kore/Debug.hs
virgil-serbanuta: Would we get benefits right now, or we would need to do something else to use child stack traces?

ttuegel: As written, any function wrapped with `trace...` that calls `error` will produce a call stack that ends at `trace...`. Depending on the error, that may capture enough information, or it may not.

virgil-serbanuta: Ah, I see, thanks!

PR #430, opened 2019-01-27
Dockerfile, Jenkinsfile: containerized dockerized build

ehildenb:
###### Reviewer checklist

---


ehildenb:
It seems the failure may be because of multiple invocations of `stack` running at the same time. I'll see if I can debug it later. https://github.com/commercialhaskell/stack/issues/3496

ttuegel:
Multiple concurrent `stack` invocations should be OK if they use different `--work-dir=` settings.

ehildenb:
Yeah, the bug was something else. Now that the latest nightly K has the fixed string quoting, I expect this to probably pass. We'll see.

ehildenb:
@ttuegel please re-review.

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+        libgmp-dev libjemalloc-dev libmpfr-dev libstdc++6 libtool libxml2    \
+        libyaml-cpp-dev llvm-6.0 m4 make maven opam openjdk-8-jdk pandoc     \
+        pkg-config python3 python-jinja2 python-pygments python-recommonmark \
+        scala stylish-haskell time libtinfo-dev unifdef zlib1g-dev
+
+RUN update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
+
+RUN curl -sSL https://get.haskellstack.org/ | sh
+
+RUN    git clone 'https://github.com/z3prover/z3' --branch=z3-4.8.4 \
```
ttuegel: Could we use a pre-built release of Z3?

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ARG GROUP_ID=1000
+RUN    groupadd --gid $GROUP_ID user                                        \
+    && useradd --create-home --uid $USER_ID --shell /bin/sh --gid user user
+
+USER $USER_ID:$GROUP_ID
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.28.0
+
+RUN    cd /home/user                                                          \
+    && git clone 'https://github.com/kframework/k' --branch=nightly-0f3835d3a \
```
ttuegel: Are we going to have to update this for every new release of K? Or is this just to setup OPAM?

scripts/check.sh
```diff
@@ -1,8 +1,15 @@
 #!/bin/sh
 
+set -e
```
ttuegel: Probably also want `set -o pipefail`, right?

scripts/check.sh
```diff
@@ -1,8 +1,15 @@
 #!/bin/sh
```
ttuegel: The meanings of the `set` options are particular to Bash, so we should change the shebangs to `#!/usr/bin/env bash`.

Jenkinsfile
```diff
@@ -0,0 +1,35 @@
+pipeline {
```
ttuegel: I don't know enough about `Jenkinsfile`s to judge if this is really correct, but no part of it seems unreasonable.

Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM ubuntu:bionic
```
ttuegel: I don't know enough about `Dockerfile`s to properly review this, but I guess if it runs the :+1:.

Dockerfile
ehildenb: This is just to setup opam. If we ever change compiler switches again, then we will need to update this. I don't anticipate a compiler switch for at least a year though (we just recently went from 4.03.0+k to 4.06.0+k), so I don't think it will need to change before we sunset the ocaml backend anyway.

Dockerfile
ehildenb: Well, the prebuilt installed version wasn't recent enough (at least, I thought so). Prebuilt is 4.4.0 I believe. VSC and KEVM require == 4.6.0, so we build it that way. I can test the prebuilt version for this job, see if it works.

scripts/check.sh
ehildenb: Done.

scripts/check.sh
ehildenb: Done.

Dockerfile
ehildenb: I'm testing the prebuilt version instead `apt install z3 libz3-dev`. I've also trimmed the list of Dockerfile `apt` dependencies.

ttuegel: There are pre-built releases on GitHub, too.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG GROUP_ID=1000
+RUN    groupadd --gid $GROUP_ID user                                        \
+    && useradd --create-home --uid $USER_ID --shell /bin/sh --gid user user
+
+USER $USER_ID:$GROUP_ID
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.28.0
+
+RUN    cd /home/user                                                          \
+    && git clone 'https://github.com/kframework/k' --branch=nightly-0f3835d3a \
```
dwightguth: why are we even setting up opam on this particular dockerfile, anyway? it shouldn't be at all necessary for this repo, and setting it up this way instead of via an ADD is very fragile.

ehildenb: I disagree that it's fragile, but I agree it's not needed. I'll pull it out.

ehildenb: Done, I also pulled out the `rustup`.

ehildenb:
@dwightguth and @ttuegel this is ready for re-review. If you want to merge @ttuegel remember to switch the branch protection over to the new status check, and we'll disable the other test-set.

scripts/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o
```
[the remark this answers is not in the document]
ehildenb: Fixed.

PR #442, opened 2019-02-06
Decide how to evaluate symbols when the program starts.

virgil-serbanuta:
I left a TODO about moving some functions to the new EvaluationStrategy.hs file, I plan to move them in a follow-up PR.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Function/Evaluator.hs
+++ b/src/main/haskell/kore/src/Kore/Step/Function/Evaluator.hs
@@ -8,7 +8,13 @@ Stability   : experimental
 Portability : portable
 -}
 module Kore.Step.Function.Evaluator
-    ( evaluateApplication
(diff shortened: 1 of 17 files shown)


src/main/haskell/kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -72,6 +73,16 @@ import           Kore.Step.StepperAttributes
 import           Kore.Unparser
 import           Kore.Variables.Fresh
 
+{-|Describes whether simplifiers are allowed to return multiple results or not.
+-}
+data AcceptsMultipleResults = WithMultipleResults | OnlyOneResult
```
ttuegel: Should we derive the standard instances here?
```suggestion
data AcceptsMultipleResults = WithMultipleResults | OnlyOneResult
    deriving (Eq, Ord, Show)
```

PR #444, opened 2019-02-08
Disjoin predicates with inequal substitutions

ttuegel:
`Kore.Step.Simplification.Equals.makeEvaluateFunctionalOr` requires that `Kore.Step.Simplification.Or.disjoinPredicates` disjoins predicates with inequal substitutions, so we do that by promoting the substitution to the predicate.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Simplification/Or.lhs
+++ b/src/main/haskell/kore/src/Kore/Step/Simplification/Or.lhs
@@ -19,6 +19,7 @@ module Kore.Step.Simplification.Or
 
 import           Control.Applicative
                  ( Alternative (..) )
+import qualified Data.Function as Function
 import qualified Data.Functor.Foldable as Recursive
 
 import           Kore.AST.Pure
@@ -26,6 +27,7 @@ import           Kore.Predicate.Predicate
                  ( makeOrPredicate )
 import           Kore.Step.ExpandedPattern
                  ( ExpandedPattern, Predicated (..) )
+import qualified Kore.Step.ExpandedPattern as Predicated
 import           Kore.Step.OrOfExpandedPattern
                  ( OrOfExpandedPattern )
 import qualified Kore.Step.OrOfExpandedPattern as OrOfExpandedPattern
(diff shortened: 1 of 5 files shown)


src/main/haskell/kore/src/Kore/Step/Simplification/Or.lhs
```diff
@@ -187,21 +189,47 @@ disjoinPredicates
...
+them by promoting the substitutions into the predicate,
+```
+(t, p₁     , s₁) ∨ (t, p₂     , s₂)
+(t, p₁ ∧ s₁, ⊤ ) ∨ (t, p₂ ∧ s₂, ⊤ )
+```
+and following the algorithm described above for disjoining predicates when the
+substitutions are equal. This is not strictly a simplification: simplifying
+predicates extracts substitutions into the corresponding field of the
+configuration. Nevertheless, this simplification is required by
+`Kore.Step.Simplification.Equals.makeEvaluateFunctionalOr`.
```
virgil-serbanuta: This comment has a few questions for @ttuegel , but also has some things for @traiansf and whoever is concerned about how this should work in the future. This is not a request to change ttuegel's solution right now.

As I said in the status meeting, I had to take took some (unplanned) vacation, so I didn't manage to fully follow Slack and I missed half of your explanation about what you're doing in this PR. Sorry if you already answered the things below.

I think I also asked on Slack why is this happening, and I don't remember getting an answer. I wouldn't normally expect functions on symbolic variables during normal execution, so I wouldn't expect function evaluation to return multiple values, which I would guess is the only thing that would require simplifying something like `phi = psi1 \/ psi2`. I guess it could happen when evaluating an `owise` condition, since that has an existential quantifier. Is this what happened?

Assuming that the above is correct, then, although this usually works, I am a bit concerned. An owise condition has the form `exists x1...xn . predicate(x1,... xn)`, and we expect that the predicate is fully evaluated to top or bottom when we simplify the exist.

Exist simplification applies the substitution for the quantified variable (if any), resimplifies the pattern, and checks if the result still contains the quantified variable outside the substitution. If it does not, then the exists disappears. We expect this to always happen when reaching the outermost exists.

Now, if we don't have substitutions (because we encoded them as predicates), then our only hope is that the SMT will be able to tell that the predicate is bottom. Which means that it should be able to handle a fairly large subset of kore and builtins / user-defined symbols. As an example, we might not be able to do map unification without these substitutions, so the SMT should be able to handle it.

Of course, this problem might become even more interesting when doing symbolic execution.

The obvious solution would be to just split the configuration. I previously suggested in on Slack and/or the status meeting, but I guess that there was a reason for not doing that (Too difficult for the benefit? It did not actually work? It did work, but it's likely that it would increase execution time by too much? I did not give a good reason for doing that? I'm not sure.)

traiansf: After  discussing this with @virgil-serbanuta today, I tend to believe that a specialized simplification for existentials would be good to have anyway, and would avoid the necessity for this fix (and the questions it raises) altogether.

PR #448, opened 2019-02-11
Feature/logging part 1

vladciobanu:
###### Reviewer checklist

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,4 @@ ENV LC_ALL=C.UTF-8
 ADD --chown=user:user stack.yaml /home/user/.tmp-haskell/
 ADD --chown=user:user src/main/haskell/kore/package.yaml /home/user/.tmp-haskell/src/main/haskell/kore/
 RUN    cd /home/user/.tmp-haskell \
-    && stack build --only-dependencies --test --bench --haddock --library-profiling
+    && stack build --only-dependencies --no-haddock-deps --test --bench --haddock --library-profiling
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,4 @@ ENV LC_ALL=C.UTF-8
 ADD --chown=user:user stack.yaml /home/user/.tmp-haskell/
 ADD --chown=user:user src/main/haskell/kore/package.yaml /home/user/.tmp-haskell/src/main/haskell/kore/
 RUN    cd /home/user/.tmp-haskell \
-    && stack build --only-dependencies --no-haddock-deps --test --bench --haddock --library-profiling
+    && stack build --only-snapshot --no-haddock-deps --test --bench --haddock --library-profiling
--- a/src/main/haskell/kore/README.md
+++ b/src/main/haskell/kore/README.md
@@ -18,6 +18,9 @@ To regenerate the golden data for regression tests:
 `stack test --no-keep-going --ta --accept`
 
 To generate documentation: `stack build --haddock`.
+Note that because of `co-log`'s `typerep-map` dependency (which fails haddock),
+we are currently building with `--no-haddock-deps`. This should be reverted
+once `typerep-map`'s haddock is fixed.
 
 To test parsing performance:
 
--- a/src/main/haskell/kore/README.md
+++ b/src/main/haskell/kore/README.md
@@ -19,8 +19,9 @@ To regenerate the golden data for regression tests:
 
 To generate documentation: `stack build --haddock`.
 Note that because of `co-log`'s `typerep-map` dependency (which fails haddock),
-we are currently building with `--no-haddock-deps`. This should be reverted
-once `typerep-map`'s haddock is fixed.
+we are currently building with `--no-haddock-deps`. This seems to be a problem
+with `stack` and libraries that have internal libraries in the same project,
+see https://github.com/commercialhaskell/stack/issues/3989 for details.
 
 To test parsing performance:
 
--- a/src/main/haskell/kore/bench/exec/Main.hs
+++ b/src/main/haskell/kore/bench/exec/Main.hs
@@ -183,7 +183,7 @@ execBenchmark root kFile definitionFile mainModuleName test =
         -> IO (CommonStepPattern Object)
     execution (verifiedModule, purePattern) =
         SMT.runSMT SMT.defaultConfig
-        $ evalSimplifier
+        $ evalSimplifier mempty
         $ exec verifiedModule strategy purePattern
       where
         unlimited :: Limit Natural
--- a/src/main/haskell/kore/bench/exec/Main.hs
+++ b/src/main/haskell/kore/bench/exec/Main.hs
@@ -28,6 +28,8 @@ import           Kore.Exec
 import           Kore.IndexedModule.IndexedModule
                  ( IndexedModule (..), VerifiedModule,
                  makeIndexedModuleAttributesNull, mapIndexedModulePatterns )
+import           Kore.Logger.Output
+                 ( emptyLogger )
 import           Kore.Parser.Parser
                  ( parseKoreDefinition, parseKorePattern )
 import           Kore.Step.AxiomPatterns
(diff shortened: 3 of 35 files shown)


vladciobanu:
A few things that I did not decide on yet, but we might want to wait until there are more use-cases, although I could be convinced otherwise:
- do we want the `log` functions to take `Builder` instead of `Text`?
- what do we do about `Text.pack (show pattern)`? Do we need a `Text` (or `Builder`) based `Show`?
- no filtering by `scope` yet
- should `scope` be a `[Text]` instead of `Text`, in `LogMessage`?
- do we want the helper `log*` functions or should we just say `log Debug`? I like the idea of importing qualified and using `Log.logDebug`, in which case they are needed

Dockerfile
```diff
@@ -30,4 +30,4 @@ ENV LC_ALL=C.UTF-8
 ADD --chown=user:user stack.yaml /home/user/.tmp-haskell/
 ADD --chown=user:user src/main/haskell/kore/package.yaml /home/user/.tmp-haskell/src/main/haskell/kore/
 RUN    cd /home/user/.tmp-haskell \
-    && stack build --only-dependencies --test --bench --haddock --library-profiling
+    && stack build --no-haddock-deps --test --bench --haddock --library-profiling
```
ttuegel: You still need `--only-dependencies` because we are only building the Docker image right now, the source is not checked out.
```suggestion
    && stack build --only-dependencies --no-haddock-deps --test --bench --haddock --library-profiling
```

ttuegel:
> what do we do about `Text.pack (show pattern)`? Do we need a `Text` (or `Builder`) based `Show`?

I would recommend using the [`formatting`](http://hackage.haskell.org/package/formatting) library for this. Summary: you can define reusable, typed format strings and instantiate them to `Text`, `Builder`, or anything you can imagine.

> should `scope` be a `[Text]` instead of `Text`, in `LogMessage`?

I would recommend `newtype`-ing `Text` (as `Scope`?) and taking a `[Scope]`.

> I like the idea of importing qualified and using `Log.logDebug`, in which case they are needed

:+1:, import ALL THE THINGS qualified.

src/main/haskell/kore/README.md
```diff
@@ -18,6 +18,9 @@ To regenerate the golden data for regression tests:
 `stack test --no-keep-going --ta --accept`
 
 To generate documentation: `stack build --haddock`.
+Note that because of `co-log`'s `typerep-map` dependency (which fails haddock),
+we are currently building with `--no-haddock-deps`. This should be reverted
+once `typerep-map`'s haddock is fixed.
```
ttuegel: Could you open an issue on our tracker for this, and include a link to the upstream issue?

src/main/haskell/kore/bench/exec/Main.hs
```diff
@@ -183,7 +183,7 @@ execBenchmark root kFile definitionFile mainModuleName test =
         -> IO (CommonStepPattern Object)
     execution (verifiedModule, purePattern) =
         SMT.runSMT SMT.defaultConfig
-        $ evalSimplifier
+        $ evalSimplifier mempty
```
ttuegel: Could we define a value, say `emptyLogger = mempty`, just to keep this clear?

src/main/haskell/kore/src/Kore/Logger.hs
```diff
@@ -0,0 +1,128 @@
+{-|
+Module      : Kore.Logger
+Description : Logging functions.
+Copyright   : (c) Runtime Verification, 2018
+License     : NCSA
+Maintainer  : vladimir.ciobanu@runtimeverification.com
+Stability   : experimental
```
ttuegel: For this and all the other new modules, I would be inclined to leave off `Stability` and `Portability`. `Stability` doesn't mean anything, except it makes a tiny note in the upper right-hand corner of the generated documentation that nobody looks at. All of our `Portability` fields are wrong, anyway: they should say `GHC-only`, which is what *everyone's* say, or should say, so it's just noise.

src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -123,7 +126,9 @@ transitionRule tools substitutionSimplifier simplifier =
             $ stepWithRewriteRule tools substitutionSimplifier config a
         case result of
             Left _ -> pure []
-            Right results -> return $ mapMaybe (patternFromResult proof) results
+            Right results ->
+                Log.withLogScope "transitionRule" $
```
ttuegel: Can we easily get a source location here? There are already multiple things called `transitionRule`.

src/main/haskell/kore/src/SMT.hs
```diff
@@ -7,7 +7,7 @@ Maintainer  : thomas.tuegel@runtimeverification.com
 -}
 
 module SMT
-    ( SMT, getSMT
+    ( SMT (..)
```
ttuegel: I specifically want to avoid exporting the constructor here:
```suggestion
    ( SMT, getSMT
```

src/main/haskell/kore/src/Kore/Logger.hs
```diff
@@ -0,0 +1,128 @@
...
+    -- ^ Used for odd/unusual cases which are recoverable.
+    | Error
+    -- ^ Used for application errors, unexpected behaviors, etc.
+    | Critical
+    -- ^ Used before shutting down the application.
+    deriving (Show, Read, Eq, Ord)
+
+-- | This type should not be used directly, but rather should be created and
+-- dispatched through the `log` functions.
+data LogMessage = LogMessage
```
ttuegel: The field names here do not conform to the [style guide](https://github.com/kframework/kore/wiki/Coding-Standards#records).

dwightguth:
While you are at it, can you replace --only-dependencies with --only-snapshot?

src/main/haskell/kore/README.md
vladciobanu: I have, and they closed it and linked to this bug in `stack`: https://github.com/commercialhaskell/stack/issues/3989 ; I'll add this to the docs here.

vladciobanu:
@ttuegel I addressed all comments. I actually need a way to go from the `Simplifier :: ReaderT (..., MVar Solver)` to `SMT :: ReaderT (MVar Solver)` so I created `withSolver'`. Is that better than exporting the constructor directly?

Also, I did not add the formatters library yet. I think formatting the log message is not necessarily the main problem here (it might become soon, I agree). I was actually concerned by `Show pat`  being show because (I think?) it does a lot of concats on strings, which are slow...? So I was hoping there was a `Show`-like class that uses `Text` -- or better yet, `Builder`?

ttuegel:
> Is that better than exporting the constructor directly?

I think so.

>  I was actually concerned by `Show pat` being show because (I think?) it does a lot of concats on strings, which are slow...?

`Show` is actually based on `shows`, which returns a `String`-builder (`String -> String`), so it's not too bad actually.

> So I was hoping there was a `Show`-like class that uses `Text` -- or better yet, `Builder`?

`shows` is still string-based, but I recommended `formatting` because you can use the same format declaration to send output to `Text`, `Builder`, `String`, even directly to a `Handle`. It does not provide a class, but I'm not sure how valuable a class would be for logging: We will always know the concrete type of the thing we are logging (no runtime instance resolution), and it is likely that different types of logging will want to represent the same information in different ways (no single-instance per type).

vladciobanu:
Ok, now I get it. I think I might add `formatting` once I get into the meat of adding debugging logs with some real use-cases, along with perhaps some variants of the log functions that take a `Format` as an argument. Thanks for the review!

src/main/haskell/kore/src/Kore/Logger.hs
```diff
@@ -0,0 +1,134 @@
+{-|
```
virgil-serbanuta: You should include documentation about the expected usage patterns.

src/main/haskell/kore/src/Kore/Logger.hs
```diff
@@ -0,0 +1,134 @@
...
+Module      : Kore.Logger
+Description : Logging functions.
+Copyright   : (c) Runtime Verification, 2018
```
virgil-serbanuta: 2019

PR #453, opened 2019-02-13
Parametric hooks

traiansf:
###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -149,6 +151,9 @@ unificationWithAppMatchOnTop tools substitutionSimplifier first second
                         tools
                         substitutionSimplifier
                         (zip firstChildren secondChildren)
+                else if symbolOrAliasConstructor firstHead
```
ttuegel: Because we are growing a longer list of cases here, I think the body of this `case` branch would be better written as a sequence of guards.

src/main/haskell/kore/src/Kore/Step/Function/Matcher.hs
```diff
@@ -149,6 +151,9 @@ unificationWithAppMatchOnTop tools substitutionSimplifier first second
...
+                        == symbolOrAliasConstructor secondHead
+                then throwError (UnificationError UnsupportedPatterns)
```
ttuegel: It wasn't clear to me at first *why* this should be an error. Maybe a comment like the following would be clearer?
```suggestion
                -- The application heads have the same symbol or alias constructor with
                -- different parameters, but we do not handle unification of symbol parameters.
                then throwError (UnificationError UnsupportedPatterns)
```

PR #456, opened 2019-02-14
specs with functions in requires

traiansf:
I've created a test to reproduce some stack-related behavior in `KEVM`.  It's a version of IMP with a separate evaluation stack limited to size 8, and two versions of the specs we were proving for imp: one in which we start with an empty stack (e.g., `src/main/k/working/smc/prove/max-spec.k`), and one in which we start with a symbolic stack, which must be constrained to have enough items left to fit the remaining computation (e.g., `src/main/k/working/smc/prove/max-stack-spec.k`).  The ones with an empty stack are passing.  The others not, consistent with the behavior observed in `KEVM`. All (except `double-sum-stack-spec.k` pass with the Java backend)

To reproduce, run `make all` in the `kore` root, then 
`make test-prove` in `src/main/k/working/smc/prove`.

Whoever has time, please start looking into this.  @virgil-serbanuta maybe add a few words about why is this happening to put people in the right direction.

I'll next move to attempt kernelc testing.

###### Reviewer checklist

---


ttuegel:
When the test fails, the generated `vdefinition.kore` file gets deleted; how can I keep it around? I could be off the trail entirely, but it's weird that the configuration which fails doesn't have a cell at the top level, and likewise the failing rule.

traiansf:
@ttuegel what I'm getting is as follows:
```
kore/src/main/k/working/smc$ make prove/add-stack-spec.kprove 2>&1 | ../../../python/debugFilter.py 
/home/traian/Documents/kore/.build/k/bin/kprove --haskell-backend-command "/home/traian/Documents/kore/.stack-work/install/x86_64-linux/lts-13.4/8.6.3/bin/kore-exec " -d . -m VERIFICATION prove/add-stack-spec.k
Executing (/home/traian/Documents/kore/.stack-work/install/x86_64-linux/lts-13.4/8.6.3/bin/kore-exec, /home/traian/Documents/kore/src/main/k/working/smc/./.kprove-2019-02-15-17-29-50-798-f74f133d-8696-4557-8af1-0217bf7970c6/vdefinition.kore, --module, VERIFICATION, --prove, /home/traian/Documents/kore/src/main/k/working/smc/./.kprove-2019-02-15-17-29-50-798-f74f133d-8696-4557-8af1-0217bf7970c6/spec.kore, --spec-module, ADD-STACK-SPEC, --output, /home/traian/Documents/kore/src/main/k/working/smc/./.kprove-2019-02-15-17-29-50-798-f74f133d-8696-4557-8af1-0217bf7970c6/result.kore)
Reading the input file TimeSpec (sec = 0, nsec = 66186)
Parsing the file TimeSpec (sec = 0, nsec = 166780735)
Verifying the definition TimeSpec (sec = 0, nsec = 148177041)
Reading the input file TimeSpec (sec = 0, nsec = 36813)
Parsing the file TimeSpec (sec = 0, nsec = 316482)
Verifying the definition TimeSpec (sec = 0, nsec = 132002557)
kore-exec: While applying axiom:
RulePattern (left = PurePattern ("LblstackSize(_)_SMC__ValStack"((),("Lbl_Coln__SMC__Val_ValStack"((),(Variable ("VarV":("SortVal"())),Variable ("VarS":("SortValStack"()))))))), right = PurePattern ("Lbl_+Int_"((),("1":("SortInt"()),"LblstackSize(_)_SMC__ValStack"((),(Variable ("VarS":("SortValStack"()))))))), requires = GenericPredicate (PurePattern (T)), attributes = AxiomPatternAttributes (heatCool = Normal, productionID = ProductionID (getProductionID = Nothing), assoc = Assoc (isAssoc = False), comm = Comm (isComm = False), unit = Unit (isUnit = False), idem = Idem (isIdem = False), trusted = Trusted (isTrusted = False), concrete = Concrete (isConcrete = False), simplification = Simplification (isSimplification = False)))
to configuration:
Predicated (term = PurePattern ("LblstackSize(_)_SMC__ValStack"((),(Variable ("VarS":("SortValStack"()))))), predicate = GenericPredicate (PurePattern (Not ("_PREDICATE"(),Equals ("SortValStack"(),"_PREDICATE"(),Variable ("VarS":("SortValStack"())),"Lbl.ValStack_SMC_"((),()))))), substitution = NormalizedSubstitution ())
Unexpected non-false predicate:
GenericPredicate (PurePattern (T))
when substitutions:
fromList ()
do not cover all variables in left axiom:
fromList ("VarS":("SortValStack"()),"VarV":("SortVal"()))

CallStack (from HasCallStack),:
  error,,, called at src/Kore/Step/BaseStep.hs:534:14 in kore-0.0.1.0-JTsul9IiXrRJDhxh4KFeZS:Kore.Step.BaseStep
(Error), Critical: I/O Error while executing
(./.kprove-2019-02-15-17-29-50-798-f74f133d-8696-4557-8af1-0217bf7970c6/result.kore)
(No such file or directory),)
(Warning), Critical: Haskell backend returned non-zero exit code
Makefile:13: recipe for target 'prove/add-stack-spec.kprove' failed
make: *** (prove/add-stack-spec.kprove), Error 113
```

My understanding is that it's trying to apply a function axiom  on `stackSize`, which is part of the predicate of the input configuration of the spec, and this is not implemented.

@virgil-serbanuta has a better idea about how that could be fixed.

virgil-serbanuta:
Assuming that this bug is how I remember it, the issue is the following:

First, and we want to apply an axiom on a pattern during normal execution, we unify the left-hand-side with the pattern, then we expect the resulting substitution to cover all variables in the right hand side. If not, we raise an error. This uncovered many bugs in the past, and it's still probably useful to have.

However, that does not work as well when doing symbolic execution. Let's say that we have a function defined like this:
```
unwrap(Wrapper(x)) = x
```
and let's say that we try to apply it to the pattern `unwrap(y)`. Then we would get [y=Wrapper(x)] as a substitution, but we wouldn't be able to substitute x in the right-hand-side, so we would trigger the error mentioned above.

There are at least two possible ways to solve this:
1. Never expand symbolic variables with patterns containing other variables: detect this case (we are trying to expand a symbolic variable, which is defined over the variables we are missing) in the axiom applying code and, instead of returning an error, signal that we failed to apply the axiom.
2. Allow replacing a symbolic variable with a pattern containing other variables, which become symbolic (in the same case as above).

1 is the easy fix, but we will probably need to replace with something better later. 2 is not trivial to do, since we may get into an infinite expansion loop. Let's say we have a function over Peano natural numbers defined by
```
pred(S(x)) = x
pred(Z) = Z
```
Note that we could expand `pred(y)` forever.

This is not a hypothetical case, Traian discovered this issue in a kevm proof, when evaluating the length function on a symbolic list variable, `len(cons(x, y))=1+len(y)`.

Here's my attempt at describing a possible solution: https://github.com/kframework/kore/blob/master/docs/2018-12-21-Function-Evaluation.md

I think @jerry.jung may have looked into this in the past.

I would personally start implementing parts of the document above, starting with the safe stuff, and maybe stop when we can move past the current bug.

dwightguth:
You can keep the temp directory that k creates by passing the --debug flag.

traiansf:
@ttuegel, the rationale for making the variableCounter also depend on the variableSort is that we have no power on how the users choose to name their symbolic variables in the configuration, and the chance to have a collision is quite probable.  In fact, the debugging me and Virgil started on Friday and I continued today led to the fact that refreshVariables on `True andBool B:Bool => B:Bool` yielded
something of the form `True andBool B_0:Int => B_0:Int` because in the configuration we had a variable `B:Int`.
I'm asking for your review because I want to make sure I haven't violated any assumptions on `refreshVariable` with my change.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -802,13 +816,21 @@ stepWithRemainders
...
+  = do
+    resultOrError <- runExceptT
+        $ stepWithRemaindersForUnifier
+            tools
+            (UnificationProcedure unificationProcedure)
+            substitutionSimplifier
+            (map (\ (RewriteRule rule) -> rule) rules)
+            patt
+    case resultOrError of
+        Left _ -> error $
```
ttuegel: Do I understand correctly, that this is a "not implemented" case? It would be clearer if the error message had those words in it.

PR #457, opened 2019-02-14
Remove the global counter

ttuegel:
The global counter used to generate fresh variable names is removed. Instead, we generate locally-unique names by updating a counter attached to `Variable`. `Variable` gains a field to hold the counter so that we no longer need to go munging through strings to find it; the counter can be updated to avoid capturing variables in time logarithmic in the number of free variables.

###### Reviewer checklist

---


src/main/haskell/kore/src/Data/Sup.hs
```diff
@@ -0,0 +1,52 @@
...
+    = Element !a
+    | Sup  -- ^ least upper bound (supremum)
+    deriving (Data, Functor, Generic, Read, Show, Typeable)
+
+instance Eq a => Eq (Sup a) where
+    (==) Sup         = \case { Sup -> True        ; _ -> False }
+    (==) (Element a) = \case { Element b -> a == b; _ -> False }
+
+instance Ord a => Ord (Sup a) where
+    compare Sup         = \_ -> GT
```
virgil-serbanuta: Hm, shouldn't `Sup` be equal to `Sup`?

src/main/haskell/kore/src/Kore/Step/Pattern.hs
```diff
@@ -327,14 +335,80 @@ may appear in the right-hand side of any substitution, but this is not checked.
...
+        head
+        $ dropWhile (\var -> Set.member var freeVariables')
+        $ iterate nextVariable variable
+
+    underBinder freeVariables' variable child = do
+        let variable' = safeVariable freeVariables' variable
+        child' <- Reader.local (Map.insert variable variable') child
+        return (variable', child')
+
+    fromFreshVariablesWorker (valid :< patt) = do
```
virgil-serbanuta: Having type signatures for some of these functions would make the code easier to read.

src/main/haskell/kore/src/Data/Sup.hs
ttuegel: Thanks, you're quite right! I pushed some tests for the instance laws.

PR #461, opened 2019-02-18
Internalize integer and bool builtins

njohnwalker:
PR to add internal representation of Integer and Boolean domain values. Following verification, all Integer and Boolean domain values are assumed to be internal. This means that the correct verifiers must be used for verification prior to execution.

- New constructors were added to the Domain.Builtin.Builtin type to contain Haskell Integer and Booleans
- Verification now converts Integer and Boolean domain values to that internal representation.
- Simplification was modified to add support for the new constructors.
- Tests on Integer and Boolean domain values were modified to operate on the internal representations.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Builtin/Bool.hs
```diff
@@ -104,11 +112,24 @@ symbolVerifiers =
...
+    :: Text -- ^ error message Context
+    -> DomainValue Object Domain.Builtin child
+    -> Bool
+extractBoolDomainValue
+    ctx
+    dv@DomainValue { domainValueChild }
+  =
+    case domainValueChild of
+        Domain.BuiltinBool bool -> bool
+        _ -> Builtin.verifierBug $ Text.unpack ctx ++ ": "
```
virgil-serbanuta: I think that this line should also be split after '$'.

PR #462, opened 2019-02-18
Ignore overload axioms

ttuegel:
@ehildenb This pull request would fix the first bug we found with KWASM today.

###### Reviewer checklist

---


dwightguth:
I expect in the long run you will need to do the same thing with overload axioms you do with the injection triangle axiom in order for it to truly work, fwiw.

ttuegel:
Explicit injections and overloading will go away entirely if we implement unsorted Matching-Mu Logic, so I'm not worried about that.

PR #466, opened 2019-02-20
Replace `Builtin.Bool` tests with tabular tests.

marick:
###### Reviewer checklist

---


src/main/haskell/kore/test/Test/Terse.hs
```diff
@@ -199,6 +207,81 @@ f_2_expected
...
+  -> String       -- ^ a description of the intention of the comparison
+  -> TestTree
+wrapped_maker_expected_name_intention wrapper maker expected name intention =
+    wrapper $ \getResource ->
+        testCase name $ do
+            resource <- getResource
+            maker resource >>= assertEqualWithExplanation intention expected
+
+wrapped_maker_expected_name
+  :: (HasCallStack, Eq b, Show b, EqualWithExplanation b)
```
virgil-serbanuta: The style guide requires 4-space indentation except in some specific cases (e.g. the "where" keyword). The list of allowed cases may not be comprehensive, and our code contains indentation levels that do not fit these requirements, but which make sense in context. However, most of the two-spaces indentation in this pull request seems to be used for normal-looking code, and it does not seem to me that it would be worth an exception from the style guide.

marick: Done.

src/main/haskell/kore/test/Test/Terse.hs
```diff
@@ -122,16 +130,16 @@ has_ value tuples = has value tuples "Has properties"
 gives :: forall a . HasCallStack => (a -> Bool) -> [(a, Bool)] -> String -> TestTree
 gives predicate tuples name =
   testCase name (traverse_ checkOne tuples)
-  where
-    checkOne :: (a, Bool) -> Assertion
-    checkOne (value, expected) =
-      assertEqual "" expected (predicate value)
+    where
```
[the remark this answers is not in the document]
marick: Merged.

PR #470, opened 2019-02-23
Review ch. 7 definition of #substitute in 'Semantics of K'

ammkrn:
I'm pretty sure this is just a typo, but it's in a fairly key definition so I thought it might be worth a second look. Both 'x' and 'v' are used twice to reference the same variable in the definition of the meta theory's substitution operation. I just made them all 'v' since that seems to be the default variable name throughout the paper.

Thank you!

###### Reviewer checklist

---


ehildenb:
Closing because fork does not exist anymore (or is private).

PR #472, opened 2019-02-25
Refactoring - make ceil evaluation return MultiOr

virgil-serbanuta:
* Fix TODO - we now throw errors for x=f(x) substitutions.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
+++ b/src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
@@ -31,6 +31,7 @@ module Kore.Step.OrOfExpandedPattern
     , crossProductGenericF
     , extractPatterns
     , toExpandedPattern
+    , toPredicate
     , traverseWithPairs
     , traverseFlattenWithPairs
     , traverseFlattenWithPairsGeneric
@@ -47,7 +48,8 @@ import           GHC.Generics
 import           Kore.AST.Pure
 import           Kore.AST.Valid
 import           Kore.Predicate.Predicate
-                 ( makeTruePredicate )
+                 ( Predicate, makeFalsePredicate, makeOrPredicate,
+                 makeTruePredicate )
 import           Kore.Step.ExpandedPattern
                  ( ExpandedPattern, Predicated (..) )
 import qualified Kore.Step.ExpandedPattern as ExpandedPattern
@@ -95,6 +97,11 @@ type OrOfExpandedPattern level variable
 -}
 type OrOfPredicateSubstitution level variable
     = OrOfPredicated level variable ()
+
+{-| 'OrOfPredicate' is a 'MultiOr' of 'Predicate'.
+-}
+type OrOfPredicate level variable =
+    MultiOr (Predicate level variable)
 
 {-| 'CommonOrOfExpandedPattern' particularizes 'OrOfExpandedPattern' to
 'Variable', following the same convention as the other Common* types.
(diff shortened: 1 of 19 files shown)


src/main/haskell/kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -705,27 +787,48 @@ variableFunctionAndEquals
...
+                -- be careful to not just drop the term.
+                return ExpandedPattern.topPredicate
+            SimplificationType.Equals -> do
+                (resultOr, _proof) <-
+                    Ceil.makeEvaluateTerm tools substitutionSimplifier second
+                case OrOfExpandedPattern.extractPatterns resultOr of
+                    [] -> return ExpandedPattern.bottomPredicate
+                    [resultPredicateSubstitution] ->
+                        return resultPredicateSubstitution
+                    _ -> error
```
traiansf: could you please add a comment / example here on when/ why this could be happening.

src/main/haskell/kore/src/Kore/Step/Substitution.hs
```diff
@@ -292,3 +303,96 @@ normalizePredicatedSubstitution
...
+    -> PredicateSubstitutionMerger level variable m
+createPredicatesAndSubstitutionsMerger tools substitutionSimplifier =
+    PredicateSubstitutionMerger worker
+  where
+    worker
+        :: [Predicate level variable]
+        -> [Substitution level variable]
+        -> m (PredicateSubstitution level variable)
+    worker predicates substitutions = do
+        (merged, _proof) <- mergePredicatesAndSubstitutions
```
ttuegel: Does this do anything extra, beyond combining predicates with `\and` and normalizing the merged substitution? Do we need to normalize the substitution at this point?

src/main/haskell/kore/src/Kore/Unification/Procedure.hs
```diff
@@ -86,21 +89,25 @@ unificationProcedure tools substitutionSimplifier p1 p2
...
-            ( OrOfExpandedPattern.make
-                [ Predicated
-                    { term = ()
-                    , predicate = makeAndPredicate predicate pred'
-                    , substitution
-                    }
-                ]
-            , EmptyUnificationProof
-            )
+        else Monad.Trans.lift $ do
```
ttuegel: Is the `else` case here checking that the returned unifier is defined, by wrapping the term in `\ceil`? Why would `termUnification` ever return an undefined unifier that is not simply `\bottom`?

src/main/haskell/kore/src/Kore/Step/Substitution.hs
virgil-serbanuta: This is exactly what this does. This is called at the end of unification/matching, which, I think, should return a normalized substitution, but we could argue about this if you disagree.

src/main/haskell/kore/src/Kore/Step/OrOfExpandedPattern.hs
```diff
@@ -483,3 +490,16 @@ toExpandedPattern (MultiOr patts) =
                 , predicate = makeTruePredicate
                 , substitution = mempty
                 }
+
+{-| Transforms an 'OrOfPredicate' into a 'Predicate'. -}
+toPredicate
```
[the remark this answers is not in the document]
ttuegel: It was caused by any simplification that would return an `\or` inside a `Predicate`, because this gets unpacked into a `MultiOr`, then re-packed, then unpacked... by the simplifier. Once I finish my current task of implementing branching in the predicate simplifier it won't matter because we will never re-pack such things.

src/main/haskell/kore/src/Kore/Step/Substitution.hs
ttuegel: > This is called at the end of unification/matching, which, I think, should return a normalized substitution

It wasn't clear to me that this is only called at the end of unification and matching. In that case, I agree.

PR #474, opened 2019-02-25
Changed evaluation of KEQUALS builtin to support discharing nontrivia…

u--:
…l side conditions with smt. Added integration test lem.k

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`
- [ ] Style conformance: `stylish-haskell`

---

PR #479, opened 2019-02-28
Unrevert "Changed evaluation of KEQUALS builtin to support disc…

ehildenb:
…haring nontrivia… (#474)" (#478)"

This reverts commit 07e85b4683e2a8a9aa8bf61b1c8fcf323f1bce8c.

###### Reviewer checklist

---

This needs to behave consistently between our machines and the CI before it is merged.

Steps to reproduce:

```
stack build
cd kore/src/main/k/working/imp
make tests/impossible-branch.search.final.output
```

Changing the z3 timout in `SMT.hs` has little effect:

```
        , timeOut = TimeOut (Limit 500)
```

Perhaps we should try out DappHub's `--z3-tactic`:

```
(or-else (using-params smt :random-seed 3 :timeout 1000) (using-params smt :random-seed 2 :timeout 2000) (using-params smt :random-seed 1))
```


ehildenb:
@ttuegel the last commit attempts to switch to DappHub's strategy to no avail on my Arch machine.

It tries 3 different hardcoded random seeds, and gives each one successively longer to try and discharge the query.

ttuegel:
I don't have any objection to changing the default tactic, but I would like to split up the un-revert into two pull requests because the original commit had two different threads of changes.

ehildenb:
That sounds good to me.

ehildenb:
Subsumed by #482.

PR #482, opened 2019-02-28
Un-revert SMT lemma translation

ttuegel:
We rely on Z3 to reject attempts to redefine builtin symbols; version 4.4.1
rejects such declarations, but versions 4.8.x do not reject them. Therefore,
this commit disables sending new declarations to the solver. At the moment, we
do not send any useful declarations anyway. We are working with the frontend to
develop a blacklist of SMT-LIB symbols shared by all backends; this commit
should be reverted after the blacklist is available.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Function/EvaluationStrategy.hs
```diff
@@ -163,24 +165,24 @@ evaluateBuiltin
...
         AttemptedAxiom.NotApplicable
           | isPattConcrete
-          , Just hook <- getAppHookString ->
+          , App_ appHead children <- patt
+          , Just hook <- getAppHookString appHead
+          , all isValue children ->
             error
                 (   "Expecting hook " ++ hook
-                ++  " to reduce concrete pattern\n\t"
+               ++  " to reduce concrete pattern\n\t"
```
virgil-serbanuta: This needs one more space to be aligned with the reminder of this expression.

PR #483, opened 2019-03-01
decidePredicate: Also check negation of predicates

ttuegel:
I forgot to fix this in #482. This would run a few more things by Z3.

###### Reviewer checklist

---


ehildenb:
I think this may be a premature optimization. I'd be willing to bet that this makes us take a performance hit, and only very rarely actually leads to the constraint predicate being simplified to true. Do we have any cases which we can't verify with this, or cases where the performance has been measured and shown favorable? If not, we should think twice about including it.

ttuegel:
I tested on a long-running KWASM proof. The difference is less than 1%, within the standard deviation of runtimes. But, I also don't care.

PR #485, opened 2019-03-01
Kore.Step.Step.transitionRule: Do not silence unification failures

ttuegel:
Unification failures should produce an error, not silently attempt to execute
down other paths.

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
+++ b/src/main/haskell/kore/src/Kore/Step/BaseStep.hs
@@ -29,8 +29,6 @@ module Kore.Step.BaseStep
 import           Control.Monad.Except
 import           Control.Monad.Trans.Except
                  ( throwE )
-import           Data.Either
-                 ( partitionEithers )
 import qualified Data.Hashable as Hashable
 import           Data.List
                  ( foldl' )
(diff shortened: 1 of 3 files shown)


src/main/haskell/kore/src/Kore/Step/Step.hs
```diff
@@ -139,7 +139,11 @@ transitionRule tools substitutionSimplifier simplifier axiomIdToSimplifier =
                 config
                 a
         case result of
-            Left _ -> pure []
+            Left _ -> error $
```
vladciobanu: Might be a good idea to also print the rule and configuration for debug purposes. Other than that, failing earlier is always better.

src/main/haskell/kore/src/Kore/Step/BaseStep.hs
```diff
@@ -381,7 +379,7 @@ stepWithRule
         proof = renamingProof <> unificationProof'
         attachProof result = (result, proof)
     results <-
-        keepGoodResults $ return $ map
+        traverse
```
virgil-serbanuta: If not too difficult, I would add a test that failed before this change, but passes now.

PR #486, opened 2019-03-04
Fix some leftover thing from the last ceil-as-or refactoring.

virgil-serbanuta:
* Make And usage explicit when computing and-of-ceil
* Throw an error when or-of-ceil contains a substitution

###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
+++ b/src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
@@ -51,7 +51,8 @@ instance TopBottom child => TopBottom (MultiAnd child)
 {-| 'AndBool' is an some sort of Bool data type used when evaluating things
 inside an 'MultiAnd'.
 -}
--- TODO(virgil): Refactor, this is the same as OrBool
+-- TODO(virgil): Refactor, this is the same as OrBool. Make it a
+-- Top | Bottom | Other or a Maybe Bool.
 data AndBool = AndTrue | AndFalse | AndUnknown
 
 {-|Does a very simple attempt to check whether a pattern
@@ -86,7 +87,7 @@ extractPatterns = getMultiAnd
 {- | Simplify the conjunction.
 
 The arguments are simplified by filtering on @\\top@ and @\\bottom@. The
-idempotency property of disjunction (@\\and(φ,φ)=φ@) is applied to remove
+idempotency property of conunction (@\\and(φ,φ)=φ@) is applied to remove
 duplicated items from the result.
 
 See also: 'filterUnique'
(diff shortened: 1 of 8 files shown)


virgil-serbanuta:
This is not 100% done, I'll add/fix tests tomorrow.

src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
```diff
@@ -0,0 +1,136 @@
...
+    :: (Ord term, TopBottom term)
+    => MultiAnd term
+    -> MultiAnd term
+filterAnd =
+    filterGeneric patternToAndBool . filterUnique
+
+
+{- | Simplify the disjunction by eliminating duplicate elements.
+
+The idempotency property of disjunction (@\\or(φ,φ)=φ@) is applied to remove
```
ttuegel:
```suggestion
The idempotency property of conjunction (@\\and(φ,φ)=φ@) is applied to remove
```
Actually, I'm not sure if this should be done here or in the simplifier. Although it doesn't require the full machinery of the simplifier, removing duplicate terms is a matter of policy and not representation. Also, we can only apply this simplification to certain types of arguments. I guess the same comment would apply to `MultiOr` though; this is consistent.

src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
```diff
@@ -0,0 +1,136 @@
...
+  where
+    isTop (MultiAnd []) = True
+    isTop _ = False
+    isBottom (MultiAnd [child]) = isBottom child
+    isBottom _ = False
+
+{-| 'AndBool' is an some sort of Bool data type used when evaluating things
+inside an 'MultiAnd'.
+-}
+-- TODO(virgil): Refactor, this is the same as OrBool
```
ttuegel: You don't need to address it in this pull request, but why not `Maybe Bool` for both?

src/main/haskell/kore/src/Kore/Step/Simplification/Equals.hs
```diff
@@ -676,12 +676,24 @@ makeEvaluateTermsToPredicateSubstitution
                     axiomIdToSimplfier
                     second
             let
+                toPredicateSafe
```
ttuegel: Do you think this definition should be extracted?

src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
virgil-serbanuta: I usually prefer using specific types whenever that does not make other things too hard. In particular, this type is actually  supposed to be `Top, Bottom, OtherPattern`, which, IMO, says much more than a `Maybe Bool`, for historical reasons, the names are a bit different. I added a TODO.

src/main/haskell/kore/src/Kore/Step/MultiAnd.hs
virgil-serbanuta: I agree that the design of this is suboptimal. This also applies to the `makePredicate*` functions which follow the same pattern. However, IIRC, this made testing easier. I don't know if this has any influence on execution, I would guess that it does not.

src/main/haskell/kore/src/Kore/Step/Simplification/Equals.hs
virgil-serbanuta: I think this is a temporary hack (hence the error below) and we should replace it with something else. I think that we will not be able to replace it with another function, we actually need to rewrite the caller. As such, I feel uncomfortable making a library function and hiding that error inside of it, I would prefer to have the error where we need to do the actual fix.

PR #492, opened 2019-03-05
Extracted a semilattice from MultiAnd and MultiOr.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/src/main/haskell/kore/src/Kore/Step/Representation/Semilattice.hs
+++ b/src/main/haskell/kore/src/Kore/Step/Representation/Semilattice.hs
@@ -1,0 +1,68 @@
+{-|
+Module      : Kore.Step.Representation.Semilattice
+Description : Data structures and functions for handling semilattice terms.
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+Maintainer  : virgil.serbanuta@runtimeverification.com
+Stability   : experimental
+Portability : portable
+
+This is intended to be imported qualified:
+
+@
+import           Kore.Step.Representation.Semilattice
+                 ( Semilattice )
+import qualified Kore.Step.Representation.Semilattice as Semilattice
+@
+-}
+module Kore.Step.Representation.Semilattice
+    ( ElementType (..)
+    , Semilattice (..)
+    ) where
+
+import           Data.Proxy
+                 ( Proxy (..) )
+import qualified Data.Set as Set
+
+{-| 'ElementType' identifies element types in a semilattice.
+-}
+data ElementType
+  = Absorbing
+  -- ^ An element a such that ∀x . x * a = a
+  | Neutral
+  -- ^ An element e such that ∀x . x * e = x
+  | Other
+  -- ^ Any element which is not identified by another ElementType case.
+
+{-| A semilattice with an absorbing and neutral elements.
+
+The absorbing and neutral elements are assumed to be different (i.e. the
+semilattice should have more than one element).
+-}
+class Ord element => Semilattice term element where
+    -- | Returns the type of an element.
+    elementType :: Proxy (term element) -> element -> ElementType
+    -- | Transforms a term into a list of elements.
+    toList :: term element -> [element]
+    -- | Transforms a list of elements into a term.
+    fromList :: [element] -> term element
+    -- | Simplifies a term by removing neutral elements, reducing everything
+    -- to an absorbing element if any is present and applying idempotency.
+    simplify :: term element -> term element
+    simplify =
+        fromList . simplifyIdempotency . simplifyAbsorbingNeutral . toList
+      where
+        simplifyIdempotency :: [element] -> [element]
+        simplifyIdempotency = Set.toList . Set.fromList
+        simplifyAbsorbingNeutral :: [element] -> [element]
+        simplifyAbsorbingNeutral elements = go [] elements
+          where
+            go filtered [] = reverse filtered
+            go filtered (element:unfiltered) =
+                case elementType (Proxy :: Proxy (term element)) element of
+                    Absorbing -> [element]
+                    Neutral -> go filtered unfiltered
+                    Other -> go (element:filtered) unfiltered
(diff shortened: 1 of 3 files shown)


src/main/haskell/kore/src/Kore/Step/Representation/Semilattice.hs
```diff
@@ -0,0 +1,68 @@
...
+{-| A semilattice with an absorbing and neutral elements.
+
+The absorbing and neutral elements are assumed to be different (i.e. the
+semilattice should have more than one element).
+-}
+class Ord element => Semilattice term element where
+    -- | Returns the type of an element.
+    elementType :: Proxy (term element) -> element -> ElementType
+    -- | Transforms a term into a list of elements.
+    toList :: term element -> [element]
```
ttuegel: The list conversion functions are defined in the class `GHC.Exts.IsList` already.

src/main/haskell/kore/src/Kore/Step/Representation/Semilattice.hs
```diff
@@ -0,0 +1,68 @@
...
+  -- ^ An element e such that ∀x . x * e = x
+  | Other
+  -- ^ Any element which is not identified by another ElementType case.
+
+{-| A semilattice with an absorbing and neutral elements.
+
+The absorbing and neutral elements are assumed to be different (i.e. the
+semilattice should have more than one element).
+-}
+class Ord element => Semilattice term element where
```
ttuegel: I feel a little weird about tying `term` and `element` together through the `Semilattice` class. Usually we would call `element` by itself a (join or meet) semilattice. The `term` is not really a part of the semilattice structure of `element`, it's just a container that respects the semilattice structure of `element`. It's also a little weird to me that the `Semilattice` class doesn't know if it's a join or meet semilattice. How would you feel about the following instead?

~~~ haskell
-- | Semigroup operation respects the meet (@\\and@) semilattice of @child@.
instance TopBottom child => Semigroup (MultiAnd child) where
    ...

-- | Identity is @\\top@.
instance TopBottom child => Monoid (MultiAnd child) where
    ...
~~~

...and likewise for `MultiOr`. The `Semigroup` definition of `(<>)` would both merge and simplify. We would need to add functions to `TopBottom` for constructing the `top` and `bottom` of each type. (Then we would be able to write laws for that class!) The implementation of `simplify` could still be a single function, just specialized in each module for `MultiAnd` and `MultiOr`:

~~~ haskell
import GHC.Exts (IsList, fromList, toList)

simplifySemilattice :: IsList term => (element -> ElementType) -> term element -> term element
simplifySemilattice elementType =
    fromList . simplifyIdempotency . simplifyAbsorbingNeutral . toList
  where
    simplifyIdempotency :: [element] -> [element]
    simplifyIdempotency = Set.toList . Set.fromList
    simplifyAbsorbingNeutral :: [element] -> [element]
    simplifyAbsorbingNeutral elements = go [] elements
      where
        go filtered [] = reverse filtered
        go filtered (element:unfiltered) =
            case elementType element of
                Absorbing -> [element]
                Neutral -> go filtered unfiltered
                Other -> go (element:filtered) unfiltered

joinSemilattice :: TopBottom element => element -> ElementType
joinSemilattice e
  | isTop e    = Absorbing
  | isBottom e = Neutral
  | otherwise  = Other

meetSemilattice :: TopBottom element => element -> ElementType
meetSemilattice e
  | isTop e    = Neutral
  | isBottom e = Absorbing
  | otherwise  = Other

simplifyMultiAnd = simplifySemilattice meetSemilattice

simplifyMultiOr = simplifySemilattice joinSemilattice
~~~

virgil-serbanuta:
A few comments before we chat and decide how to continue:

* MultiAnd is used in a rather narrow setting right now, but may become more popular when handling, say, predicates. By default I would say that it will stay in one form or another, and it will not be influenced by the branching-on-predicates stuff.

* A large part of MultiOr usage will probably be replaced by the branching-on-predicates stuff, but I would guess that it will not be completely removed. However, I'm unsure about how it will work.

We should chat about these, but I'll probably not have the time today after the iteration meeting, so I'll postpone this PR. Other comments below, I'm still thinking about how to solve those issues.

* I should have used `expression` and `atomicTerm` (or something like that) instead of `term` and `element`. Not sure if that would make it less weird, I'd still have to use, say, `Proxy`.

* The semilattice algebraic structure is common to both join and meet ones, so I should be able to just say "this is a semilattice" and, later, perhaps, say "and, btw, it's a meet/join one". We have an order relation here given by term inclusion, so we could naturally assign join/meet adjectives to it, but I would still want to have the basic data structure. Also, unlike regular lattices, this one is about building expressions in that lattices, not about fully evaluating expressions, which means that it's more than just defining its three operations (absorbing, neutral, composition).

* The monoid instance is tempting (and I thought about it), but it only captures part of the actual algebraic structure. Which is better than nothing, but somehow I wanted more. I'll consider it again.

* I'm not that knowledgeable about portability stuff, but I thought we were avoiding using GHC.Exts in our code. I may not be remembering correctly.

* Constructing top/bottom for each class - IIRC that was impossible for one of the instances of this class ( `instance TopBottom ()` ), and that's why we don't have it yet. It may work with two classes, one for construction, the other one for identification.

ttuegel:
> I'm not that knowledgeable about portability stuff, but I thought we were avoiding using GHC.Exts in our code. I may not be remembering correctly.

It's true that `GHC.Exts` is un-portable, but:

1. The project is already un-portable because we use `LANGUAGE` extensions.
1. There has not been an actively-developed open-source compiler other than GHC for about ten years.
1. Stack only supports GHC, and the project is already Stack-only.

I didn't know that we had consciously avoided using `GHC.Exts`, but if that's so, I think being able to write `deriving (IsList)` is a reason to revisit that decision.

You are right that `MultiAnd` is not going to interact much with the branching work that I'm doing. We can discuss the rest later.

marick:
Is this PR still alive?

PR #496, opened 2019-03-06
Kore.Exec.prove: Return all failing results

ttuegel:
Returning all failing results is very useful for debugging and exploration.

Instead of this implementation, we may want to consider printing a flat list of
patterns. The result is always a disjunction at the top, but because it is a
list of patterns, that disjunction can quickly become very deeply nested.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Step/Representation/OrOfExpandedPattern.hs
```diff
@@ -122,6 +123,26 @@ toExpandedPattern multiOr
...
+-}
+toStepPattern
+    ::  ( MetaOrObject level
+        , SortedVariable variable
+        , Ord (variable level)
+        , Show (variable level)
+        , Unparse (variable level)
+        )
+    => OrOfExpandedPattern level variable -> StepPattern level variable
+toStepPattern multiOr =
```
virgil-serbanuta: Why not `ExpandedPattern.toMLPattern . toExpandedPattern`?

src/main/haskell/kore/test/Test/Kore/OnePath/Verification.hs
```diff
@@ -105,7 +108,7 @@ test_onePathVerification =
             , simpleClaim Mock.a Mock.b
             ]
         assertEqualWithExplanation ""
-            (Left $ ExpandedPattern.fromPurePattern Mock.a)
+            (Left $ MultiOr.make [ExpandedPattern.fromPurePattern Mock.a])
```
virgil-serbanuta: Could you add a test for a Left result with multiple ExpandedPatterns?

src/main/haskell/kore/src/Kore/Step/Representation/OrOfExpandedPattern.hs
ttuegel: If there are multiple patterns, won't that give two extra `\and`s?

PR #499, opened 2019-03-07
Implement getExitCode

njohnwalker:
Added a simplification pass, after normal execution, to extract the pattern in the cell of the configuration marked "exit".  
    If there is no such cell, the `kore-exec` executable returns 0 for success, 
    if there is such a cell and simplification returns an integer, it will be returned as the exit code
    otherwise, the program returns exit code 111.

###### Reviewer checklist

---


src/main/haskell/kore/src/Kore/Attribute/Exit.hs
```diff
@@ -0,0 +1,72 @@
...
+import           GHC.Generics
+                 ( Generic )
+
+import           Kore.AST.Kore
+import           Kore.Attribute.Parser
+                 ( ParseAttributes (..) )
+import qualified Kore.Attribute.Parser as Parser
+
+-- | @Exit@ represents the @exit@ attribute for symbols.
+newtype Exit = Exit { isDeclaredExit :: Bool }
```
ttuegel: The field can just be `isExit`.
```suggestion
newtype Exit = Exit { isExit :: Bool }
```
The distinction between `isFoo` and `isDeclaredFoo` is important for properties like functionality; a symbol may be declared functional, but the pattern is only functional if its arguments are also functional.

src/main/haskell/kore/test/Test/Kore/Exec.hs
```diff
@@ -280,3 +285,88 @@ applyToNoArgs sort name =
             , symbolOrAliasParams = []
             }
         []
+
+test_execGetExitCode :: TestTree
+test_execGetExitCode =
+    testGroup "execGetExitCode"
+    [ makeTestCase "No getExitCode symbol => ExitSuccess"
```
ttuegel: The list here should be indented by four spaces. Please also check your function argument indentation in the `where` block below.

src/main/haskell/kore/src/Kore/Attribute/Exit.hs
```diff
@@ -0,0 +1,72 @@
+{-|
+Module      : Kore.Attribute.Exit
+Description : Exit symbol attribute
+Copyright   : (c) Runtime Verification, 2018
```
virgil-serbanuta: 2019

src/main/haskell/kore/src/Kore/Attribute/Exit.hs
```diff
@@ -0,0 +1,72 @@
...
+License     : NCSA
+Maintainer  : thomas.tuegel@runtimeverification.com
```
virgil-serbanuta: By default, this should be the author's e-mail.

src/main/haskell/kore/src/Kore/Exec.hs
```diff
@@ -131,6 +138,29 @@ exec indexedModule strategy purePattern = do
...
+execGetExitCode
+    :: VerifiedModule StepperAttributes Attribute.Axiom
+    -- ^ The main module
+    -> ([Rewrite] -> [Strategy (Prim Rewrite)])
+    -- ^ The strategy to use for execution; see examples in "Kore.Step.Step"
+    -> CommonStepPattern Object
+    -- ^ The final pattern (top cell) to extract the exit code
+    -> Simplifier ExitCode
+execGetExitCode indexedModule strategy' purePattern =
+    case resolveSymbol indexedModule $ noLocationId "LblgetExitCode" of
```
virgil-serbanuta: Why use a hardcoded symbol name here?

src/main/haskell/kore/src/Kore/Attribute/Exit.hs
```diff
@@ -0,0 +1,72 @@
...
+
+-}
+module Kore.Attribute.Exit
```
virgil-serbanuta: Is this used anywhere besides its own test?

src/main/haskell/kore/src/Kore/Exec.hs
njohnwalker: It's not defined anywhere else

src/main/haskell/kore/src/Kore/Attribute/Exit.hs
njohnwalker: No, my first implementation depended on it, but it turns out you don't need to know the exit cell.  Should I remove it?

virgil-serbanuta: Unused code tends to break with time, so I would remove it unless someone else has some reason to keep it.

src/main/haskell/kore/src/Kore/Exec.hs
virgil-serbanuta: What is LblgetExitCode? Is it a builtin, a function defined by the kore prelude, a function which is sometimes defined? I assume it's the latter.

Anyway, there are two issues here: 
1. It seems to be dependent on the algorithm the frontend uses to generate names; in the other similar cases, i.e. builtin functions, we used attributes to identify them.
2. It seems specific to a given semantics.

I suggest either making an attribute for this, or sending it as a command line argument. If none of these is reasonable, I would define a string constant in this module.

dwightguth: This function is always going to be given this name by the k frontend if it exists because that is how it is implemented in the llvm backend. The decision was made so that static source files that are not dynamically generated can be statically linked against this function without the need for dynamically generated glue code. I'm not opposed to an attribute if it makes you feel more comfortable, but it's not specific to a semantics; it's generated by the k compiler as an implementation of the "exit" cell attribute. There is no advantage to making it configurable in one backend but not the other, and it was made this way in the llvm backend deliberately to simplify the design.

virgil-serbanuta: Ok, I see. Given this context, it makes a bit more sense, although I don't know why we need the static linkage you mentioned. My preference for an attribute was based on:
* we use attributes instead of relying on raw names for builtins, and this seems similar enough;
* an attribute would be within the compiler's control, while the symbol name space belongs to both the k compiler and to the user. Yes, I know that users can use random attributes, but it's expected that some of them are reserved for the compiler, and there are reasonable (IMO) ways to have namespace-like stuff there.

Anyway, given that the LLVM backend already uses this name, I guess it's acceptable to also use it here. I would still define that string constant I mentioned earlier. I wonder if @traiansf has any thoughts about this.

dwightguth: For reference, the getExitCode function needs to be called by the main function in the llvm backend. We could of course have simply introduced a dynamic glue function, but this design allows us to simply declare a function prototype for it and call it from the source file for the main function. Which would not be possible if we did not know the name of the function statically.

src/main/haskell/kore/src/Kore/Attribute/Exit.hs
njohnwalker: Okay, I'll get rid of it.

PR #501, opened 2019-03-11
Feature/repl step n

vladciobanu:
I'm sorry the commit history is not of much use. Notable changes / points of interest:
- new executable `kore-repl` (currently takes parameters similar to `kore-exec` for the `prove` part; it also currently ignores `--output`)
- bit of refactoring around `kore-exec/Main` (extracted code to `GlobalMain`) since I also needed some of that functionality in `kore-repl`
- added necessary/alternative functions for `repl` mode in the `Exec`, `Verification` and `Strategy` modules
- `Repl.hs` changed to reflect the new way we do repl
- I removed the old repl from everywhere, which is why there's a lot of changed tests (that used to take `noRepl`)

You can easily test the repl:

```
$ cd src/main/k/woking/imp
$ make prove/max-spec.krepl
```

Note that you need `dot`/`graphviz` installed for the `graph` command to work.

###### Reviewer checklist

---


kore/app/repl/Main.hs
```diff
@@ -0,0 +1,177 @@
...
+    { timeOut :: !SMT.TimeOut
+    , prelude :: !(Maybe FilePath)
+    }
+
+-- | Options for the kore repl.
+data KoreReplOptions = KoreReplOptions
+    { definitionModule :: !KoreModule
+    , specModule       :: !KoreModule
+    , smtOptions       :: !SmtOptions
+    , outputFileName   :: !(Maybe FilePath)
```
ttuegel: Since we don't plan to do anything with this option for the REPL, I would rather omit the field. We can still have the parser accept and ignore the option by joining the option parser with `<*` instead of `<*>`.

kore/app/repl/Main.hs
```diff
@@ -0,0 +1,177 @@
...
+parseKoreReplOptions =
+    KoreReplOptions
+    <$> parseMainModule
+    <*> parseSpecModule
+    <*> parseSmtOptions
+    <*> optional
+            (strOption
+                (  metavar "PATTERN_OUTPUT_FILE"
+                <> long "output"
+                <> help "Output file to contain final Kore pattern."
```
ttuegel: `help` should say that this option is ignored.

kore/app/repl/Main.hs
```diff
@@ -0,0 +1,177 @@
...
+                )
+            )
+  where
+    parseMainModule :: Parser KoreModule
```
ttuegel: Could we share the particular option parser definitions between `kore-exec` and `kore-repl` so that the common options remain synchronized?

kore/src/Kore/OnePath/Verification.hs
```diff
@@ -235,5 +237,104 @@ verifyClaim
...
+        -> Simplifier [(StrategyPattern (CommonExpandedPattern level), StepProof level Variable)]
+    transitionRule' =
+        OnePath.transitionRule
+            metadataTools
+            substitutionSimplifier
+            simplifier
+            axiomIdToSimplifier
+
+-- | TODO: Docs.
+type Configuration level = StrategyPattern (CommonExpandedPattern level)
```
ttuegel: Could you add documentation to this type synonym? I think you could also use it above.

kore/src/Kore/OnePath/Verification.hs
```diff
@@ -235,5 +237,104 @@ verifyClaim
...
+
```
ttuegel:
```suggestion
```

kore/src/Kore/OnePath/Verification.hs
```diff
@@ -235,5 +237,104 @@ verifyClaim
...
+    -- Given a default proof, pass it as a default to the transitionRule and
+    -- discard the proof part of its result.
+    stripProof
+        :: forall prim strategy f g proof
+        .  (Functor f, Functor g)
+        => proof
+        -> (prim -> (strategy, proof) -> f (g (strategy, proof)))
+        -> prim -> strategy -> f (g strategy)
+    stripProof defaultProof fn prim =
```
ttuegel: Is the `defaultProof` ever anything but `mempty`?

kore/src/Kore/Repl.hs
```diff
@@ -6,64 +6,428 @@ License     : NCSA
 Maintainer  : vladimir.ciobanu@runtimeverification.com
 -}
 
+{-# LANGUAGE TemplateHaskell #-}
+{-# OPTIONS_GHC -Wno-unused-top-binds    #-}
```
ttuegel: I strongly dislike turning off warnings by-module. I think it would be appropriate to export `ReplState`; that would encourage developing modular REPL features, and remove the warning.

kore/src/Kore/Repl.hs
```diff
@@ -6,64 +6,428 @@ License     : NCSA
...
+import           Kore.Step.StepperAttributes
+                 ( StepperAttributes )
+import qualified Kore.Step.Strategy as Strategy
+import           Kore.Unparser
+                 ( unparseToString )
+
+-- Type synonym for the actual type of the execution graph.
+type ExecutionGraph level
+    = Strategy.ExecutionGraph
+        (StrategyPattern (CommonExpandedPattern level))
```
ttuegel:
```suggestion
        (Configuration level)
```

kore/src/Kore/Repl.hs
```diff
@@ -6,64 +6,428 @@ License     : NCSA
...
+                    context = Graph.context graph node
+                case Graph.suc' context of
+                    [] -> pure (NoChildNodes, node)
+                    [configNo] -> do
+                        lensNode .= configNo
+                        pure (Success, node)
+                    neighbors -> pure (Branch neighbors, node)
+            else pure (NodeAlreadyEvaluated, node)
+
+    performStepNoBranching
```
ttuegel: It isn't clear to me what this does, could you add a comment?

kore/src/Kore/Repl.hs
```diff
@@ -6,64 +6,428 @@ License     : NCSA
...
+                        = emptyExecutionGraph claim
+                lensGraph .= graph
+                lensClaim .= claim
+                lensNode  .= root
+                putStrLn' "Execution Graph initiated"
+
+    showGraph0 :: StateT (ReplState level) Simplifier ()
+    showGraph0 = do
+        Strategy.ExecutionGraph { graph } <- Lens.use lensGraph
+
```
ttuegel:
```suggestion
```

kore/src/Kore/Repl.hs
```diff
@@ -6,64 +6,428 @@ License     : NCSA
...
+        liftIO
+            . (flip Graph.runGraphvizCanvas') Graph.Xlib
```
ttuegel: I think we could extract the combination of this line and the next line into a nice helper function. It would make this easier to read by separating the high-level of the REPL interface from the low-level of display frameworks.

Also, I see `Xlib`: does this work outside Linux?

kore/src/Kore/Step/Strategy.hs
```diff
@@ -276,6 +280,72 @@ constructExecutionHistory transit instrs0 config0 =
...
+    -> Strategy prim
+    -- ^ Primitive strategy
+    -> ExecutionGraph config
+    -- ^ execution graph so far
+    -> Graph.Node
+    -- ^ current "selected" node
+    -> m (ExecutionGraph config)
+    -- ^ graph with one more step executed for the selected node
+executionHistoryStep transit prim ExecutionGraph { graph, history } node
+    | nodeIsNotLeaf = error "Node has already been evaluated"
```
ttuegel: Guards should be indented two spaces only.

kore/src/Kore/OnePath/Verification.hs
vladciobanu: This is intended to be internal in the module and only as an alias to avoid ridiculously long types; not sure I have any docs to write to it.

vladciobanu: Okay, I ended up defining this right below `StrategyPattern` as `CommonStrategyPattern` and replacing all occurrences.

kore/src/Kore/Repl.hs
vladciobanu: I'll see if I can test it on a Mac tomorrow, but I think we can consider this as an "extra" feature for demo purposes. It will definitely change in the future.

PR #504, opened 2019-03-13
Remove level distinction

ttuegel:
This (draft) pull request removes the distinction between levels by setting `type Meta = Object`.

The library and programs build, but the test suite does not. The main barrier is the `AsAst` class used in the `ProofAssistant` test.

@virgil-serbanuta Is the `ProofAssistant` test still relevant? If we can remove it, then this pull request will be simple to finish up. Otherwise, we may need to rewrite those tests to use `CommonPurePattern`. I think this is the only significant change to the test suite.

###### Reviewer checklist

---


virgil-serbanuta:
I think that, if the `Logic` package is still needed, then we want to keep the ProofAssistant test. I also think that we don't want to remove the `Logic` package, at least not right now, since we may still want to do ML proofs.

@traiansf Is the above accurate enough?

traiansf:
> I think that, if the `Logic` package is still needed, then we want to keep the ProofAssistant test. I also think that we don't want to remove the `Logic` package, at least not right now, since we may still want to do ML proofs.
> 
> @traiansf Is the above accurate enough?

I think it is; but why is it AsAst hard to translate?

PR #507, opened 2019-03-14
Remove dead code

ttuegel:
This pull request removes about 8,000 lines of dead code in support of #504.

###### Reviewer checklist

---


ttuegel:
@virgil-serbanuta During standup @traiansf told me this would be OK to remove, but now I wonder why given your comment on #504 (which I had not seen yet).

traiansf:
Well, as I said in the meeting, the proof assistant can be used to check ML proofs.  However (1) it will take a long time until we get to generate anything we would need to check, as proofs are not on our critical path anymore; (2) the proof system itself changed at least twice since there hasn't been any work on this part of the code, and might change in the future; and (3) if we will build an interactive theorem prover, it will be more likely be built around the things we're currently developing, as the REPL; moreover, as a proof checker, it might be worthwhile to be a project on its own.  Hence, I'm ok for us to tag it for future reference and remove it from our current master.

virgil-serbanuta:
If Traian says it's ok to tag+remove, then I won't object.

PR #508, opened 2019-03-14
Tests for branch pruning

ttuegel:
After #484 I indicated I would write tests for branch pruning in the new simplification functions. I actually found one error where a branch was not pruned, although this matched the behavior of the code I was refactoring, i.e. the branch was never pruned before, but now it is. The pull request also contains other various cleanup.

###### Reviewer checklist

---


ttuegel:
Disregard this pull request for now. I need to split it up because one part is causing the KEVM tests to fail.

kore/src/Kore/Unification/UnifierImpl.hs
```diff
@@ -141,15 +135,15 @@ simplifyAnds
                 (result, _) <-
                     termUnification
                         tools
-                        substitutionSimplifier
+                        emptyPredicateSubstitutionSimplifier
```
virgil-serbanuta: I was trying to figure these out, but I guess it's easier to ask: Why is it good enough to use an empty simplifier here and below? Also, even if it works now, why would this also work in the future? Why not an `crash the program` simplifier?

ttuegel: > Why is it good enough to use an empty simplifier here and below? Also, even if it works now, why would this also work in the future?

The normalizer is calls the simplifier in a loop until the result stabilizes. It isn't necessary for the simplier to call itself recursively here because it is being called within that loop.

> Why not an `crash the program` simplifier?

Well, we don't want to crash the program. I expect that `termUnification` will try to call the simplifier. By returning its argument unsimplified, `emptyPredicateSubstitutionSimplifier` is delaying simplification until the next iteration of the loop in the normalizer.

virgil-serbanuta: Copy-pasting from Slack:

As far as I see it (I hope I'm not wrong), we're talking about different things:
1. `normalizeSubstitutionDuplication` will take duplicate variable assignments, will unify them, and will produce a predicate+substitution out of those, it will rejoin the substitutions and repeat until it stabilizes.
2. `substitutionSimplifier` is used when unification needs to merge predicates/substitutions that are produced when unifying. Say, we want to build a `ceil`, and we want to apply a simplification axiom to that, and it turns out that we need to simplify predicate-substitutions when doing that. At that moment we would need this substitutionSimplifier to do its work.

I'm not 100% sure about my answer, I would need a bit more time to answer that in a more precise way, and I don't have it right now.

PR #509, opened 2019-03-15
Interrupt stepper with Ctrl-C without exiting the repl

vladciobanu:
I tested it by adding a `sleep` in the stepper function. It only works when *not* running through `kprove`. I'm not sure we want to fix that right now, since that's a hack already. Normally, whenever we want to actually port this to the front-end, it will probably be something entirely different.

###### Reviewer checklist

---


vladciobanu:
I added an entry in the wiki page about interrupting, see https://github.com/kframework/kore/wiki/Kore-REPL#interrupting

kore/src/Kore/Repl.hs
```diff
@@ -174,6 +180,13 @@ runRepl tools simplifier predicateSimplifier axiomToIdSimplifier axioms' claims'
                 pure True
             else pure False
 
+    catchInterruptWithDefault :: a -> Simplifier a -> Simplifier a
+    catchInterruptWithDefault def sa =
```
ttuegel: I think the lens-y (optical?) way of doing this is actually more complicated. Why not this?
```.haskell
catchInterruptWithDefault def =
    handle $ \UserInterrupt -> do
        liftIO $ putStrLn "Step evaluation interrupted."
        pure def
```

PR #511, opened 2019-03-15
Refactor rule application for branching

ttuegel:
###### Reviewer checklist

---


ttuegel:
@vladciobanu One reason I marked you to review this pull request is that I would like your feedback about what we should expose from `applyRewriteRule` to the REPL. (I will do that in the next pull request though.)

kore/src/Kore/Step/AxiomPatterns.hs
```diff
@@ -72,6 +77,17 @@ data RulePattern level variable = RulePattern
     }
     deriving (Eq, Show)
 
+instance Unparse (variable level) => Pretty (RulePattern level variable) where
+    pretty RulePattern { left, right, requires } =
```
virgil-serbanuta: We use it everywhere, so I'm not necessarily asking for a change: with this type of extracting the fields it's easy to forget to update the Pretty instance when adding fields to the RulePattern. In many cases this would not be an issue, since we would actually want to use the new field in the same PR, but here we will only notice it only when we use the repl or something else that's not correlated with adding the field.

One suggestion would be to use `RulePattern _ _ _ _` for matching and to use the field getters inside.

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -887,3 +860,528 @@ unwrapPredicateVariables
...
+wrapUnificationOrSubstitutionError
+    :: Functor m
+    => ExceptT (UnificationOrSubstitutionError level variable) m a
+    -> ExceptT (StepError                      level variable) m a
+wrapUnificationOrSubstitutionError =
+    withExceptT unificationOrSubstitutionToStepError
+
+{- | Lift an action from the unifier into the stepper.
+ -}
+fromUnification
```
virgil-serbanuta: `liftFromUnification`, maybe, to signal that this is a boilerplate-like transformation?

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -887,3 +860,528 @@ unwrapPredicateVariables
...
+instantiateRule
+    metadataTools
+    predicateSimplifier
+    patternSimplifier
+    axiomSimplifiers
+
+    unified@Predicated { term = axiom@RulePattern { left, right, requires } }
+  = do
+    let unifier = unified { term = () }
+        merged = PredicateSubstitution.fromPredicate requires *> unifier
```
virgil-serbanuta: It's late so I'm stopping the review here (sorry, I wanted to finish it today, but I can't).

I don't have time to look it up, so how does this thing work? *> is supposed to discard the first action, so I guess it must be something about some side effects? Anyway, is there a more obvious way of doing this?

ttuegel:
> How would this interact with regular `transitionRule` steps from the verification module? Ideally, `OnePath.transitionRule` would also return a list of applied rules (which, in turn seems to use `stepWithRemainders` so maybe that's the connection).

@vladciobanu 
What's an "applied rule"? Or maybe a better question is "when": the rule undergoes unification, then its variables are instantiated by applying the unification substitution, then the initial conditions are applied. Which of these pieces of information should we expose through the REPL?

kore/src/Kore/Step/BaseStep.hs
vladciobanu: Had to look it up myself when reviewing. I *think* it uses the `Predicated/Monad` semantics, which is:

given a `p :: Predicated term` (with `term1 /\ predicate1 /\ substitution1` and a kleisli arrow `f :: term -> Predicated`
apply `f p` and obtain `term2 /\ predicate2 /\ substitution2`
return a Predicated like `term2 /\ (predicate1 /\ predicate2) /\ (substitituion1 <> substitution2)`

So basically, it is assumed that the function `f` only takes into account the `term` part of `p`.

Which in turn means that `p *> q` would mean:
`qterm /\ (ppredicate /\ qpredicate) /\ (psubstitution <> qsubstitution)`

Considering a `Predicate term` is `(term, predicate, substitution)`, this lines up with the tuple/pair Monad instance, which requires a `Monoid` instance on `fst`.

ttuegel: @vladciobanu is right; the `Applicative` and `Monad` instances for `Predicated` combine the side conditions (predicates and substitutions) using `\and`. `p *> q` means "combine the predicates and substitutions of `p` and `q`, but take only the `term` from `q`". Actually, in this case the term of both `p` and `q` is `()`, so it wouldn't matter which one is taken.

vladciobanu:
I think I was wrong, we definitely want one rule not multiple. The rule should be the rule that was used for unification for each step.

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -890,3 +863,526 @@ unwrapPredicateVariables
...
+applyRule
+    metadataTools
+    predicateSimplifier
+    patternSimplifier
+    axiomSimplifiers
+
+    initial
+    instantiated
+  =
+    applyInitialConditions' initial (right <$> instantiated)
```
virgil-serbanuta: Doesn't `instantiated` already contain the initial predicate+substitution?

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -890,3 +863,526 @@ unwrapPredicateVariables
...
+        , Show    (variable Object)
+        , Unparse (variable Object)
+        , SortedVariable variable
+        )
+    => MultiOr (MultiAnd (Predicate Object variable))
+    -> MultiOr (Predicate Object variable)
+negateCoverage = negateCoverageWorker . Foldable.toList
+  where
+    negateCoverageWorker [] = MultiOr.singleton Predicate.makeTruePredicate
+    negateCoverageWorker (x : xs) =
```
virgil-serbanuta: Would this be more clear if it was build with one of the three `MultiOr.*crossProduct*` functions?

virgil-serbanuta:
One more thing: I'm not sure I am right, but I had to spend some time to find out that ensures is used in the remainder, and that the initial conditions are added twice. I find the current flow a bit hard to follow. If I was right, then that's a reason to think that it's hard not only for me. I would be in favor of making clear which piece of data is used for what, e.g. you could pass only `right` and the predicate-substitution to `applyRule`

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -890,3 +863,526 @@ unwrapPredicateVariables
...
+    ::  ( Ord     (variable Object)
+        , Show    (variable Object)
+        , Unparse (variable Object)
+        , SortedVariable variable
+        )
+    => PredicateSubstitution Object (StepperVariable variable)
+    -- ^ Unification solution
+    -> MultiAnd (Predicate Object (StepperVariable variable))
+unificationCoverage Predicated { predicate, substitution } =
+    predicateCoverage predicate <|> substitutionCoverage substitution'
```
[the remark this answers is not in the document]
virgil-serbanuta: BTW, regardless of whether this is right or not, could you add a test that shows where `ensures` is used by the step function?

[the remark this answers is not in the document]
ttuegel: I think `ensures` should only constrain the right-hand side of the rule. In principle, the `ensures` is on the right-hand side under an `\exists`, so we couldn't include it in the remainder. For reference, this is what we do now: https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/BaseStep.hs#L480

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -59,33 +73,38 @@ import qualified Kore.Predicate.Predicate as Predicate
 import           Kore.Step.Axiom.Data
                  ( BuiltinAndAxiomSimplifierMap )
 import           Kore.Step.AxiomPatterns
-                 ( RewriteRule (RewriteRule), RulePattern (RulePattern) )
+                 ( RewriteRule (RewriteRule), RulePattern (..) )
```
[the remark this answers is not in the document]
ttuegel: I don't think we ever updated the style guide to say that, but it's not a bad idea.

kore/src/Kore/Step/BaseStep.hs
ttuegel: Do you mean like this?
```.hs
    negateCoverageWorker [] = MultiOr.singleton Predicate.makeTruePredicate
    negateCoverageWorker (x : xs) =
        MultiOr.crossProductGeneric
            Predicate.makeAndPredicate
            (mkNotMultiAnd x)
            (negateCoverageWorker xs)
```

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -890,3 +863,526 @@ unwrapPredicateVariables
...
+    patternSimplifier
+    axiomSimplifiers
+
+    unified@Predicated { term = axiom }
+  = do
+    let unifier = unified { term = () }
+        RulePattern { requires, ensures } = axiom
+        merged =
+            PredicateSubstitution.fromPredicate requires
+            *> PredicateSubstitution.fromPredicate ensures
```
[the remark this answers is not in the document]
ttuegel: > Why is it ok to drop the term with `<*` and `*>` without knowing that it's `()`?

We drop terms all the time; for example, when we apply a rule we drop the initial term and replace it with the right-hand side of the rule.

> The Applicative instance lets us do that, so we must be very careful with the types we're handling (i.e. if we ever change a function to return an `ExpandedPattern` instead of a `PredicateSubstitution`, we would have to check all places where its result might be used with, say, `*>`).

This is a problem with or without using `Applicative`. There are numerous places—even without counting the uses of `*>`—where we assume we have been returned a `PredicateSubstitution` and drop the `term`.

> Wouldn't the code be more readable and safe if we were using named functions to do some of these operations?

Are you arguing against using `Applicative` for `Predicated` everywhere, or only here? I will only address this specific use: I do not think it would be more readable. The ways this usage could be unsafe are: 1. we change the return type of `PredicateSubstitution.fromPredicate` (that would be deliberately evil and I expect we will not do that) or 2. we change the type of `instantiateRule` to be more permissive (in which case we would need to inspect the body of the function to see if it can operate over the more permissive type—which is true no matter if we use `Applicative` or less permissively-typed functions).

kore/src/Kore/Step/BaseStep.hs
ttuegel: No, but I will refactor this section of code to make that clearer.

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -868,25 +845,466 @@ unwrapStepErrorVariables =
...
+    finalCondition <- normalize (initial <> unification <> ensuresCondition)
+    -- Apply the normalized substitution to the right-hand side of the axiom.
+    let
+        Predicated { substitution } = finalCondition
+        substitution' = Substitution.toMap substitution
+        RulePattern { right = finalTerm } = renamedRule
+        finalTerm' = Pattern.substitute substitution' finalTerm
+    return finalCondition { ExpandedPattern.term = finalTerm' }
+  where
+    normalize =
```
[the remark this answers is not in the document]
ttuegel: The only thing we could do with a helper function is the `liftFromUnification`; we would still have to pass in the `MetadataTools` and all three simplifiers.

kore/src/Kore/Step/Representation/Predicated.hs
```diff
@@ -0,0 +1,276 @@
...
+{- | @Predicated@ represents a value conditioned on a predicate.
+
+@Predicated level variable child@ represents a @child@ conditioned on a
+@predicate@ and @substitution@ (which is a specialized form of predicate).
+
+The 'Applicative' instance conjoins the predicates and substitutions so that the
+result is conditioned on the combined predicates of the inputs. The combined
+predicate and substitution are /not/ normalized.
+
+There is intentionally no 'Monad' instance; such an instance would have
```
virgil-serbanuta: There is a monad instance, see below.

kore/test/Test/Kore/Comparators.hs
```diff
@@ -1554,3 +1557,149 @@ instance EqualWithExplanation (AxiomIdentifier level)
...
+        ]
+
+instance EqualWithExplanation Attribute.Axiom where
+    compareWithExplanation = structCompareWithExplanation
+    printWithExplanation = show
+
+instance StructEqualWithExplanation Attribute.Axiom where
+    structConstructorName _ = "Axiom"
+    structFieldsWithNames expect actual =
+        [ Function.on (EqWrap "heatCool = "       ) Attribute.heatCool       expect actual
```
virgil-serbanuta: Line too long. Also below.

kore/src/Kore/Step/BaseStep.hs
virgil-serbanuta: I meant something like
```hs
negateCoverageWorker xs =
    fmap
        Predicate.makeMultipleAndPredicate 
        (MultiOr.fullCrossProduct
            (map makeNotOnEachPieceOfAnd xs)
        )
```

kore/src/Kore/Step/BaseStep.hs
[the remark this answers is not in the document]
virgil-serbanuta: Well, it's almost there, but maybe it should be explicit:

```
Do not use prefixes on exported names to replace namespaces unless that helps readability. Even then, consider using a separate import with an appropriate alias. As an example, ‘delay’ below is a generic name, which may be harder to understand locally, so you should provide some local disambiguation.
```

kore/src/Kore/Step/BaseStep.hs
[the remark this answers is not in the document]
virgil-serbanuta: I'm saying that it's easy to drop a term by mistake, so, as a rule, we should have something explicit for dropping them. We used `term=()` in several places and that's sort of fine because it's explicit. It could be better if we would return something without the term, or we would check that we can drop the term (i.e. the entire thing is top, bottom, or the term is top) and crash otherwise.

`*>` makes it very easy to implicitly drop the term. That's only a small part of `Applicative`, I'm not against using `<*>` since that applies a function which, supposedly, wants to handle things that have terms.

kore/src/Kore/Step/BaseStep.hs
```diff
@@ -868,25 +845,466 @@ unwrapStepErrorVariables =
...
+    -> BuiltinAndAxiomSimplifierMap level
+    -- ^ Map from symbol IDs to defined functions
+
+    -> ExpandedPattern level variable
+    -- ^ Configuration being rewritten.
+    -> RewriteRule level variable
+    -- ^ Rewriting axiom
+    -> ExceptT (StepError level variable) Simplifier
+        (OrStepResult level variable)
+applyRewriteRule
```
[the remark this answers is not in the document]
ttuegel: Just documenting our conversation: Yes, that is coming in the next pull request.

kore/src/Kore/Step/Representation/Predicated.hs
ttuegel: The Monad instance is removed now.

PR #514, opened 2019-03-18
Show Axiom Location

vladciobanu:
###### Reviewer checklist

---


kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+    def = Location Nothing Nothing
+
+-- | Kore identifier representing the @location@ attribute symbol.
+locationId :: Id Object
+locationId = "org'Stop'kframework'Stop'attributes'Stop'Location"
+
+instance ParseAttributes Location where
+    parseAttribute = Parser.withApplication locationId parseApplication
+      where
+
```
virgil-serbanuta: Why the empty line?

The same for the file below.

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
+{-|
+Module      : Kore.Attribute.Location
+Description : Line/column location attribute
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+Maintainer  : vladimir.ciobanu@runtimeverification.com
```
virgil-serbanuta: Could you update docs/attributes.md with both attributes?

kore/src/Kore/Repl.hs
```diff
@@ -54,6 +54,9 @@ import qualified Kore.AST.Common as Kore
                  ( Variable )
 import           Kore.AST.MetaOrObject
                  ( MetaOrObject )
+import           Kore.Attribute.Axiom
+                 ( LineColumn (..), Location (..), Source (..), location,
```
virgil-serbanuta: Locatiuon and source are fields that do not use the type name as prefix, so, according to our style guide, they should be in a qualified import.

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+        parseApplication params args _ = do
+            Parser.getZeroParams params
+            case args of
+                [_] -> do
+                    arg <- Parser.getOneArgument args
+                    StringLiteral str <- Parser.getStringLiteral arg
+                    pure . maybe def id . parseMaybe locationParser $ Text.unpack str
+                _ -> pure def
+
+type Parser = Parsec String String
```
ttuegel: It's a little confusing that we are using two `Parser` types here. Could you add a comment to this definition explaining what it is used for?

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+
+locationParser :: Parser Location
```
ttuegel: Could you leave a comment describing what the expected format is?

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+locationParser =
+    Location
```
ttuegel: This parser would accept
```
Location(2,3
```
and
```
,5,7)
```
as input; I don't think that's right, is it?

kore/src/Kore/Attribute/Source.hs
```diff
@@ -0,0 +1,70 @@
...
+import           Text.Megaparsec
+                 ( Parsec, noneOf, parseMaybe )
+import           Text.Megaparsec.Char
+
+import           Kore.AST.Kore
+import           Kore.Attribute.Parser
+                 ( ParseAttributes (..) )
+import qualified Kore.Attribute.Parser as Parser
+
+data Source = Source
```
ttuegel: Why not a `newtype`?

kore/src/Kore/Repl.hs
```diff
@@ -417,12 +439,35 @@ replInterpreter =
...
     unparseStrategy =
         \case
-            Bottom -> "Reached goal!"
+            Bottom -> "Reached bottom"
             Stuck pat -> "Stuck: \n" <> unparseToString pat
             RewritePattern pat -> unparseToString pat
 
-    indexNotFound :: String
-    indexNotFound = "Variable or index not found"
+    formatSourceAndLocation :: (Source, Location) -> Maybe String
```
ttuegel: Maybe we could give instances of `Pretty` for `Source` and `Location` instead?

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+import qualified Kore.Attribute.Parser as Parser
+
+data LineColumn = LineColumn
+    { line   :: Int
+    , column :: Int
+    } deriving (Eq, Ord, Show, Generic)
+
+instance NFData LineColumn
+
+instance Default LineColumn where
```
ttuegel: I'm not sure this `Default` instance makes sense. It doesn't seem to be used, is that right?

kore/src/Kore/Attribute/Location.hs
```diff
@@ -0,0 +1,93 @@
...
+import           Text.Megaparsec.Char.Lexer
+                 ( decimal )
+
+import           Kore.AST.Kore
+import           Kore.Attribute.Parser
+                 ( ParseAttributes (..) )
+import qualified Kore.Attribute.Parser as Parser
+
+data LineColumn = LineColumn
+    { line   :: Int
```
ttuegel: These fields should be strict.

kore/src/Kore/Repl.hs
vladciobanu: The problem with that is they are not really useful on their own, especially `Location`. Ideally, I would change this attribute in the frontend to be a single attribute, although I'm not really sure it's worth the effort since it's really not that important.

ttuegel: We could make it a single attribute in the backend.

vladciobanu: I thought about that, but don't we parse attributes sequentially? So when trying to merge two attributes, you'd have to "guess" the order? Or maybe I am missing something, because the naive combined parser fails:

```
data SourceLocation = SourceLocation
    { location :: !Location
    , source   :: !Source
    } deriving (Eq, Ord, Show, Generic)

instance ParseAttributes SourceLocation where
    parseAttribute patt attrs =
        SourceLocation
            <$> parseAttribute patt def
            <*> parseAttribute patt def
```

kore/src/Kore/Attribute/Location.hs
vladciobanu: Done.

kore/src/Kore/Attribute/Location.hs
vladciobanu: Fixed.

kore/src/Kore/Repl.hs
vladciobanu: Nevermind, that was obviously the wrong approach. I figured out how to do it and updated the PR (see latest commit).

kore/src/Kore/Attribute/SourceLocation.hs
```diff
@@ -0,0 +1,78 @@
...
+    , source   :: !(Maybe Source)
+    } deriving (Eq, Ord, Show, Generic)
+
+instance NFData SourceLocation
+
+instance Default SourceLocation where
+    def = SourceLocation Nothing Nothing
+
+instance ParseAttributes SourceLocation where
+    parseAttribute patt SourceLocation { location, source } = do
```
ttuegel: Other "aggregate" attributes define their parser as
```.hs
    lensLocation (parseAttribute patt)
    >=> lensSource (parseAttribute patt)
```
which handles duplicate attributes correctly. Could we do that here, too?

kore/src/Kore/Repl.hs
```diff
@@ -147,38 +155,34 @@ runRepl tools simplifier predicateSimplifier axiomToIdSimplifier axioms' claims'
...
-                        node
+                graph' <-
+                    lift
+                        . catchInterruptWithDefault graph
+                        $ verifyClaimStep
+                            tools
+                            simplifier
+                            predicateSimplifier
+                            axiomToIdSimplifier
+                            claim
```
ttuegel: Could we pass the `ReplState` directly to `verifyClaimStep` instead of passing its fields one by one?

docs/attributes.md
```diff
@@ -215,6 +215,26 @@ A symbol is functional if it is given the `functional` attribute or the
 
 ## misc.
 
+### Location
+
+```
+org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(sl,sc,el,ec)")
```
ttuegel: Are any of `sl`, `sc`, `el`, or `ec` optional?

vladciobanu: No, it's all-or-nothing from the front-end.

kore/src/Kore/Repl.hs
vladciobanu: I don't like the idea of having any module depend on anything related to the REPL right now, so I would avoid this.

vladciobanu:
@ttuegel I either addressed or replied to every comment. Let me know what you think.

kore/src/Kore/Repl.hs
ttuegel: I didn't observe that `verifyClaimStep` comes from outside the module. Sounds good to me!

ttuegel:
@vladciobanu The only thing I'm left wondering about is the cases where the attribute parser doesn't fail on malformed input, such as: if the `Location` attribute or `Source` attribute has no arguments, or if the `Location(_,_,_,_)` string is missing either pair.

vladciobanu:
I have seen examples where either Source or Location are empty, and I don't think we want to FAIL on these. Maybe we want to produce a message with Pretty saying "missing X", just so it's obvious? Either way, I don't think I want to spend the time investigating this issue in the front-end quite yet.

Although the type doesn't quite say the same story, its parser only allows "all-or-nothing": https://github.com/kframework/kore/blob/62b1fb0a87459b9b1d50e14b155b707fd6a66ef8/kore/src/Kore/Attribute/Location.hs#L86

I guess the correct representation for `Location` would be `Maybe (LineColumn, LineColumn)`, but again, I'm not sure it's really a relevant distinction. Whether the attribute is present and empty, or missing is really not that important.

PR #518, opened 2019-03-20
Design decision for 'ensures' clauses in rewrite axioms.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/design-decisions/2019-03-20-Ensures-for-Axioms.md
+++ b/design-decisions/2019-03-20-Ensures-for-Axioms.md
@@ -75,7 +75,7 @@ Now, note that this is an axiom, so the entire expression must be a
 
 But `ensures` is a predicate.
 1. If `ensures` is `⊥`, then `left ∧ requires` must be `⊥`, which means that either
-   `left` or `requires` are `⊥`, which means that `⌈left⌉` is `⊥`.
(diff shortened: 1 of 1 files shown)


design-decisions/2019-03-20-Ensures-for-Axioms.md
```diff
@@ -0,0 +1,122 @@
...
+```
+left ∧ requires ⇒ ∃ Z . right ∧ ensures
+left ∧ requires → • ∃ Z . right ∧ ensures  -- expand ⇒
+left ∧ requires → ∃ Z . • right ∧ ensures  -- ∃ commutes with •
+∃ Z . left ∧ requires → • right ∧ ensures  -- ∃ commutes with ∨ and →
+∃ Z . ( (left ∧ requires → • right)        -- a → b ∧ c iff (a → b) ∧ (a → c)
+      ∧ (left ∧ requires → ensures)
+      )
+```
+
```
traiansf: I think the argument below could have probably made easier, by arguing that 
`∃ Z . (A /\ B ) -> (∃ Z . A) /\ (∃ Z . B )` which means that, in particular `(∃ Z . B )` is a theorem.

Also, I don't think you can discard the forall quantifier when checking UNSAT, because when checking satisfiability all vars are existentially quantified by default.

PR #524, opened 2019-03-25
Apply rules in sequence or parallel

ttuegel:
## API changes

* `Kore.Step.Step` moves to `Kore.Step`: This module is the high-level interface to everything under `Kore.Step`.
* `Kore.Step.BaseStep` moves to `Kore.Step`: This module is the low-level interface.
* Factored `Kore.Step.Proof` out of `Kore.Step.Step`.
* Renamed `StepperVariable` to `Target` and moved to `Kore.Variables.Target`: The new name describes **what** the type is used for—targeting specific variables for unification and substitution—instead of **where** it is used. It moves out of `Kore.Step` because it is not specific to it.
* `Kore.Step.AxiomPatterns` moves to `Kore.Step.Rule`: "Rule" is part of the domain vocabulary, "AxiomPattern" is not.
* Factored `Kore.Step.Remainder` out of `Kore.Step.Step`.
* `Kore.Step.StepperAttributes` moves to `Kore.Attribute.Symbol`: `StepperAttributes` is the type of attributes attached to symbols, so it goes to live with the other attribute types and gets a more-specific name.

###### Reviewer checklist

---


kore/src/Kore/Step/Step.hs
```diff
@@ -1,200 +1,803 @@
...
+            { rewrittenPattern = rewrittenPattern1
+            , remainder = remainder1
+            }
+        OrStepResult
+            { rewrittenPattern = rewrittenPattern2
+            , remainder = remainder2
+            }
+      =
+        OrStepResult
+            { rewrittenPattern = rewrittenPattern1 <> rewrittenPattern2
```
virgil-serbanuta: This does not apply idempotency laws for 'or', is that appropriate here?

PR #525, opened 2019-03-26
Added back smt lemma generation. Added smt-hooks

traiansf:
I think it still needs at least one integration test.  Still figuring that out.

###### Reviewer checklist

---


kore/src/Kore/Step/TranslateSMT.hs
```diff
@@ -213,16 +214,19 @@ translatePredicate translateUninterpreted predicate =
...
-            Nothing -> empty
+      = case getSmtlib (symAttributes smtlibTools applicationSymbolOrAlias)
+         <|> getSmthook (symAttributes smthookTools applicationSymbolOrAlias)
+         of Nothing -> empty
             Just sExpr -> shortenSExpr <$>
                 applySExpr sExpr
                     <$> zipWithM translatePattern
                         applicationChildrenSorts
                         applicationChildren
         where
```
virgil-serbanuta: I'm fairly sure there's something wrong about indentation around this line. Since this is a revert of the revert, you can leave it as-is, but you can also fix it.

src/main/k/working/tests/smt-lemma/tests/1.t
```diff
@@ -0,0 +1 @@
+test(1)
```
virgil-serbanuta: Could you add some comments about what these things are checking?

PR #528, opened 2019-03-27
show unevaluated or stuck leafs

ana-pantilie:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Repl/Interpreter.hs
+++ b/kore/src/Kore/Repl/Interpreter.hs
@@ -185,12 +185,12 @@ showLeafs = do
     let result =
             groupSort
                 . catMaybes
-                . fmap (f graph)
+                . fmap (getNodeState graph)
                 $ leafs
 
     putStrLn' $ foldr ((<>) . showPair) "" result
   where
-    f graph node =
+    getNodeState graph node =
         maybe Nothing (\x -> Just (x, node))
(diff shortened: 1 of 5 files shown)


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -166,6 +172,34 @@ showConfig configNode = do
...
+    let leafs = filter (\x -> (Graph.outdeg graph x) == 0) nodes
+    let result =
+            groupSort
+                . catMaybes
+                . fmap (f graph)
+                $ leafs
+
+    putStrLn' $ foldr ((<>) . showPair) "" result
+  where
+    f graph node =
```
vladciobanu: I know I helped with this refactor, but we forgot to rename `f` to something sensible. How about `getNodeState`?

kore/test/Test/Kore/Repl/Parser.hs
```diff
@@ -138,6 +138,12 @@ configTests =
                               \expecting end of input, integer, or white space\n"
     ]
 
+leafsTests :: [ParserTest ReplCommand]
```
vladciobanu: You forgot to add this to the big list on top, so it's not executing yet. A good rule of thumb is to try to make the test so that it fails, for example:

```
"this should fail" `parsesTo` ShowLeafs
```

then run the tests and make sure it fails. You can then change the test to something that works and re-run to make sure the tests pass.

PR #529, opened 2019-03-27
Smt sorts

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/SMT/Sorts.hs
+++ b/kore/src/Kore/Step/SMT/Sorts.hs
@@ -1,0 +1,416 @@
+{-|
+Module      : Kore.Step.SMT.Sorts
+Description : Declares sorts to the SMT solver.
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+Maintainer  : virgil.serbanuta@runtimeverification.com
+-}
+
+module Kore.Step.SMT.Sorts
+    ( translateSort
+    , declareSorts
+    ) where
+
+import           Control.Monad
+                 ( zipWithM )
+import qualified Data.Map.Strict as Map
+import           Data.Maybe
+                 ( mapMaybe )
+import           Data.Reflection
+                 ( Given, given )
+import qualified Data.Set as Set
+import           Data.Text
+                 ( Text, isPrefixOf, pack )
+
+import           Kore.AST.Common
+                 ( SymbolOrAlias (SymbolOrAlias), Variable (Variable) )
+import           Kore.AST.Common as SymbolOrAlias
+                 ( SymbolOrAlias (..) )
+import qualified Kore.AST.Common as Variable
+                 ( Variable (..) )
+import           Kore.AST.Identifier
+                 ( Id (getId) )
+import           Kore.AST.Kore
+                 ( KorePattern, VerifiedKorePattern )
+import           Kore.AST.MetaOrObject
+                 ( Object (Object), Unified )
+import           Kore.AST.Sentence
+                 ( SentenceAxiom (SentenceAxiom), SentenceSort (SentenceSort) )
+import qualified Kore.AST.Sentence as SentenceSort
+                 ( SentenceSort (..) )
+import qualified Kore.AST.Sentence as SentenceAxiom
+                 ( SentenceAxiom (..) )
+import           Kore.AST.Valid
+                 ( pattern App_, pattern Bottom_, pattern Exists_, pattern Or_,
+                 Valid, pattern Var_ )
+import qualified Kore.Attribute.Axiom as Attribute
+import qualified Kore.Attribute.Axiom.Constructor as Axiom.Constructor
+import           Kore.Attribute.Hook.Hook
+                 ( Hook (Hook) )
+import qualified Kore.Attribute.Hook.Hook as Hook
+import           Kore.Attribute.Smtlib
+                 ( applySExpr )
+import           Kore.Attribute.Smtlib.Smtlib
+                 ( Smtlib (Smtlib) )
+import qualified Kore.Attribute.Smtlib.Smtlib as Smtlib
+import qualified Kore.Attribute.Sort as Attribute
+import           Kore.Attribute.Symbol
+                 ( Symbol )
+import qualified Kore.Builtin.Bool as Bool
+import qualified Kore.Builtin.Int as Int
+import qualified Kore.Domain.Builtin as Domain
+import           Kore.IndexedModule.IndexedModule
+                 ( IndexedModule, recursiveIndexedModuleAxioms,
+                 recursiveIndexedModuleSortDescriptions )
+import           Kore.IndexedModule.MetadataTools
+                 ( MetadataTools (sortAttributes) )
+import           Kore.Sort
+                 ( Sort (SortActualSort), SortActual (SortActual) )
+import qualified Kore.Sort as SortActual
+                 ( SortActual (..) )
+import           Kore.Step.Pattern
+                 ( CommonStepPattern, fromKorePattern )
+import           Kore.Step.SMT.Encoder
+                 ( encodeName )
+import           Kore.Unparser
+                 ( unparseToString )
+import qualified SMT
+
+data SMTConstructor =
+    SMTConstructor
+        { name :: Text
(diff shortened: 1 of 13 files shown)


virgil-serbanuta:
In its current form, this crashes z3 4.8.0 because the following crashes 4.8.0:
```
(push 1 )
(declare-datatypes ()
    (
        (HB_S
            HB_C
            (HB_D (HB_D1 HB_T))
        )
        (HB_T HB_E )
    )
)
(declare-fun x () HB_S )
(assert (not (= x HB_C ) ) )
(assert (not (= x (HB_D HB_E ) ) ) )
(check-sat )
(pop 1 )
```

Still, some feedback would be nice.

ttuegel:
That `declare-datatypes` command is not valid according to the SMT-LIB 2 standard because the first argument (list of datatypes) does not have the same length as the second argument (list of declarations).

ttuegel:
I think the problem might be that Z3 interprets plain `assert`s about a function as being a definition. The following correctly returns `unsat`:
```
(declare-datatypes
    ( (HB_S 0) (HB_T 0) )
    (
      (
        ; Constructors of HB_S
        (HB_C)
        (HB_D (HB_D1 HB_T))
      )
      (
        ; Constructors of HB_T
        (HB_E)
      )
    )
)
(declare-fun x () HB_S )
;(assert (not (= x HB_C ) ) )
;(assert (not (= x (HB_D HB_E ) ) ) )
(assert (and (= x HB_C) (= x (HB_D HB_E))))
(check-sat )
```

kore/src/Kore/Step/SMT/Sorts.hs
```diff
@@ -0,0 +1,416 @@
...
+        (recursiveIndexedModuleAxioms indexedModule)
+
+parseSMTDataType
+    ::  ( Attribute.Axiom
+        , SentenceAxiom param VerifiedKorePattern
+        )
+    -> Maybe SMTDataType
+parseSMTDataType (attributes, SentenceAxiom {sentenceAxiomPattern})
+  | Axiom.Constructor.isConstructor (Attribute.constructor attributes)
+  = case fromKorePattern Object sentenceAxiomPattern of
```
ttuegel: A good refactoring would be to make this function (or a variant of it) not return an `Either` anymore.

kore/src/Kore/Step/SMT/Sorts.hs
```diff
@@ -0,0 +1,416 @@
...
+    Left err -> (error . unlines)
+        [ "Unexpected error transforming kore pattern to pure pattern:"
+        , " all patterns should be pure."
+        , "err=" ++ show err
+        , "pattern=" ++ show sentenceAxiomPattern
+        ]
+    Right patt -> parseSMTDataTypePattern patt
+  | otherwise = Nothing
+
+parseSMTDataTypePattern :: CommonStepPattern Object -> Maybe SMTDataType
```
ttuegel: Is this pattern always supposed to come from a no-junk axiom?

kore/src/Kore/Step/SMT/Sorts.hs
```diff
@@ -0,0 +1,416 @@
...
+    checkOnlyQuantifiedVariablesOnce
+        :: Set.Set (Variable Object)
+        -> [CommonStepPattern Object]
+        -> Maybe [Variable Object]
+    checkOnlyQuantifiedVariablesOnce
+        allowedVars
+        []
+      | Set.null allowedVars = Just []
+      | otherwise = Nothing
+    checkOnlyQuantifiedVariablesOnce
```
ttuegel: What is this checking?

kore/src/SimpleSMT.hs
```diff
@@ -944,6 +1006,12 @@ store :: SExpr {- ^ array -}     ->
          SExpr
 store x y z = fun "store" [x,y,z]
 
+-- | Quantifiers: forall
```
ttuegel: Did you change the default theory in `SMT.defaultConfig` to allow quantifiers?

kore/src/Kore/Step/SMT/Sorts.hs
```diff
@@ -0,0 +1,416 @@
...
+                sortName = getId sortActualName
+            addToMap axiom _ = error
+                ("Unexpected no-junk axiom for sort variable: " ++ show axiom)
+
+        sortDeclarations
+            :: [(Attribute.Sort, SentenceSort Object VerifiedKorePattern)]
+        sortDeclarations =
+            Map.elems (recursiveIndexedModuleSortDescriptions indexedModule)
+
+        (declarationsWithoutNoJunk, declarationsWithNoJunk) =
```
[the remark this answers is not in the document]
virgil-serbanuta: The no-junk ones do not have constructors, so I'm declaring them as sorts, not as datatype. When I attempted to declare them as datatype I received an error about them not having constructors.

kore/src/Kore/Step/SMT/Sorts.hs
virgil-serbanuta: Yes.

kore/src/Kore/Step/SMT/Sorts.hs
virgil-serbanuta: I'm doing some sanity checks that I'm parsing what I think I should be parsing.

In particular, this is part of a check that axiom is an `or` of patterns of the type `exists x1 x2 .. xn . sigma(y1, y2, ... ym)`, and this function checks that `m=n` and `{x1, x2, ..., xn} = {y1, y2, ..., yn}`.

kore/src/SimpleSMT.hs
virgil-serbanuta: No, but for some reason it seems to work. I have no clue why. I was reluctant at first to add this, but, since I was only using it in a test, I thought it should be fine.

virgil-serbanuta:
@ttuegel I should have used the official documentation.

However, note that I've been using the declare-datatypes command as recommended on this z3 documentation/tutorial, which should be official enough given that it's created by Microsoft:
https://rise4fun.com/Z3/tutorial/guide
In this, the list immediately after declare-datatypes contains the sort parameters.

kore/src/Kore/Step/SMT/Sorts.hs
traiansf: Can we leave us for a subsequent PR, to prevent delaying this PR any further?

ttuegel:
Instead of using SMT-LIB datatypes, we could just declare the symbols and sorts and send the constructor axioms ourselves. For example, Z3 can handle this without errors:
```.smt
(declare-sort |HB_S| 0)
(declare-sort |HB_T| 0)

(declare-fun |HB_C| () |HB_S|)
(declare-fun |HB_D| (HB_T) |HB_S|)

(declare-fun |HB_E| () |HB_T|)

(assert (forall ((x |HB_T|)) (= x |HB_E|)))

; No junk in HB_S
(assert
  (forall ((x |HB_S|))
    (or
      (= x |HB_C|)
      (exists ((y |HB_T|)) (= x (|HB_D| y)))
    )
  )
)

; No confusion, different constructors in HB_S
(assert
  (forall ((x |HB_T|)) (distinct |HB_C| (|HB_D| x)))
)

; No confusion, same constructors in HB_S
(assert
  (forall ((x |HB_T|) (y |HB_T|))
    (=>
      (= (|HB_D| x) (|HB_D| y))
      (= x y)
    )
  )
)

(declare-fun x () |HB_S| )
(assert (distinct x |HB_C| ))
(check-sat)
(assert (distinct x (|HB_D| |HB_E| ) ))
(check-sat)
```

traiansf:
I've been thinking about this, too, but I was hoping we don't need to go there...

ttuegel:
Well, we already have the axioms in the Kore definition, so the translation should be pretty mechanical. It's not ideal, but it would let us support the widest possible range of versions of Z3, and even other solvers.

virgil-serbanuta:
I tried translating the axioms to asserts previously, and it seemed to work fine, but, it also seemed that there was a better option available. Oh well.

I'll try to use asserts+sorts. I'll probably change the SimpleSMT.declareDatatypes function.

PR #530, opened 2019-03-27
Test matching vs. unification

ttuegel:
I expect this test to pass on the basis that it uses unification rather than matching, but it does not pass.

###### Reviewer checklist

---


ttuegel:
Never mind, I understand what the problem is; it's this check: https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/BaseStep.hs#L600

PR #532, opened 2019-03-28
Modularize tests by moving common functionality into `include.mk`

ehildenb:
###### Reviewer checklist

---
 Moves a lot of common testing functionality into `include.mk`.


vladciobanu:
This definitely looks like the right approach, but some tests seem to be failing. Have you had a chance to investigate?

ehildenb:
@ttuegel yeah, it's the case that `test` targets are not run in CI. Pretty sure there are many of them not passing, but most of them are slow running tests I believe (which is why they are not in `test-k`).

ttuegel:
> Pretty sure there are many of them not passing, but most of them are slow running tests I believe (which is why they are not in `test-k`).

I know that's true if the `Makefile` defines `test` and `test-k`, but some of the files don't seem to define `test-k`. After this is merged, one of us (probably me) should go through and make sure that is intentional. For example, it looks like all of our unification-modulo-builtins tests are disabled, and I'm almost certain that is not intentional.

PR #537, opened 2019-03-28
Don't forget substitutions when simplifying \or

ttuegel:
1. In `disjoinPredicates`, don't attempt to combine predicates with different substitutions.
2. In `simplifyPartial`, remove an unforced error.
3. In `transitionRemoveDestination`, do not mix predicates with the term.
4. Remove some unused code.

Bonus: 35% speed-up on the slowest verification test!

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/Or.lhs
+++ b/kore/src/Kore/Step/Simplification/Or.lhs
@@ -19,7 +19,6 @@ module Kore.Step.Simplification.Or
 
 import           Control.Applicative
                  ( Alternative (..) )
-import qualified Data.Function as Function
 import qualified Data.Functor.Foldable as Recursive
 
 import           Kore.AST.Pure
@@ -27,7 +26,6 @@ import           Kore.Predicate.Predicate
                  ( makeOrPredicate )
 import           Kore.Step.Representation.ExpandedPattern
                  ( ExpandedPattern, Predicated (..) )
-import qualified Kore.Step.Representation.Predicated as Predicated
 import           Kore.Step.Representation.OrOfExpandedPattern
                  ( OrOfExpandedPattern )
 import qualified Kore.Step.Representation.MultiOr as MultiOr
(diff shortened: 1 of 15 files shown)


kore/src/Kore/Step/Simplification/Or.lhs
```diff
@@ -189,47 +187,18 @@ disjoinPredicates
...
-                        makeOrPredicate
-                        Predicated.toPredicate
-                        predicated1
-                        predicated2
-                , substitution = mempty
-                }
-    in Just (result, SimplificationProof)
-
+  | term1 == term2, substitution1 == substitution2 =
+    Just (result, SimplificationProof)
```
virgil-serbanuta: Why not make an `or` of predicates when the substitutions are equal, but the predicates are different?

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -52,15 +51,22 @@ simplifyPartial
...
-            ]
+    rewrapPredicate predicated =
+        predicated
+            { term = ()
+            , predicate =
+                -- It is safe to re-wrap the simplified term as a predicate:
+                -- simplifying a predicate must produce a predicate because
+                -- simplification preserves sorts.
+                Predicate.makeAndPredicate
+                    (Predicate.wrapPredicate simplified)
```
virgil-serbanuta: It should be wrong to wrap `simplified`. If it is really a predicate, simplification should have already put it in the `predicate` field. Why does this make sense?

kore/src/Kore/Step/Simplification/Or.lhs
ttuegel: That's exactly what this does. The change is: it no longer tries to combine predicates when the substitutions are inequal.

kore/src/Kore/Step/Simplification/Predicate.hs
ttuegel: > If it is really a predicate, simplification should have already put it in the predicate field.
That is true only if it is something we know how to simplify, but there are patterns which are not handled, such as `\not(...)`.

ttuegel:
I am closing this pull request and splitting it up, while I figure out what is going on.

kore/src/Kore/Step/Simplification/Predicate.hs
virgil-serbanuta: That should _always_ be true, including not-handled patterns like not. Simplification over, say, a `not (a == b)` should identify that `a == b` is a predicate, and applying `not` on top of it should not change that. Simplification over `(not a) == b` should also identify that `==` produces predicates.

The only case in which this does not apply is for things like `a or (not a)` where `a` is not a predicate, but, if we have those as predicates, we should just fail here since those are not handled properly by most of our code, including the SMT.

@traiansf since I may be wrong about this.

BTW, that's why the documentation for `wrapPredicate` says `This is intended for predicate evaluation and tests and should not be used outside of that.` The message could have been better.

PR #540, opened 2019-03-29
First version of all path reachability design doc

traiansf:
###### Reviewer checklist

---


docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,240 @@
...
+Recommended reading: [proving One-Path Reachability
+claims](2018-11-08-One-Path-Reachability-Proofs.md).
+
+
+Definitions
+-----------
+
+### Weak always
+
+Given a formula `φ`, let `[w]φ` denote the formula “weak always” `φ`,
```
yzhang90: weak always  = all path eventually?

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,240 @@
...
+
+If `φ'` is `⊥`, then the implication holds and we are done.
+
+### Simplifying `φ ∧ ¬∃x.ψ` where x does not appear in `φ`
+
+```
+φ ∧ ¬∃x.ψ                                           is equivalent with
+φ ∧ ¬∃x.ψ /\ (∃x.⌈φ∧ψ⌉ ∨ ¬∃x.⌈φ∧ψ⌉)                 is equivalent with
+(φ ∧ ¬∃x.ψ /\ ∃x.⌈φ∧ψ⌉) ∨ (φ ∧ ¬∃x.ψ ∧ ¬∃x.⌈φ∧ψ⌉)   is equivalent with
+(φ ∧ ∃x.ψ ∧ ¬∃x.ψ) ∨ (φ ∧ ¬∃x.ψ ∧ ¬∃x.⌈φ∧ψ⌉)        is equivalent with
```
yzhang90: assuming \phi is partial functional?

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: essentially, yes.  This is how @hreada named it.

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: Yes, I should probably say somewhere that we are under the same assumption as for the one-path reachability design doc

marick:
If this had examples of the important cases, we could use them to produce tests - which we'll need anyway.

This could potentially be used as a case study of iteratively developing a major new feature via a well-chosen sequence of examples (tests). Done well, that would discover mistakes soon, rather than after a whole bunch of code is written. Done poorly, it wouldn't be worse than discovering mistakes after a whole bunch of code is written.

dlucanu:
Since the algorithm may not terminate, it's good to have a bound of the number of while iterations. If the bound is reached, then it returns `Unproved`.

traiansf:
@marick I totally agree with you!  However, I'm not sure myself or Virgil are the best qualified to provide the "important cases". I think these would best come from the users of this feature.

I can write *some* cases which *I* find important, and I'll follow your advice and do so, but we should probably try to get feedback from our intended users and a ranking for them, and maybe have them propose new tests.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+* goal: `∀x.φ → [w]∃z.ψ`
+* Either claims or axioms:
+    * claims `∀x₁.φ₁ → [w]∃z₁.ψ₁`, `∀x₂.φ₂ → [w]∃z₂.ψ₂`, …, `∀xₙ.φₙ → [w]∃zₙ.ψₙ`
+    * axioms `∀x₁.φ₁ → •∃z₁.ψ₁`, `∀x₂.φ₂ → •∃z₂.ψ₂`, …, `∀xₙ.φₙ → •∃zₙ.ψₙ`
+    * we will not consider the form, but just the patterns and variables involved
+
+__Output:__ `(Goals, goalᵣₑₘ)`
+
+* Let `goalᵣₑₘ := ∀x.(φ ∧ ¬∃x₁.⌈φ∧φ₁⌉ ∧ …  ∧ ¬∃xₙ.⌈φ∧φₙ⌉) → [w]∃z.ψ`
+* Let `Goals := { ∀x∪z₁.(∃x₁.ψ₁ ∧ ⌈φ∧φ₁⌉) → [w]∃z.ψ, … , ∀x∪zₙ.(∃xₙ.ψₙ ∧ ⌈φ∧φₙ⌉) → [w]∃z.ψ }`
```
yzhang90: I'd like to have a function to specifically compute {∀x∪z₁.(∃x₁.ψ₁ ∧ ⌈φ∧φ₁⌉), ..., ∀x∪zₙ.(∃xₙ.ψₙ ∧ ⌈φ∧φₙ⌉)} for a given φ

traiansf: I guess that could be arranged.  It would be good to let @marick know that you need that, to include it in our weekly planning.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+Assume `∀xᵢ.φᵢ →  •∃zᵢ.ψᵢ, 1 ≤ i ≤ n`  are all the one-step axioms
+in the definition.
+
+Using the same reasoning as in when applying all claims in parallel,
+`∀x.φ → α` is equivalent with
+```
+(∀x.φ ∧ ∃x₁.φ₁ → α) ∧ … ∧ (∀x.φ ∧ ∃xₙ.φₙ → α) ∧ (∀x.φ ∧ ¬∃x₁.φ₁ ∧ … ∧ ¬∃xₙ.φₙ → α)
+```
+
+Now, for the first thing to check, take `α := ⊤`.
```
yzhang90: α := ⊤ should be α :=• ⊤?

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+```
+we can deduce that `∀x.φ ∧ ∃xᵢ.φᵢ → [w]∃z.ψ`.
+
+This would allow us to replace goal `∀x.φ ∧ ∃xᵢ.φᵢ → [w]∃z.ψ`
+with goal '∀x∪zᵢ.ψᵢ ∧ ∃xᵢ.⌈φ ∧ φᵢ⌉ → [w]ψ'.
+
+### Applying axioms
+
+We're back now to `∀x.φ →  (∃z.○[w]ψ) ∧ •⊤`, which is equivalent to 
+`(∀x.φ →  ○[w]ψ) ∧ (∀x.φ →  •⊤)`
```
yzhang90: why ∃z is omitted here?

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+∀x∪zᵢ.ψᵢ ∧ ∃xᵢ.⌈φ ∧ φᵢ⌉ → [w]ψ
+```
+we can deduce that `∀x.φ ∧ ∃xᵢ.φᵢ → [w]∃z.ψ`.
+
+This would allow us to replace goal `∀x.φ ∧ ∃xᵢ.φᵢ → [w]∃z.ψ`
+with goal '∀x∪zᵢ.ψᵢ ∧ ∃xᵢ.⌈φ ∧ φᵢ⌉ → [w]ψ'.
+
+### Applying axioms
+
+We're back now to `∀x.φ →  (∃z.○[w]ψ) ∧ •⊤`, which is equivalent to
```
yzhang90: why ∃z comes out of  ○[w]ψ? can ∃ propagate out of \mu? If not, what we need to prove is `∀x.φ →  (○[w]∃z.ψ) ∧ •⊤`

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+Definitions
+-----------
+
+### Weak always
+
+Given a formula `φ`, let `[w]φ` denote the formula “weak always” `φ`,
+defined by:
+
+```
+[w]φ := νX.φ ∨ (○X ∧ •⊤)
```
ttuegel: What is the role/purpose of `∧ •⊤`?

hreada: Intuitively, it is to make sure that when `φ` doesn't hold at current state, *then the current state doesn't get stuck*, and [w]φ holds on all next states.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,253 @@
...
+Given a set of all-path reachability claims, of the form `∀x.φ(x) → [w]∃z.ψ(x,z)`,
+we are trying to prove all of them together.
+
+
+Algorithms
+----------
+(detailed in the next section)
+
+
+### Algorithm `proveAllPath`
```
hreada: It looks like the algorithm derivates from the classic all-path reachability logic prover, where circularities (claims) are regarded as semantic axioms, and (Circurlarity) rule is forbidden. Is it intended?

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: yes. thanks!

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: by mistake.  thanks for spotting it.

traiansf: I did not have `∃z` initially, but then I realized we need them, as they introduce variables which become universally quantified in the derived goals.  I tried to refactor and add them everywhere, but it seems I missed some spots.

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: yes, ∃z was on the wrong side there.

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: I added a explanatory paragraph concerning this.  Thanks @ttuegel and @hreada

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: > Since the algorithm may not terminate, it's good to have a bound of the number of while iterations. If the bound is reached, then it returns `Unproved`.

@dlucanu : I added a note about this close to the description of the algorithm.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+
+__Input:__ claims `∀x₁.φ₁ → [w]∃z₁.ψ₁`, `∀x₂.φ₂ → [w]∃z₂.ψ₂`, …, `∀xₙ.φₙ → [w]∃zₙ.ψₙ`
+
+__Output:__ Proved or Unproved
+
+* For each claim `∀x.φ → [w]∃z.ψ`
+    * Let `Goals := { ∀x.φ → [w]∃z.ψ }`
+    * While `Goals` is not empty:
+        * Extract and remove `goal` of the form `∀x.φ → [w]∃z.ψ` from `Goals`
+        * Let `goalᵣₑₘ := ∀x. (φ ∧ ¬∃z.⌈φ ∧ ψ⌉) → [w]∃z.ψ`
```
hreada: As a remark, note that `φ ∧ ¬∃z.⌈φ ∧ ψ⌉` equals `φ ∧ ¬∃z.ψ`, given that `φ` is a function-like pattern.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Algorithm `derivePar`
+
+__Input:__: goal and set of claims/axioms
+
+* goal: `∀x.φ → [w]∃z.ψ`
+* Either claims or axioms:
+    * claims `∀x₁.φ₁ → [w]∃z₁.ψ₁`, `∀x₂.φ₂ → [w]∃z₂.ψ₂`, …, `∀xₙ.φₙ → [w]∃zₙ.ψₙ`
+    * axioms `∀x₁.φ₁ → •∃z₁.ψ₁`, `∀x₂.φ₂ → •∃z₂.ψ₂`, …, `∀xₙ.φₙ → •∃zₙ.ψₙ`
+    * we will not consider the form, but just the patterns and variables involved
```
hreada: what is "the form"?

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+
+__Output:__ `(Goals, goalᵣₑₘ)`
+
+* Let `goalᵣₑₘ := ∀x.(φ ∧ ¬∃x₁.⌈φ∧φ₁⌉ ∧ …  ∧ ¬∃xₙ.⌈φ∧φₙ⌉) → [w]∃z.ψ`
```
hreada: Similarly, we can prove that `φ ∧ ¬∃x₁.⌈φ∧φ₁⌉ ∧ …  ∧ ¬∃xₙ.⌈φ∧φₙ⌉` equals `φ ∧ ¬∃x₁.φ₁ ∧ …  ∧ ¬∃xₙ.φₙ`, given that `φ` is a function-like pattern.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+__Note:__ Since the derivation process can continue indefinitely, one could add
+a bound on the total number of (levels of) expansions attempted before
+returning `Unprovable`.
+
+### Algorithm `derivePar`
+
+__Input:__: goal and set of claims/axioms
+
+* goal: `∀x.φ → [w]∃z.ψ`
+* Either claims or axioms:
```
hreada: Are claims and axioms going to have the same algorithm? I thought they will have different algorithms, because axioms are about one-path reachability, while claims are all-path reachability.

docs/2019-03-28-All-Path-Reachability-Proofs.md
hreada: For clarity, I suggest we write `∀x.(∃x₁.ψ₁ ∧ ⌈φ∧∃z₁.φ₁⌉) → [w]∃z.ψ`, and then simplify it to its equivalent form `∀x∪x₁∪z₁.(ψ₁ ∧ ⌈φ∧φ₁⌉) → [w]∃z.ψ`.

docs/2019-03-28-All-Path-Reachability-Proofs.md
hreada: I think it suffices to use `derivePar` for axioms and `deriveSeq` for (all-path reachability) claims.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+
+__Input:__: goal and set of claims
+
+* goal: `∀x.φ → [w]∃z.ψ`
+* claims `∀x₁.φ₁ → [w]∃z₁.ψ₁`, `∀x₂.φ₂ → [w]∃z₂.ψ₂`, …, `∀xₙ.φₙ → [w]∃zₙ.ψₙ`
+
+__Output:__ `(Goals, goalᵣₑₘ)`
+
+* Let `goalᵣₑₘ := ∀x.(φ ∧ ¬∃x₁.⌈φ∧φ₁⌉ ∧ …  ∧ ¬∃xₙ.⌈φ∧φₙ⌉) → [w]∃z.ψ`
+* Let `Goals := { ∀x∪z₁.(∃x₁.ψ₁ ∧ ⌈φ₁ʳᵉᵐ ∧φ₁⌉) → [w]∃z.ψ, … , ∀x∪zₙ.(∃xₙ.ψₙ ∧ ⌈φₙʳᵉᵐ φ∧φₙ⌉) → [w]∃z.ψ }`
```
hreada: Typo at `⌈φₙʳᵉᵐ φ∧φₙ⌉`.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,273 @@
...
+
+```
+∀x.φ → [w]∃z.ψ                                              is equivalent to
+∀x.φ ∧ (∃xᵢ.φᵢ ∨ ¬∃xᵢ.φᵢ) → [w]∃z.ψ                         which is equivalent to
+∀x.(φ ∧ ∃xᵢ.φᵢ) ∨ (φ ∧ ¬∃xᵢ.φᵢ) → [w]∃z.ψ                   which is equivalent to
+∀x.((φ ∧ ∃xᵢ.φᵢ) → [w]∃z.ψ) ∧ ∀x. ((φ ∧ ¬∃xᵢ.φᵢ) → [w]∃z.ψ) (1)
+```
+
+Note that the remainder `∀x.φ ∧ ¬∃xᵢ.φᵢ → [w]∃z.ψ` can be rewritten as
+`∀x.φ ∧ ¬∃xᵢ.⌈φ∧φᵢ⌉ → [w]ψ`, as detailed above.
```
hreada: missing `∃z`.

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: Form in the sense that the algorithm does not need distinguish between `∀x₁.φ₁ → [w]∃z₁.ψ₁` and  `∀x₁.φ₁ → •∃z₁.ψ₁`; It just works with tupples of the form `(x₁,φ₁,z₁,ψ₁)`.
I wanted to say that the algorithm works for both claims and axioms in a similar fashion. It's just the justifications for why it works that are different.

docs/2019-03-28-All-Path-Reachability-Proofs.md
traiansf: I agree with you that it suffices;  however, using `derivePar` for claims is also sound (although it might cover the same space multiple times), and there were some discussions in the team that it might be easier to implement, so I decided to leave it here, and let a design decision argue which is the best for us to use at the moment.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+one consequence of the above is that `[w]φ = φ ∨ (○[w]φ ∧ •⊤)`.
+
+Given this definition of weak always, an all-path reachability claim
+```
+∀x.φ(x) → [w]∃z.ψ(x,z)
+```
+basically states that if `φ(x)` holds for a configuration `γ`, for some `x`,
+then `P(γ)`, where `P(G)` is recursively defined on configurations as:
```
ttuegel: Maybe it would be clearer to say,
```suggestion
then `P(γ)` holds, where `P(G)` is recursively defined on configurations as:
```

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,346 @@
...
+used in this document to be _extended function-like patterns_, that is patterns
+which can be written as `t ∧ p` where the interpretation of `t` contains at most
+one element and `p` is a predicate.
+
+_Extended constructor patterns_ will be those extended function-like patterns
+for which `t` is a functional term, composed out of constructor-like symbols
+and variables.
+
+
+__Note:__
```
ttuegel: I would be inclined to call the two immediately-following statements "Lemma 1a" and "Lemma 1b" and refer back to them as needed below.

docs/2019-03-28-All-Path-Reachability-Proofs.md
```diff
@@ -0,0 +1,346 @@
...
+
+Moreover, suppose all free variables in the above formulae are from `x`,
+we will assume that the unification condition `⌈φ ∧ ψ⌉` can always be
+computed to be of the form `z = t ∧  p`, where
+
+* `t`s are functional patterns with no free variables from `z`
+    * i.e., [t / z] is a substitution.
+* `p` is a predicate over `x ∪ z`
+
+Under this assumption, `∃z.⌈φ ∧ ψ⌉` can be rewritten without the existential
```
ttuegel: I have the same comment as above, but calling this "Lemma 2".

PR #544, opened 2019-04-02
Feature/repl dump output to file

vladciobanu:
###### Reviewer checklist

---


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -75,39 +77,48 @@ import qualified Kore.Step.Strategy as Strategy
 import           Kore.Unparser
                  ( unparseToString )
 
+type ReplM level a = RWST () String (ReplState level) Simplifier a
```
ttuegel: The writer part of `RWST` leaks space just like `WriterT`. We should never use it.

vladciobanu: I knew something to that effect, but I thought it's fine for brief uses? We're concatenating a couple of strings before resolving the `RWST` to a `StateT` on every command. Its use is confined to this module, and its lifetime to a single command. Would that still be a problem?

ttuegel: > Its use is confined to this module, and its lifetime to a single command.

Could you explain how that works more? It looks like several commands have acquired `MonadWriter` constraints.

ttuegel:
Approved; we addressed the comments during stand-up.

PR #548, opened 2019-04-02
Jenkinsfile: post-build action which sends a slack notification

ehildenb:
###### Reviewer checklist

---

This should enable sending emails on failed jobs to `master` or `[kevm-integration]` tagged commits.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,6 +30,11 @@ pipeline {
           '''
         }
       }
+      post {
+        always {
+          junit 'kore/test-results.xml'
+        }
+      }
     }
     stage('K Integration Tests') {
       steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -56,6 +56,13 @@ pipeline {
               '''
             }
           }
+          post {
+            failure {
```
dwightguth: I think you probably want regression, not failure here.

ehildenb: done, though I think we actually want `failure` because I want it to be annoying if it keeps failing. But we can change it later.

dwightguth: Alright. Note that failure will not run if the build is unstable; not sure if you ever get unstable builds, but something to consider.

ehildenb: Changed back to `failure`, we can change to `regression` if it becomes annoying.

PR #549, opened 2019-04-02
Update K Nightly version used for integration testing

ehildenb:
PR generated by './scripts/update-k-nightly-pr.sh'


ttuegel:
This seems to be failing still, so evidently I just can't replicate the failure locally.

PR #553, opened 2019-04-03
Use the same K for all integration tests

ttuegel:
###### Reviewer checklist

---

### DIFF
--- a/scripts/kevm-integration.sh
+++ b/scripts/kevm-integration.sh
@@ -17,14 +17,10 @@ git clone --recurse-submodules 'https://github.com/kframework/evm-semantics' $EV
 
 cd $EVM_SEMANTICS
 
-(   cd .build/k
-    (   cd haskell-backend/src/main/native/haskell-backend
-        git fetch $TOP
-        git checkout FETCH_HEAD
-    )
-    git add haskell-backend/src/main/native/haskell-backend
-    ci_git commit --message '!!! haskell-backend/src/main/native/haskell-backend: integration testing haskell backend'
-)
+# Use the K Nightly build from the Kore integration tests.
+rm -rf .build/k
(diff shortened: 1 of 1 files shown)


scripts/kevm-integration.sh
```diff
@@ -17,14 +17,10 @@ git clone --recurse-submodules 'https://github.com/kframework/evm-semantics' $EV
...
-        git fetch $TOP
-        git checkout FETCH_HEAD
-    )
-    git add haskell-backend/src/main/native/haskell-backend
-    ci_git commit --message '!!! haskell-backend/src/main/native/haskell-backend: integration testing haskell backend'
-)
+# Use the K Nightly build from the Kore integration tests.
+rm -rf .build/k
+mkdir -p .build/k/k-distribution/target/release
+ln -s $TOP/.build/k .build/k/k-distribution/target/release
```
ehildenb: We can see if it works, but we may want to instead `cp -r` the directory into place, because the scripts sometimes depend on things like `$0`, which may not resolve correctly.

scripts/kevm-integration.sh
```diff
@@ -33,7 +29,6 @@ make clean
 git submodule update --init --recursive
```
ehildenb: Probably don't want to either commit the submodule, run `make clean` or `git submodule update --init --recursive`. It will wipe out the `.build/k` directory and re-clone KEVM's version.

scripts/kevm-integration.sh
ttuegel: This is how I always set up my `evm-semantics` tree locally (`.build/k/k-distribution/target/release/k` is a symlink) but I can change it if you want.

ehildenb: If it works locally for you, sounds good to me :+1:

PR #558, opened 2019-04-04
Non-deterministic one-path proofs

ttuegel:
@marick @ehildenb 
We were right in thinking that one of these proofs would fail.

###### Reviewer checklist

---


ttuegel:
I removed the `bug` tag because this limitation is intended.

ttuegel:
@traiansf Could you take a look at the `NOTE:` I added, and make sure I understand this correctly?

I'm also tagging @virgil-serbanuta because I renamed two constructors of `StrategyPattern`. The names were a source of confusion for us (me, @marick, and @ehildenb).

kore/src/Kore/OnePath/Verification.hs
```diff
@@ -77,6 +77,38 @@ import qualified Kore.TopBottom as TopBottom
 import           Numeric.Natural
                  ( Natural )
 
+{- NOTE: Non-deterministic semantics
+
+The current implementation of one-path verification assumes that the definition
+semantics is deterministic, that is: at most one rule may apply to any
```
virgil-serbanuta: Well, it's a bit more complicated, non-determinism is allowed as long as you have confluence and your final state not in the middle of a non-confluent thing. As an example, you can have non-deterministic rewriting for heating and cooling, but usually you can reach the proof goal only after the non-deterministic part ended, so all is fine.

It's up to you whether to rewrite this in a more precise way.

kore/src/Kore/OnePath/Step.hs
```diff
@@ -112,28 +112,31 @@ if x then phi else psi
...
 data StrategyPattern patt
     = RewritePattern !patt
     -- ^ Pattern on which a normal 'Rewrite' can be applied. Also used
     -- for the start patterns.
-    | Stuck !patt
+    | Remainder !patt
     -- ^ Pattern which can't be rewritten anymore.
-    | Bottom
-    -- ^ special representation for a bottom rewriting/simplification result.
+    | Goal
```
virgil-serbanuta: This is not always a goal, e.g. while rewriting you can get a state split which cannot be reduced immediately to bottom, but which becomes bottom in the simplification step. You might be replacing a confusing thing with something equally confusing.

Not sure what's the best way to solve this, you might try splitting the state.

kore/src/Kore/OnePath/Verification.hs
ttuegel: I think this is an important distinction that I did not fully understand, so I will try to be more precise.

kore/src/Kore/OnePath/Step.hs
ttuegel: I can see how this might also be confusing. Maybe it would be better to call this `GoalReached` (which is what I meant) or even just `Done`?

ttuegel: I decided to revert this particular change. We can revisit it later if necessary.

PR #560, opened 2019-04-05
Allow branching in And.makeEvaluate

ttuegel:
This is not critical, but it removes an unforced `error` call and some dead code.

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/And.hs
+++ b/kore/src/Kore/Step/Simplification/And.hs
@@ -13,6 +13,8 @@ module Kore.Step.Simplification.And
     , simplifyEvaluated
     ) where
 
+import           Control.Applicative
+                 ( Alternative (empty) )
 import qualified Control.Monad.Trans as Monad.Trans
 import           Data.List
                  ( foldl1', nub )
@@ -31,7 +33,7 @@ import           Kore.Step.Pattern
 import           Kore.Step.Representation.ExpandedPattern
                  ( ExpandedPattern )
 import qualified Kore.Step.Representation.ExpandedPattern as ExpandedPattern
-                 ( bottom, isBottom, isTop )
+                 ( isBottom, isTop )
 import qualified Kore.Step.Representation.MultiOr as MultiOr
                  ( make )
 import           Kore.Step.Representation.OrOfExpandedPattern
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -251,56 +252,39 @@ makeEvaluateNonBool
...
             firstTerm
             secondTerm
-    (   Predicated
-            { predicate = mergedPredicate
-            , substitution = mergedSubstitution
-            }
-        , _proof
-        ) <- mergePredicatesAndSubstitutions
+    let firstCondition = Predicated.withoutTerm first
+        secondCondition = Predicated.withoutTerm second
```
virgil-serbanuta: BTW, in the past we used `condition`, then we moved towards `predicate` as a synonym for `condition`, and now it seems to be moving again towards using `condition` for something that is either a `predicate` or something close.

So: Why are we restarting to use `condition`? Is a `condition` just a predicate, or is it a special kind of predicate that has a separate substitution part (i.e. a `PredicateSubstitution`), or something else?

ttuegel: At @marick's suggestion, I've been trying to think about the "ubiquitous language" of our domain. I had been getting questions about `PredicateSubstitution` because it wasn't clear to other people that it is usually a unification solution. Brian suggested calling `PredicateSubstitution` a `SideCondition` instead, but @traiansf pointed out that is not quite right. I suggested `UnificationCondition` instead because it seems to be that the condition—one way or another—is almost always arising from unification. I didn't rename the type, but I started to gradually introduce this terminology in the hope it would clarify for other people looking at our code; if this isn't the right vocabulary for the domain I will definitely change it.

virgil-serbanuta: I don't care much either way, but I like consistency.

Both "predicate" and "condition" are used in the semantics-of-k document, though the latter is usually used with certain qualifications (side condition/welformedness condition). I guess that one reason why we switched to "predicate" because it is a stand-alone-ish term, while condition seems to imply that it's a condition for something, which makes it slightly awkward to read code that handles stand-alone conditions.

Anyway, that does not matter much, if we're switching to "condition", then maybe we should consider doing a search+replace renaming in our entire code.

virgil-serbanuta: BTW, if this is used only for the PredicateSubstitution stuff, then the tiny objection in my previous message does not apply.

ttuegel: It is only intended to be used for `PredicateSubstitution`, yes.

PR #561, opened 2019-04-05
No function remainders

ttuegel:
Return `NotApplicable` for a function evaluation if the definition does not cover the entire input, i.e. if there are any remainders. This is safe to do because the other backends do not allow functions to branch at all. (This still allows branching, but only if the branches can be proven to cover the entire input.) This is useful to do because it postpones evaluating functions until more information is available; concrete execution of KEVM is up to 40% faster! This also stops the problem where we would loop forever trying to evaluate stuck functions in a remainder: this allows one new KWASM proof to succeed, and all remaining failing proofs get stuck instead of looping!

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -359,6 +359,9 @@ evaluateWithDefinitionAxioms
     patt
   = do
     let
+        -- TODO (thomas.tuegel): Take an 'ExpandedPattern' as input so that its
+        -- conditions can be used to remove remainder branches and evaluate more
+        -- functions.
         expanded :: ExpandedPattern level variable
         expanded = ExpandedPattern.fromPurePattern patt
 
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -135,7 +135,7 @@ totalDefinitionEvaluation rules =
         axiomSimplifiers
         term
       = do
-        (result, proof) <- evaluate
+        (result, proof) <- evaluate term
         if AttemptedAxiom.hasRemainders result
             then return (AttemptedAxiom.NotApplicable, proof)
             else return (result, proof)
@@ -147,7 +147,6 @@ totalDefinitionEvaluation rules =
                 predicateSimplifier
                 termSimplifier
                 axiomSimplifiers
-                term
 
 {-| Creates an evaluator that choses the result of the first evaluator that
 returns Applicable.
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -359,9 +359,8 @@ evaluateWithDefinitionAxioms
   =
     AttemptedAxiom.exceptNotApplicable $ do
     let
-        -- TODO (thomas.tuegel): Take an 'ExpandedPattern' as input so that its
-        -- conditions can be used to remove remainder branches and evaluate more
-        -- functions.
+        -- TODO (thomas.tuegel): Figure out how to get the initial conditions
+        -- and apply them here, to remove remainder branches sooner.
         expanded :: ExpandedPattern level variable
         expanded = ExpandedPattern.fromPurePattern patt
 
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -91,7 +91,7 @@ definitionEvaluation rules =
     BuiltinAndAxiomSimplifier
         (evaluateWithDefinitionAxioms rules)
 
-{- | Creates an evaluator for a function all the rules that define it.
+{- | Creates an evaluator for a function from all the rules that define it.
 
 The function is not applied (@totalDefinitionEvaluation@ returns
 'AttemptedAxiom.NotApplicable') if the supplied rules do not match the entire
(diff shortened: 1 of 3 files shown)


traiansf:
@ttuegel Do you have a plan for sending the condition down the pattern during the simplification phase?
It is logically correct to do so over symbols and most logical connectives (though negation and quantifiers are trickier).

I remember me and @virgil-serbanuta thought about it and decided it's easier to do a case split and propagate the disjunction up.  @virgil-serbanuta do you remember what we were worrying about w.r.t. propagating the predicate downwards?

In some sense this is a natural follow-up on the decision to not evaluate functions if they have symbolic input so I am not opposing this change.  

W.r.t. the code itself  I'll let @virgil-serbanuta handle that.

virgil-serbanuta:
@traiansf Not works just fine, `P and (not phi) = P and (not (P and phi))`. Moving a predicate above a `not` is slightly trickier, but doable.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -92,6 +91,64 @@ definitionEvaluation rules =
     BuiltinAndAxiomSimplifier
         (evaluateWithDefinitionAxioms rules)
 
+{- | Creates an evaluator for a function all the rules that define it.
```
virgil-serbanuta: I think you're missing one or more words here (from?).

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -92,6 +91,64 @@ definitionEvaluation rules =
...
+            else return (result, proof)
+      where
+        evaluate =
+            evaluateWithDefinitionAxioms
+                rules
+                tools
+                predicateSimplifier
+                termSimplifier
+                axiomSimplifiers
+                term
```
virgil-serbanuta: Why not leave 'term' as an argument to 'evaluate'? You did something similar in most of the last PRs I reviewed, and I think that the `evaluate` (or equivalent) call looked better that way.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -299,17 +356,20 @@ evaluateWithDefinitionAxioms
     simplifier
     axiomIdToSimplifier
     patt
-  = do
+  =
+    AttemptedAxiom.exceptNotApplicable $ do
     let
+        -- TODO (thomas.tuegel): Take an 'ExpandedPattern' as input so that its
```
virgil-serbanuta: Why an `ExpandedPattern` and not a `Predicate` or a `PredicateSubstitution`?

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -92,6 +91,64 @@ definitionEvaluation rules =
...
+            )
+    totalDefinitionEvaluationWorker
+        tools
+        predicateSimplifier
+        termSimplifier
+        axiomSimplifiers
+        term
+      = do
+        (result, proof) <- evaluate
+        if AttemptedAxiom.hasRemainders result
```
virgil-serbanuta: Why is it reasonable to not simplify the remainders' predicates before checking whether there are any remainders?

ttuegel: The remainders are normalized during rule application by [`Kore.Step.Step.applyRemainder`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Step.hs#L374), so the predicates should already be simplified.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
ttuegel: That's a good point; I reworded the comment to leave that as an open question.

ttuegel:
> I remember me and @virgil-serbanuta thought about it and decided it's easier to do a case split and propagate the disjunction up. @virgil-serbanuta do you remember what we were worrying about w.r.t. propagating the predicate downwards?

Yeah, I agree it's tricky both ways. We need to avoid re-evaluating function remainders somehow; we could also solve that with a pattern annotation that says, "this application pattern is a remainder, do not re-evaluate it." However, the big speed-up we get for concrete execution indicates that we were probably splitting into a disjunction too early; it's not clear to me if a pattern annotation solves that case. I think this solution has the advantage that it tends to seek out the path of least-branching for function evaluation, but maybe it's too complicated to get the initial conditions where they are needed?

ttuegel:
Actually, since we simplify patterns from the bottom up, isn't it the case that when evaluating functions we only need to consider initial conditions from simplifying the application children? The function evaluator [already has that information](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Function/Evaluator.hs#L95), it only needs to be passed on though to the axiom evaluator.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
virgil-serbanuta: I so wish that we could have a type-safe way (or any way, really) of telling that this is, indeed the case. Preferably a way that's stable through refactoring or other code changes.

Anyway, that's not related to this PR.

virgil-serbanuta:
So, a few things:
1. We only propagated conditions up because it was easier, and we could always add the down-propagation later.
2. Propagating top conditions down makes sense. Consider the following (c) code:
```
if (p != NULL) {
  value = p->x;
  <other code that relies on p not being NULL>
}
```
We might solve everything inside that `if` statement through rewriting, but each time we apply a function that depends on `p` not being null, propagating the top predicate should help (if pointers are not convincing, consider a test that something is not zero, followed by division / modulo operations).
3. As long as we pay attention to quantifiers and variable clash, we can directly use top conditions anywhere in a formula, the only real problem is passing them through our code. I.e. if `C[x]` is a context (I hope that's the right terminology), and P is a predicate whose variables are not quantified in `C[x]`, then `P and C[phi] == P and C[phi and P]` (not sure what's the right syntax for applying a context).
4. Using child predicates helps, but, as mentioned above, if we're interested, we can do more.

PR #562, opened 2019-04-07
Simplifying one-path tests for use in upcoming all-paths work

marick:
###### Reviewer checklist

---


kore/test/Test/Kore/Step.hs
```diff
@@ -516,7 +372,108 @@ runSteps metadataTools stepLimit configuration axioms =
...
+
+-- Builders
+
+-- | Create a function pattern from a function name and list of argnames.
+applyConstructorToVariables :: Text -> [Text] -> TestPattern
+applyConstructorToVariables name arguments =
+    mkApp patternMetaSort symbol $ fmap var arguments
+  where
+    symbol = SymbolOrAlias -- can this be more abstact?
+        { symbolOrAliasConstructor = Id name AstLocationTest
```
ttuegel: Why not `testId`?
```suggestion
        { symbolOrAliasConstructor = testId name
```

kore/test/Test/Kore/Step.hs
```diff
@@ -516,7 +372,108 @@ runSteps metadataTools stepLimit configuration axioms =
...
+        }
+
+-- | Do the busywork of converting a name into a variable pattern.
+var :: Text -> TestPattern
+var name =
+    mkVar $ (Variable (testId name) mempty) patternMetaSort
+    -- can the above be more abstract?
+
+
+rewritesTo
```
ttuegel: We can concretize `variable` as `Variable` in the signature of `rewritesTo` and then replace `StepPattern ...` with `TestPattern`.

kore/test/Test/Kore/Step.hs
```diff
@@ -516,7 +372,108 @@ runSteps metadataTools stepLimit configuration axioms =
...
+  where
+    symbol = SymbolOrAlias -- can this be more abstact?
+        { symbolOrAliasConstructor = Id name AstLocationTest
+        , symbolOrAliasParams = []
+        }
+
+-- | Do the busywork of converting a name into a variable pattern.
+var :: Text -> TestPattern
+var name =
+    mkVar $ (Variable (testId name) mempty) patternMetaSort
```
ttuegel: Because the sort doesn't really matter for these tests, and because `patternMetaSort` will go away, I would suggest creating an arbitrary sort for this module's tests.

kore/test/Test/Kore/Step.hs
marick: Done

kore/test/Test/Kore/Step.hs
marick: Done

kore/test/Test/Kore/Step.hs
marick: Done.

PR #564, opened 2019-04-09
Feature/monad unify

vladciobanu:
###### Reviewer checklist

---


kore/src/Control/Monad/Unify.hs
```diff
@@ -0,0 +1,106 @@
+{-|
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+-}
+
+module Control.Monad.Unify where
+
+import           Colog
```
ttuegel: If we need this here, let's export it from `Kore.Logger` instead; I would like that module to be the "single source of truth" for logging-related items.

kore/src/Control/Monad/Unify.hs
```diff
@@ -0,0 +1,106 @@
...
+
+    mapVariable f (Unifier e) =
+        Unifier $ withExceptT (mapUnificationOrSubstitutionErrorVariables f) e
+
+instance MonadReader Environment (Unifier variable) where
+    ask = liftSimplifier ask
+    local f (Unifier ma) = Unifier $ local f ma
+
+instance Log.WithLog Log.LogMessage (Unifier variable) where
+    askLogAction = do
```
ttuegel: This instance can be delegated to the instance of `ExceptT`:
```.hs
askLogAction = Unifier Log.askLogAction
withLog f = Unifier . Log.withLog f . getUnifier
```

kore/src/Control/Monad/Unify.hs
```diff
@@ -0,0 +1,106 @@
+{-|
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+-}
+
+module Control.Monad.Unify where
```
ttuegel: I think this should be in the `Kore.Unification` tree because it depends on other things there, perhaps `Kore.Unification.Unifier`?

PR #565, opened 2019-04-09
repl: add index attribute + print index in showRule + bug fix in try

ana-pantilie:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Attribute/RuleIndex.hs
+++ b/kore/src/Kore/Attribute/RuleIndex.hs
@@ -11,12 +11,11 @@ module Kore.Attribute.RuleIndex
     , ruleIndexId, ruleIndexSymbol, ruleIndexAttribute
     ) where
 
-import           Control.DeepSeq
-                 ( NFData )
-import qualified Control.Monad as Monad
-import           Data.Default
-import           GHC.Generics
-                 ( Generic )
+import Control.DeepSeq
+       ( NFData )
+import Data.Default
+import GHC.Generics
+       ( Generic )
 
 import           Kore.AST.Kore
 import           Kore.Attribute.Parser
@@ -56,10 +55,9 @@ ruleIndexAttribute =
 
 instance ParseAttributes RuleIndex where
     parseAttribute =
-        withApplication $ \params args RuleIndex { getRuleIndex } -> do
+        withApplication $ \params args _ -> do
             Parser.getZeroParams params
             Parser.getZeroArguments args
             return RuleIndex { getRuleIndex = Nothing }
       where
         withApplication = Parser.withApplication ruleIndexId
-        failDuplicate = Parser.failDuplicate ruleIndexId
--- a/kore/src/Kore/Attribute/RuleIndex.hs
+++ b/kore/src/Kore/Attribute/RuleIndex.hs
@@ -14,11 +14,6 @@ import Data.Default
 import GHC.Generics
        ( Generic )
 
-import           Kore.AST.Kore
-import           Kore.Attribute.Parser
-                 ( ParseAttributes (..) )
-import qualified Kore.Attribute.Parser as Parser
-
 {- | This attribute is used in the REPL for tagging
     and uniquely identifiying axioms and claims.
  -}
--- a/kore/src/Kore/Attribute/RuleIndex.hs
+++ b/kore/src/Kore/Attribute/RuleIndex.hs
@@ -1,14 +1,11 @@
 {-|
-Module      : Kore.Attribute.RuleIndex
 Description : Rule index attribute
 Copyright   : (c) Runtime Verification, 2018
 License     : NCSA
-Maintainer  : ???@runtimeverification.com
 
 -}
 module Kore.Attribute.RuleIndex
     ( RuleIndex (..)
-    , ruleIndexId, ruleIndexSymbol, ruleIndexAttribute
     ) where
 
 import Control.DeepSeq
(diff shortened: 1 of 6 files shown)


kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
+{-|
+Module      : Kore.Attribute.RuleIndex
```
vladciobanu: The new template only has the Description, Copyright,and License fields.

kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
...
+import Data.Default
+import GHC.Generics
+       ( Generic )
+
+import           Kore.AST.Kore
+import           Kore.Attribute.Parser
+                 ( ParseAttributes (..) )
+import qualified Kore.Attribute.Parser as Parser
+
+{- | @RuleIndex@ represents the @RuleIndex@ attribute.
```
vladciobanu: I would probably just say "This attribute is used in the REPL for tagging and uniquely identifying axioms and claims".

kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
...
+newtype RuleIndex = RuleIndex { getRuleIndex :: Maybe Int }
+    deriving (Eq, Ord, Show, Generic)
+
+instance NFData RuleIndex
+
+instance Default RuleIndex where
+    def = RuleIndex Nothing
+
+-- | Kore identifier representing the @ruleIndex@ attribute symbol.
+ruleIndexId :: Id Object
```
vladciobanu: Not sure this is required since we don't really parse or generate this attribute.

kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
...
+ruleIndexId = "ruleIndex"
+
+-- | Kore symbol representing the @ruleIndex@ attribute.
+ruleIndexSymbol :: SymbolOrAlias Object
```
vladciobanu: Not sure this is required since we don't really parse or generate this attribute.

kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
...
+ruleIndexSymbol =
+    SymbolOrAlias
+        { symbolOrAliasConstructor = ruleIndexId
+        , symbolOrAliasParams = []
+        }
+
+-- | Kore pattern representing the @ruleIndex@ attribute.
+ruleIndexAttribute :: CommonKorePattern
```
vladciobanu: Not sure this is required since we don't really parse or generate this attribute.

kore/src/Kore/Attribute/RuleIndex.hs
```diff
@@ -0,0 +1,63 @@
...
+ruleIndexAttribute =
+    (asCommonKorePattern . ApplicationPattern)
+        Application
+            { applicationSymbolOrAlias = ruleIndexSymbol
+            , applicationChildren = []
+            }
+
+instance ParseAttributes RuleIndex where
```
vladciobanu: Would this still work without this instance and without adding it to the big `ParseAttributes` list? Since we're not really parsing it, we only care about it having its default `Nothing` value. Try removing this and the extra parser line in `Axiom.hs`

kore/src/Kore/Repl.hs
```diff
@@ -101,6 +106,46 @@ runRepl tools simplifier predicateSimplifier axiomToIdSimplifier axioms' claims'
...
             }
 
+    addIndexesToAxioms
+        :: [Axiom level]
+        -> [Axiom level]
+    addIndexesToAxioms axs =
+        fmap (Axiom . addIndex) (zip (fmap unAxiom axs) [0..(length axs)])
+
+    addIndexesToClaims
+        :: [Axiom level]
```
vladciobanu: This is a bit confusing. I would rather just pass `length axioms` and take an `Int` here.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -288,6 +290,7 @@ showRule
     -> m ()
 showRule configNode = do
     Strategy.ExecutionGraph { graph } <- Lens.use lensGraph
+    ReplState { axioms } <- get
```
vladciobanu: You could probably combine this line with the above into something like

`ReplState { axioms, graph = Strategy.ExecutionGraph { graph } } <- get`

... I think?

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -297,7 +300,31 @@ showRule configNode = do
                 . Graph.inn'
                 . Graph.context graph
                 $ node'
+            let mrule = getRewriteRuleFromLabel
+                            . Graph.inn'
+                            . Graph.context graph
+                            $ node'
+            case mrule of
+                (Just rule) -> do
```
vladciobanu: `Just rule` not need parentheses

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -297,7 +300,31 @@ showRule configNode = do
...
+                        "Error: identifier attribute wasn't initialized."
+                        (axiomOrClaim (length axioms))
+                        mid
+                Nothing ->
+                    putStrLn' "No rule was applied."
         else putStrLn' "Invalid node!"
+  where
+    axiomOrClaim :: Int -> Int -> String
+    axiomOrClaim len iden =
+        if iden < len
```
vladciobanu: We prefer guards to if/then/else whenever possible:
```
  | iden < len = "Rule is axiom " <> show iden
  | otherwise = "Rule is claim " <> show (iden - len)
```

PR #566, opened 2019-04-09
Use constructor symbols in the SMT.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/SMT/Translate.hs
+++ b/kore/src/Kore/Step/SMT/Translate.hs
@@ -37,21 +37,22 @@ import           Data.Reflection
 import           Kore.AST.Kore
 import           Kore.AST.Valid
 import           Kore.Attribute.Hook
-import           Kore.Attribute.Smthook
 import           Kore.Attribute.Smtlib
 import qualified Kore.Attribute.Sort as Attribute
 import           Kore.Attribute.Symbol
-import qualified Kore.Attribute.Symbol as StepperAttributes
 import qualified Kore.Builtin.Bool as Builtin.Bool
 import qualified Kore.Builtin.Int as Builtin.Int
 import           Kore.IndexedModule.MetadataTools
 import           Kore.Predicate.Predicate
 import           Kore.Step.Pattern
+import           Kore.Step.SMT.Symbols
+                 ( SymbolTranslation (SymbolTranslation), translateSymbol )
+import qualified Kore.Step.SMT.Symbols as SymbolTranslation
+                 ( SymbolTranslation (..) )
+import           Kore.Unparser
 import           SMT
                  ( SExpr (..), SMT )
 import qualified SMT
-
-import Kore.Unparser
 
 -- ----------------------------------------------------------------
 -- Predicate translation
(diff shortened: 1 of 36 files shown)


kore/test/Test/Kore/Step/SMT/Builders.hs
```diff
@@ -0,0 +1,280 @@
...
+    $ mkApp (makeSort "Attribute") (makeHead "constructor") []
+
+smtlib :: Text -> Attribute
+smtlib value =
+    Attribute
+    $ eraseAnnotations
+    $ patternPureToKore
+    $ mkApp (makeSort "Attribute") (makeHead "smtlib") [mkStringLiteral value]
+
+hook :: Text -> Attribute
```
ttuegel: Except for the outer `Attribute` wrapper, this is [`Kore.Attribute.Hook.hookAttribute`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Attribute/Hook.hs#L44), and likewise for the other attribute definitions here.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -214,23 +215,14 @@ translatePredicate translateUninterpreted predicate =
             { applicationSymbolOrAlias
             , applicationChildren
             }
-      = case
-            getSmtlib (symAttributes smtlibTools applicationSymbolOrAlias)
-            <|>
-            getSmthook (symAttributes smthookTools applicationSymbolOrAlias)
-        of
+      = case translateSymbol applicationSymbolOrAlias of
```
ttuegel: Would it be better to use `translateName` here because the sorts are not needed? That would eliminate the need for `applicationSorts` in `MetadataTools`: the sorts are only used by `declareSymbol`, which already has the full symbol declaration sentence. I noticed a lot of test changes because of that addition which are otherwise not related to the SMT changes.

kore/test/Test/Kore/Step/SMT/Builders.hs
virgil-serbanuta: Fixed.

kore/src/Kore/Step/SMT/Translate.hs
virgil-serbanuta: No. I added a note in the `translateSymbol` documentation to explain why:

```
Note that it's not enough to call `translateName` to check whether a symbol
was/can be declared since it only checks that the symbol itself is declarable,
but it does not check that the symbol's sort is declarable.
```

ttuegel: > check whether a symbol was/can be declared

Could `declareSymbols` update `MetadataTools` with *that* information ("symbol was/was not declared") instead? I'm satisfied with the not, but just trying to think of ways to modularize the change. You can consider that, and I will approve the pull request.

traiansf: I've talked with Virgil and he agreed to look into this when he gets back on Monday.  Hence, I will merge this as-is and we'll move the card into Cleanup.

PR #567, opened 2019-04-09
All-path reachability claims

ttuegel:
This is the work-in-progress branch for implementing all-path reachability proofs.

###### Reviewer checklist

---


kore/test/Test/Kore/AllPath.hs
```diff
@@ -0,0 +1,347 @@
...
+
+type Goal = (K, K)
+
+type ProofState = AllPath.ProofState Goal
+
+type Rule = (K, K)
+
+ruleAC :: Rule
+ruleAC = (A, C)
+
```
marick: I'm not sure the name is preferable to the definition - especially since you later use plain tuples like  `unmodified (AllPath.Goal    (A, B))`.

kore/test/Test/Kore/AllPath.hs
```diff
@@ -0,0 +1,347 @@
...
+ruleEF = (E, F)
+
+noRules :: [Rule]
+noRules = []
+
+oneRule :: [Rule]
+oneRule = [ruleAB]
+
+concurrentRules :: [Rule]
+concurrentRules = [ ruleAB, ruleAC ]
```
marick: I'd probably prefer to see [ (A, B), (A, C) ] used explicitly in code like

```haskell
proves    concurrentRules (A, BorC) "concurrent rules"
```

That is:

```haskell
proves    [(A, B),  (A, C)] (A, BorC) "concurrent rules"
``` 
... so that I only have to look one place to see *which* concurrent rules.  I might also use infix to make it more clear which are the rules and which is the goal:

```haskell
[(A, B),  (A, C)] `proves` (A, BorC) $ "concurrent rules"
``` 

But that's the English major in me again.

kore/test/Test/Kore/AllPath.hs
```diff
@@ -0,0 +1,347 @@
...
+
+differentLengthPaths :: [Rule]
+differentLengthPaths =
```
marick: Took me some time to figure this out.  Maybe:

```haskell
[ AF

-- second path to F
, AB
, BC
, CF

-- Third path
-- AB again
BF
...
]

I don't know how many paths are needed to get the point across.

File: kore/test/Test/Kore/AllPath.hs [hunk 4]
```diff
@@ -0,0 +1,347 @@
...
+transitionRule
+    :: Prim
+    -> ProofState
+    -> Strategy.TransitionT Rule Identity ProofState
+transitionRule =
+    AllPath.transitionRule
+        removeDestination
+        triviallyValid
+        derivePar
+
```
marick: I'd be tempted to define things like `removeDestination` and `triviallyValid` after this definition. I'd forgotten what role they played until I saw how they were used.

That is, `where`-style rather than `let`-style.

I might actually recommend putting the `test_*` functions before the supporting definitions, assuming that people will want to read them first.

kore/test/Test/Kore/AllPath.hs
```diff
@@ -0,0 +1,347 @@
...
+    , AllPath.unprovenNodes
+        (emptyExecutionGraph (AllPath.Goal 1))
+        `equals`
+        (MultiOr.MultiOr [1])
+        $  "returns single unproven node"
+    , AllPath.unprovenNodes
+        (emptyExecutionGraph (AllPath.Goal 0)
+            & insNode (1, AllPath.Goal 1)
+            & insNode (2, AllPath.Proven)
+        )
```
marick: Is such a case possible?

kore/test/Test/Kore/AllPath.hs
```diff
@@ -0,0 +1,347 @@
...
+        `equals_`
+        (MultiOr.MultiOr [0, 1])
+    , AllPath.unprovenNodes
+        (emptyExecutionGraph (AllPath.Goal 0)
+            & insNode (1, AllPath.Goal 1)
+            & insEdge (0, 1)
+            & insNode (2, AllPath.Proven)
+            & insEdge (0, 2)
+        )
```
marick: Maybe

```haskell
goal 0 
  & (subgoal 0 1 $ AllPath.Goal 1)
  & (subgoal 0 2  AllPath.Proven)
```

kore/test/Test/Kore/AllPath.hs
ttuegel: I think it obscures whether the tuple is a `Rule` or a `Goal`, but I can use a `newtype` for that instead.

kore/test/Test/Kore/AllPath.hs
ttuegel: Such an `ExecutionGraph` would not arise from executing any of the strategies we employ, but I would like to make sure that `AllPath.unprovenNodes` behaves nicely over any graph to avoid dynamic entanglements.

PR #573, opened 2019-04-12
Unification with non-variable select pattern

traiansf:
###### Reviewer checklist

---


kore/src/Kore/Builtin/Set.hs
```diff
@@ -612,11 +613,26 @@ unifyEquals
           | otherwise = case Set.toList set1 of
             [fromConcreteStepPattern -> key1] ->
                 Reflection.give tools $ do
+                    let emptySetPat = asInternal tools sort1 Set.empty
                     (elemUnifier, _proof) <-
                         unifyEqualsChildren key1 key2
+                    -- when subunification problems fail, halt execution
```
traiansf: Extract Monad.When as a function

kore/src/Kore/Builtin/Set.hs
```diff
@@ -700,3 +706,22 @@ unifyEquals
       where
         Domain.InternalSet { builtinSetSort } = builtin1
         Domain.InternalSet { builtinSetChild = set1 } = builtin1
+
+-- Check whether the term part of an expanded pattern
+-- is identical to the expected term and error if not.
+errorIfNotUnifying
```
ttuegel: I'm still not sure *what* this is checking for, but please give it a `HasCallStack` constraint (from `GHC.Stack`) and use `unparse` on the arguments so that the first person who triggers this error has some hope of understanding what has happened.

virgil-serbanuta: +1 to what Thomas said, though it's likely that one would also need the other unification term in order to understand what happens.

Also:
1. It seems to me that this might work here, but it might not work in a list. To be specific I'm not sure it will work if the `expected` pattern contains (symbolic) variables. Since part of the builtin code was developed by copy-pasting, you may want to add a comment about that somewhere, or you may want to check whether the unified pattern or its predicate contains `and` instead (or something close to that, I'm not sure what). Actually, now that I'm thinking about it, what would be an example of something that triggers this error?

2. Given how confusing some of my similar `error`s were in the past, I would add (either in the error message or in a code comment) an explanation about why we expect unification to succeed. Mentioning the above example would probably help. I assume that this would be a combination of "set unification not fully implemented" and "some unknown part of generic unification didn't know how to handle something".

PR #578, opened 2019-04-15
Fix some of the unanswered comments in #524

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -17,7 +17,7 @@ module Kore.Step.Step
     , unwrapRule
     , applyUnifiedRule
     , applyRule
-    , applyRules
+    , applyRulesInParallel
     , applyRewriteRule
     , applyRewriteRules
     , sequenceRules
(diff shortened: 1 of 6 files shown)


kore/src/Kore/Step/Step.hs
```diff
@@ -135,6 +137,12 @@ deriving instance Ord (variable Object) => Ord (Result variable)
 
 deriving instance Show (variable Object) => Show (Result variable)
 
+instance TopBottom (Result variable) where
+    isTop (Result {unifiedRule, result}) =
+        isTop (Predicated.withoutTerm unifiedRule) && isTop result
```
ttuegel: The `PredicateSubstitution` from `unifiedRule` is always included in `result`, so we only need to check the `result`.

virgil-serbanuta: Yes, but there is nothing forcing them to be in sync, so I guess that checking both is better.

ttuegel: Then we should throw an `error` if they are not in sync, because if they are not kept in sync then the result will be wrong.

PR #582, opened 2019-04-16
Simplify \exists with bound substitutions

ttuegel:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/Exists.hs
+++ b/kore/src/Kore/Step/Simplification/Exists.hs
@@ -24,11 +24,11 @@ import           Kore.Attribute.Symbol
 import           Kore.IndexedModule.MetadataTools
                  ( MetadataTools )
 import           Kore.Predicate.Predicate
-                 ( Predicate, makeExistsPredicate, makeTruePredicate,
-                 unwrapPredicate )
+                 ( Predicate, makeExistsPredicate, makeTruePredicate )
+import qualified Kore.Predicate.Predicate as Predicate
 import           Kore.Step.Axiom.Data
                  ( BuiltinAndAxiomSimplifierMap )
-import           Kore.Step.Pattern
+import           Kore.Step.Pattern as Pattern
 import           Kore.Step.Representation.ExpandedPattern
                  ( ExpandedPattern, Predicated (..) )
 import qualified Kore.Step.Representation.ExpandedPattern as ExpandedPattern
@@ -48,8 +48,6 @@ import           Kore.Unification.Substitution
                  ( Substitution )
 import qualified Kore.Unification.Substitution as Substitution
 import           Kore.Unparser
-import           Kore.Variables.Free
-                 ( freePureVariables )
 import           Kore.Variables.Fresh
 
 
@@ -240,10 +238,9 @@ makeEvaluateNoFreeVarInSubstitution
   =
     (MultiOr.make [simplifiedPattern], SimplificationProof)
   where
-    termHasVariable =
-        Set.member variable (freePureVariables term)
-    predicateHasVariable =
-        Set.member variable (freePureVariables $ unwrapPredicate predicate)
+    hasVariable = Set.member variable
+    termHasVariable = hasVariable (Pattern.freeVariables term)
+    predicateHasVariable = hasVariable (Predicate.freeVariables predicate)
     simplifiedPattern = case (termHasVariable, predicateHasVariable) of
         (False, False) -> patt
         (False, True) ->
--- a/kore/src/Kore/Step/Simplification/Exists.hs
+++ b/kore/src/Kore/Step/Simplification/Exists.hs
@@ -12,6 +12,8 @@ module Kore.Step.Simplification.Exists
     , makeEvaluate
     ) where
 
+import           Control.Applicative
+                 ( Alternative ((<|>)) )
 import           Data.Map.Strict
                  ( Map )
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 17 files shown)


kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -236,45 +235,50 @@ makeEvaluateNoFreeVarInSubstitution
...
-                        (ExpandedPattern.toMLPattern
-                            Predicated
-                                { term = term
-                                , predicate = predicate
-                                , substitution = mempty
-                                }
-                        )
-                , predicate = makeTruePredicate
-                , substitution = substitution
+                    $ Maybe.fromMaybe (mkTop patternSort)
```
virgil-serbanuta: I know that this `mkTop patternSort` is not actually used, but, since it's here, locally it looks weird that we're adding an `exists` on something that might be a `top`. Fmapping the mkExists before calling `fromMaybe` might look better.

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -236,45 +235,50 @@ makeEvaluateNoFreeVarInSubstitution
...
+                    $ mkAndMaybe (Just term)
+                    $ Predicate.fromPredicate patternSort <$> boundCondition
+                }
+          | otherwise =
+            (ExpandedPattern.topOf patternSort)
+                { term = term
+                , predicate =
+                    Predicate.makeAndPredicate freePredicate
+                    $ Predicate.makeExistsPredicate variable
+                    $ Maybe.fromMaybe Predicate.makeTruePredicate boundCondition
```
virgil-serbanuta: The same as above: why add a quantifier on a `top` value?

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -198,16 +199,16 @@ makeEvaluate
...
-        _ -> do
+        Just boundTerm -> do
             (substitutedPat, _proof) <-
                 substituteTermPredicate
                     term
                     predicate
-                    localSubstitutionList
-                    (Substitution.wrap globalSubstitution)
+                    (Map.singleton variable boundTerm)
+                    (Substitution.unsafeWrap $ Map.toList freeSubstitution)
```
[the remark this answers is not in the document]
ttuegel: This step will only work if the substitution is normalized, otherwise we can't apply the substitution correctly. Should we call `normalize` here to ensure that it is normalized, or fall back to the second case if it is not?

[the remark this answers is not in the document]
virgil-serbanuta: I'm worried that there is no simple and obvious way (preferably one that's also stable through code changes) to see that `freeSubstitution` is, indeed, normalized.

Calling `normalize` here should be fine. Using `Kore.Unification.Substitution.partition` or something close to it (but declared in the same module) that produces a wrapped substitution would also be fine.

ttuegel:
@virgil-serbanuta This is ready for your re-review. I changed `Kore.Step.Simplification.Exists.makeEvaluate` to normalized the substitution before doing anything that depends on the normalization state. I also discovered that we were doing some things with `Kore.Unification.Substitution.toMap` which were not quite safe; I changed a few types and added some error cases to catch uses of `toMap` on denormalized substitutions.

kore/src/Kore/Unification/Substitution.hs
```diff
@@ -35,71 +38,108 @@ import           Data.Set
...
+A collection of substitutions @[xᵢ=φᵢ]@ is /normalized/ if, for all @xⱼ=φⱼ@ in
+the collection, @xⱼ@ is unique among all @xᵢ@ and none of the @xᵢ@ (including
+@xⱼ@) occur free in @φⱼ@.
+
+ -}
 data Substitution level variable
+    -- TODO (thomas.tuegel): Instead of a sum type, use a product containing the
+    -- normalized and denormalized parts of the substitution together. That
+    -- would enable us to keep more substitutions normalized in the Semigroup
+    -- instance below.
```
virgil-serbanuta: Are you sure it's worth doing that? What would be the benefit? And it might not be trivial - consider `x=f(y)`: it's ok to merge it with `z=g(y)`, but not with any of `x=g(z)`, `y=g(z)`, `z=g(x)`, or `z=sigma(y) /\ z = sigma(g(t))`.

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -196,194 +195,199 @@ makeEvaluate
...
-@
-(∃ y. φ ∧ boundP ∧ boundS) ∧ freeP ∧ freeS
-@
-
-otherwise (if @y@ does not occur in @φ@) the result is
-
-@
-φ ∧ (freeP ∧ (∃ y. boundP ∧ boundS)) ∧ freeS.
-@
+See also: 'quantifyExpandedPattern'
```
virgil-serbanuta: This function relies on `boundTerm` not containing `variable` as a free variable, it may be worth adding that constraint here (another way would be to use a normalized substitution, but that may not be worth the effort).

ttuegel: I documented this fact and added a check.

kore/src/Kore/Unification/Substitution.hs
ttuegel: I think it's worth it if we keep substitution deduplication as a separate step, because we can keep the substitution normalized (well, semi-normalized: cycle-free) incrementally. The most expensive thing we do during execution is merging and re-normalizing substitutions; it's expensive in large part because merging/re-normalization always begins by assuming nothing is normalized. I have in mind a representation such as `data Substitution = Substitution { normal = Map variable (MultiAnd pattern), denormal = Map variable (MultiAnd pattern) }` with the invariant that no variable occurs as a key in both `normal` and `denormal` and the keys of `normal` do not occur in any values of `normal` or `denormal`.

I would merge those examples in the following ways, where each substitution is represented as a pair as described above:

1. `([x=f(y)], []) ∧ ([x=g(z)], [])` = `([x=f(y)∧g(z)], [])` (postponing de-duplication)
1. `([x=f(y)], []) ∧ ([y=g(z)], [])` = `([x=f(g(z)), y=g(z)], [])` (applying substitution, which has no cycles)
1. `([x=f(y)], []) ∧ ([z=g(x)], [])` = `([x=f(y), z=g(f(y))], [])` (as above)
1. `([z=σ(y)], []) ∧ ([z=σ(g(t))], [])` = ([z=σ(y)∧σ(g(t))], [])` (postponing de-duplication, as above)
1. An interesting case not mentioned: `([x=f(y), []) ∧ ([y=g(x)], [])` = `([x=f(y), y=g(f(y))])` (breaking a cycle by applying the `normal` substitution and denormalizing; there is an choice about which way we break the cycle)

Anyway, we can discuss this in more detail when performance becomes an issue, and we can decide to if and how to do this, or not. For now, I mainly wanted to document that we have the option of doing this.

PR #585, opened 2019-04-17
Distinguish \bottom unifier and \bottom result

ttuegel:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -13,6 +13,8 @@ module Kore.Step.Step
     , UnifiedRule
     , Results (..)
     , Result (..)
+    , gatherResults
+    , withoutRemainders
     , unifyRule
     , unwrapRule
     , applyUnifiedRule
@@ -34,6 +36,9 @@ import qualified Data.Function as Function
 import qualified Data.Map.Strict as Map
 import           Data.Semigroup
                  ( Semigroup (..) )
+import           Data.Sequence
+                 ( Seq )
+import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 import qualified Data.Text.Prettyprint.Doc as Pretty
 import           GHC.Generics as GHC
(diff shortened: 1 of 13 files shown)


kore/src/Kore/Step/Step.hs
```diff
@@ -340,6 +301,75 @@ applyUnifiedRule
             axiomSimplifiers
             condition
 
+{- | Produce the final configurations of an applied rule.
+
+The rule's 'ensures' clause is applied to the conditions and normalized. The
+substitution is applied to the right-hand side of the rule to produce the final
+configurations.
+
+Because the rule as known to apply, @finalizeAppliedRule@ always returns exactly
```
virgil-serbanuta: as known?

kore/src/Kore/Step/Step.hs
```diff
@@ -303,34 +268,30 @@ applyUnifiedRule
...
     -> PredicateSubstitution Object variable
     -- ^ Initial conditions
-    -> UnifiedRule variable
-    -- ^ Non-normalized final configuration
-    -> BranchT unifier
-        (ExpandedPattern Object variable)
-applyUnifiedRule
+    -> PredicateSubstitution Object variable
+    -- ^ Unification conditions
+    -> BranchT unifier (OrOfPredicateSubstitution Object variable)
```
virgil-serbanuta: I didn't think carefully about this, but why do you need BranchT here? Is it only to avoid a `lift` in the caller?

The same question for `finalizeAppliedRule`. Actually, since we're restarting to work with `OrOfStuff`, do we still need BranchT in this file?

ttuegel: > Is it only to avoid a `lift` in the caller?

Yes. This keeps all the code in the caller at the same level of abstraction.

> Actually, since we're restarting to work with `OrOfStuff`, do we still need BranchT in this file?

Yes. Your comments on another recent pull request made me realize that we cannot always satisfy the internal invariants of `MultiOr`. For example, when applying rules, we need to branch on things that might not be `TopBottom`, and we may need to keep branches "alive" even when they return something that `isBottom`, such as when the right-hand side of a rule is bottom.

ttuegel: > This keeps all the code in the caller at the same level of abstraction.

Replying to my own comment: This won't be necessary anymore when we clean up the issues around defining an `instance MonadUnify (BranchT unifier)`.

PR #589, opened 2019-04-18
Parameterized Repl and Prover on claims. Added OnePath and AllPath claims

traiansf:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -23,6 +23,8 @@ import qualified Control.Monad as Monad
 import           Control.Monad.Trans.Except
                  ( runExceptT )
 import qualified Data.Bifunctor as Bifunctor
+import           Data.Coerce
+                 ( coerce )
 import qualified Data.Map.Strict as Map
 import           System.Exit
                  ( ExitCode (..) )
@@ -47,8 +49,7 @@ import           Kore.IndexedModule.Resolvers
                  ( resolveSymbol )
 import qualified Kore.Logger as Log
 import           Kore.OnePath.Verification
-                 ( Axiom (Axiom), Claim, defaultStrategy, toClaim, unClaim,
-                 verify )
+                 ( Axiom (Axiom), Claim, defaultStrategy, verify )
 import qualified Kore.OnePath.Verification as Claim
 import           Kore.Predicate.Predicate
                  ( pattern PredicateTrue, makeMultipleOrPredicate,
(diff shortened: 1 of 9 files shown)


kore/src/Kore/Exec.hs
```diff
@@ -275,26 +279,34 @@ proveWithRepl definitionModule specModule = do
         axioms
         claims
 
-makeClaim :: (Attribute.Axiom, Rewrite) -> Claim Object
+assertSomeClaims :: Monad m => [claim] -> m ()
+assertSomeClaims claims =
+    Monad.when (null claims) . error
+        $   "Unexpected empty set of claims.\n"
+        ++  "Possible explanation: the frontend and the backend don't agree "
+        ++  "on the representation of claims."
```
ttuegel: Would this happen because of using the wrong frontend and backend versions together?

kore/src/Kore/OnePath/Verification.hs
```diff
@@ -113,14 +114,22 @@ decision is subject to change without notice.
 
  -}
 
-{- | Wrapper for a rewrite rule that should be used as a claim.
+{- | Class for claim-like rules
 -}
-newtype Claim level = Claim
-    { unClaim :: RewriteRule level Variable
-    }
+class Claim claim where
```
ttuegel: I don't like making classes that expose the underlying representation like this, because we can never change the representation. The only types that can correctly implement this class are `newtype` wrappers around `RewriteRule`; why not use [`Coercible`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Coerce.html#t:Coercible) instead?

By the way, because `RewriteRule`, `OnePathRule`, and `AllPathRule` each have a distinct Kore representation, I think we should use `RulePattern` instead of `RewriteRule` here.

kore/src/Kore/Step/Rule.hs
```diff
@@ -184,6 +197,52 @@ extractRewriteAxiomFrom level sentence =
   where
     koreSentence = constructUnifiedSentence SentenceAxiomSentence sentence
 
+-- | Extracts all One-Path claims from a verified definition.
```
ttuegel:
```suggestion
-- | Extracts all One-Path claims from a verified module.
```

kore/src/Kore/Exec.hs
traiansf: Yes. What I wanted to achieve was a quick way to fail the build (instead of returning #True for verifying the empty list of claims) as it happened before and also was happening in this PR

PR #594, opened 2019-04-19
Keep \not predicates

ttuegel:
###### Reviewer checklist

---


virgil-serbanuta:
Could you add, in the PR description, an example where it's better to use `and(not)` instead of `not(or)`? I know that the PR title provides a hint, but it would be nice to also have something that says what this PR solves instead of just hinting at it.

ttuegel:
> an example where it's better to use `and(not)` instead of `not(or)`

Ah, that's not even the point of this pull request, so probably we should discuss this more!

Before this pull request, we constructed the `not(or)` of multiple terms by combining the predicate and substitution with the term. We don't want to do that because it forgets information learned during simplification, and it violates the predicate-simplifier's assumption that the result of simplifying a predicate has a `\top` term. I only made it return `and(not)` because that was the most obvious way to keep terms and predicates distinct using the existing `Not.makeEvaluate` and `And.makeEvaluate`. An `and(not)` condition is stronger (easier to refute) but this also produces more branches because `Not.makeEvaluate` often returns multiple results.

kore/src/Kore/Step/Simplification/Not.hs
```diff
@@ -77,31 +99,37 @@ besides the pattern sort, which will make it even more useful to carry around.
...
-        _ ->
-            ( makeFromSinglePurePattern
-                (mkNot
-                    (ExpandedPattern.toMLPattern
-                        (OrOfExpandedPattern.toExpandedPattern simplified)
-                    )
-                )
-            , SimplificationProof
-            )
+    Foldable.foldrM mkAnd ExpandedPattern.top (simplified >>= makeEvaluate)
```
virgil-serbanuta: For some reason, I don't find the MultiOr monad intuitive enough... Oh well, I guess I need to get used to it.

PR #596, opened 2019-04-23
Added variableType field to Variable

traiansf:
###### Reviewer checklist

---


kore/src/Kore/AST/Common.hs
```diff
@@ -112,6 +112,20 @@ instance Unparse (SymbolOrAlias level) where
       =
         unparse symbolOrAliasConstructor <> parameters symbolOrAliasParams
 
+
+{-|@VariableType@ corresponds to the `EVar` and `SVar` categories of variables,
+in the Matching mu-Logic paper, used to distinguish between element variables,
+which can only be instantiated by an singleton, i.e., a functional pattern,
```
virgil-serbanuta: s/an/a/

kore/test/Test/Kore/Parser/Parser.hs
```diff
@@ -380,9 +382,10 @@ applicationPatternParserTests =
     parseTree korePatternParser
         [ success "#v:#Char"
             ( asCommonKorePattern $ VariablePattern Variable
-                { variableName = testId "#v" :: Id Meta
+                { variableName = testId "v" :: Id Meta
```
virgil-serbanuta: s/Meta/Object?

kore/src/Kore/AST/Common.hs
```diff
@@ -132,19 +146,27 @@ data Variable level = Variable
     { variableName :: !(Id level)
     , variableCounter :: !(Maybe (Sup Natural))
     , variableSort :: !(Sort level)
+    , variableType :: !VariableType
```
ttuegel: I would like to suggest that element and set variables should be different types of `Variable`. These have completely different semantics. Parts of the code that operate on element variables will not necessarily operate on set variables, and vice versa. Using distinct types would remove the need to check the `variableType` field in potentially every function which takes a `Variable` as input. We could also add support for set variables in a completely incremental fashion, without changing any of the code that currently handles element variables.

traiansf:
Closing this in favor of PR #598

kore/src/Kore/AST/Common.hs
virgil-serbanuta: Me and Traian had a long discussion about this (I was originally suggesting the same thing as you, for the same reasons). The problem is that their semantics are also similar: we can unify them (although the patterns don't have to be functional anymore) and we can substitute them, and both of these probably work in almost the same way as for element variable. Of course, they don't use the same quantifiers and I think I found a few other differences, but using the same type was tempting.

Well, I guess that adding another `variable` parameter to our types can't be that bad, so let's go with the safer solution.

traiansf: In the end we did not add another variable parameter, but instead we added another type of pattern and a SetVariable wrapper over the variable argument.

We'll think about how to handle them for unification when we get there.

PR #601, opened 2019-04-24
Fixes for comparators

virgil-serbanuta:
* Fix argument matching for StructEqualWithExplanation and WrapperEqualWithExplanation in order to let people know they should change the comparators when they change the structures. As you'll notice from this PR, they don't always remember to do it. In case you think that these should be automatically generated: I fully agree, I sent a PR that did it, but it seems that I used a reflection library that wasn't nice enough. Maybe I'll try to do it again, but I won't mind if someone else does it before.

* Fix line length issues.

###### Reviewer checklist

---


virgil-serbanuta:
For whatever it's worth, here's my previous PR:
https://github.com/kframework/kore/pull/382

PR #606, opened 2019-04-26
Early termination

yzhang90:
@ttuegel 
I implemented the early termination as you suggested. please take a look.

###### Reviewer checklist

---


kore/src/Kore/ModelChecker/Step.hs
```diff
@@ -115,34 +119,34 @@ transitionRule
...
     proofState
-  =
-    case strategyPrim of
-        CheckProofState -> transitionCheckProofState proofState
-        Simplify -> transitionSimplify proofState
-        Unroll goalrhs -> transitionUnroll goalrhs proofState
-        ComputeWeakNext rewrites -> transitionComputeWeakNext rewrites proofState
+  = do
+    execState <- Monad.Trans.lift State.get
+    case execState of
```
ttuegel: I think this would be clearer, with less creeping indentation:
```.hs
  = do
    execState <- Monad.Trans.lift State.get
    Foldable.traverse_ (const empty) execState  -- End early if any unprovable state was reached
    case strategyPrim of
        ...
```
You could also use a `Bool` instead of `Maybe ()` if you don't want to return the unprovable state itself, then it would look like
```.hs
  = do
    unprovable <- Monad.Trans.lift State.get
    Control.Monad.when unprovable empty  -- End early if any unprovable state was reached.
    case strategyPrim of
        ...
```

kore/src/Kore/ModelChecker/Step.hs
```diff
@@ -185,7 +188,7 @@ transitionRule
       = case modalOp of
             "ag" -> do
                 result <-
-                    Monad.Trans.lift
+                    (Monad.Trans.lift . Monad.Trans.lift)
```
ttuegel: The parentheses are actually redundant,
```suggestion
                    Monad.Trans.lift . Monad.Trans.lift
```

kore/src/Kore/ModelChecker/Step.hs
```diff
@@ -117,16 +125,17 @@ transitionRule
...
     tools
     predicateSimplifier
     patternSimplifier
     axiomSimplifiers
     strategyPrim
     proofState
-  =
+  = do
+    execState <- Monad.Trans.lift State.get
+    when (isJust execState) empty     -- End early if any unprovable state was reached
```
yzhang90: @ttuegel I used `when` instead of `traverse_` here.

kore/src/Kore/ModelChecker/Step.hs
yzhang90: done

kore/src/Kore/ModelChecker/Step.hs
yzhang90: done

PR #608, opened 2019-04-28
Spring Cleaning: Mass renaming

ttuegel:
This is the "mass renaming" portion of clean-up. It doesn't cover any of the other refactoring, just renaming, and sometimes pinning `level ~ Object`. A few things still have old names involving `ExpandedPattern` to remind me that they actually need to be deleted instead of renamed because they shouldn't exist in the first place.

###### Reviewer checklist

---


kore/src/Kore/Step/Condition/Evaluator.hs
```diff
@@ -13,19 +13,17 @@ module Kore.Step.Condition.Evaluator
...
-import           Kore.AST.Pure
 import           Kore.Attribute.Symbol
 import           Kore.IndexedModule.MetadataTools
-import           Kore.Predicate.Predicate
-import           Kore.Step.Representation.ExpandedPattern
-                 ( ExpandedPattern, PredicateSubstitution, Predicated (..) )
-import qualified Kore.Step.Representation.ExpandedPattern as ExpandedPattern
-import qualified Kore.Step.Representation.OrOfExpandedPattern as OrOfExpandedPattern
-                 ( isFalse, isTrue, toExpandedPattern )
+import qualified Kore.Predicate.Predicate as Syntax
```
virgil-serbanuta: I'm curious: why "Syntax"?

kore/src/Kore/Step/Pattern.hs
```diff
@@ -1,317 +1,241 @@
...
 import           Kore.Unparser
-import           Kore.Variables.Fresh
-
-type StepPattern level variable =
-    PurePattern level Domain.Builtin variable (Valid (variable level) level)
+import           Kore.Variables.Free
+                 ( pureAllVariables )
 
-type CommonStepPattern level = StepPattern level Variable
+{- | The conjunction of a pattern, predicate, and substitution.
```
virgil-serbanuta: Now that the terminology has changed, you should probably s/pattern/term-like/

PR #611, opened 2019-04-29
Spring cleaning: Remove MetaOrObject

ttuegel:
Today I removed many level parameters, and the `MetaOrObject` class itself. I also started to split up the module `Kore.AST.Common`. Most other modules that import this one only need a tiny subset of its definitions. It also takes a long time to compile (up to tens of seconds) because it is so large; GHC can build in parallel if it is split up. Finally, some of the syntax types have their `level` parameter removed and replaced with a `sort` parameter. This is important so we can use a `Sort` type in the syntax checker with more information. It will also be important when we decide to support Applicative Kore, which has no sort information in the parsed source at all. I will finish breaking up `Kore.AST.Common` tomorrow, so it is not essential that this be merged immediately.

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`
- [ ] Style conformance: `stylish-haskell`

---

PR #613, opened 2019-04-30
Spring cleaning: Break up Kore.AST.Common

ttuegel:
Follow-up to #611.

###### Reviewer checklist

---


kore/src/Kore/Syntax/Not.hs
```diff
@@ -0,0 +1,52 @@
...
+import Kore.Unparser
+
+{-|'Not' corresponds to the @\not@ branches of the @object-pattern@ and
+@meta-pattern@ syntactic categories from the Semantics of K,
+Section 9.1.4 (Patterns).
+
+'notSort' is both the sort of the operand and the sort of the result.
+
+-}
+data Not sort child = Not
```
virgil-serbanuta: It's not obvious why the `sort` argument is needed. Is it to allow the Applicative stuff?

PR #615, opened 2019-05-01
Sort module sentences before indexing

ttuegel:
Sort module sentences before indexing.

The order of sentences in a module is not formally significant, so the order produced by the frontend varies depending on the build, the operating system, the phases of the moon, non-local quantum variables, ...

Although the order of sentences is not formally significant, it affects debugging: if either of two axioms would produce an error, applying them in one order or the other changes which error is actually produced. Therefore, we sort all the sentences before beginning so that errors are more readily-reproducible.

###### Reviewer checklist

---

### DIFF
--- a/kore/test/Test/Kore/IndexedModule/Resolvers.hs
+++ b/kore/test/Test/Kore/IndexedModule/Resolvers.hs
@@ -1,14 +1,13 @@
-module Test.Kore.IndexedModule.Resolvers
-where
+module Test.Kore.IndexedModule.Resolvers where
 
 import Test.Tasty
 import Test.Tasty.HUnit
--- import Test.Terse
 
 import           Data.Default
+import qualified Data.List as List
+import           Data.Map
+                 ( Map )
 import qualified Data.Map as Map
-import           Data.Maybe
-                 ( fromMaybe )
 import qualified Data.Set as Set
 
 import           Kore.Annotation.Valid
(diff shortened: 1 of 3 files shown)


kore/test/Test/Kore/IndexedModule/Resolvers.hs
```diff
@@ -37,6 +36,18 @@ objectS1 = simpleSort (SortName "s1")
 objectA :: SentenceSymbol Object (TermLike Variable)
 objectA = mkSymbol_ (testId "a") [] objectS1
 
+-- Two variations on a constructor axiom for 'objectA'.
```
marick: It would be nice for the names, or the comment, to describe what's different in the two variants. I can't tell by reading.

kore/test/Test/Kore/IndexedModule/Resolvers.hs
```diff
@@ -285,6 +295,11 @@ test_resolvers =
                 (getSentenceSymbolOrAliasHead objectB [])
             )
         )
+    , testCase "sort indexed axioms"
```
marick: You have to do a lot of digging to see what axioms are sorted *by*.

marick: Though I suppose it doesn't actually matter, so long as the order is always the same. Still.

kore/test/Test/Kore/IndexedModule/Resolvers.hs
```diff
@@ -285,6 +295,11 @@ test_resolvers =
...
+        (assertEqual ""
+            (reverse $ List.sort [axiomA, axiomA'])
+            (getIndexedSentence <$> indexedModuleAxioms testIndexedObjectModule)
+        )
```
marick: I'd prefer something like:

```haskell
axiomsSortedBackward = [axiomSortedLast, axiomSortedFirst]
-- axiomSortedBackward is use to insert into `testObjectModel`

assertEqual
   assertEqual (reverse axiomSortedBackward)
   ....
```

I would also typically, if it's easy, have a test that shows that the naive way of extracting the elements would result in them being in the opposite order. Otherwise, there's a too high chance of the particular test being only coincidentally correct.

This also works nicely with TDD. A typical sequence would be:

1. Write a test that confirms that the function returns the elements in the order inserted (if it doesn't exist).
2. Preserve that test.
3. Write a new test that inserts the elements in the opposite order but it expects the same result as the earlier test.
4. That fails, forcing you to add sorting code. 
5. The two tests guard against coincidental correctness.

PR #616, opened 2019-05-01
Postpone normalizing in some cases

ttuegel:
Postpone normalizing in some cases where we cannot handle multiple results.

Depends on: #615

###### Reviewer checklist

---


ttuegel:
In the stand-up this morning, I said we don't *need* this because of #627, but upon further reflection I still want to merge it: it begins to achieve the goal of separating the "simplification" and "evaluation" stages of execution. Separate (structural) simplification from evaluation will improve the architecture of the code, and it avoids the repeated evaluation calls which hinder our performance.

PR #620, opened 2019-05-02
Spring cleaning: Final reorganization

ttuegel:
###### Reviewer checklist

---


ttuegel:
In addition to renaming, my final pull request begins to organize the module structure to reflect the layers of the backend by adding two new groups: `Kore.Syntax` and `Kore.Internal`.

At the bottom are the modules under `Kore.Syntax`, which are types for representing the syntax of Kore as written. Values in these types are produced by `Kore.Parser`.

Upward from the parser are internal representations of Kore patterns under `Kore.Internal`. These came from `Kore.Step` for historical reasons, but they don't belong to that hierarchy; they are the internal representations common throughout most of the backend. The `Kore.Builtin` modules probably belong in `Kore.Internal` also. (At the moment, the internal representation of patterns is still partially tied to the parsed representation, but we need to break that dependency eventually.)

The `Kore.Syntax` types are taken to `Kore.Internal` by the `Kore.ASTVerifier` hierarchy. The `Kore.IndexedModule` modules help here, but these probably should be moved into the syntax checker because neither is independent of the other.

The `Kore.Step` hierarchy is still a big monolithic chunk which is consumed by `Kore.Exec`, `Kore.Repl`, `Kore.OnePath`, etc. Next we might want to move the simplifier out of `Kore.Step` because it is also used independently by the others also; rule application done in `Kore.Step` is just another client of the simplifier. @virgil-serbanuta suggested recently to divide simplification into two stages; I think that's a good idea and it would be productive to split the simplifier between "structural" (pure functions) and "normalizing" (call out to Z3) lines.

marick:
Actually, not a stylish error. A `--pedantic` error.

PR #623, opened 2019-05-07
Haskell linting

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-exit_code=0
+set -exuo pipefail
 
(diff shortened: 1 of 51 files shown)


scripts/lint.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+exit_code=0
+
+for f in `find ./kore -name '*.hs'`; do
```
ttuegel: I think we have a more general invocation to find Haskell source files in one of the other scripts (`check.sh`?). Should we extract it for consistency?

virgil-serbanuta: Thanks, done (it was `stylish.sh`).

PR #626, opened 2019-05-07
repl: check if graphviz is installed

ana-pantilie:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Repl/Interpreter.hs
+++ b/kore/src/Kore/Repl/Interpreter.hs
@@ -52,6 +52,7 @@ import           GHC.Exts
                  ( toList )
 import           GHC.IO.Handle
                  ( hGetContents, hPutStr )
+import           Numeric.Natural
 import           System.Directory
                  ( findExecutable )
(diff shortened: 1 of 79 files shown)


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -201,16 +201,20 @@ prove cindex = do
...
     => m ()
 showGraph mfile = do
     graph <- getInnerGraph
     axioms <- Lens.use lensAxioms
-    liftIO $ maybe
-            (showDotGraph (length axioms) graph)
-            (saveDotGraph (length axioms) graph)
-            mfile
+    installed <- liftIO Graph.isGraphvizInstalled
+    case installed of
```
ttuegel: I would probably use `if ... then ... else` instead of `case ... of ...` for a `Bool`.

ana-pantilie: Yeah, you're right. Thanks!

PR #629, opened 2019-05-07
Fail early during one-path reachability proofs

ttuegel:
This pull request causes one-path reachability proofs to fail at the first stuck remainder. Such proofs can loop if we fail to apply a circularity (for example) and this makes such a failure easier to debug.

###### Reviewer checklist

---


ttuegel:
This pull request now also does not perform extra simplifications of nodes which are `Stuck` because it would not change the result. This produces an 18x speed-up.

PR #632, opened 2019-05-09
Implement SET.intersection

ttuegel:
I implemented this because I thought I needed it. In fact, I don't need it, but it would be good to have anyway.

###### Reviewer checklist

---


ttuegel:
See also: https://github.com/kframework/kore/wiki/Builtin-symbols

PR #633, opened 2019-05-09
Test.Kore.Step.Substitution: Check for normalization

ttuegel:
Follow-up to #616. As discussed, this explicitly checks that the substitution is normalized by the functions which specifically normalize the substitution.

###### Reviewer checklist

---


kore/test/Test/Kore/Step/Substitution.hs
```diff
@@ -1,22 +1,21 @@
-module Test.Kore.Step.Substitution
+ module Test.Kore.Step.Substitution
```
virgil-serbanuta: Unintended space, I guess.

PR #634, opened 2019-05-09
patternToAxiomPattern: Recognize \ceil axioms

ttuegel:
In the future, we expect to need many such defined-ness axioms.

It would also be nice if we could specify these axioms in the frontend, but for now this will do.

###### Reviewer checklist

---


ttuegel:
> I'm ok with us supporting adding ceil rules manually, but I would rather try to add them through the frontend, because hardcoded rules are harder to present to the user.

Of course! At this point I'm only adding one rule as a work-around until we can do so in the frontend.

PR #637, opened 2019-05-09
Pattern.simplify: Use filterOr

ttuegel:
This fixes a bug in `Pattern` simplification related to a use of `traverse`. I also added a card to the story board to clean up other possibly-problematic uses.

Related: #635 

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/Pattern.hs
+++ b/kore/src/Kore/Step/Simplification/Pattern.hs
@@ -15,6 +15,7 @@ import           Kore.IndexedModule.MetadataTools
                  ( SmtMetadataTools )
 import           Kore.Internal.OrPattern
                  ( OrPattern )
+import qualified Kore.Internal.OrPattern as OrPattern
 import           Kore.Internal.Pattern
                  ( Conditional (..), Pattern )
 import qualified Kore.Internal.Pattern as Pattern
@@ -58,7 +59,7 @@ simplify
     Conditional {term, predicate, substitution}
   = do
     simplifiedTerm <- simplifyTerm' term
-    traverse
+    OrPattern.filterOr <$> traverse
         (give tools $ Pattern.mergeWithPredicate
             tools
             substitutionSimplifier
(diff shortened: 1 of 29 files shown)


kore/src/Kore/Step/Simplification/Pattern.hs
```diff
@@ -58,7 +59,7 @@ simplify
     Conditional {term, predicate, substitution}
   = do
     simplifiedTerm <- simplifyTerm' term
-    traverse
+    OrPattern.filterOr <$> traverse
```
virgil-serbanuta: Two questions:
1. Exactly why is this needed here (I looked at #635 , which was referenced in the PR description, but the connection is not obvious)?
2. Does this mean that we should never fully convert this function to using `BranchT Simplifier`? If yes, then you should add a comment explaining why this is the case. A test would be even better.

Actually, especially since this fixes a bug, we should add a (regression?) test regardless of the answer to 2 above.

ttuegel: >     2\. Does this mean that we should never fully convert this function to using `BranchT Simplifier`?

Actually, we would never have this bug if we used `BranchT Simplifier`! The problem is: the signature of `traverse` is too polymorphic to allow us to filter the results. (The results might not be patterns, for example.) If we used `BranchT`, we would have to use `fmap MultiOr.make . gather` before we could get an `OrPattern`, which would do the filtering. I think a reasonable solution to this is to make `OrPattern` a `newtype` and give it a [monomorphic `traverse` function](https://hackage.haskell.org/package/mono-traversable-1.0.11.0/docs/Data-MonoTraversable.html#v:otraverse) which also filters, which is why I added [this card](https://miro.com/app/board/o9J_kyKH9fw=/?moveToWidget=3074457346625013517) 

>     1. Exactly why is this needed here

The `traverse` can construct a non-empty `OrPattern` whose only element is a `\bottom` pattern, and then `isBottom` incorrectly returns `False`. I didn't think it made sense to have `isBottom` call `filterOr` every time because this is supposed to be an invariant of `OrPattern`.

ttuegel:
@virgil-serbanuta I added tests as you suggested.

PR #639, opened 2019-05-10
Repl: alias with arguments

vladciobanu:
Limitations:
~- no parsing error when attempting to add an alias for an existing command (although it will have no real effect)~ fixed
- we don't have great errors (but this is a problem with the repl in general since we hide parsing errors)
~- no checking until we actually try running the alias, so the alias could be completely malformed~
- no checking that the alias is malformed, but we do check that at least the "verb" (first word in the command) is defined
~- you can create infinite loops:~

You can't create infinite loops because of the check above (first line would error because `ln` is not defined
```
> alias l0 = loop
Error: unknown command.
```

### DIFF
--- a/kore/src/Kore/Repl.hs
+++ b/kore/src/Kore/Repl.hs
@@ -100,7 +100,7 @@ runRepl tools simplifier predicateSimplifier axiomToIdSimplifier axioms' claims'
         str <- prompt
         let command = maybe ShowUsage id $ parseMaybe commandParser str
         when (shouldStore command) $ lensCommands Lens.%= (Seq.|> str)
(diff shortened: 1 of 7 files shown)


ttuegel:
If I defined this alias,
```
alias c n = config n > "file n.kore"
```
how would `c 3` expand?

vladciobanu:
> If I defined this alias,
> 
> ```
> alias c n = config n > "file n.kore"
> ```
> 
> how would `c 3` expand?

It would expand to `config 3 > "file n.kore"`, this is specifically why the `QuotedArgument` constructor exists.

PR #645, opened 2019-05-14
Distinguish internal and external representations of patterns

ttuegel:
Depends on: #628 

###### Reviewer checklist

---


ttuegel:
> Looks good to me, but some things (like Binder and lenses for it) might deserve being presented to the team.

Good idea!

> Also, now that you've added String builtins, maybe we can finally get rid of StringLiteral and CharLiteral and simply use String for domain values of non-builtin sort.

I don't want to do that: string and character literals (without semantic interpretation) are present in the concrete syntax of Kore, and I would like to reserve `InternalString` for domain values that have the _semantics_ of `STRING.String`.

PR #654, opened 2019-05-20
exists x. y=x simplification

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Unification/Substitution.hs
+++ b/kore/src/Kore/Unification/Substitution.hs
@@ -23,6 +23,7 @@ module Kore.Unification.Substitution
     , Kore.Unification.Substitution.filter
     , Kore.Unification.Substitution.freeVariables
     , partition
+    , reverseIfRhsIsVar
     ) where
 
 import           Control.DeepSeq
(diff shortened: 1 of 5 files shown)


kore/src/Kore/Unification/Substitution.hs
```diff
@@ -219,6 +226,98 @@ partition criterion (NormalizedSubstitution substitution) =
...
+      | var == substitutionVar = (substitutionVar, mkVar substitutedVar)
+      | otherwise = original
+    reversePairIfRhsVar _ original = original
+reverseIfRhsIsVar variable original@(NormalizedSubstitution substitution) =
+    case reversableVars of
+        [] -> original
+        (v:vs) ->
+            let
+                replacementVar :: variable
+                replacementVar = foldr max v vs
```
ttuegel: We should probably use `foldl'` here because the result is a summary of the input and it will be immediately used.

kore/test/Test/Kore/Step/Simplification/Exists.hs
```diff
@@ -37,12 +37,20 @@ import           Test.Tasty.HUnit.Extensions
 
 test_simplify :: [TestTree]
 test_simplify =
-    [ [plain10, plain11] `simplifies` [plain10', plain11']            $ "\\or distribution"
-    , [top]              `simplifies` [top]                           $ "\\top"
-    , []                 `simplifies` []                              $ "\\bottom"
-    , [equals]           `simplifies` [quantifyPredicate equals]      $ "\\equals"
-    , [substForX]        `simplifies` [top]                           $ "discharge substitution"
-    , [substOfX]         `simplifies` [quantifySubstitution substOfX] $ "substitution"
+    [ [plain10, plain11] `simplifiesTo` [plain10', plain11']
```
ttuegel: This is completely stylistic, but it might make sense to use `(&)` here instead of `($)`; then the cases would read like
```.hs
"\\or distribution"
    & [plain10, plain11] `simplifiesTo` [plain10', plain11']
```

virgil-serbanuta:
Oops, I'm sorry, for some reason I thought there are no comments so I merged. I'll fix them in a follow-up PR.

PR #657, opened 2019-05-21
Fix a \not simplifier bug

ttuegel:
This fixes a small bug in the \not simplifier. I can't seem to trigger this bug "in production," so I'm calling it "small".

In addition to adding a test and fixing the bug, I wrote the simplifier in a more type-ful way, so that I can make changes more confidently in the future.

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/Not.hs
+++ b/kore/src/Kore/Step/Simplification/Not.hs
@@ -18,6 +18,12 @@ import qualified Data.Foldable as Foldable
 import qualified Kore.Attribute.Symbol as Attribute
 import           Kore.IndexedModule.MetadataTools
                  ( SmtMetadataTools )
+import qualified Kore.Internal.Conditional as Conditional
+import           Kore.Internal.MultiAnd
+                 ( MultiAnd )
+import qualified Kore.Internal.MultiAnd as MultiAnd
+import           Kore.Internal.MultiOr
+                 ( MultiOr )
 import           Kore.Internal.OrPattern
                  ( OrPattern )
 import qualified Kore.Internal.OrPattern as OrPattern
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Step/Simplification/Not.hs
```diff
@@ -110,14 +116,14 @@ simplifyEvaluated
...
     simplified
-  | OrPattern.isFalse simplified = return (OrPattern.fromPatterns [Pattern.top])
-  | OrPattern.isTrue  simplified = return (OrPattern.fromPatterns [])
-  | otherwise =
-    fmap OrPattern.fromPatterns . gather
-    $ Foldable.foldrM mkAnd Pattern.top (simplified >>= makeEvaluate)
+  =
+    fmap OrPattern.fromPatterns $ gather $ do
+        let not' = Not { notChild = simplified, notSort = () }
+        andPattern <- scatterAnd (evaluateNotPattern <$> distributeNot not')
```
virgil-serbanuta: This is reasonable, but you seem to want to change the implicit conventions for the simplification code. To be more precise, usually the simplification code unwraps the various data structures, solves some of the issues that should be solving at that unwrapping level, and sends whatever is left to lower lever simplification stuff (e.g. `simplify` takes a `Not(MultiOr)`, sends the children to `simplifyEvaluated`, which does whatever is needed at the `MultiOr` level (i.e. `not(or)=and(not)`, sending the individual patterns inside the `or` to `makeEvaluate` to negate them, then `and`ing the results to produce the final thing).

Now, it seems that you prefer to work with unwrapped stuff. I think this is reasonable in itself, but I wanted to remind you that the unification code becomes less uniform this way. Up to you to fix or ignore.

ttuegel: I acknowledge that one downside of this is that the unification code becomes less uniform. However, I really messed up this simplifier the last time I worked on it, and I think the wrappers (or, the types of the wrappers) are a useful guide. Specifically, I messed up the order in which I distributed the `\and` over the `\or` and applied de Morgan's identity, and with the types of the wrappers in place, it becomes much more difficult to do that.

PR #664, opened 2019-05-23
Unification modulo equations proposal

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/docs/2019-05-23-Unification-Modulo-Equations.md
+++ b/docs/2019-05-23-Unification-Modulo-Equations.md
@@ -51,11 +51,19 @@ this is the most general unifier, i.e. there is no way of showing that we
 truly expanded the term in all possible ways. Second, this would be really
 inefficient.
 
+### Actual proposal
+
 The second attempt, however, would be to define some unification axioms (I
 think that they are, indeed, axioms, but they could be theorems if we find
 a way to infer them), specifying how to solve certain unification cases
-in away that can be used almost directly by the Haskell backend (some small
-changes needed). See the use-cases below for examples and explanations.
+in a way that can be used almost directly by the Haskell backend (some
+changes needed).
+
+These axioms would allow us to reduce the hard to handle
+unification cases to unambiguous ones (e.g. reduce AC unification to
+constructor-based unification).
+
+See the use-cases below for examples and explanations.
 
 We could write these in a way that we don't duplicate unification cases.
 Then we could apply these as simplification axioms, defining our unification
@@ -71,10 +79,10 @@ Use-case: lists
 
 For lists, we could have the following axioms
 ```
-⌈elem(x)∧unit⌉=⊥
-⌈elem(x)∧elem(y)⌉=⌈x∧y⌉
-⌈concat(x,y)∧unit⌉=⌈x∧unit⌉∧⌈y∧unit⌉
-⌈concat(elem(x),y)∧concat(elem(z),t)⌉=⌈x∧z⌉∧⌈y∧t⌉
+⌈elem(x)∧unit⌉ = ⊥
+⌈elem(x)∧elem(y)⌉ = ⌈x∧y⌉
+⌈concat(x,y)∧unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x),y) ∧ concat(elem(z),t)⌉ = ⌈x∧z⌉ ∧ ⌈y∧t⌉
 ```
 This is missing some axioms (e.g. the ones unifying at the end of the list,
 or splitting the list), but it looks like these, together with the usual list
@@ -91,12 +99,30 @@ with `concat(concat(unit, elem(ψ_1)), concat(elem(ψ_2), ψ_3))`. We first use
 the list axioms to rewrite the first term as
 `concat(elem(φ_1), concat(elem(φ_2), φ_3))`, the second one as
 `concat(elem(ψ_1), concat(elem(ψ_2), ψ_3))`, then we apply the
-`⌈concat(elem(x),y)∧concat(elem(z),t)⌉=⌈x∧z⌉∧⌈y∧t⌉` axiom repeatedly to compute
+`⌈concat(elem(x),y) ∧ concat(elem(z),t)⌉ = ⌈x∧z⌉ ∧ ⌈y∧t⌉` axiom repeatedly
+to compute
 ```
 ⌈ concat(elem(φ_1), concat(elem(φ_2), φ_3))
 ∧ concat(elem(ψ_1), concat(elem(ψ_2), ψ_3)) ⌉
-    = ⌈φ_1∧ψ_1⌉∧⌈concat(elem(φ_2), φ_3)∧concat(elem(ψ_2), ψ_3)⌉
-    = ⌈φ_1∧ψ_1⌉∧⌈φ_2∧ψ_2⌉∧⌈φ_3∧ψ_3⌉
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈concat(elem(φ_2), φ_3) ∧ concat(elem(ψ_2), ψ_3)⌉
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈φ_2∧ψ_2⌉ ∧ ⌈φ_3∧ψ_3⌉
+```
+
+Use-case: sets
+--------------
+
+Let us try to define the axioms for set lookup through unification, i.e.
+for unifying `concat(elem(x), y)` with a set. Assuming that our terms are
+normalized (elems moved to the left, concats represented as
+`concat(a, concat(b, concat(...)))`), then the following axioms should be
+enough:
+
+```
+⌈concat(x, y) ∧ unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ elem(a)⌉ = ⌈x∧a⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ concat(elem(a), b)⌉
+   = ⌈x∧a⌉ ∧ ⌈y∧b⌉
+     ∨ ∃ z . ⌈concat(elem(x),z)∧b⌉ ∧ ⌈y∧concat(elem(a),z)⌉
 ```
 
 Use case: heating and cooling
@@ -121,11 +147,13 @@ we need to unify `x + y ~> z` with `((1 + 2) + (3 + 4))`. While we could expand
 the expression in all possible ways and apply matching for the top `~>`, we
 could do this in a more principled way if we had the following axiom:
 ```
-⌈(a+b)~>c ∧ (d+e)~>f⌉ =
-    (⌈a∧d⌉ ∧ ⌈b∧e⌉ ∧ ⌈c∧f⌉)
-    ∨ (⌈(a+b)~>g ∧ d⌉ ∧ ⌈c ∧ (g~>[]+e~>f)⌉)
-    ∨ (⌈(a+b)~>g ∧ e⌉ ∧ ⌈c ∧ (g~>d+[]~>f)⌉)
+⌈x~>y ∧ (a+b)~>c⌉ =
+    (⌈x ∧ (a+b)⌉ ∧ ⌈y∧c⌉)
+    ∨ ∃ d . (⌈x~>d ∧ a⌉ ∧ ⌈y ∧ (d~>[]+b~>c)⌉)
+    ∨ ∃ d . (⌈x~>d ∧ e⌉ ∧ ⌈y ∧ (d~>a+[]~>c)⌉)
+  if isKItem(x)
 ```
+
 This would give us a clear way of doing unification, in which we'd be sure that
 we're not missing any terms when expanding, and which would make it easy to
 prove that we're doing the right thing.
--- a/docs/2019-05-23-Unification-Modulo-Equations.md
+++ b/docs/2019-05-23-Unification-Modulo-Equations.md
@@ -49,7 +49,7 @@ the same sort, then use unification for the remainders.
 However, this has a few problems. First, there is no way of proving that
 this is the most general unifier, i.e. there is no way of showing that we
 truly expanded the term in all possible ways. Second, this would be really
-inefficient.
+inefficient. Also, the no-confusion axioms might be hard to write.
 
 ### Actual proposal
 
@@ -112,11 +112,13 @@ to compute
     = ⌈φ₁∧ψ₁⌉ ∧ ⌈φ₂∧ψ₂⌉ ∧ ⌈φ₃∧ψ₃⌉
 ```
 
-Use-case: sets
---------------
+Use-case: multi-sets
+--------------------
 
-Let us try to define the axioms for set lookup through unification, i.e.
-for unifying `concat(elem(x), s)` with a set. Assuming that our terms are
+These are associative, commutative, with neutral element.
+
+Let us try to define the axioms for multi-set lookup through unification, i.e.
+for unifying `concat(elem(x), s)` with a multi-set. Assuming that our terms are
 normalized (elems moved to the left, concats represented as
 `concat(a, concat(b, concat(...)))`), then the following axioms should be
 enough:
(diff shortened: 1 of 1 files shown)


docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,159 @@
...
+used more frequently in Haskell.
+
+Or, if that ever becomes desirable (e.g. if we want it to make proof
+certificates easier to generate), we could give up on performance and define
+all simplification cases as axioms/theorems.
+
+Use-case: lists
+---------------
+
+For lists, we could have the following axioms
```
ttuegel: I like this proposal! I have a lot of questions about how we should implement it. Maybe these are premature, but I will ask them anyway. How would we apply these axioms? By matching on `⌈_∧_⌉`? Should we use Kore's `\in` connective instead, so we do not just apply any random axiom which has a similar form? Should we have a `unification` attribute? Are the variables in the equations element or set variables? It seems to me like we could have them be set variables to avoid generating lots of extra `\ceil`s.

virgil-serbanuta: Here are my preferences: apply them by matching, but we should normalize the terms before; use an `unification` attribute.

Yes, we should certainly use set variables! That also came up during a discussion with @traiansf . I've updated the document to reflect that.

grosu:
There is absolutely no doubt that this is the right way to go!  Why have we even remotely considered doing things any other way?  :)  I can only think of bringing it up again that you guys should try to stay in synch with the papers describing the underlying logic of K.  In this case, the LICS'19 paper, or the TR:http://fsl.cs.illinois.edu/FSL/papers/2019/chen-rosu-2019-tr/chen-rosu-2019-tr-public.pdf.  Please check Proposition 22.  Besides these no-junk and no-confusion axioms you mentioned, @virgil-serbanuta , we will also need axioms using `mu`, stating the actual lfp nature of the constructors.  That is, no-junk no-confusion are not even going to be sufficient for all the proofs we'll need.

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+need to make sure that the axioms still hold).
+
+Use-case: lists
+---------------
+
+For lists, we could have the following axioms
+```
+⌈elem(x)∧unit⌉ = ⊥
+⌈elem(x)∧elem(y)⌉ = ⌈x∧y⌉
+⌈concat(x,y)∧unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
```
yzhang90: `⌈concat(x,y)∧unit⌉` can directly equal to `⊥` right?

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+for that sort (e.g. any list can be built with `unit`, `elem` and `concat`),
+and then expand the two terms in all possible ways using the available
+equations, then match the two terms at the top as long as we stay within
+the same sort, then use unification for the remainders.
+
+However, this has a few problems. First, there is no way of proving that
+this is the most general unifier, i.e. there is no way of showing that we
+truly expanded the term in all possible ways. Second, this would be really
+inefficient.
+
```
xc93: Another problem is that the (No Confusion) axioms for symbols with underlying axioms might take complex forms (e.g., conder the (No Confusion) axioms for `concat`).

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: Not if x and y are unit.

xc93: No. One can take `x=unit` and `y=unit`. Suggestion: use `x,y,z,...` for variables of `Nat` and `l1,l2,...` for variables of `List`.

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+Use-case: lists
+---------------
+
+For lists, we could have the following axioms
+```
+⌈elem(x)∧unit⌉ = ⊥
+⌈elem(x)∧elem(y)⌉ = ⌈x∧y⌉
+⌈concat(x,y)∧unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x),y) ∧ concat(elem(z),t)⌉ = ⌈x∧z⌉ ∧ ⌈y∧t⌉
```
xc93: How would you write the corresponding axiom for set `union`, which is similar to `concat` but is assoc/comm?

docs/2019-05-23-Unification-Modulo-Equations.md
yzhang90: yes...you're right. empty list is not bottom

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+unification.
+
+Note that applying these axioms allows us to reduce complex expressions to
+simpler ones, without worrying about how many terms are equivalent to the
+one we have: we just have to bring the term in one of the forms required for
+one of these axioms, then we apply the axiom and that’s it.
+
+As an example, say we need to unify `concat(concat(elem(φ_1), elem(φ_2)), φ_3)`
+with `concat(concat(unit, elem(ψ_1)), concat(elem(ψ_2), ψ_3))`. We first use
+the list axioms to rewrite the first term as
```
xc93: The "list axioms" here, I guess, refer to the assoc axiom of `concat`. Such an axiom can be applied in both directions. How do we know which direction should we use? In general, the approach here looks like the one that uses theory axioms (e.g., assoc, comm, etc.) to transform terms to some sort of canonical forms, and then use unification axioms (after orientation) as rewriting rules to solve the unification problem. Indeed, the unification rules seem to require that both terms have `concat` associative to the right. 
If my above understanding is correct, then the problem is that (almost) all theory axioms are not simplification rules, and they do not yield canonical forms in general because they can be applied in both directions, and no one direction is more preferable than the other.

grosu:
For the record, please also take a look at Section 5.7 in the original ML paper, to see how unification = proof using such axioms: http://fsl.cs.illinois.edu/FSL/papers/2017/rosu-2017-lmcs/rosu-2017-lmcs-public.pdf.  The claim there is that unification algorithms, like Montanari's, can be turned into proof searchers.  When we generate proof objects later on, this will be unavoidable.

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈concat(elem(φ_2), φ_3) ∧ concat(elem(ψ_2), ψ_3)⌉
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈φ_2∧ψ_2⌉ ∧ ⌈φ_3∧ψ_3⌉
+```
+
+Use-case: sets
+--------------
+
+Let us try to define the axioms for set lookup through unification, i.e.
+for unifying `concat(elem(x), y)` with a set. Assuming that our terms are
+normalized (elems moved to the left, concats represented as
```
xc93: hmm... I have a feeling that assuming that terms are all normalized is a very strong assumption. In my understanding, (which could be very wrong, so please let me know if I am wrong), the reason why people study unification modulo assoc/comm is because of the mind setting of a high-level separation of concern. Given a set of equations (that form an equational theory), we divide all equations into _simplification equations_ and _background/theory equations_. For simplification equations, we oriente them as simplification rules and use them to compute the canonical form of a term. For background/theory equtions, such as assoc/comm/identity, we cannot regard them as simplification rules. Instead, we develop special matching/unification algorithm for these background theories.

xc93: So, if the assumption that all terms are already in some sort of normalized form too strong?

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+`concat(a, concat(b, concat(...)))`), then the following axioms should be
+enough:
+
+```
+⌈concat(x, y) ∧ unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ elem(a)⌉ = ⌈x∧a⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ concat(elem(a), b)⌉
+   = ⌈x∧a⌉ ∧ ⌈y∧b⌉
+     ∨ ∃ z . ⌈concat(elem(x),z)∧b⌉ ∧ ⌈y∧concat(elem(a),z)⌉
```
xc93: I can imagin why line127 -> line128 \/ line129, but the other direction seems more involved.

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+`⌈concat(elem(x),y) ∧ concat(elem(z),t)⌉ = ⌈x∧z⌉ ∧ ⌈y∧t⌉` axiom repeatedly
+to compute
+```
+⌈ concat(elem(φ_1), concat(elem(φ_2), φ_3))
+∧ concat(elem(ψ_1), concat(elem(ψ_2), ψ_3)) ⌉
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈concat(elem(φ_2), φ_3) ∧ concat(elem(ψ_2), ψ_3)⌉
+    = ⌈φ_1∧ψ_1⌉ ∧ ⌈φ_2∧ψ_2⌉ ∧ ⌈φ_3∧ψ_3⌉
+```
+
+Use-case: sets
```
xc93: sets => multi-sets

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+`concat(a, concat(b, concat(...)))`), then the following axioms should be
+enough:
+
+```
+⌈concat(x, y) ∧ unit⌉ = ⌈x∧unit⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ elem(a)⌉ = ⌈x∧a⌉ ∧ ⌈y∧unit⌉
+⌈concat(elem(x), y) ∧ concat(elem(a), b)⌉
+   = ⌈x∧a⌉ ∧ ⌈y∧b⌉
+     ∨ ∃ z . ⌈concat(elem(x),z)∧b⌉ ∧ ⌈y∧concat(elem(a),z)⌉
+```
```
xc93: How about sets (not multi-sets) where `concat(x,x) = x`? What would the last axiom look like?

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: You can find it in the next "use-case" section (although I called it `concat` instead of `union`).

The trick is that we assume these to be normalized by the Haskell backend, so if the term is initially `concat(concat(elem(x), z), elem(y))`, it will be transformed to `concat(elem(x), concat(elem(y), z))`.

I'll also try to answer some of the other questions here.

First, for the usual data structures (list, set, map), we can always implement this norrmalization in the backend. That does not allow other people to define their own operations, with their own normalization rules, but it's still a step forward.

Second, if we want to be able to define normalization axioms/theorems, we probably need some meta-level conditions (being able to add `x does not match thing` where thing can be something like `elem(_)` would probably be good enough; alternately, `x`'s head is/is not `elem` would also work). But, if we have those, even if we keep the AC axioms separate, we can define simplification theorems, which we can use for this purpose (the theorem can restate the axiom with the orientation needed for applying it and perhaps with an additional side condition to make sure we're applying it on the right things).

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: I hope the answer above covers this. Please tell me if it doesn't.

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: I hope the answer above covers this. Please tell me if it doesn't.

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+to something like
+```
+((1 + 2) + (3 + 4))
+```
+we need to unify `x + y ~> z` with `((1 + 2) + (3 + 4))`. While we could expand
+the expression in all possible ways and apply matching for the top `~>`, we
+could do this in a more principled way if we had the following axiom:
+```
+⌈x~>y ∧ (a+b)⌉ =
+    (⌈x ∧ (a+b)⌉ ∧ ⌈y ∧ unit⌉)
```
xc93: How about the case when `x = unit`?

docs/2019-05-23-Unification-Modulo-Equations.md
```diff
@@ -0,0 +1,163 @@
...
+    ∨ ∃ d . (⌈x~>d ∧ a⌉ ∧ ⌈y ∧ (d~>[]+b)⌉)
```
xc93: How to solve `⌈y ∧ (d~>[]+b)⌉`?

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: So, I thought that, for maps, `concat(elem(a, b), elem(a, ?))` is bottom. I assumed that, for sets, it's the same. Now I realize that while `concat(elem(a, b), elem(a, c))` is probably bottom, `concat(elem(a, b), elem(a, b))` might actually be `elem(a, b)`. So, just to make sure, how is this supposed to work?

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: Thanks, I replaced the list variables.

docs/2019-05-23-Unification-Modulo-Equations.md
xc93: I don't know actually :) That's why I ask. The axiom in line 127 looks very interesting, and it takes me quite a while to digest its intuition.

docs/2019-05-23-Unification-Modulo-Equations.md
xc93: " for the usual data structures (list, set, map), we can always implement this norrmalization in the backend. That does not allow other people to define their own operations, with their own normalization rules, but it's still a step forward". 

It is OK. I see. Thanks for explaining it. 

So here's my understanding about this propsal and why we want it:

This proposal extracts some part of the unification algorithms for lists/sets/maps to K's rewriting engine. This means that, we can assume *less* about the backends. We don't need all backends to properly implement all common unification algorithms for the common background theories. Instead, we just assume that backends have a reasonably powerful rewriting enging, some sort of ML reasoning power, and have an agreement on how to normalize terms. Given these, the rest of the unification algorithms are actually implemented in K, so that they are backend-independent.

Does it sound right to you? I like this idea, too.

docs/2019-05-23-Unification-Modulo-Equations.md
ttuegel: K's `Set` has the property that `concat(X, X) = \bottom` so that `concat` is commutative if its arguments are distinct, but share keys.

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: I added a proof for RHS -> LHS. I'm not sure yet how the other implication works, but seems doable, I'll try it on Monday.

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: Done.

docs/2019-05-23-Unification-Modulo-Equations.md
virgil-serbanuta: Added a section for sets.

PR #669, opened 2019-05-24
IMP: Clean-up and extra specifications

ttuegel:
This is some clean-up and an extra specification I wrote for my demo.

###### Reviewer checklist

---

### DIFF
--- a/src/main/k/working/imp/prove/sum-div-spec.k
+++ b/src/main/k/working/imp/prove/sum-div-spec.k
@@ -15,38 +15,36 @@ module SUM-DIV-SPEC
   imports VERIFICATION
 
 
-rule
+  rule
     <k>
       while (!(n <= 0)) {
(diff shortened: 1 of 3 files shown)


src/main/k/working/imp/prove/sum-div-spec.k
```diff
@@ -15,38 +15,36 @@ module SUM-DIV-SPEC
...
     </k>
-  <state>
-    .Map
-  =>
-      n   |-> 0
-      sum |-> (((N +Int 1) *Int N) /Int 2)
-  </state>
-requires N >=Int 0
+    <state>
+      .Map =>
```
virgil-serbanuta: You can leave it as-is, but it seems harder to spot the arrow if it's on the same line as `.Map`. I'm not sure about others, but that's usually one of the first things I do when looking at a spec file.

PR #675, opened 2019-05-28
repl: trusted claims bug fixes

ana-pantilie:
###### Reviewer checklist

---


kore/src/Kore/Repl.hs
```diff
@@ -200,7 +203,9 @@ runRepl
     makeRuleIndex n _ = RuleIndex (Just n)
 
     firstClaim :: Claim claim => claim
-    firstClaim = maybe (error "No claims found") id $ listToMaybe claims'
+    firstClaim =
+        maybe (error "No claims found") id
```
vladciobanu: I would rather use `(!!)` with the index inside `firstClaimIndex` to avoid duplication.

PR #680, opened 2019-05-29
Overloaded productions

ttuegel:
###### Reviewer checklist

---


ttuegel:
This example is actually much worse than the example with sort injections: we have three constructors in a single sort which do not obey the "no confusion" axiom between different constructors, due to the overloading axioms.

ttuegel:
I'm reviewing the Kore definition for this. It looks like the top-most symbol gets the attribute `anywhere` to indicate that it came from a K constructor. So, the first step should be to recognize that `anywhere` symbols are injective. We should also recognize that overloaded `anywhere` symbols are `\bottom` *except* where they have a defined overloading constructor; the frontend should generate such an axiom or axioms, but for now we could write the logic in the backend.

virgil-serbanuta:
Is it really true that `anywhere` symbols are injective? I can't find an intuitive link between the name and the property. @traiansf , with whom I previously heard talking about `anywhere`, though it might have been in a completely different context.

ttuegel:
> Is it really true that `anywhere` symbols are injective?

Yes, but maybe we should give them a different attribute (`injective`?) in the frontend.

The overloaded symbols are considered [constructors in K, but not in Kore](https://github.com/kframework/k/blob/513d15d125b47ef2be32cbbe579d93743359a2aa/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L874). Therefore, they are given "no confusion, same constructor" [axioms](https://github.com/kframework/k/blob/513d15d125b47ef2be32cbbe579d93743359a2aa/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L355), in other words: injectivity axioms.

virgil-serbanuta:
Ok. There is something weird about this, I would just add the `injective` attribute (which we can also add for normal constructors, it can't hurt, and it will probably simplify the attribute inference we do when loading kore definitions), but I guess that we can also use `anywhere` if we document it.

ttuegel:
See also: kframework/k#513

ttuegel:
I am closing this pull request because supporting KWASM is not a priority right now.

Issue #677, opened 2019-05-28
Multiple uses of existential/question-mark variables in function definitions do not work as expected 


nishantjr:
Minimal example below:

```
$ cat foo.k
module FOO
  imports INT
  imports BOOL
  
  syntax Seq ::= List{Int, ":"}
  syntax Seq ::= "foo" "(" Int ")" [function]
  rule foo(N) => ?_:Int : foo(N -Int 1)
    requires N >Int 0
  rule foo(0) => .Seq

  syntax X ::= "foo"
  rule <k> foo => foo(3) </k>
endmodule

$ ./ext/k/k-distribution/target/release/k/bin/kompile foo.k --backend haskell
[Warning] Compiler: Could not find main syntax module with name FOO-SYNTAX in
definition.  Use --syntax-module to specify one. Using FOO as default.

$ cat foo
foo

$ ./ext/k/k-distribution/target/release/k/bin/krun foo
Reading the input file TimeSpec {sec = 0, nsec = 84834}
Parsing the file TimeSpec {sec = 0, nsec = 45296501}
Verifying the definition TimeSpec {sec = 0, nsec = 25025880}
Reading the input file TimeSpec {sec = 0, nsec = 31449}
Parsing the file TimeSpec {sec = 0, nsec = 121358}
Verifying the pattern TimeSpec {sec = 0, nsec = 126137}
Executing TimeSpec {sec = 0, nsec = 27401608}
<k>
  'Unds'12 : 'Unds'12 : 'Unds'12 : .Seq
</k>
```

Expected output is 

```
<k>
  'Unds'12 : 'Unds'13 : 'Unds'14 : .Seq
</k>
```

virgil-serbanuta:
Summary: at first sight, it seems to be both a frontend issue (universal quantification instead of existential) and a backend issue (i.e. not renaming conflicting variables). I think that we implicitly assumed existential quantification for new variables in rules (assumption that should go away at some point, I guess), so the main problem should be in the backend.

The `rule foo(N) => ?_:Int : foo(N -Int 1)` axiom gets translated to
```
axiom{R} \implies{R} (
    \equals{SortBool{},R}(
        Lbl'Unds-GT-'Int'UndsUnds'INT-COMMON'UndsUnds'Int'Unds'Int{}(
            VarN:SortInt{},
            \dv{SortInt{}}("0")
        ),
        \dv{SortBool{}}("true")),
    \and{R} (
        \equals{SortSeq{},R} (
[...]

PR #684, opened 2019-05-30
Existential quantification for new step variables

virgil-serbanuta:
This should fix the ML issues in #677 , but some further work is needed to make new variables work properly (i.e. lifting quantifiers to the top).

###### Reviewer checklist

---


kore/src/Kore/Step/Simplification/Pattern.hs
```diff
@@ -38,6 +41,42 @@ import           Kore.Syntax.Variable
...
+    patt
+  = do
+    simplified <-
+        simplify
+            tools
+            substitutionSimplifier
+            termSimplifier
+            axiomIdToSimplifier
+            patt
+    return (removeTopExists <$> simplified)
```
ttuegel: Should we remove the `\exists` before or after simplification? `\exists` simplification can leave predicates in the term, so could they be left in the term here?

kore/src/Kore/Step/Step.hs
```diff
@@ -139,16 +140,45 @@ unwrapRule = Rule.mapVariables Target.unwrapVariable
...
-    Pattern.mapVariables Target.unwrapVariable
-        config { Pattern.substitution = substitution' }
+unwrapAndQuantifyConfiguration config@Conditional { substitution } =
+    if List.null targetVariables
+    then unwrappedConfig
+    else
+        Pattern.fromTermLike
+            (List.foldl'
+                quantify
+                (Pattern.toTermLike unwrappedConfig)
```
ttuegel: Doesn't this discard the predicate and substitution?

kore/src/Kore/Step/Simplification/Pattern.hs
virgil-serbanuta: I asked myself the same question. I'm not sure, I think that both are somewhat broken:

* If we want to remove the existentials for rewrite rules that we just added in this PR, then we can remove them both at the beginning and at the end, and removing them at the beginning works better, since we keep the term-predicate-substitution split.

* If we want to remove other existentials, then we need to lift them. In principle we should be able to lift and remove them before simplification, but we're not using quantifiers in a consistent way (yet), so we're going to break things. Here's the issue:

In symbolic execution we're usually trying to prove a `phi->psi` formula. The symbolic variables are quantified universally, i.e. `forall x . (phi -> psi)`, and we don't have explicit quantifiers for them. The new variables are quantified existentially in `phi`, i.e. `(exists y . phi)->psi`, which can also be written as `forall y . (phi -> psi)`, which means that they are similar enough to the symbolic variables to remove quantifiers. But the problem is that, sometimes, we prefer to have them as existential quantifiers, since `exists x . (phi and [x=alpha])`, where `x` does not occur in `alpha`, can be transformed into a formula that does not contain `x` at all, and we use that for `owise` clauses. But if we remove the quantifier, i.e. we transform the variable in a symbolic one, we'll never be able to fully get rid of it.

So, maybe a better solution would have been to make unification work with quantifiers (at least with the top ones) and to not remove them, but that would have made everything somewhat more complicated. So, short-term, removing exists before simplification is better, since currently we don't lift quantifiers. Medium term, removing it after simplification is better. Long-term, there's a good chance that both are bad. I just picked the medium term solution.

kore/src/Kore/Step/Step.hs
virgil-serbanuta: No, it merges them into the term.

PR #685, opened 2019-05-30
Anywhere symbol attribute

ttuegel:
###### Reviewer checklist

---


kore/src/Kore/Attribute/Symbol/Anywhere.hs
```diff
@@ -0,0 +1,56 @@
+{- |
```
virgil-serbanuta: Could you add this to `attributes.md`? I don't know what the `anywhere` attribute means, and, from its name, I can't figure out why it would have anything to do with injectivity.

ttuegel:
@virgil-serbanuta You're right, actually: the overloaded constructors are marked `anywhere` because the overloading axioms are considered "anywhere" rules. The frontend should be marking the overloaded symbols injective; the proper fix is there.

ttuegel:
@virgil-serbanuta This is ready for re-review. I'm keeping the `anywhere` attribute, but it no longer implies `injective`. We won't use it for anything yet, but I'd hate to throw this away.

PR #707, opened 2019-06-06
Separate internal representation of symbols and aliases during execution

ttuegel:
We can reduce the use of `MetadataTools` by giving distinct internal representations to symbols and aliases. The representation for symbols will also cache the symbol attributes. Ultimately this will make it easier to keep the pattern attributes up-to-date.

###### Reviewer checklist

---


ttuegel:
### TODO

- [ ] Extract `Symbol` helpers from `Test.Kore.Unification.Unifier`

ttuegel:
The unit tests pass, but the integration tests will still fail because we are bypassing the `constructorFunctions` special cases, particularly for the `kseq` and `dotk` symbols. I have [inquired](https://runtimeverification.slack.com/archives/C7E701MFG/p1560085206113200) about giving those symbols the `constructor` attribute.

PR #733, opened 2019-06-12
evaluateWithDefinitionAxioms: Use matching instead of unification

ttuegel:
Using matching instead of unification tends to postpone function application
until there are fewer branching points, which simplifies execution. Before, we
would use unification to apply the rule, but we used a heuristic to reject
common symbolic rewriting solutions. However, there are several (extremely)
common cases which cannot be detected this way. A more robust solution is to use
matching instead of unification in the first place.

@yzhang90 Does this fix the problem you experienced with the expansion of `andBool` rules?

###### Reviewer checklist

---


yzhang90:
it does solve the `andBool` issue

ttuegel:
This is waiting for #740.

The only remaining test failures are due to not having implemented matching on builtins, particularly maps.

@virgil-serbanuta There is a note in `Kore.Step.Axiom.Matcher` about returning `Left` versus `\bottom`. I think we should return `\bottom` whenever we have a known case of two patterns that do not match, and `Left` only when we have a case we do not recognize. There was a reason that matching did not originally do that, but I'm not sure I understand why.

virgil-serbanuta:
First, I agree that right now we can use `bottom` instead of `Left` and have no bad consequences, and, probably, we'll need less changes when using matching instead of unification. However, I'm not sure that this is the right thing long term. The main problem is that `\bottom` does not have a well defined meaning here, and, at least for some attempts to give it a well defined-meaning, we should keep using `Left`.

For unification things are clear: The unifier of `phi` and `psi` is `phi and ceil(phi and psi)`, so `\bottom` means that the `ceil` expression is bottom.

Matching is a weirder thing. It might mean "The structure of these terms is identical", and we started with that. If we take it to mean that and nothing else, then maybe it can be argued that using bottom instead of left makes sense (or maybe not, I'm not sure).

However, we probably want to bend that rule (e.g. by using "ac matching" or sometimes falling back to equality when we can't solve something). It's not clear how much we want to bend it (having an "identical structure"-only matching certainly has its appeal), or how many matching algorithms we want to have at the same time in our code, but we can see matching as something on a continuum between "identical structure" and "full unification". This means that using bottom as "These things can never match, regardless of how we define matching" and using "Left" as "maybe it could work, but we don't know right now" makes sense.

However, I can't see how we can give a ML reason for using `bottom` when we don't handle something. If that's indeed the case and it's not just me, then it seems like a hack to make our code easier to write, which raises the question: Why don't we check for `Left` results when applying function axioms (or anything else that would use matching) and treat that as we would treat bottom (i.e. just skip the axiom)?

ttuegel:
> Why don't we check for `Left` results when applying function axioms (or anything else that would use matching) and treat that as we would treat bottom (i.e. just skip the axiom)?

When applying multiple rules using the machinery in `Kore.Step.Step`, any `Left` result from unification or matching aborts all the other rules. That makes sense when doing unification: if we encounter one unknown case, we cannot continue. But matching uses `Left` for "unknown" and "doesn't match". So, when a function is defined by several axioms, function application by matching only works if the first rule matches, and it will never match subsequent rules. If it doesn't make sense to use `\bottom` for "doesn't match" then we must give up on re-using `Kore.Step.Step` between unification and matching.

ttuegel:
I suppose what we could do to keep the matching algorithm the same (i.e. continue to return `Left` everywhere) is to have the function evaluator wrap the matching function to `catch` the `Left` and convert it to `\bottom`. I agree that maybe `\bottom` doesn't make sense in every context, but it does make sense _in the context of function rule application_.

virgil-serbanuta:
Yep, I agree that we should change the `Kore.Step.Step` behaviour, but why not add a new `applyRulesSequence` function that treats `Left` as `bottom`?

ttuegel:
> Yep, I agree that we should change the `Kore.Step.Step` behaviour, but why not add a new `applyRulesSequence` function that treats `Left` as `bottom`?

The function that actually needs to be changed is `Kore.Step.Step.unifyRule`, so we would need to have a second version of everything above that function in the call graph, not only `applyRulesSequence`. But, we can achieve the desired result by wrapping the `UnificationProcedure` and using all the same machinery in `Kore.Step.Step` instead.

virgil-serbanuta:
Ah, I see, so you'd also need a duplicate `unifyRules` function that does the error->bottom transformation. (`unifyRule` can stay the same, I guess)...

Wrapping the unifier/matcher seems to solve most of my objections, so, I guess, it's probably better than duplicating two functions.

vladciobanu:
I tried merging master into this and there are two `map` tests that are failing. One needed removal of `mkEvaluated` (not sure if that's correct) to pass (the test *map function, non-matching* from `Simplification/Integration`).

The other doesn't seem to match things properly (*map function, matching* from the same file). I think that in order to get the second test working we need #770 to be merged.

PR #741, opened 2019-06-18
Add stack lock file to repo

vladciobanu:
It seems the `stack.yaml.lock` file should not actually be ignored (according to https://docs.haskellstack.org/en/stable/lock_files/#lock-files )

###### Reviewer checklist

---


ttuegel:
Do you know why Git thinks the `stack.yaml.lock` file is dirty?

ttuegel:
I think the lock file was coming up dirty because there was an unlocked dependency, which Stack dutifully locked on Jenkins, modifying the file in the process. I ran `stack freeze` locally to lock everything, and it seems stable.

Issue #700, opened 2019-06-06
Unification with evaluated remainders


ttuegel:
We do not currently do unification (even against variables) of evaluated function remainders.

### Instructions

```
$ git clone https://github.com/kframework/k

$ cd k; mvn package -DskipTests -Dllvm.backend.skip -Dhaskell.backend.skip

$ cd k-distribution/tests/regression-new

$ make KOMPILE_BACKEND=haskell --keep-going -C spec-rule-application def11-spec.k
make: Entering directory '/home/ttuegel/k/k-distribution/tests/regression-new/spec-rule-application'
/home/ttuegel/k/k-distribution/bin/kompile --syntax-module DEF --backend haskell  def.k -d .
/home/ttuegel/k/k-distribution/bin/kprove def11-spec.k   -d . | diff - def11-spec.k.out
Reading the input file TimeSpec {sec = 0, nsec = 42049}
Parsing the file TimeSpec {sec = 0, nsec = 41585080}
Verifying the definition TimeSpec {sec = 0, nsec = 27363519}
Reading the input file TimeSpec {sec = 0, nsec = 36479}
Parsing the file TimeSpec {sec = 0, nsec = 680192}
Verifying the definition TimeSpec {sec = 0, nsec = 18872930}
kore-exec: Not implemented error:
    Unsupported patterns: Unknown unification case.
    first=Lbl'Unds'-Int'UndsUnds'INT-COMMON'UndsUnds'Int'Unds'Int{}(
        /* evaluated: */
        Lbl'UndsPlus'Int'Unds'{}(VarY:SortInt{}, \dv{SortInt{}}("1")),
        \dv{SortInt{}}("1")
    )
    second=VarX:SortInt{}
    
while applying a \rewrite axiom to the pattern:
    \and{_}(
        /* term: */
        Lbl'-LT-'generatedTop'-GT-'{}(
            Lbl'-LT-'T'-GT-'{}(
                Lbl'-LT-'k'-GT-'{}(
                    kseq{}(
                        inj{SortTerm{}, SortKItem{}}(
                            Lblmid'UndsUnds'DEF'UndsUnds'Int{}(VarX:SortInt{})
                        ),
                        dotk{}()
                    )
                ),
                Lbl'-LT-'var'-GT-'{}(Var'Unds'0:SortTerm{})
            ),
            VarDotVar0:SortGeneratedCounterCell{}
        ),
        \and{_}(
            /* predicate: */
[...]

PR #743, opened 2019-06-18
Synthetic pattern attributes

ttuegel:
Fixes: #700 

#### TODO

- [x] Test unifying `x:S{} ~ /* evaluated: */ φ`
- [x] Test matching `x:S{} ~ /* evaluated: */ φ`

###### Reviewer checklist

---


kore/src/Kore/Attribute/Synthetic.hs
```diff
@@ -6,45 +6,104 @@ License     : NCSA
...
+    Recursive.fold worker
   where
-    synthesize1 (a :< ft) =
-        Recursive.embed (synthetic (a :< fb) :< ft)
-      where
-        fb = Cofree.headF . Recursive.project <$> ft
+    worker (_ :< ft) = synthesizeAux synth ft
+{-# INLINE resynthesizeAux #-}
 
+{- | @/synthesize/@ an attribute @a@ bottom-up along a tree @s@.
```
virgil-serbanuta: Is this really bottom-up, or does it do stuff only at the top?

kore/src/Kore/Step/Rule.hs
```diff
@@ -5,13 +5,18 @@ License     : NCSA
 
 -}
 
+{-# LANGUAGE TemplateHaskell #-}
```
virgil-serbanuta: We are adding this to more and more files. Is it true that it increases compilation time by a lot?

kore/src/Kore/Syntax/And.hs
```diff
@@ -73,3 +77,14 @@ instance Unparse child => Unparse (And Sort child) where
...
             ])
+
+instance Ord variable => Synthetic (And sort) (FreeVariables variable) where
+    synthetic = Foldable.fold
+    {-# INLINE synthetic #-}
+
+instance Synthetic (And Sort) Sort where
+    synthetic And { andSort, andFirst, andSecond } =
+        andSort
+        & seq (matchSort andSort andFirst)
```
virgil-serbanuta: So, just to make sure: these `matchSort` calls are only used as sanity checks, right?

kore/test/Test/Kore/Attribute/Pattern/Defined.hs
```diff
@@ -0,0 +1,87 @@
...
+    , testGroup "ForallF" $ map (isn't . ForallF) (Forall sort Mock.x <$> range)
+    , testGroup "VariableF" [ is $ VariableF Mock.x ]
+    , testGroup "MuF" $ map (isn't . MuF) (Mu (SetVariable Mock.x) <$> range)
+    , testGroup "NuF" $ map (isn't . NuF) (Nu (SetVariable Mock.x) <$> range)
+    , testGroup "SetVariableF" [ isn't $ SetVariableF (SetVariable Mock.x) ]
+    -- Interesting cases
+    , testGroup "ApplySymbolF" $ do
+        x <- range
+        y <- range
+        [ expect (x <> y) $ ApplySymbolF $ Application sigma [x, y] ]
```
virgil-serbanuta: A test for a non-total symbol would be nice.

kore/test/Test/Kore/Attribute/Pattern/Function.hs
```diff
@@ -0,0 +1,83 @@
...
+
+import qualified Test.Kore.Step.MockSymbols as Mock
+
+test_instance_Synthetic :: [TestTree]
+test_instance_Synthetic =
+    [ testGroup "AndF" $ map (isn't . AndF) (And sort <$> range <*> range)
+    , testGroup "ApplySymbolF" $ do
+        x <- range
+        y <- range
+        [ expect (x <> y) $ ApplySymbolF $ Application sigma [x, y] ]
```
virgil-serbanuta: A test with a non-function symbol would be nice.

kore/test/Test/Kore/Attribute/Pattern/Functional.hs
```diff
@@ -0,0 +1,83 @@
...
+
+import qualified Test.Kore.Step.MockSymbols as Mock
+
+test_instance_Synthetic :: [TestTree]
+test_instance_Synthetic =
+    [ testGroup "AndF" $ map (isn't . AndF) (And sort <$> range <*> range)
+    , testGroup "ApplySymbolF" $ do
+        x <- range
+        y <- range
+        [ expect (x <> y) $ ApplySymbolF $ Application sigma [x, y] ]
```
virgil-serbanuta: A test with a non-functional symbol would be nice.

kore/src/Kore/Attribute/Synthetic.hs
```diff
@@ -6,45 +6,104 @@ License     : NCSA
...
+with attributes @inh@ along the nodes, @synthesize@ produces a tree @t@ with
+attributes @syn@ along the nodes using the given @('Base' s)@-algebra from the
+bottom up.
 
 See also:
 <https://en.wikipedia.org/wiki/Attribute_grammar#Synthesized_attributes>
 
  -}
-synthesize
+resynthesize
```
virgil-serbanuta: One more thing: Does this synthesize stuff work with things that contain concrete stuff, e.g. set builtins?

kore/src/Kore/Attribute/Synthetic.hs
ttuegel: Ah, you're right, it only does things at the top. I think I copied this from `resynthesize`, which calls `synthesize` at every level.

kore/src/Kore/Step/Rule.hs
ttuegel: Yes, significantly. It's not necessary for using lenses, I'm just being lazy. I will think about how we should handle this in general.

kore/src/Kore/Syntax/And.hs
ttuegel: That's right, it just throws an exception if the sorts don't match. Ordinarily that can't happen, but it might when we are constructing patterns in the code.

kore/src/Kore/Attribute/Synthetic.hs
ttuegel: Right now, the instance for builtins ignores the concrete stuff, but we can fix that. I didn't think about it because the concrete parts don't usually change during execution.

PR #745, opened 2019-06-19
Remove config environment from ReplState

vladciobanu:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Repl.hs
+++ b/kore/src/Kore/Repl.hs
@@ -21,6 +21,10 @@ import           Control.Monad.Catch
                  ( MonadCatch, catch )
 import           Control.Monad.IO.Class
                  ( MonadIO, liftIO )
+import           Control.Monad.Reader
+                 ( ReaderT (..) )
+import           Control.Monad.RWS.Strict
+                 ( RWST, execRWST )
 import           Control.Monad.State.Strict
                  ( MonadState, StateT, evalStateT )
 import           Data.Coerce
@@ -31,7 +35,6 @@ import           Data.List
 import qualified Data.Map.Strict as Map
 import qualified Data.Sequence as Seq
 import           Kore.Attribute.RuleIndex
-import           System.Exit
 import           System.IO
                  ( hFlush, stdout )
 import           Text.Megaparsec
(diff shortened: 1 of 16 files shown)


kore/src/Kore/Repl.hs
```diff
@@ -81,39 +84,39 @@ runRepl
...
-    evaluateScript rs =
-        maybe
-            (pure . pure $ state)
-            (parseEvalScript state)
-            (unReplScript rs)
+    evaluateScript :: ReplScript -> RWST (Config claim m) String (ReplState claim) m ()
+    evaluateScript = maybe (pure ()) parseEvalScript . unReplScript
 
-    repl0 :: StateT (ReplState claim m) m ()
+    repl0 :: ReaderT (Config claim m) (StateT (ReplState claim) m) ()
```
ttuegel: Why do we have a `ReaderT _ (StateT _ m)` stack in some places, but use `RWST` in others? I recall there was a reason we used `StateT` instead of `RWST` in some places, but I don't remember what it was. Anyway, to avoid some of the awkwardness of chaining `runReaderT` with `runStateT` you could also use `RWST (Config claim m) () (ReplState claim) m` because `()` is a `Monoid`.

kore/test/Test/Kore/Builtin/Builtin.hs
```diff
@@ -162,7 +162,7 @@ testEnv =
 evaluate :: TermLike Variable -> SMT (Pattern Variable)
 evaluate = evalSimplifier testEnv . TermLike.simplify
 
-evaluateT :: Trans.MonadTrans t => TermLike Variable -> t SMT (Pattern Variable)
+evaluateT :: Trans.MonadTrans t => TermLike Variable -> t (SmtT IO) (Pattern Variable)
```
ttuegel: What do we need this for? The same thing happens in a few other places below.

vladciobanu: I totally forgot to change them back after you suggested the type alias that would make this work without running into partial application. I'll revert these!

kore/src/Kore/Repl.hs
vladciobanu: Would `RWST _ () _` have the `WriterT` performance problems if we use `()` as the write type? I wasn't sure, which is why I used the `ReaderT ... StateT`. See a few lines up, the `forever repl0`, which is the worrying bit.

PR #753, opened 2019-06-21
Set unit unification

ttuegel:
@virgil-serbanuta I was able to re-create the integration test failure in a unit test. It only appears using `termUnification`, it doesn't happen when simplifying a `mkAnd` pattern, so I guess (?) we stopped running something through the simplifier before unification.

###### Reviewer checklist

---


virgil-serbanuta:
Okay, so here's how this was supposed to work:

1. The `concat(X, unit)` term should have been simplified to just `X`
2. Unification should have handled that as "variable vs functional".

I'm assuming that the first step is skipped in the test, and, if the `concat` is part of an axiom, it may also be skipped in actual execution. I'll try to fix it.

Issue #681, opened 2019-05-29
What is a sort injection?


ttuegel:
What is a sort injection?

Given these syntax definitions,
```
syntax BC ::= "B" | "C"
syntax A ::= "A" | BC
```
the sort `A` in Kore contains the constructor `A`, and we have subsort axioms for injections of  `inj{BC, A}(B)` and `inj{BC, A}(C)`. 

**Q1(a)**: Are these injections in the image of some constructor, or are they themselves constructor-like?

**Q1(b)**: If the injections are in the image of some constructor, which constructor?

The Haskell backend currently treats them as constructor-like; for example,
```
\ceil(A ∧ inj{BC, A}(B)) = ⊥
```

On the other hand, the no-junk constructor axioms for sort `A` do not mention them; that is, we have
```
axiom A
```
instead of
```
axiom A ∨ (∃ x. inj{BC, A}(x))
```

The answer to **Q1** has implications for how we treat overloaded symbols. Consider these definitions, adapted from #680:
```
syntax Number ::= Int | Float

syntax IType ::= "i32" | "i64"
syntax FType ::= "f32" | "f64"
syntax  Type ::= IType | FType

syntax IVal ::= overloaded(IType, Int) [klabel(overloaded)]
syntax FVal ::= overloaded(FType, Float) [klabel(overloaded)]

syntax Val ::= IVal | FVal
             | overloaded(Type, Number) [klabel(overloaded)]
```
Each `overloaded` symbol is regarded as a constructor in its own sort. Because all three `overloaded` symbols have the same `klabel`, the frontend generates overloading axioms such as
```
axiom overloaded_Val(inj{IType, Type}(T), inj{Int, Number}(X)) =
        inj{IVal, Val}(overloaded_IVal(T, X))
```
If `inj{IVal, Val}` is constructor-like, this axiom violates the no-confusion axiom for constructors in `Val`! The intent of such overloading is that the `overloaded` symbol in `Val` is merely an abstraction for the injection of the `IVal` and `FVal` symbols, but all three symbols are given constructor axioms. We could regard `overloaded_Val` as a (partial) function instead, but the other backends cannot handle functions on the left-hand side of rules.

**Q2**: What is `overloaded_Val`?

xc93:
These injections are constructor-like. The set of all constructors of sort `A` contains the terminal `A` and the injection `inj{BC, A}`. This means that when stating the (No Junk) axioms, we need to take the injection into account. In addition, these injections are also injective functions. 

[...]

PR #758, opened 2019-06-24
Use NormalizedSet as the internal set representation

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Attribute/Pattern/Functional.hs
+++ b/kore/src/Kore/Attribute/Pattern/Functional.hs
@@ -13,6 +13,7 @@ import qualified Data.Foldable as Foldable
 import           Data.Functor.Const
 import           Data.Hashable
 import           Data.Monoid
+import qualified Data.Set as Set
 import qualified Generics.SOP as SOP
 import qualified GHC.Generics as GHC
 
@@ -123,7 +124,34 @@ instance Synthetic (Rewrites sort) Functional where
 
 -- | A 'Builtin' pattern is 'Functional' if its subterms are 'Functional'.
 instance Synthetic (Builtin key) Functional where
-    synthetic = Foldable.fold
+    synthetic (BuiltinSet InternalSet {builtinSetChild}) =
+        normalizedSetDefined builtinSetChild
+      where
+        normalizedSetDefined :: NormalizedSet key Functional -> Functional
+        normalizedSetDefined
+            NormalizedSet
+                { elementsWithVariables = []
+                , sets = []
+                }
+          = Functional True
+        normalizedSetDefined
+          NormalizedSet
+              { elementsWithVariables = [withVariable]
+              , concreteElements
+              , sets = []
+              }
+          | Set.null concreteElements
+          = withVariable
+        normalizedSetDefined
+          NormalizedSet
+              { elementsWithVariables = []
+              , concreteElements
+              , sets = [set]
+              }
+          | Set.null concreteElements
+          = set
+        normalizedSetDefined _ = Functional False
+    synthetic builtin = Foldable.fold builtin
     {-# INLINE synthetic #-}
 
 instance Synthetic (Top sort) Functional where
--- a/kore/src/Kore/Attribute/Pattern/Functional.hs
+++ b/kore/src/Kore/Attribute/Pattern/Functional.hs
@@ -27,7 +27,7 @@ import           Kore.Syntax
 {- | A pattern is 'Functional' if it matches exactly one element.
  -}
 newtype Functional = Functional { isFunctional :: Bool }
-    deriving (Eq, GHC.Generic, Show)
+    deriving (Eq, GHC.Generic, Ord, Show)
     deriving (Semigroup, Monoid) via All
 
 instance SOP.Generic Functional
--- a/kore/src/Kore/Builtin/Set.hs
+++ b/kore/src/Kore/Builtin/Set.hs
@@ -18,14 +18,18 @@ module Kore.Builtin.Set
     ( sort
     , assertSort
     , sortDeclVerifiers
+    , isSetSort
     , symbolVerifiers
     , builtinFunctions
     , Domain.Builtin
+    , returnConcreteSet
     , returnSet
     , asInternal
+    , asInternalConcrete
     , asPattern
     , asTermLike
     , expectBuiltinSet
+    , expectConcreteBuiltinSet
       -- * Symbols
     , lookupSymbolIn
     , lookupSymbolDifference
@@ -48,7 +52,7 @@ module Kore.Builtin.Set
 import           Control.Applicative
                  ( Alternative (..) )
 import           Control.Error
-                 ( MaybeT (MaybeT), partitionEithers, runMaybeT )
+                 ( MaybeT (MaybeT), fromMaybe, partitionEithers, runMaybeT )
 import           Control.Error.Util
                  ( note )
 import           Control.Monad
@@ -57,6 +61,7 @@ import qualified Control.Monad.Trans as Monad.Trans
 import qualified Data.Foldable as Foldable
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.List as List
+import qualified Data.List
 import           Data.Map.Strict
                  ( Map )
 import qualified Data.Map.Strict as Map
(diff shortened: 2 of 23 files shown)


kore/src/Kore/Builtin/Set.hs
```diff
@@ -658,65 +782,24 @@ addToSetDisjoint = foldM addElementDisjoint
...
-      | otherwise = False
-    isEmptySet _ = False
-
-    elementSymbol = Builtin.lookupSymbolElement tools sort1
-    concatSymbol = Builtin.lookupSymbolConcat tools sort1
-    element elem' = mkApplySymbol elementSymbol [elem']
-    concat' set1 set2 = mkApplySymbol concatSymbol [set1, set2]
-
-    hookTools = Attribute.Symbol.hook <$> tools
+    addElementDisjoint :: Ord a => ([a], Set a) -> a -> Maybe ([a], Set a)
```
vladciobanu: I might be mistaking, but I think I saw this function in a different where clause. Would it be worth it to make it a helper function in the module that both functions could reuse?

kore/src/Kore/Step/Simplification/Not.hs
```diff
@@ -111,17 +112,35 @@ makeEvaluateNot
...
+makeEvaluatePredicate
+    Conditional
+        { term = ()
+        , predicate
+        , substitution
+        }
+  = Conditional
+        { term = ()
+        , predicate =
+            makeNotPredicate
```
vladciobanu: Why is there a negation here? I'm asking because this would confuse me just by reading the function name.

kore/src/Kore/Builtin/Set.hs
```diff
@@ -409,74 +520,92 @@ asInternal
...
+    else
+        let baseC = normalizedConcrete
+            baseS = addSets baseC filteredSets
+        in addElements baseS elementsWithVariables
   where
-    Domain.InternalSet { builtinSetChild = set } = builtin
+    normalizedConcrete :: TermLike variable
+    normalizedConcrete =
+        case Set.toList concreteElements of
+            [] -> mkApplySymbol unitSymbol []
```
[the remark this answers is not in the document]
virgil-serbanuta: The top comment was misleading (will be fixed in the next commit), and the name may also be misleading (but I'm not keen to fix it as it seems to be the same for all Builtins), but the basic idea is that `asTermLike` externalizes the pattern, so we should be able to represent its result in textual kore. Builtin_ does not exist in kore, and we don't have domain values for sets (which would be the next obvious option).

kore/src/Kore/Builtin/Set.hs
```diff
@@ -270,66 +330,99 @@ evalUnit =
...
+                        } = _set1
+                    Domain.NormalizedSet
+                        { elementsWithVariables = withVariable2
+                        , concreteElements = concrete2
+                        , sets = sets2
+                        } = _set2
+
+                    maybeResult :: Maybe (MaybeT m (AttemptedAxiom variable))
+                    maybeResult = do
+                        variableSetPartial <-
```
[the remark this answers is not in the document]
virgil-serbanuta: Well, `variableSetPartial` should be equal to `Set.fromList withVariable1`, but there is no type level guarantee that `withVariable1` has disjoint elements (although, unless I missed something, the current code enforces that), so I used `addToSetDisjoint`.

kore/src/Kore/Builtin/Set.hs
```diff
@@ -270,66 +330,99 @@ evalUnit =
...
+
+            emptyResult :: AttemptedAxiom variable
+            emptyResult = AttemptedAxiom.Applied mempty
+
+        case (toNormalizedSet tools _set1, toNormalizedSet tools _set2) of
+            (Bottom, _) -> return emptyResult
+            (_, Bottom) -> return emptyResult
+            -- The NormalizedSet matching is useful only for getting
+            -- notified when new fields are being added.
+            ( Normalized _set1@(Domain.NormalizedSet _ _ _)
```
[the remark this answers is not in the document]
virgil-serbanuta: We used it in the pretty-printing code (i.e. the haskell-like pretty-printing, not the unparsing one). Right now we use it in most of `StructEqualWithExplanation` instances in `Test.Kore.Comparators`, maybe in other places, too.

I agree with adding it to the style guide. I would automatically remove `_ _ _` for spurious compile errors, but I would recommend reconsidering its usefulness in that place.

kore/src/Kore/Builtin/Set.hs
```diff
@@ -270,66 +330,99 @@ evalUnit =
...
+              , Normalized _set2
```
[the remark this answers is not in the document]
virgil-serbanuta: Fixed. I wonder if we should allow this kind of half-indentation for `case` matchers that span more than one line. Maybe not worth the trouble, since it works nicely only for tuples and lists.

kore/src/Kore/Builtin/Set.hs
virgil-serbanuta: Fixed.

kore/src/Kore/Step/Simplification/Not.hs
virgil-serbanuta: Added function documentation, is it better now?

kore/src/Kore/Attribute/Pattern/Functional.hs
```diff
@@ -26,7 +27,7 @@ import           Kore.Syntax
 {- | A pattern is 'Functional' if it matches exactly one element.
  -}
 newtype Functional = Functional { isFunctional :: Bool }
-    deriving (Eq, GHC.Generic, Show)
+    deriving (Eq, GHC.Generic, Ord, Show)
```
ttuegel: What is the `Ord` instance for?

kore/src/Kore/Step/Simplification/Equals.hs
```diff
@@ -39,7 +39,7 @@ import qualified Kore.Step.Simplification.And as And
                  ( simplifyEvaluated )
 import qualified Kore.Step.Simplification.AndTerms as AndTerms
                  ( termEquals )
-import qualified Kore.Step.Simplification.Ceil as Ceil
+import {-# SOURCE #-} qualified Kore.Step.Simplification.Ceil as Ceil
```
ttuegel: The proliferation of cyclic dependencies is a burden I'd like to avoid. Without going into all the reasons why, I think that one of the next things we should look at is paying off this technical debt.

kore/test/Test/Kore/Unification/Unifier.hs
```diff
@@ -646,81 +560,119 @@ injUnificationTests :: [TestTree]
...
             term1
             term2
             []
+    , testCase "matching injections" $
+        andSimplifySuccess
+            (UnificationTerm (Mock.sortInjectionSubSubToTop Mock.aSubSubsort))
+            (UnificationTerm (Mock.sortInjectionSubToTop (mkVar Mock.xSubSort)))
+            [ UnificationResult
+                { term =
+                    -- TODO (virgil): Fix this in unification, injection
```
ttuegel: Not relevant to this pull request directly, but: In light of #681, we see that injections are actually regarded as constructors in the sorts where they appear. Couldn't we normalize sort injections by fully expanding them, i.e. so that in `inj{from, to}` the sorts `from` and `to` are always in a direct subsort relation?

virgil-serbanuta: It depends. Are we going to handle diamond-like subsort relations (a is a subsort of b and c, which are subsorts of c)? There was a long discussion about it a while ago, and I don't remember the outcome. My guess is that we want to handle those, at least to some extent, and, if so, I think unification would become more complicated if we were using only direct subsort relations and we'd be forced to pick one of the two branches of the diamond.

kore/src/Kore/Step/Simplification/Equals.hs
virgil-serbanuta: LGTM++

kore/src/Kore/Attribute/Pattern/Functional.hs
virgil-serbanuta: Well, it's for testing. To be precise, it's for this `With` instance:
https://github.com/kframework/kore/blob/master/kore/test/Test/Kore/With.hs#L322
used here:
https://github.com/kframework/kore/blob/master/kore/test/Test/Kore/Attribute/Pattern/Functional.hs#L62

kore/test/Test/Kore/Unification/Unifier.hs
virgil-serbanuta: b and c are subsorts of d above, sorry.

ttuegel: If I understand the result of #681, then diamond relations should be invalid, but I will bring it up there so we get an explicit answer.

PR #759, opened 2019-06-25
repl: commands which write to files crash repl when directory does not exist

ana-pantilie:
###### Reviewer checklist

---


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -663,10 +665,15 @@ redirect
 redirect cmd path = do
     config <- ask
     get >>= runInterpreter config >>= put
-    putStrLn' "File created."
   where
     redirectToFile :: String -> IO ()
-    redirectToFile = writeFile path
+    redirectToFile str = do
+        isCorrectPath <- checkIfCorrectFilePath path
```
vladciobanu: This could be refactored into something like
```
whenPathIsReachable :: FilePath -> (FilePath -> IO ()) -> IO ()
whenPathIsReachable path action =
    ifM
        (checkIfCorrectFilePath path)
        (action path)
        $ putStrLn "Directory does not exist."
```

Then you could replace all the copies with this, I think.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1200,3 +1217,18 @@ execStateReader config st action =
     flip execStateT st
         $ flip runReaderT config
         $ action
+
+checkIfCorrectFilePath :: MonadIO m => FilePath -> m Bool
+checkIfCorrectFilePath =
+    liftIO . doesDirectoryExist . fst . splitFileName
+
+whenPathIsReachable
```
ttuegel: Functions written in continuation passing style like this are often named `withSomething`. We might consider calling this `withExistingDirectory`.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1200,3 +1217,18 @@ execStateReader config st action =
     flip execStateT st
         $ flip runReaderT config
         $ action
+
+checkIfCorrectFilePath :: MonadIO m => FilePath -> m Bool
```
ttuegel: The name of this function is vague: what makes a `FilePath` "correct"? Maybe we could call it something like, `doesParentDirectoryExist`.

PR #761, opened 2019-06-26
repl: add weakExistsFinally to implications in repl output

ana-pantilie:
###### Reviewer checklist

---


kore/src/Kore/Step/Rule.hs
```diff
@@ -360,10 +359,28 @@ onePathRuleToPattern (OnePathRule rulePatt) =
...
-        )
+       ( mkApplyAlias
+            wEF
+            [( mkAnd
+                (Predicate.unwrapPredicate . ensures $ rulePatt)
+                (right rulePatt)
+            )]
+       )
+  where
+    wEF :: Alias
```
ttuegel: Because this definition is special, maybe we should promote it to the top-level.

PR #763, opened 2019-06-27
Matching builtins

### LINKED ISSUE #749: Matching builtins
We need to implement matching on builtins for #733:

- [x] Bool
- [x] Int
- [x] List
- [ ] Set
- [x] Map
- [x] DomainValues

This should be simpler than unification because we need only implement matching against canonical forms.

vladciobanu:
Fixes #749 

Current status:
- [x] Bool
- [x] Int
- [x] List
- [x] Set
- [x] Map
- [x] DomainValues

Tests and implementation should be working for `Bool`, `Int`, and `List`. There is also a working implementation for `Map`, but that will change when #758 is merged (and I will also have to add code/tests for `Set`).

###### Reviewer checklist

---


kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1040,195 @@ test_unificationWithAppMatchOnTop =
...
+        actual <- matchConcrete True True
+        assertEqualWithExplanation "" expect actual
+    , testCase "concrete bottom" $ do
+        let expect = Nothing
+        actual <- matchConcrete True False
+        assertEqualWithExplanation "" expect actual
+    , testCase "variable" $ do
+        let expect = substitution [(Mock.xBool, True)]
+        actual <- matchVariable Mock.xBool True
+        assertEqualWithExplanation "" expect actual
```
ehildenb: Let's add a test that ensures that `match True Mock.xBool` _doesn't_ work.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1040,195 @@ test_unificationWithAppMatchOnTop =
...
+        actual <- matchConcrete 1 1
+        assertEqualWithExplanation "" expect actual
+    , testCase "concrete bottom" $ do
+        let expect = Nothing
+        actual <- matchConcrete 1 0
+        assertEqualWithExplanation "" expect actual
+    , testCase "variable" $ do
+        let expect = substitution [(Mock.xInt, 1)]
+        actual <- matchVariable Mock.xInt 1
+        assertEqualWithExplanation "" expect actual
```
ehildenb: Similarly, can we add a test that `match 1 Mock.xInt` doesn't work?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1040,195 @@ test_unificationWithAppMatchOnTop =
...
+            , Right 3
+            , Left Mock.yInt
+            ]
+            [ 1, 2, 3, 4 ]
+        assertEqualWithExplanation "" expect actual
+    , testCase "no AC" $ do
+        let expect = Nothing
+        actual <- matchVariable
+            [ Right 1, Left Mock.xInt] [ 2, 1 ]
+        assertEqualWithExplanation "" expect actual
```
ehildenb: Add `match [ 1 , 2 ] [ Left Mock.xInt , Left Mock.yInt ]`.

Also can we do `match Mock.xList [ 1 , 2 ]`, which should have substitution `Mock.xList -> [ 1 , 2 ]`?

And we should be able to match the tail of a list too, something like `match [ Mock.xInt , Mock.yList ] [ 1 , 2 , 3 ]` where `Mock.xInt -> 1` and `Mock.yList -> [ 2 , 3 ]`.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1040,195 @@ test_unificationWithAppMatchOnTop =
...
+            , (2, Left Mock.xInt)
+            , (3, Right 12)
+            , (4, Left Mock.yInt)
+            ]
+            [ (2, 11)
+            , (1, 10)
+            , (4, 13)
+            , (3, 12)
+            ]
+        assertEqualWithExplanation "" expect actual
```
ehildenb: We should be able to handle the case `match [ Mock.xMap , ( 1 , 2 ) ] [ (1, 2) , (3, 4), (5, 6) ]` which should get substitution `Mock.xMap -> [ (3, 4), (5, 6) ]`.

ehildenb:
I added cases I would like to see handled, but they might not all be feasible right now.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,814 @@ test_unificationWithAppMatchOnTop =
...
+        let expect =
+                substitution
+                    [ (Mock.xInt, 2)
+                    , (Mock.yInt, 4)
+                    ]
+        actual <-
+            matchVariable
+                [Concrete 1, Concrete 3, ElemVar Mock.xInt, ElemVar Mock.yInt]
+                [1, 2, 3, 4]
+        assertEqualWithExplanation "" expect actual
```
ehildenb: This test seems potentially fragile. Are we sure that (based on the order things are stored in and such) it will always be `x |-> 2    y |-> 4`, or could it ever be `y |-> 2    x |-> 4`?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,814 @@ test_unificationWithAppMatchOnTop =
...
+                        , substitution = Substitution.unsafeWrap
+                            [ (Mock.xSet, mkConcreteSet [mkKey 2, mkKey 3])
+                            , (Mock.xInt, mkKey 1)
+                            ]
+                        }
+                    ]
+        actual <-
+            matchVariable
+                [SetVar Mock.xSet, ElemVar Mock.xInt]
+                [1, 2, 3]
```
ehildenb: This one too could be fragile.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,814 @@ test_unificationWithAppMatchOnTop =
...
+                        , substitution = Substitution.unsafeWrap
+                            [ (Mock.xSet, mkConcreteSet [mkKey 3])
+                            , (Mock.xInt, mkKey 2)
+                            ]
+                        }
+                    ]
+        actual <-
+            matchVariable
+                [SetVar Mock.xSet, ElemVar Mock.xInt, Concrete 1]
+                [1, 2, 3]
```
ehildenb: potentially fragile?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,814 @@ test_unificationWithAppMatchOnTop =
...
+                    ]
+        actual <-
+            matchVariable
+                [ SetVar Mock.xSet
+                , ElemVar Mock.xInt
+                , ElemVar Mock.yInt
+                , Concrete 1
+                , Concrete 3
+                ]
+                [1, 2, 3, 4, 5]
```
ehildenb: potentially fragile

kore/test/Test/Kore/Step/Axiom/Matcher.hs
vladciobanu: It could definitely change in the future, so I agree these tests are brittle. I will add an issue to refactor these tests into property tests, when we have time or when this becomes a priority.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,813 @@ test_unificationWithAppMatchOnTop =
...
+    matchVariable var val =
+        matchDefinition (mkVar var) (mkStr val)
+
+matchingList :: [TestTree]
+matchingList =
+    [ testCase "concrete top" $ do
+        let expect = top
+        actual <- matchConcrete [1, 2] [1, 2]
+        assertEqualWithExplanation "" expect actual
+    , testCase "concrete bottom" $ do
```
virgil-serbanuta: I'd also add a bottom test for `[1,2]` vs `[1,2,3]`

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,813 @@ test_unificationWithAppMatchOnTop =
...
+        actual <- matchConcrete [1, 2, 3] []
+        assertEqualWithExplanation "" expect actual
+
+    , testCase "elementVar vs singleton" $ do
+        let expect = substitution [(Mock.xInt, 1)]
+        actual <- matchVariable [ElemVar Mock.xInt] [1]
+        assertEqualWithExplanation "" expect actual
+    , testCase "{elementVar, rest} vs {set}" $ do
+        let expect = substitution [(Mock.xInt, 2)]
+        actual <- matchVariable [Concrete 1, Concrete 3, ElemVar Mock.xInt] [1, 2, 3]
```
virgil-serbanuta: 80 chars?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,813 @@ test_unificationWithAppMatchOnTop =
...
+        matchConcreteSet `on` fmap mkKey
+    matchVariable var val =
+            matchDefinition
+                (mkSet
+                    (mkKey <$> concrete var)
+                    (mkVar <$> elemVars var)
+                    (mkVar <$> setVars  var)
+                )
+                (mkConcreteSet $ fmap mkKey val)
+    -- cons = Mock.concatMap
```
virgil-serbanuta: Do you need these?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,813 @@ test_unificationWithAppMatchOnTop =
...
+    --         . mkMap
+    --         $ fmap (mapWithKey mkVal) m2
+
+matchingMap :: [TestTree]
+matchingMap =
+    [ testCase "concrete top" $ do
+        let expect = top
+        actual <- matchConcrete [(1, 11), (2, 12)] [(1, 11), (2, 12)]
+        assertEqualWithExplanation "" expect actual
+    , testCase "concrete bottom" $ do
```
virgil-serbanuta: I'd also add a bottom test for values not matching.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,813 @@ test_unificationWithAppMatchOnTop =
...
+    mkMap = Map.asInternal Mock.metadataTools Mock.mapSort . Map.fromList
+    mapWithKey = Bifunctor.bimap mkKey
+    matchMap = matchDefinition `on` mkMap
+    matchConcrete =
+        matchMap `on` fmap (mapWithKey mkVal)
+    matchVariable var val =
+            matchMap
+            (mapWithKey (either mkVar mkVal) <$> var)
+            (mapWithKey mkVal <$> val)
+    cons = Mock.concatMap
```
virgil-serbanuta: Just curious: Why cons instead of concat/concat'/whatever?

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -435,3 +452,159 @@ checkVariableEscape vars predSubst
...
+                (Builtin.BuiltinMap prefix2)
+        suffix <- match qv x (mkBuiltin $ Builtin.BuiltinMap suffix2)
+        pure $ prefix <> suffix
+
+    splitMap
+        :: Builtin.InternalMap (TermLike Concrete) (TermLike variable)
+        ->  ( Builtin.InternalMap (TermLike Concrete) (TermLike variable)
+            , Builtin.InternalMap (TermLike Concrete) (TermLike variable)
+            )
+    splitMap m =
```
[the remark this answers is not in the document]
vladciobanu: I went with `splitM2By` and refactored the implementation a bit. Let me know if it's better now.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -435,3 +452,159 @@ checkVariableEscape vars predSubst
...
+    fmap Foldable.fold
+        . traverse match'
+        $ Seq.zip seq1 seq2
+  where
+    seq1 = Builtin.builtinListChild l1
+    seq2 = Builtin.builtinListChild l2
+    match' = uncurry (match qv)
+matchBuiltins qv (Builtin.BuiltinSet s1) (Builtin.BuiltinSet s2)
+  | varElems2 == [] && varSets2 == [] && length varSets1 <= 1 = do
+    Monad.guard $ concrete1 `Set.intersection` concrete2 == concrete1
```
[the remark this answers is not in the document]
vladciobanu: Good point, I updated the code.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
vladciobanu: Forgot to rename it. I started with it as cons, decided concat is better but forgot to rename. Fixed.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -435,3 +452,160 @@ checkVariableEscape vars predSubst
...
+matchBuiltins qv (Builtin.BuiltinSet s1) (Builtin.BuiltinSet s2)
+  | varElems2 == [] && varSets2 == [] && length varSets1 <= 1 = do
+    let remainder = Set.toList $ concrete2 Set.\\ concrete1
+    Monad.guard $ length remainder + length concrete1 == length concrete2
+    Monad.guard $ length remainder >= length varElems1
+
+    variableMatches <-
+        fmap Foldable.fold
+            . traverse match'
+            . zip varElems1
```
ttuegel: `zip` discards the tail of the longer list; is that OK here? It isn't clear to me.

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -1029,6 +1043,816 @@ test_unificationWithAppMatchOnTop =
...
+    Foldable.fold
+        [ matchingBool
+        , matchingInt
+        , matchingString
+        , matchingList
+        , matchingSet
+        , matchingMap
+        ]
+
+matchingBool :: [TestTree]
```
ttuegel: Could we make this a `TestTree` using `testGroup` so that `test_matchingBuiltins` above isn't flat? I think it will make the test output clearer.

kore/src/Kore/Step/Axiom/Matcher.hs
vladciobanu: Yes, I think it's precisely what we need:
- if we have more `varElems`, code is not reached due to the `guard` above
- if we have more concrete elements than `varElems`, that is fine, since we check the `setRemainder` below

PR #764, opened 2019-06-28
Updated set variables and removed meta ids

traiansf:
###### Reviewer checklist

---


src/test/resources/user-meta-nat.kore
```diff
@@ -1,7 +1,7 @@
 []
     module Nat
-            sort #Nat{}[]
```
virgil-serbanuta: Shouldn't you also rename the file?

kore/src/Kore/Parser/Parser.hs
```diff
@@ -84,41 +82,34 @@ BNF definition:
...
             { sortActualName = identifier
             , sortActualSorts = sorts
             }
 
-{-|'validateMetaSort' checks that a @meta-sort@ is well-formed.
 
-Relevant BNF definitions:
+metaSortParser :: Parser Sort
+metaSortParser = do
+    sort <- keywordBasedParsers
```
ttuegel: Could we extract this, so that `metaSortParser` is all written at one level of detail?
```.hs
metaSortKeywordParser :: Parser Sort
metaSortKeywordParser =
    keywordBasedParsers
        (map
            (show &&& return . metaSort)
            metaSortsListWithString
        )
```

kore/src/Kore/Parser/Parser.hs
```diff
@@ -84,41 +82,34 @@ BNF definition:
...
+        (map
+            (show &&& return . metaSort)
+            metaSortsListWithString
+        )
+    sorts <- inCurlyBracesListParser objectSortParser
```
ttuegel: By the way, I think we could replace `objectSortParser` on this line with the `fail` from the next line and get the same error with a more useful location.

kore/src/Kore/Parser/Parser.hs
```diff
@@ -200,7 +189,7 @@ unaryOperatorRemainderParser
     -> Parser result
 unaryOperatorRemainderParser childParser constructor =
     constructor
-    <$> inCurlyBracesRemainderParser sortParser
+    <$> inCurlyBracesRemainderParser objectSortParser
```
ttuegel: Why is this only an `objectSortParser`?

kore/src/Kore/Parser/Parser.hs
```diff
@@ -411,21 +388,21 @@ BNF definitions:
 ⟨set-variable⟩ ::= ⟨set-variable-identifier⟩ ‘:’ ⟨sort⟩
 @
 
-The @set-@ version always starts with @#@, while the regular one does not.
+The @set-@ version always starts with @\@@, while the regular one does not.
 -}
 variableParser :: Parser Variable
 variableParser = idParser >>= variableRemainderParser
 
 singletonVariableParser :: Parser Variable
```
ttuegel: I know it wasn't introduced in this pull request, but doesn't _The Semantics of K_ use the terminology, "singular variable," not "singleton variable"?

PR #765, opened 2019-06-30
return the counterexample and enforce one rule in the spec module

yzhang90:
###### Reviewer checklist

---
1) We return the counterexample to the frontend instead of using Debug.Trace to print it to the std output.
2) Currently we enforce only one rule in the spec module. If a spec module contains several rules to check, it is hard to present the final results to the user.

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -400,8 +400,13 @@ mainWithOptions
                           where
                             failure pat = (ExitFailure 1, pat)
                             success = (ExitSuccess, mkTop $ mkSortVariable "R")
-                            unknown = (ExitSuccess,
-                                       mkTop $ mkSortVariable "Unknown")
+                            unknown =
+                                ( ExitSuccess
+                                , mkVar
+                                    $ varS
+                                        "Unkown"
+                                        (mkSort $ noLocationId "SortUnknown")
+                                )
                 )
         let unparsed = (unparse . externalizeFreshVariables) finalPattern
         case outputFileName of
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -40,6 +40,8 @@ import           Kore.Internal.Pattern
 import           Kore.Internal.TermLike
 import           Kore.Logger.Output
                  ( KoreLogOptions (..), parseKoreLogOptions, withLogger )
+import qualified Kore.ModelChecker.Bounded as Bounded
+                 ( CheckResult (..) )
 import           Kore.Parser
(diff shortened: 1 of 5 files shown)


kore/app/exec/Main.hs
```diff
@@ -395,6 +400,8 @@ mainWithOptions
                           where
                             failure pat = (ExitFailure 1, pat)
                             success = (ExitSuccess, mkTop $ mkSortVariable "R")
+                            unknown = (ExitSuccess,
+                                       mkTop $ mkSortVariable "Unknown")
```
virgil-serbanuta: According to the style guide, indentation should use a number of spaces divisible by 4 except in certain cases where it makes sense to use half-indentation (e.g. for the `where` keyword). I think that this line uses 4n+3 spaces.

yzhang90: done

kore/src/Kore/ModelChecker/Step.hs
```diff
@@ -183,13 +180,7 @@ transitionRule
...
                 else do
                     (Monad.Trans.lift . State.put) (Just ())
-                    trace
-                        (show . Pretty.vsep
-                            $ [ "config failed to prove the invariant:"
-                              , Pretty.indent 4 (unparse config)
-                              ]
-                        )
-                        return Unprovable
+                    return (Unprovable config)
```
virgil-serbanuta: Could you add a test for returning unprovable configurations?

yzhang90: Currently I don't have unit tests for KBMC. But I'll turn all the examples under https://github.com/kframework/kore/tree/master/src/main/k/in-progress/bmc/example1/check into integration tests. This can be done after this PR and https://github.com/kframework/k/pull/545 are merged.

virgil-serbanuta: I'll approve this as-is, but I'll open a discussion about unit-tests on our main channel.

kore/src/Kore/Step/Rule.hs
```diff
@@ -161,6 +161,15 @@ deriving instance Eq variable => Eq (ImplicationRule variable)
 deriving instance Ord variable => Ord (ImplicationRule variable)
 deriving instance Show variable => Show (ImplicationRule variable)
 
+instance
+    (Ord variable, SortedVariable variable, Unparse variable)
+    => Unparse (ImplicationRule variable)
+  where
+    unparse (ImplicationRule RulePattern { left, right } ) =
```
ttuegel: What about the `requires` and `ensures`?

Issue #767, opened 2019-07-01
Conditional datatype doesn't unparse to correct Kore


ana-pantilie:
`Kore.Internal.Conditional`: unparse instance ignores sorts by replacing them with `_` or `_PREDICATE`

ttuegel:
The `Unparse` instance for `Conditional` should be removed; we can only unparse a `Kore.Internal.Pattern.Pattern`, not an arbitrary `Conditional`. The `Predicate` should be assigned the same sort as the `TermLike`, otherwise the frontend will not be happy. Anything sent to the frontend needs to be run through [`Kore.Internal.TermLike.externalizeFreshVariables`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Internal/TermLike.hs#L896), so it might be best to call that function as part of the `Unparse` instance.

PR #777, opened 2019-07-10
remove Conditional unparse instance

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/767


kore/src/Kore/Internal/Pattern.hs
```diff
@@ -207,3 +209,21 @@ toPredicate = Conditional.toPredicate
 
 splitTerm :: Pattern variable -> (TermLike variable, Predicate variable)
 splitTerm = Conditional.splitTerm
+
+unparsePattern
```
ttuegel: You can use this as the definition of `instance Unparse (Pattern variable)`, and avoid most of the other code changes. It would also be good to have it use `Kore.Internal.TermLike.externalizeFreshVariables`.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -578,8 +578,21 @@ instance
     (SortedVariable variable, Unparse variable) =>
```
ttuegel: The `Unparse variable` constraint is redundant now because of using `toVariable` below.

kore/src/Kore/Internal/Conditional.hs
```diff
@@ -146,54 +145,6 @@ instance TopBottom term
...
         isBottom term || isBottom predicate || isBottom substitution
 
-instance
-    ( SortedVariable variable
-    , Ord variable
-    , Show variable
-    , Unparse variable
-    , Unparse child
-    ) =>
-    Unparse (Conditional variable child)
```
ttuegel: I would like to keep the structure of this instance instead of using `toTermLike`, because it highlights which parts of the output are the `term`, `predicate`, and `substitution`. What we should do here is replace `child` with `TermLike variable`, and below...

kore/src/Kore/Internal/Conditional.hs
```diff
@@ -146,54 +145,6 @@ instance TopBottom term
...
-  where
-    unparse Conditional { term, predicate, substitution } =
-        unparseAnd
-            (below "/* term: */" (unparse term))
-            (unparseAnd
-                (below
-                    "/* predicate: */"
-                    (unparse predicate)
```
ttuegel: ...we should use `Kore.Syntax.Predicate.fromPredicate` on the `predicate` to set the sort (to the value of `termLikeSort term`)...

kore/src/Kore/Internal/Conditional.hs
```diff
@@ -146,54 +145,6 @@ instance TopBottom term
...
-                )
-                (below
-                    "/* substitution: */"
-                    (unparse $ Predicate.fromSubstitution substitution)
```
ttuegel: ...and the same thing on this predicate.

Issue #769, opened 2019-07-02
Differentiate `Not Matching` and `Not implemented` when matching for function rules


yzhang90:
Now matching algorithm will return `bottom` for both cases (not matching and not implemented). We may want to differentiate them and log some information when `Not implemented` case is encountered.

ttuegel:
In `master` "not implemented" is `Left` and "not matching" is `\bottom`. In #733, I first planned to make both `\bottom`, but we decided not to do that; see the discussion on that ticket for all the details.

PR #789, opened 2019-07-16
Evaluate functions by matching instead of unification

ttuegel:
Supersedes: #733, #781 
After: #800, #810

###### Reviewer checklist

---


ttuegel:
Fixes: #769

ttuegel:
There is some problem with Jenkins that I can't get around right now.

This is still broken anyway: it applies the `andBool` rule to get a non-matching substitution (somehow) and generates a remainder that Z3 can't handle.

ttuegel:
> This is still broken anyway: it applies the `andBool` rule to get a non-matching substitution (somehow) and generates a remainder that Z3 can't handle.

This is not actually directly related to this pull request, and has to do with matching `#as` patterns. I will deal with that in a second pull request.

ttuegel:
There is a test failure in `src/main/k/working/tests/collections/list-requires` due to the backend exhausting available memory.

ttuegel:
It looks like we are expanding symbolic functions infinitely (again). I'm writing tests now to check that the relevant cases do not match.

kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -255,100 +327,185 @@ instance TermWrapper Domain.NormalizedSet Domain.NoValue where
...
+flattenOpaque matchBuiltin (Domain.unwrapAc -> normalized) = do
+    let opaque = Domain.opaque normalized
+        (builtin, opaque') = partitionEithers (extractBuiltin <$> opaque)
+        transparent = Domain.wrapAc normalized { Domain.opaque = opaque' }
+    Foldable.foldrM concatNormalized transparent builtin
+  where
+    extractBuiltin termLike =
+        maybe (Right termLike) Left (matchBuiltin termLike)
+
+matchBuiltinMap
```
virgil-serbanuta: I'd put this near matchBuiltinSet

PR #795, opened 2019-07-17
Added setSubstitution field to Conditional

traiansf:
###### Reviewer checklist

---


virgil-serbanuta:
> Instead of adding the `setSubstitution` field to every `Conditional` in the tests, I think we would be better off refactoring the tests so that they don't depend closely on the representation of `Conditional`.

I'd like to advertise https://github.com/kframework/kore/blob/master/kore/test/Test/Kore/With.hs, which could be extended to handle `Conditional`.

Issue #720, opened 2019-06-11
Print timing information to the log


ttuegel:
By default, `kore-exec` is pretty noisy about printing timing information. This information belongs in the log, and probably should not be displayed at the default log level.

PR #797, opened 2019-07-18
Print timing info to the log

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/720

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -2,6 +2,8 @@ module Main (main) where
 
 import           Control.Applicative
                  ( Alternative (..), optional, (<$) )
+import           Control.Monad.Trans
+                 ( lift )
 import           Control.Monad.Trans.Reader
                  ( runReaderT )
 import qualified Data.Bifunctor as Bifunctor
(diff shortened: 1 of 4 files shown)


kore/app/exec/Main.hs
```diff
@@ -306,54 +308,73 @@ mainWithOptions
...
-                    do
-                        searchPattern <-
-                            mainParseSearchPattern indexedModule searchFileName
-                        let searchConfig = Search.Config { bound, searchType }
-                        (return . Just) (searchPattern, searchConfig)
-        proveParameters <-
-            case koreProveOptions of
+        withLogger koreLogOptions (\logger -> do
+            parsedDefinition <-
+                runReaderT
```
ttuegel: Instead of wrapping each individual call in `runReaderT`, why not put one `runReaderT` at the top of the block?

kore/app/parser/Main.hs
```diff
@@ -175,7 +187,8 @@ mainVerify willChkAttr definition =
             else DoNotVerifyAttributes
     in do
       verifyResult <-
-        clockSomething "Verifying the definition"
+          (flip runReaderT $ Logger.emptyLogger)
```
ttuegel:
```suggestion
          (flip runReaderT Logger.emptyLogger)
```

kore/app/share/GlobalMain.hs
```diff
@@ -75,6 +83,8 @@ import qualified Kore.Verified as Verified
 import qualified Paths_kore as MetaData
                  ( version )
 
+type GlobalMainIO = ReaderT (LogAction IO Logger.LogMessage) IO
```
ttuegel: I would prefer to use a `newtype` here, so that we don't accidentally depend on the implementation details. You might have to `derive` several instances; at least `MonadReader` and `MonadIO` (maybe others), but if it gets much harder than deriving instances then you can forget I said anything.

We might also simply name this `Main` instead, unless @vladciobanu feels differently.

Issue #752, opened 2019-06-21
Evaluating functions with undefined arguments


ttuegel:
Symbols with ⊥ arguments evaluate to ⊥, but we do not account for this when constructing remainders. For example, given an application pattern
```
f(φ)  // f is any symbol, φ is a possibly-undefined pattern
```
and an axiom such as
```
axiom f(x) = A
```
we will produce
```
f(φ) = A ∨ (f(φ) ∧ ¬⌈φ⌉)  // EDIT: This is not quite right, see below.
```
even though the condition attached to the latter makes the second disjunct ⊥. Consequently, we leave a mess of remainders around whenever partial patterns are involved.

I propose to solve this by adding definedness conditions to the initial condition during function application. We recognize that
```
f(φ) = f(φ) ∧ ⌈φ⌉
```
so that we can add `⌈φ⌉` to the initial conditions during rule application, which will prevent the messy remainders from ever being formed. When `φ` is a defined pattern, this condition becomes trivial, so in most cases this will not affect evaluation at all.

CC: @yzhang90 
See also: #729, #748

grosu:
@ttuegel This seems pretty reasonable.  In fact, that also happens in separation logic proofs, when people prove that `HeapPattern1 -> HeapPattern2`, then assume `HeapPattern1` defined.

What concerns me, though, is the consistency of function definitions.  For example, if one says that `_/Int_` is a `[function]` and then that `I /Int 0 = \bot`, then we got an inconsistent definition.  Many FOL provers, including Z3, assume that all the functions are total (and implicitly also terminate, not only totally defined).  For example, `3/0 = 0` in Z3.  I think we may want to do the same with symbols declared as `[function]`.  The attribute is a hint, so the function can be used as such, but when we generate proof objects later on we'll need to check the claim and reject functions which are not total.

ttuegel:
In _The Semantics of K_, what you are calling `function` is actually `functional`. `function` indicates a partial function, so I think the definition above is fine.

vladciobanu:
Does this still hold if `f` is a `const` function?

ttuegel:
@vladciobanu Yes, the semantics of symbols are such that `symbol(φ) = ⊥` if `φ = ⊥`.

ttuegel:
@virgil-serbanuta pointed out during the meeting this morning that the result of applying the axiom for `f(x)` is actually:
```
[...]

PR #800, opened 2019-07-19
Evaluating functions with undefined arguments

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/752


yzhang90:
Can we have a test like `x / x` is evaluated to `1 /\ x != 0`?

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -336,3 +351,50 @@ evaluateWithDefinitionAxioms
...
+    .  ( SortedVariable variable
+       , FreshVariable variable
+       , Unparse variable
+       , Show variable
+       )
+    => OrPredicate variable
+    -> Pattern variable
+    -> Pattern variable
+introduceDefinedness cond result =
+    foldl andCondition result cond
```
ttuegel:
```suggestion
    foldr andCondition result cond
```

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -300,29 +312,32 @@ evaluateWithDefinitionAxioms
     _axiomSimplifiers
     patt
   =
-    AttemptedAxiom.maybeNotApplicable $ do
-    let
-        -- TODO (thomas.tuegel): Figure out how to get the initial conditions
-        -- and apply them here, to remove remainder branches sooner.
-        expanded :: Pattern variable
-        expanded = Pattern.fromTermLike patt
+    ceilChildOfApplication patt >>= \ceilChild ->
```
ttuegel: Why is this introduced at the top?

kore/test/Test/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -79,13 +94,8 @@ test_definitionEvaluation =
...
                     , remainders =
                         OrPattern.fromPatterns
-                        $ map (fmap mkEvaluated)
-                            [ Conditional
-                                { term = Mock.functionalConstr10 Mock.a
-                                , predicate = makeNotPredicate requirement
-                                , substitution = mempty
-                                }
-                            ]
+                        $ map (fmap mkEvaluated . introduceDefinedness ceil)
```
ttuegel: This is just a duplication of the code in the tests. Please state the arguments-defined conditions explicitly in the expected test result. For most tests, there should not be an additional condition.

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -470,4 +541,6 @@ evaluateWithAxioms axioms =
             ,   ( AxiomIdentifier.Application Mock.elementMapId
                 , builtinEvaluation Map.evalElement
                 )
+            ,   ( AxiomIdentifier.Application Mock.tdivIntId
+                , builtinEvaluation (Int.builtinFunctions Map.! Int.tdivKey))
```
ttuegel:
```suggestion
                , builtinEvaluation (Int.builtinFunctions Map.! Int.tdivKey)
                )
```

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -336,3 +351,50 @@ evaluateWithDefinitionAxioms
...
+       , MonadSimplify simplifier
+       )
+    => TermLike variable
+    -> simplifier (OrPredicate variable)
+ceilChildOfApplication =
+    withCeilChildOfApplication
+        Ceil.makeEvaluateTerm
+        AndPredicates.simplifyEvaluatedMultiPredicate
+
+withCeilChildOfApplication
```
ttuegel: Let's move the three new functions in this module into `Kore.Step.Remainder` because they are all related to building the remainder.

kore/src/Kore/Step/Remainder.hs
```diff
@@ -153,3 +167,50 @@ substitutionConditions subst =
...
+    .  ( SortedVariable variable
+       , FreshVariable variable
+       , Unparse variable
+       , Show variable
+       )
+    => OrPredicate variable
+    -> Pattern variable
+    -> Pattern variable
+introduceDefinedness cond result =
+    foldr (flip andCondition) result cond
```
virgil-serbanuta: This seems strange. It seems that you have the definedness condition as `P1 or P2 or ...` and the result `R` and you're transforming it into `R and P1 and P2 and ...`. Why does that make sense? If it makes sense, could you add a comment explaining it?

From the function signature, I would have expected something like `(R and P1) or (R and P2) or ...`, but I'm not sure what you're trying to do.

Is this a generic function that should go to `OrPredicate.hs` (or `OrPattern.hs` or somehting similar), or is it something specific to definedness?

kore/src/Kore/Step/Remainder.hs
```diff
@@ -153,3 +167,50 @@ substitutionConditions subst =
...
+    .  ( FreshVariable variable
+       , SortedVariable variable
+       , Show variable
+       , Unparse variable
+       , MonadSimplify simplifier
+       )
+    => TermLike variable
+    -> simplifier (OrPredicate variable)
+ceilChildOfApplication =
+    withCeilChildOfApplication
```
virgil-serbanuta: Why is having the `with...` function better than using its body here?

kore/src/Kore/Step/Remainder.hs
ana-pantilie: > This seems strange. It seems that you have the definedness condition as P1 or P2 or ... and the result R and you're transforming it into R and P1 and P2 and .... Why does that make sense? If it makes sense, could you add a comment explaining it?

The `ceil` of the child of the function application is an `OrPredicate`, and that function is intended to be used when mapping over the remainders, so this will result in `(remainder1 and ceilChild) or (remainder2 and ceilChild) ...`. I think the error is that `ceilChild` should be folded to a disjunction not a conjunction (since it's an `OrPredicate`). Would this be correct?

kore/src/Kore/Step/Remainder.hs
ana-pantilie: I used to need that function for some failing tests, but after making the changes suggested by you last week the tests started working again.

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -239,6 +241,70 @@ test_simplificationIntegration =
...
+                          )
+                        , (AxiomIdentifier.Ceil (AxiomIdentifier.Application Mock.tdivIntId)
+                          , [ EqualityRule RulePattern
+                                { left =
+                                    mkCeil testSortVariable
+                                    $ Mock.tdivInt
+                                        (mkVar Mock.xInt)
+                                        (mkVar Mock.yInt)
+                                , right =
+                                    mkCeil testSortVariable
```
virgil-serbanuta: Do you still need mkCeil here? It looks like it's being applied to a predicate, so it will be removed on the first line of `makeEvaluateNonBoolCeil` anyway.

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -239,6 +241,70 @@ test_simplificationIntegration =
                     , substitution = mempty
                     }
         assertEqualWithExplanation "" expect actual
+    , testCase "function application introduces definedness condition" $ do
```
virgil-serbanuta: Could you add a comment with the initial configuration, the rules being applied and the expected result? This should make it easier to understand what happens below.

PR #801, opened 2019-07-19
Fix AC unification loop

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Builtin/AssociativeCommutative.hs
+++ b/kore/src/Kore/Builtin/AssociativeCommutative.hs
@@ -77,7 +77,7 @@ import           Kore.Internal.Symbol
                  ( Symbol )
 import           Kore.Internal.TermLike
                  ( pattern App_, pattern Builtin_, Concrete, TermLike,
-                 mkApplySymbol, mkBuiltin, termLikeSort )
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -1197,13 +1167,23 @@ unifyEqualsElementLists
...
+        Just remainderTerm -> case opaque of
+            Var_ _ -> do
+                opaqueUnifier <- unifyEqualsChildren opaque remainderTerm
+                let
+                    (opaqueTerm, opaquePredicate) =
+                        Pattern.splitTerm opaqueUnifier
+                    result = unifier `andCondition` opaquePredicate
+
+                return (result, [opaqueTerm])
+            _ -> (error . unlines)
```
ttuegel: If I understand correctly, this throws an error if an opaque collection isn't a variable?

virgil-serbanuta: Yes. The check I removed (as you pointed out) threw an error for `Application` patterns, but whitelisting seems safer then blacklisting.

PR #807, opened 2019-07-23
Internal representations

ttuegel:
This pull request adds `Kore.Builtin.internalize` which converts external syntax to internal representations. This function is _not_ yet used by default during execution because of the need to re-normalized the builtin associative-commutative representations. I will deal with that in a subsequent pull request.

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Builtin.hs
+++ b/kore/src/Kore/Builtin.hs
@@ -30,8 +30,6 @@ module Kore.Builtin
     , internalize
     ) where
 
-import           Control.Category
-                 ( (>>>) )
 import qualified Control.Comonad.Trans.Cofree as Cofree
 import qualified Data.Functor.Foldable as Recursive
 import qualified Data.HashMap.Strict as HashMap
@@ -282,5 +280,5 @@ internalize tools =
   where
     internalize1 =
             List.internalize tools
-        >>> Map.internalize tools
-        >>> Set.internalize tools
+        .   Map.internalize tools
+        .   Set.internalize tools
--- a/kore/src/Kore/Builtin.hs
+++ b/kore/src/Kore/Builtin.hs
@@ -27,8 +27,11 @@ module Kore.Builtin
     , koreEvaluators
     , evaluators
     , externalizePattern
+    , internalize
     ) where
 
+import           Control.Category
+                 ( (>>>) )
 import qualified Control.Comonad.Trans.Cofree as Cofree
 import qualified Data.Functor.Foldable as Recursive
 import qualified Data.HashMap.Strict as HashMap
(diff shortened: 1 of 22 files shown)


kore/src/Kore/Builtin.hs
```diff
@@ -266,3 +271,16 @@ externalizePattern =
...
+    :: (Ord variable, SortedVariable variable)
+    => SmtMetadataTools Attribute.Symbol
+    -> TermLike variable
+    -> TermLike variable
+internalize tools =
+    Recursive.fold (internalize1 . Recursive.embed)
+  where
+    internalize1 =
+            List.internalize tools
+        >>> Map.internalize tools
```
vladciobanu: Am I misunderstanding this, or is `>>>` used here as `flip (.)`? And since they're all essentially `a -> a`, does the order matter? Could it not be replaced by `.`?

ttuegel: You understand correctly. I did not have a particular reason for using it here.

kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -867,32 +865,27 @@ Returns Just False if the sort is a variable.
 -}
 isSort :: Text -> SmtMetadataTools attr -> Sort -> Maybe Bool
 isSort builtinName tools sort
-  | isPredicateSort = Nothing
-  | otherwise =
-    Just (getHook hook == Just builtinName)
+  | isPredicateSort            = Nothing
+  | SortVariableSort _ <- sort = Nothing
+  | otherwise                  = Just (getHook hook == Just builtinName)
```
virgil-serbanuta: Matching on SortActualSort here should have been safer.

kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -867,32 +865,27 @@ Returns Just False if the sort is a variable.
...
     Attribute.Sort {hook} = sortAttributes sort
     isPredicateSort = sort == predicateSort
 
 
-{- | Ensure that a 'StepPattern' is a concrete, normalized term.
+{- | Ensure that a 'TermLike' is a concrete, normalized term.
 
-    If the pattern is not concrete and normalized, the function is
-    'NotApplicable'.
+If the pattern is not concrete and normalized, the function is
```
virgil-serbanuta: Unfinished sentence.

kore/src/Kore/Builtin/Map.hs
```diff
@@ -460,6 +463,39 @@ asTermLike builtin =
     element (key, Domain.Value value) =
         mkApplySymbol elementSymbol [key, value]
 
+{- | Convert a Map-sorted 'TermLike' to its internal representation.
+
+The 'TermLike' is unmodified if it is not Map-sorted. @internalize@ only
+operates at the top-most level, it does not descend into the 'TermLike' to
+internalize subterms.
```
virgil-serbanuta: This seems a bit hard to understand. By default I would assume that, if I'm trying to normalize `concat(alpha, concat(beta, gamma))`, then this won't descend into the `concat(beta, gamma)` subterm, it will only normalize the first concat, which is at the top-most level.

However, my guess is that this will descend into the inner `concat` to normalize.

PR #811, opened 2019-07-24
Test for introducing definedness condition in function application

ana-pantilie:
Will probably move this to `imp`. Requires updated K nightly.

###### Reviewer checklist

---


ttuegel:
This is so simple that it probably should be a unit test.

Issue #714, opened 2019-06-11
All-path reachability claims


ttuegel:
We have [preliminary support](https://github.com/kframework/kore/blob/master/kore/src/Kore/AllPath.hs) for proving all-path reachability claims. This has taken a lower priority for several iterations because the semantics we work with most often are deterministic and proving one-path reachability claims is sufficient. Now we need to finish this feature.

- [x] Connect all-path driver to symbolic execution engine.
- [x] Refactor one-path prover to use the same driver.
- [x] Refactor bounded model checker to use the same driver.
- ~Connect `kore-repl` to the all-path prover.~
- ~Implement an interface with the frontend to choose between one-path and all-path proofs.~

PR #813, opened 2019-07-24
Refactor one-path on new class Goal

ttuegel:
@ana-pantilie This is the beginning of the refactoring I intended to use for the one-path, all-path, and model checking modules. Each type of proof has its own goal type, and we would write an instance of the new `Goal` class for each claim type. All three modules could then use the same `transitionRule` (currently defined in `Kore.AllPath`) and just have different `Strategy`s. Right now, `Kore.OnePath.Step.transitionRule` implements the methods of `Goal` for one-path reachability claims, so you should start by writing that instance.

See also: #714

###### Reviewer checklist

---


kore/src/Kore/Repl.hs
```diff
@@ -61,12 +62,13 @@ import           Kore.Unification.Procedure
...
 -- execution of proofs. Currently works via stdin/stdout interaction.
 runRepl
-    :: forall claim m
+    :: forall claim axiom m
     .  MonadSimplify m
     => MonadIO m
     => MonadCatch m
     => Claim claim
-    => [Axiom]
+    => axiom ~ Rule claim
```
ttuegel: With `axiom ~ Rule claim`, can we remove the second parameters of `ReplState` and `Config`?

kore/src/Kore/Step/Result.hs
```diff
@@ -18,7 +18,6 @@ module Kore.Step.Result
     , mapConfigs
     , traverseConfigs
     ) where
-
```
ttuegel: The blank line should stay.

kore/src/Kore/Step/Step.hs
```diff
@@ -280,6 +284,7 @@ unifyRules
 unifyRules unificationProcedure initial rules =
     Monad.Unify.gather $ do
         rule <- Monad.Unify.scatter rules
+        -- traceM . unparseToString . RewriteRule $ rule
```
ttuegel:
```suggestion
```

kore/src/Kore/Repl.hs
ana-pantilie: This would probably mean a lot of changes in the `repl`. Maybe it would be best if I do this in a follow-up PR, so I don't block the merging of the other PRs?

kore/test/Test/Kore/AllPath.hs
```diff
@@ -15,7 +15,7 @@ import qualified Data.Sequence as Seq
 import           GHC.Stack
                  ( HasCallStack )
 
-import qualified Kore.AllPath as AllPath
+-- import qualified Kore.AllPath as AllPath
```
virgil-serbanuta: Maybe this should be removed?

PR #822, opened 2019-07-29
Add ceils to predicates before SMT

### LINKED ISSUE #748: Simplifying undefined functions
Some of the KEVM proofs get stuck on patterns of the form
```
f(x) ∧ ¬⌈f(x)⌉
```
which we should be able to simplify.

virgil-serbanuta:
Fixes #748 

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Simplification/OrPattern.hs
+++ b/kore/src/Kore/Step/Simplification/OrPattern.hs
@@ -5,11 +5,22 @@ License     : NCSA
 -}
 module Kore.Step.Simplification.OrPattern
     ( simplifyPredicatesWithSmt
+    , filterMultiOrWithTermCeil
     ) where
 
+import           Kore.Internal.Conditional
+                 ( Conditional )
+import qualified Kore.Internal.Conditional as Conditional
+import           Kore.Internal.MultiOr
+                 ( MultiOr )
 import qualified Kore.Internal.MultiOr as MultiOr
 import           Kore.Internal.OrPattern
                  ( OrPattern )
+import           Kore.Internal.Pattern
+                 ( Pattern )
+import qualified Kore.Internal.Pattern as Pattern
+import qualified Kore.Step.Simplification.Ceil as Ceil
+                 ( makeEvaluateTerm )
 import           Kore.Step.Simplification.Data
                  ( MonadSimplify )
 import qualified Kore.Step.Simplification.Data as BranchT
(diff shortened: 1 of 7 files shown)


kore/src/Kore/Step/Simplification/OrPattern.hs
```diff
@@ -36,4 +47,43 @@ simplifyPredicatesWithSmt unsimplified = do
...
+        , SortedVariable variable
+        , Unparse variable
+        )
+    => OrPattern variable -> simplifier (OrPattern variable)
+filterMultiOrWithTermCeil unfiltered = do
+    wrapped <- wrapAndAddCeils unfiltered
+    filtered <- SMT.Evaluator.filterMultiOr wrapped
+    return (fmap unwrap filtered)
+  where
+    wrapAndAddCeils
```
ttuegel: When we add the `\ceil(term)`, why doesn't the result become an `OrPredicate`?

kore/test/Test/Kore/Step/Simplification/OrPattern.hs
```diff
@@ -0,0 +1,81 @@
...
+                ]
+        actual <- filterOr input
+        assertEqualWithExplanation "" expected actual
+    ]
+
+
+filterOr
+    :: [Pattern Variable]
+    -> IO [Pattern Variable]
+filterOr patterns =
```
ttuegel: Personally, I would write this as
```.hs
    fmap MultiOr.extractPatterns
    . SMT.runSMT SMT.defaultConfig emptyLogger
    . evalSimplifier mockEnv
    . filterMultiOrWithTermCeil
    . MultiOr.make
```
but I leave that entirely up to you.

kore/src/Kore/Step/Simplification/OrPattern.hs
virgil-serbanuta: Well, when we compute `\ceil(term)`, the result is, indeed, a predicate (i.e. `ceils` below is an `OrPredicate`).

However, I need to remove the results which become bottom because of that `ceil`, so I'm transforming a `term + predicate` into a `(term + predicate) + (predicate and ceil)`, i.e. with `wrapAndAddCeil` I'm transforming a `Pattern` into a `Conditional (Pattern)`, and with `wrapAndAddCeils` I'm transforming an `OrPattern` into a `MultiOr (Conditional (Pattern))`.

I can then use the SMT to filter this `MultiOr`, then I can extract the term (which is the original `Pattern`) to get the filtered `OrPattern` that I need to return.

The other option would have been to return a `MultiOr (Pattern, Predicate)`, but I can't see an easy way to make this work with just `MultiOr Predicate`.

ttuegel: Oh, I see: the outer `Conditional` contains the `\ceil`, and also the predicates of the inner `Pattern`, is that right?

virgil-serbanuta: Yes.

PR #835, opened 2019-08-02
Incremental matching

ttuegel:
Just a unit test so far.

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Axiom/Matcher.hs
+++ b/kore/src/Kore/Step/Axiom/Matcher.hs
@@ -39,7 +39,7 @@ import qualified Data.Sequence as Seq
 import qualified GHC.Generics as GHC
 
 import           Kore.Attribute.Pattern.FreeVariables
-import qualified Kore.Builtin.AssociativeCommutative as Ac
+import qualified Kore.Builtin as Builtin
 import qualified Kore.Builtin.List as List
 import qualified Kore.Domain.Builtin as Builtin
 import           Kore.Internal.MultiAnd
@@ -364,24 +364,7 @@ substitute variable termLike = do
     isIndependent = not . any (hasFreeVariable variable)
     subst = Map.singleton variable termLike
     substitute2 = fmap substitute1
-    substitute1 = renormalize . TermLike.substitute subst
-
-{- | Renormalize builtin types after substitution.
- -}
-renormalize
-    :: MatchingVariable variable
-    => TermLike variable -> TermLike variable
-renormalize termLike =
-    case termLike of
-        BuiltinMap_ internalMap ->
-            Lens.traverseOf (field @"builtinAcChild") Ac.renormalize internalMap
-            & maybe bottom' mkBuiltinMap
-        BuiltinSet_ internalSet ->
-            Lens.traverseOf (field @"builtinAcChild") Ac.renormalize internalSet
-            & maybe bottom' mkBuiltinSet
-        _ -> termLike
-  where
-    bottom' = mkBottom (termLikeSort termLike)
+    substitute1 = Builtin.renormalize . TermLike.substitute subst
 
 occursCheck
     :: (MatchingVariable variable, MonadUnify unifier)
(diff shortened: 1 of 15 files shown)


kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -9,93 +9,494 @@ Stability   : experimental
...
+    return ()
+  where
+    isIndependent = not . any (hasFreeVariable variable)
+    subst = Map.singleton variable termLike
+    substitute2 = fmap substitute1
+    substitute1 = renormalize . TermLike.substitute subst
+
+{- | Renormalize builtin types after substitution.
+ -}
+renormalize
```
virgil-serbanuta: Shouldn't this be in `Kore.Domain.Builtin` or something like that?

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -9,93 +9,494 @@ Stability   : experimental
...
+            Lens.traverseOf (field @"builtinAcChild") Ac.renormalize internalMap
+            & maybe bottom' mkBuiltinMap
+        BuiltinSet_ internalSet ->
+            Lens.traverseOf (field @"builtinAcChild") Ac.renormalize internalSet
+            & maybe bottom' mkBuiltinSet
+        _ -> termLike
+  where
+    bottom' = mkBottom (termLikeSort termLike)
+
+occursCheck
```
virgil-serbanuta: Would it be better to put this in `Kore.Unification.SubstitutionNormalization` near `nonSimplifiableSortResult`?

kore/test/Test/Kore/Step/Function/Integration.hs
```diff
@@ -537,7 +539,7 @@ test_functionIntegration =
 test_Nat :: [TestTree]
 test_Nat =
     [ plus zero varN `matches` plus zero one  $ "plus(0, N) ~ plus(0, 1)"
-    , plus (succ varM) varN `notMatches` plus zero one  $ "plus(Succ(M), N) !~ plus(0, 1) "
+    , plus (succ varM) varN `unsolved` plus zero one  $ "plus(Succ(M), N) !~ plus(0, 1) "
```
virgil-serbanuta: Although it wasn't grammatically correct, `notMatches` was almost a verb, so it made the code readable. Why is `unsolved` better?

ttuegel: I changed the name because I got confused about halfway through this pull request about whether or not the matcher can ever return `\bottom`. It can't, and I changed the name to emphasize that we are checking for "no solution".

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -9,93 +9,494 @@ Stability   : experimental
...
+    -> MaybeT (MatcherT variable unifier) ()
+matchApplication
+    (Pair term1@(App_ symbol1 children1) term2@(App_ symbol2 children2))
+
+  -- Match identical symbols.
+  | symbol1 == symbol2 =
+    Foldable.traverse_ push (zipWith Pair children1 children2)
+
+  -- Reject non-matching constructors.
+  | Symbol.isConstructor symbol1 || Symbol.isConstructor symbol2 = empty
```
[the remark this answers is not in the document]
ttuegel: You're right, this check is totally redundant. (See my other comment about "no solution" vs `\bottom`.)

kore/src/Kore/Step/Axiom/Matcher.hs
ttuegel: Well, now I'm not so sure about the implementation. In unification, we throw an error for `x = f(x)` substitutions (`f` not a constructor) because we could, possibly, handle those. For matching, on the other hand, we probably want to say intentionally that `x` does not match `f(x)`; in that case this would be
```.hs
occursCheck variable termLike = (Monad.guard . not) (hasFreeVariable variable termLike)
```
in which case it should probably not be used outside the matcher.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -9,93 +9,494 @@ Stability   : experimental
...
-        , SortedVariable variable
-        , MonadUnify unifier
-        )
+ -}
+matchOne
+    :: (MatchingVariable variable, MonadUnify unifier)
+    => Pair (TermLike variable)
+    -> MatcherT variable unifier ()
+matchOne pair =
+    (   matchEqual       pair
```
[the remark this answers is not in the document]
ttuegel: I was planning to implement matching for (at least) `\exists` and `\equals` because my next task depends on it, so I didn't want to cover all the pattern heads, but this covers domain values, literals, and builtins. And if two patterns are literally equal, then they must match. But, I should probably single out the cases that we have tests for right now.

PR #836, opened 2019-08-05
Fix default logging options

### LINKED ISSUE #830: Fix default logging options
- [x] The default log level is too high; should be `warning`.
- [x] Most of the log entries at `info` should be at `debug` level.
- [x] The default log output should be `stderr`.
- [x] The log output `stdout` shouldn't exist because it conflicts with `--output`.

vladciobanu:
Fixes #830 

There is one more `logInfo` remaining in the `BMC`, but that looked like a summary so I thought `Info` was appropriate.

###### Reviewer checklist

---


kore/src/Kore/Logger/Output.hs
```diff
@@ -104,7 +104,7 @@ withLogger KoreLogOptions { logType, logLevel, logScopes } cont =
         LogNone     ->
             cont mempty
         LogStdOut   ->
```
ttuegel: Should `LogStdOut` change to `LogStdErr` correspondingly?

vladciobanu: Definitely. I just read it but not see it, for some reason.

PR #838, opened 2019-08-06
Avoid redundant constraints.

### LINKED ISSUE #705: Avoid redundant constraints applying rules that produce one child
When applying a rule produces only one child pattern, we know that the initial constraint subsumes the constraints on the rule; therefore, it is redundant to add the new constraints.

vladciobanu:
Fixes #705 

###### Reviewer checklist

---


vladciobanu:
This is definitely the wrong way to do it. Back to the drawing board.

Issue #812, opened 2019-07-24
Stuck on remainders of abstract cells


ttuegel:
Sometimes an intermediate step of a proof touches a cell which is abstract in the configuration and we are left with a remainder that has the form
```
¬ ∃ S. X = C(S)
```
where `C` is the only constructor in its sort (the cell constructor). The constraint is not sent to Z3.

ttuegel:
Questions:

1. Could we send such a constraint to Z3? It is a predicate, so maybe that would work.
1. Failing that, could we instantiate the configuration at initialization to replace cell variables? For some reason, I have it in my head that the Java backend does this.

ttuegel:
@virgil-serbanuta wrote a document describing how we might deal with `¬ ∃` remainders in general: https://docs.google.com/document/d/1Uo7NdZnb8xBVWlaVW1aoTsI7NzRybgyDU4XOnwQaF-c/edit?usp=sharing

ttuegel:
@yzhang90 realized that we can prove the required lemma (`∃ S. X = C(S)`) using the constructor axioms. I believe we can already apply this lemma by matching. I think that the most straightforward way to handle this would be to add these single-constructor-sort lemmas to the definition as a separate pass during startup. This solution has several benefits: it takes advantage of features that are already implemented, the implementation is modular, and it narrowly targets a specific problem. A disadvantage I see is that we may find out later that we need to solve the problem more generally, but I'm satisfied to wait until we know more to tackle the (more difficult) general problem.

ttuegel:
- [ ] Match patterns incrementally
- [ ] Match \equals patterns
- [ ] Match \exists patterns
- [ ] Apply selected simplification axioms
- [ ] Add single-constructor lemmas to definitions

PR #844, opened 2019-08-07
Match \exists and \forall

ttuegel:
See also: #812 

###### Reviewer checklist

- [x] Style conformance: `stylish-haskell`

---


kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -171,14 +183,50 @@ matchEqualHeads (Pair (DV_ _ dv1) (DV_ _ dv2)) =
...
+matchForall _ = empty
+
+matchBinder
+    :: (MatchingVariable variable, MonadUnify unifier)
+    => Binder variable (TermLike variable)
+    -> Binder variable (TermLike variable)
+    -> MaybeT (MatcherT variable unifier) ()
+matchBinder (Binder variable1 term1) (Binder variable2 term2) = do
+    Monad.guard (sort1 == sort2)
+    refreshed1 <- refreshVariable variable1
```
virgil-serbanuta: Something that is not obvious here is that `refreshVariable` generates a new variable that is globally unique, and that, because of that, bound variables are added to `avoiding` and never removed. Could you add a comment about that?

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -171,14 +183,50 @@ matchEqualHeads (Pair (DV_ _ dv1) (DV_ _ dv2)) =
...
+    matchBinder (Binder variable1 term1) (Binder variable2 term2)
+matchForall _ = empty
+
+matchBinder
+    :: (MatchingVariable variable, MonadUnify unifier)
+    => Binder variable (TermLike variable)
+    -> Binder variable (TermLike variable)
+    -> MaybeT (MatcherT variable unifier) ()
+matchBinder (Binder variable1 term1) (Binder variable2 term2) = do
+    Monad.guard (sort1 == sort2)
```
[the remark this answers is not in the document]
ttuegel: Yes, this happens any time we try to match binders with the same result sort, but different variable sorts, i.e. `\exists{S}(x:X, ...)` versus `\exists{S}(y:Y, ...)`.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -555,14 +557,17 @@ bindVariable variable = do
     field @"bound" %= Set.insert variable
     field @"avoiding" %= Set.insert variable
 
-{- | Generate a fresh name for the variable, if it shadows another name.
+{- | Lift a (bound) variable to the top level by with a globally-unique name.
```
virgil-serbanuta: by with?

PR #847, opened 2019-08-08
Remove redundant constraints take 3

### LINKED ISSUE #705: Avoid redundant constraints applying rules that produce one child

vladciobanu:
Fixes #705 

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -229,7 +229,8 @@ unifyRule
     let
         RulePattern { left = ruleLeft } = rule'
     unification <- unifyPatterns ruleLeft initialTerm
-    -- Combine the unification solution with the rule's requirement clause.
+    -- Combine the unification solution with the rule's requirement clause,
+    -- unless it's rendundant to do so.
     let
         RulePattern { requires = ruleRequires } = rule'
         requires' = Predicate.fromPredicate ruleRequires
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -41,6 +41,8 @@ import qualified Control.Monad as Monad
 import qualified Control.Monad.State.Strict as State
 import qualified Control.Monad.Trans.Class as Monad.Trans
 import qualified Data.Foldable as Foldable
+import           Data.Function
+                 ( on, (&) )
 import qualified Data.List as List
 import qualified Data.Map.Strict as Map
 import qualified Data.Sequence as Seq
@@ -67,12 +69,15 @@ import           Kore.Internal.TermLike as TermLike
 import           Kore.Logger
                  ( LogMessage, WithLog )
 import qualified Kore.Logger as Log
+import qualified Kore.Predicate.Predicate as Syntax
 import qualified Kore.Step.Remainder as Remainder
 import qualified Kore.Step.Result as Step
 import           Kore.Step.Rule
                  ( RewriteRule (..), RulePattern (RulePattern) )
 import qualified Kore.Step.Rule as Rule
 import qualified Kore.Step.Rule as RulePattern
+import           Kore.Step.SMT.Evaluator
+                 ( evaluate )
 import qualified Kore.Step.Substitution as Substitution
 import qualified Kore.TopBottom as TopBottom
 import qualified Kore.Unification.Substitution as Substitution
@@ -228,8 +233,32 @@ unifyRule
     let
         RulePattern { requires = ruleRequires } = rule'
         requires' = Predicate.fromPredicate ruleRequires
-    unification' <- Substitution.normalizeExcept (unification <> requires')
+    unification' <- removeRedundancy unification requires'
     return (rule' `Conditional.withCondition` unification')
+  where
+    removeRedundancy unificationSolution requires = do
+        let predicate =
+                (makePredicate `on` Predicate.toPredicate)
+                    unificationSolution requires
+        isPreRedundant <- evaluate predicate
+        let nonRedundantPredicate =
+                case isPreRedundant of
+                    Just False -> unificationSolution
+                    _         -> unificationSolution <> requires
+        Substitution.normalizeExcept nonRedundantPredicate
+
+    -- | ¬(⌈L ∧ φ⌉ ∧ P → Pre)
+    makePredicate unificationSolution requires =
+        Syntax.makeNotPredicate
+            $ Syntax.makeImpliesPredicate
+                (Syntax.makeAndPredicate
+                     unificationSolution
+                     initialPredicate
+                )
+                requires
+
+    initialPredicate =
+        Conditional.toPredicate $ Conditional.withoutTerm initial
 
 unifyRules
     ::  forall unifier variable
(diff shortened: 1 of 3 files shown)


vladciobanu:
It might be worth it to move the code I added in `Step.Step` to a different module, but I would like a review now that focuses on the changes to the tests.

kore/src/Kore/Step/Step.hs
```diff
@@ -224,8 +229,26 @@ unifyRule
...
+                    unificationSolution requires
+        isPreRedundant <- evaluate predicate
+        let nonRedundantPredicate =
+                case isPreRedundant of
+                    Just True -> unificationSolution
+                    _         -> unificationSolution <> requires
+        Substitution.normalizeExcept nonRedundantPredicate
+
+    -- | ⌈L ∧ φ⌉ ∧ P → Pre
+    makePredicate unificationSolution requires =
```
ttuegel: The comment helps, but the name `makePredicate` is a little vague: what kind of predicate does this make?

kore/src/Kore/Step/Step.hs
```diff
@@ -224,8 +229,26 @@ unifyRule
     let
         RulePattern { requires = ruleRequires } = rule'
         requires' = Predicate.fromPredicate ruleRequires
-    unification' <- Substitution.normalizeExcept (unification <> requires')
+    unification' <- removeRedundancy unification requires'
     return (rule' `Conditional.withCondition` unification')
+  where
+    removeRedundancy unificationSolution requires = do
```
ttuegel: The name `removeRedundancy` is unclear about what is redundant. Maybe, `removeRedundantRequires`?

kore/test/Test/Kore/OnePath/Step.hs
```diff
@@ -312,19 +312,9 @@ test_onePathStrategy =
                     $ Mock.f Mock.b
             ]
         assertEqualWithExplanation ""
-            [ Stuck Conditional
```
ttuegel: This change does not seem reasonable to me. We cannot unconditionally prove the claim `constr10(b) => a` given only the axiom `constr10(b) => a requires f(b) = c`.

kore/test/Test/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -80,18 +82,11 @@ test_definitionEvaluation =
                     { results =
                         OrPattern.fromPattern Conditional
                             { term = Mock.g Mock.a
-                            , predicate = requirement
+                            , predicate = makeTruePredicate
```
ttuegel: Where in the pre-condition does it imply that `f(a) = g(b)`?

kore/test/Test/Kore/Step/Function/Integration.hs
```diff
@@ -1028,5 +1032,42 @@ evaluate
...
+env =
+    Env
+        { metadataTools       = tools
+        , simplifierTermLike  = Mock.termLikeSimplifier
+        , simplifierPredicate = Mock.predicateSimplifier
+        , simplifierAxioms    = Mock.axiomSimplifiers
+        }
+  where
+    tools =
+        Mock.makeMetadataTools
```
ttuegel: IIRC, `Mock.makeMetadataTools` does some things that are sketchy, and I intentionally avoided it for this module. Why is this necessary?

kore/test/Test/Kore/Step/Axiom/UserDefined.hs
```diff
@@ -107,14 +107,14 @@ test_userDefinedFunction =
     , testCase "Cannot apply step with unsat axiom pre-condition" $ do
         let expect =
                 AttemptedAxiom.Applied AttemptedAxiomResults
-                    { results = OrPattern.fromPatterns []
-                    , remainders = OrPattern.fromPatterns
+                    { results = OrPattern.fromPatterns
                         [ Conditional
-                            { term = Mock.functionalConstr10 (mkVar Mock.x)
+                            { term = Mock.functionalConstr11 (mkVar Mock.x)
```
ttuegel: The axiom below cannot apply because it `requires False`.

kore/test/Test/Kore/Step/Step.hs
```diff
@@ -546,12 +546,6 @@ test_applyRewriteRule_ =
         actual <- applyRewriteRule_ initial axiomEnsuresBottom
         assertEqualWithExplanation "" expect actual
 
-    , testCase "rule a => b requires \\bottom" $ do
```
ttuegel: Why was this removed?

kore/test/Test/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -272,48 +262,6 @@ test_firstFullEvaluation =
                 )
                 (Mock.functionalConstr10 (mkVar Mock.x))
         assertEqualWithExplanation "" expect actual
-    , testCase "Error when not fully rewriting" $ do
```
ttuegel: These errors should not have gone away.

kore/test/Test/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -352,18 +300,11 @@ test_simplifierWithFallback =
                     { results = OrPattern.fromPatterns
                         [ Conditional
                             { term = Mock.g Mock.a
-                            , predicate = requirement
```
ttuegel: I don't think these tests should have changed, either.

kore/src/Kore/Step/Step.hs
virgil-serbanuta: Actually, wouldn't `addRequiresUnlessRedundant` be a better name? This does not remove anything...

kore/src/Kore/Step/Step.hs
```diff
@@ -224,8 +229,26 @@ unifyRule
...
+        let predicate =
+                (makePredicate `on` Predicate.toPredicate)
+                    unificationSolution requires
+        isPreRedundant <- evaluate predicate
+        let nonRedundantPredicate =
+                case isPreRedundant of
+                    Just True -> unificationSolution
```
virgil-serbanuta: So, I'm a bit confused by this. I think we only implemented the "Just False" case, see https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/SMT/Evaluator.hs#L143
so I would expect this to be a no-op. On the other hand, your change seems to make a difference for the tests below.

So, I'm not sure what comment to make here, but this does not seem right.

vladciobanu:
@ttuegel Yep, you are exactly right. I somehow convinced myself that those were right, but they were obviously not. The problem was that I tried to be cute with `Function.&` and accidentally reversed the implication of the predicate, thus making it `Pre → ⌈L ∧ φ⌉ ∧ P` which turns out to be true pretty often.

Additionally, @virgil-serbanuta was correct: we only have the `Just False` case for `Unsat`, so I needed to negate the predicate anyway.

I rebased/force push-ed the new version which looks a lot better now, I think (no change needed to other tests, added a test that is failing on `master` but working here; can you think of other tests I should add?).

@virgil-serbanuta also had an interesting idea I am inclined to take on: what if we change `evaluate` (this tricked me since it returns `Maybe Bool` but it only really returns `Nothing` or `Just False`) to be more precise? For example, we could add some argument that specifies whether we want to prove something's always true or always false and it would negate the predicate itself if needed...?

PR #850, opened 2019-08-09
Apply axioms early

ttuegel:
User-defined axioms are now applied at the start of `TermLike` simplification, and we alternate user axioms and internal simplification until the result stabilizes. The `StabilizeT` monad transformer is introduced to track when the computation stabilizes.

Other changes:

1. ~Added `Control.Monad.Stabilize` for the `StabilizeT` transformer.~
1. Added `Kore.Step.Function.Evaluator.evaluateOnce`.

###### Reviewer checklist

- [x] Style conformance: `stylish-haskell`

---


ttuegel:
I actually now feel like I made the stabilizer code more complicated than it needs to be for a simple state machine, and I intend to rewrite that part.

PR #854, opened 2019-08-12
matching equals patterns

andreiburdusa:
###### Reviewer checklist

---


kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -131,6 +131,22 @@ test_matcherEqualHeads =
                 (mkCeil_ (Mock.plain10 Mock.a))
         assertEqualWithExplanation "" expect actual
 
+    , testCase "Equals" $ do
```
ttuegel: Please look at `test_matching_Exists` and add a `test_matching_Equals`. We should include at least the following test cases:

1. The patterns match, with a substitution.
1. The patterns do not match because of a direct conflict, e.g. `\equals(x, a)` versus `\equals(y, b)`.
1. The patterns do not match because of an indirect conflict, e.g. `\equals(x, x)` versus `\equals(a, b)`.

PR #867, opened 2019-08-16
Normalize inputs to SMT

ttuegel:
###### Reviewer checklist

---

@vladciobanu We need to normalize the implication check which you added. However, this causes the concrete execution tests to fail and the unit tests to loop forever.

### DIFF
--- a/kore/src/Kore/Internal/Conditional.hs
+++ b/kore/src/Kore/Internal/Conditional.hs
@@ -17,6 +17,7 @@ module Kore.Internal.Conditional
     , splitTerm
     , toPredicate
     , Kore.Internal.Conditional.mapVariables
+    , isNormalized
     ) where
 
 import           Control.DeepSeq
@@ -357,3 +358,14 @@ mapVariables
 
 splitTerm :: Conditional variable term -> (term, Conditional variable ())
 splitTerm patt@Conditional { term } = (term, withoutTerm patt)
+
+{- | Is the condition normalized?
+
+The 'Substitution' must be normalized and must have been substituted into the
+'Predicate'.
+
+ -}
+isNormalized :: Ord variable => Conditional variable term -> Bool
+isNormalized Conditional { predicate, substitution } =
+    Substitution.isNormalized substitution
+    && Predicate.isFreeOf predicate (Substitution.variables substitution)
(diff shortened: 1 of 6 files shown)


ttuegel:
I've had this bouncing around in my head all weekend. :slightly_frowning_face: 

We call `normalizeExcept` all over the `Kore.Step.Step` module; it should be safe, and I can't think of any reason it would loop. There is one thing, though: We are normalizing the predicate *before* we know if the rule applies. We should not do the implication check or any other checks before we know the rule applies. At first, it seems like we might want to do the check as early as possible to avoid duplicating work. Actually, the implication check can only duplicate work if the rule applies, and _most rules don't apply_. Delaying the implication check saves work most of the time by avoiding the implication check altogether.

virgil-serbanuta:
I agree that most rules don't apply, but how about rules that unify? I would expect those to mostly apply.

ttuegel:
> I agree that most rules don't apply, but how about rules that unify? I would expect those to mostly apply.

Assuming the rule has non-trivial side conditions, I still don't think this is true. If a rule has a non-trivial side condition, I expect that usually there will be one or more other rules which cover the cases that the side condition would reject. I think, usually, if a rule with side conditions unifies, it applies at most 50% of the time. (I don't worry if the side condition is `\top` because that will be trivial anyway.)

There is another concern: Based on the errors I see when I try to enable this normalization step, I *think* that sometimes the unifier returns a partial result (one that contains unsolved `\ceil(_ ∧ _)` or `\equals(_, _)` patterns) which should be rejected by some combination of initial and side conditions; if we attempt a normalization problem involving the partial unifier, such as the implication check, then we trip over these unsolved partial results.

kore/src/Kore/Step/Step.hs
```diff
@@ -235,34 +232,9 @@ unifyRule
     -- Combine the unification solution with the rule's requirement clause,
     -- unless it's rendundant to do so.
     let
-        RulePattern { requires = ruleRequires } = rule'
-        requires' = Predicate.fromPredicate ruleRequires
-    unification' <- addRequiresUnlessRedundant unification requires'
+        RulePattern { requires } = rule'
+    unification' <- Substitution.normalizeExcept (unification <> Predicate.fromPredicate requires)
```
virgil-serbanuta: 80 chars?

PR #911, opened 2019-08-29
Reformat the world with Ormolu

ttuegel:
Adopting an automatic formatting tool is an extreme step, fraught with
trade-offs, but it mitigates the risk of accidents; specifically, it reduces the
potential damage caused by an accident. This is a soft suggestion: I'm offering
it for consideration because it's easy to implement, not because I have strong
feelings about it.

If we did adopt a formatting tool, Ormolu is the obvious choice:

- It is the only tool that will format 100% of the code without occasionally
  mangling anything.
- The dictated format is relatively diff-friendly.
- The result is just ugly enough that everyone will hate it a little, but not so
  ugly that anyone will hate it a lot.

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Control/Monad/Counter.hs
+++ b/kore/src/Control/Monad/Counter.hs
@@ -11,45 +11,55 @@ concrete implementation of the class.
 
 -}
 module Control.Monad.Counter
-    ( -- * Class
-      MonadCounter (..)
-    , incrementState
-    , findState
-    , module Numeric.Natural
-      -- * Implementation
-    , CounterT (..)
-    , runCounterT, evalCounterT
-    , Counter
-    , runCounter, evalCounter
-    ) where
+  ( -- * Class
+    MonadCounter (..),
+    incrementState,
+    findState,
+    module Numeric.Natural,
+    -- * Implementation
+    CounterT (..),
+    runCounterT,
+    evalCounterT,
+    Counter,
+    runCounter,
+    evalCounter
+    )
+where
 
-import           Control.Applicative
-                 ( Alternative )
-import           Control.Monad
-                 ( MonadPlus )
+import Control.Applicative
+  ( Alternative
+    )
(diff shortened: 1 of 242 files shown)


vladciobanu:
I skimmed it and noticed two things that bother me:
- the way imports are formatted, they no longer line up nicely (but I could live with it)
- using trailing commas instead of beginning-of-line commas -- doesn't this make diffs more noisy?

ttuegel:
> using trailing commas instead of beginning-of-line commas -- doesn't this make diffs more noisy?

Yes, it does make diffs slightly more noisy, although it makes them much nicer than what we do now. It's also a little funny: trailing commas is uncommon in "wild" Haskell code, but elsewhere the author argues against other formatting changes because _they_ are uncommon in the wild. :man_shrugging:

ttuegel:
> I assume that the size of the indentation level is configurable. Why two-space indentation? Is it because Ormolou does not use half-indentation, so we lost most of the incentive to use 4 spaces?

Nothing about Ormolu is configurable; it wants to be the `rustfmt` or `gofmt` of Haskell, the One True Style. I think the incentive to use two spaces is mostly what you said: it doesn't use half indentation, so there isn't a reason to use four.

(I looked at its source code. It seems pretty hackable. If we really wanted to commit to using a formatter, but these annoyances are too much, we could fork the project and fix them without excessive effort. But that raises the bar on, "Do we really want to commit to this?")

> I guess that this formatter does not split lines above a certain length, right?

One interesting thing about Ormolu, compared to other formatters, is that it tries to take a cue from the programmer about line splitting. So, if you write

```.hs
if ... then ... else ...
```

it will try to keep it on one line, even if it doesn't fit (I think).
But if you write

```.hs
if ...
    then ...
    else ...
```

it will not collapse it onto one line, even if it would fit.

PR #924, opened 2019-09-05
Child and top-level predicates

virgil-serbanuta:
@traiansf  said that child predicates (without the top-level predicate) reduce the SDIV time from >20 min (timeout) to ~9 min

###### Reviewer checklist

---

### DIFF
--- a/kore/src/Kore/Step/Axiom/Evaluate.hs
+++ b/kore/src/Kore/Step/Axiom/Evaluate.hs
@@ -22,6 +22,9 @@ import Kore.Internal.TermLike
     ( TermLike
     )
 import qualified Kore.Internal.TermLike as TermLike
+import qualified Kore.Predicate.Predicate as Syntax
+    ( Predicate
+    )
 import Kore.Step.Axiom.Matcher
     ( matchIncremental
     )
@@ -59,10 +62,12 @@ evaluateAxioms
     .  (SimplifierVariable variable, MonadSimplify simplifier)
     => [EqualityRule Variable]
     -> TermLike variable
+    -> Syntax.Predicate variable
     -> simplifier (AttemptedAxiom variable)
 evaluateAxioms
     definitionRules
     patt
+    predicate
   | any ruleIsConcrete definitionRules
   , not (TermLike.isConcrete patt)
   = return AttemptedAxiom.NotApplicable
@@ -91,9 +96,20 @@ evaluateAxioms
         introduceDefinedness = flip Pattern.andCondition
         markRemainderEvaluated = fmap TermLike.mkEvaluated
 
-    AttemptedAxiomResults
-        <$> OrPattern.simplifyPredicatesWithSmt (Step.gatherResults result)
-        <*> OrPattern.simplifyPredicatesWithSmt (Step.remainders result)
+    simplifiedResults <-
+        OrPattern.simplifyPredicatesWithSmt
+           predicate (Step.gatherResults result)
+
+    simplifiedRemainders <-
+        OrPattern.simplifyPredicatesWithSmt
+            predicate (Step.remainders result)
+
+    Monad.guard (not $ null simplifiedResults)
+
+    return AttemptedAxiomResults
+        { results = simplifiedResults
+        , remainders = simplifiedRemainders
+        }
 
   where
     ruleIsConcrete =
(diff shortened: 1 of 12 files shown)


kore/src/Kore/Step/Simplification/Conditional.hs
```diff
@@ -0,0 +1,51 @@
...
+-}
+simplifyPredicate
+    ::  ( SimplifierVariable variable
+        , MonadSimplify simplifier
+        , WithLog LogMessage simplifier
+        )
+    => Conditional variable term
+    -- ^ Contains the condition to be evaluated.
+    -> BranchT simplifier (Conditional variable term)
+simplifyPredicate Conditional {term, predicate, substitution} = do
```
andreiburdusa: Could you explain why you defined this function and replaced `Kore.Step.Simplification.Pattern.simplifyPredicate`? They seem to be the same function

virgil-serbanuta: The signature changed. A `Pattern` is a `Conditional` whose term is a `TermLike`, so, when I moved to a generic `term`, I thought I should move the function to a file called `Conditional`.

virgil-serbanuta:
* I also added a commit from traiansf which uses the top-level predicate, since it seemed to be fairly related to what's happening here. We can make that a separate PR if you think it does not make sense this way.

kore/src/Kore/Step/Simplification/OrPattern.hs
```diff
@@ -24,10 +39,27 @@ import qualified Kore.Step.SMT.Evaluator as SMT.Evaluator
...
+    -> OrPattern variable
+    -> simplifier (OrPattern variable)
+simplifyPredicatesWithSmt predicate' unsimplified = do
     simplifiedOrs <- traverse
-        (BranchT.gather . Pattern.simplifyPredicate)
+        (BranchT.gather . Conditional.simplifyPredicate)
         (MultiOr.extractPatterns unsimplified)
-    SMT.Evaluator.filterMultiOr (MultiOr.make (concat simplifiedOrs))
+    let newOrs :: [Conditional variable (Pattern variable)]
+        newOrs = addPredicate . conditionalAsTerm <$> concat simplifiedOrs
```
ttuegel: Could you add a comment here that says what is "new" about the `newOrs`? It looks to me like they have another `Conditional` wrapped around the original `Pattern`, and the outer conditions have the new predicate, is that right?

kore/src/Kore/Step/Simplification/Simplify.hs
```diff
@@ -246,11 +247,21 @@ simplifyTerm
...
+simplifyConditionalTermToOr
+    :: forall variable simplifier
+    .   ( SimplifierVariable variable
+        , MonadSimplify simplifier
+        )
+    => Predicate variable
+    -> TermLike variable
+    -> simplifier (OrPattern variable)
+simplifyConditionalTermToOr predicate termLike = do
+    results <- gather $ simplifyConditionalTerm predicate termLike
```
ttuegel: Does `simplifyConditionalTerm` actually use the predicate now?

traiansf: Yes, it should.

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -22,6 +22,9 @@ import Kore.Internal.TermLike
     ( TermLike
     )
 import qualified Kore.Internal.TermLike as TermLike
+import qualified Kore.Predicate.Predicate as Syntax
```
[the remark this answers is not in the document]
virgil-serbanuta: Well, I was a bit conflicted about that... The main issue was something like this: the substitution is useful when we want to actually substitute something or we want to merge it with another substitution, or something. Sending it separately was suggesting that it was going to be useful, but I actually wanted to send an opaque predicate which was going to be used only as an additional condition by the SMT (of course, it can include the substitution). Maybe I should have also wrapped it in a newtype to make that even more clear.

On the other hand, yes, this is breaking the current convention, so it feels weird. Should I make it an `Internal.Predicate`?

kore/src/Kore/Step/Simplification/OrPattern.hs
virgil-serbanuta: Right. I renamed the variables and added a comment.

kore/src/Kore/Step/Axiom/Evaluate.hs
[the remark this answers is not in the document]
ttuegel: You're right, it is awkward. In fact, we never send a substitution to the solver anyway because we abort if it is not normalized. So, I think keep it the way you have it.

PR #932, opened 2019-09-06
Add BMC tests

yzhang90:
###### Reviewer checklist

---


Makefile
```diff
@@ -51,6 +51,7 @@ coverage_report: test-kore
 
 test-k:
 	$(MAKE) --directory src/main/k/working test-k
+	$(MAKE) --directory src/main/k/in-progress/bmc/example1 test-bmc
```
ttuegel: This prevents `test-bmc` from running in parallel with `test-k`, which is probably not what we want. Could we add a top-level `test-bmc` target?

yzhang90: done

PR #938, opened 2019-09-09
Design doc for combining axioms.

virgil-serbanuta:
###### Reviewer checklist

---

### DIFF
--- a/docs/2019-09-09-Combining-Rewrite-Axioms.md
+++ b/docs/2019-09-09-Combining-Rewrite-Axioms.md
@@ -11,11 +11,11 @@ Summary
 =======
 
 If we have this chain of axioms: `α₁(X₁)⇒β₁(X₁)` … `αₙ(Xₙ)⇒βₙ(Xₙ)` in which
-all the LHS are *functional* and *injection-based* then their combined
-transition is
+all the LHS are *function-like* and *injection-based*, and all the RHS
+except `βₙ(Xₙ)` are *function-like*, then their combined transition is
 
 ```
-⌈β₁(X₁)∧α₂(X₂)⌉ ∧ ⌈β₂(X₂)∧α₃(X₃)⌉ ∧ … ∧ ⌈βₙ₋₁(Xₙ₋₁)∧αₙ(Xₙ)⌉ ∧ (α₁(X₁)→••…•βₙ(Xₙ))
+(⌈β₁(X₁)∧α₂(X₂)⌉ ∧ ⌈β₂(X₂)∧α₃(X₃)⌉ ∧ … ∧ ⌈βₙ₋₁(Xₙ₋₁)∧αₙ(Xₙ)⌉ ∧ α₁(X₁))→••…•βₙ(Xₙ)
 ```
 
 TODO: We may need to figure out how to handle maps, sets, and other structures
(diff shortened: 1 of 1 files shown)


ehildenb:
Not sure if it makes a difference or makes anything easier, but you could also start with the LHS of the first rule in the sequence instead of a single variable `GST:GeneratedTopCell`. Maybe it helps with avoiding remainder calculations.

docs/2019-09-09-Combining-Rewrite-Axioms.md
```diff
@@ -0,0 +1,172 @@
...
+symbolic configuration consisting of a single variable, or we should be able to
+combine rules into a single one. This document details the matching logic
+reasoning and assumptions behind these.
+
+Summary
+=======
+
+If we have this chain of axioms: `α₁(X₁)⇒β₁(X₁)` … `αₙ(Xₙ)⇒βₙ(Xₙ)` in which
+all the LHS are *functional* and *injection-based* then their combined
+transition is
```
ehildenb: I thought maybe we needed the LHS to be _constructors_, but this seems like a weaker condition? Is it still enough?

virgil-serbanuta: If I didn't make any mistake, the only place where we use the constructor properties is when solving `α(Y)=α(Y')` to get `Y'=Y`, and, for that, we only need the injectivity property: we have exactly the same symbols in both terms, so we don't need, say, the no-confusion-different-constructors property, or the no-junk one.

docs/2019-09-09-Combining-Rewrite-Axioms.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+Note that
+```
+P ∧ (a→•b) = (P∧ a) → •P ∧ b
+```
+And, if we also know that `P ∧ b→• P ∧ c`, then we can infer
+`(P∧ a) → ••(P ∧ c)`, i.e. `P ∧ (a→••c)`.
+
+TODO: Does the above hold? Why?
```
traiansf: it holds due to transitivity of the implication and due to the "Framing rule". Say you have `a →• b` and `b →• c`
From `b →• c` you deduce by framing  `• b →• • c`, and then from transitivity of implication and MP the conclusion follows (modulo the extraction of P which we now know it's not correct).

docs/2019-09-09-Combining-Rewrite-Axioms.md
```diff
@@ -0,0 +1,200 @@
...
+TODO: We may need to figure out how to handle maps, sets, and other structures
+based on non-injective symbols.
+
+The general transformation
+==========================
+
+Let us take a configuration `φ(X)` which is functional,
+and which can be a single variable.
+
+Let us also take a configuration `α(X)→•β(X)`, with `α(X)` being functional.
```
traiansf:
```suggestion
Let us also take a rule `α(X)→•β(X)`, with `α(X)` being functional.
```

docs/2019-09-09-Combining-Rewrite-Axioms.md
```diff
@@ -0,0 +1,200 @@
...
+        )
+   // a ∧ b → c is the same as a ∧ b → c ∧ b
+```
+
+When doing normal rewriting, we usually expect to get substitutions for the
+variables in `Y` and `Y’` when computing `⌈α(Y) ∧ φ(X)⌉`, which usually allows
+us to remove these variables. However, when combining rewriting axioms,
+we don’t always get such substitutions, so we need to take a different approach.
+
+Note that the above also work for all patterns `φ` and `α`, not only functional
```
traiansf: From line 38, in order to apply prop 5.24 you need some functionality constraints; and because you apply it both ways, probably both  `φ` and `α` need to be functional.

PR #942, opened 2019-09-10
ExpandAlias, Variable -> UnifiedVariable

### LINKED ISSUE #902: Implement alias expansion
See also: #863

vladciobanu:
###### Reviewer checklist

---

Fixes #902

### DIFF
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -65,6 +65,8 @@ import qualified Kore.Step.Simplification.Data as SimplificationType
                  ( SimplificationType (..) )
 import qualified Kore.Step.Simplification.Data as BranchT
                  ( gather, scatter )
+import           Kore.Step.Simplification.ExpandAlias
+                 ( expandAlias )
 import           Kore.Step.Simplification.NoConfusion
 import           Kore.Step.Simplification.Overloading
 import           Kore.Step.Substitution
@@ -374,7 +376,8 @@ andEqualsFunctions
         )
     => [(SimplificationTarget, TermTransformation variable unifier)]
 andEqualsFunctions = fmap mapEqualsFunctions
-    [ (AndT,    \_ _ _ _ -> boolAnd, "boolAnd")
+    [ (AndT,    \_ _ m s -> expandAlias (maybeTermAnd m s), "expandAlias")
+    , (AndT,    \_ _ _ _ -> boolAnd, "boolAnd")
     , (BothT,   \_ _ _ _ -> equalAndEquals, "equalAndEquals")
     , (EqualsT, \p _ _ _ -> bottomTermEquals p, "bottomTermEquals")
     , (EqualsT, \p _ _ _ -> termBottomEquals p, "termBottomEquals")
--- a/kore/src/Kore/Step/Simplification/ExpandAlias.hs
+++ b/kore/src/Kore/Step/Simplification/ExpandAlias.hs
@@ -1,0 +1,73 @@
+{-|
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+-}
+
+module Kore.Step.Simplification.ExpandAlias
+    ( expandAlias
+    ) where
+
+import           Control.Error
+                 ( MaybeT )
+import           Control.Error.Util
+                 ( nothing )
+import qualified Data.Map as Map
+
+import           Kore.Internal.Alias
+                 ( Alias (..) )
+import           Kore.Internal.Pattern
+                 ( Pattern )
+import           Kore.Internal.TermLike
+                 ( pattern ApplyAlias_, TermLike, Variable, substitute )
+import qualified Kore.Logger as Logger
+import           Kore.Syntax.Variable
+                 ( SortedVariable )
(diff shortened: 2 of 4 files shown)


vladciobanu:
## PR summary

Relevant new code is in the newly created file `ExpandAlias.hs`. Tests were added in `AndTerms` since the new function exported from `ExpandAlias` is used within `andEqualsFunctions`. The rest of the changes are because the `Alias` type change in its `leftAlias` member: from `[Variable]` to `[UnifiedVariable Variable]`.

The `ExpandAlias` module exports a single (unification) function which takes two terms and tries to expand an alias if there is one in either term's head. If there is no alias found, we return `Nothing`. If at least one alias was expanded, we need to continue expanding as much as possible and resume unification with the expanded arguments. In order to do so, after each expansion, we need to recursively call `maybeTermAnd` (which this function takes, partially applied with the simplifiers/etc it requires).

In all tests, I used trivial aliases that ignore their parameter and replace it with `\top`. We already test our substitution algorithm with different tests, so it felt these tests shouldn't fail because of it. Tests that were added:
- a trivial `alias` vs `\top` to test the most basic feature
- expanding aliases simultaneously (both left and right) via `alias` vs `alias`
- expanding nested aliases, `alias1(alias2)` vs `alias`

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -337,7 +339,8 @@ andEqualsFunctions
         )
     => [(SimplificationTarget, TermTransformation variable unifier)]
 andEqualsFunctions = fmap mapEqualsFunctions
-    [ (AndT,    \_ _ _ _ -> boolAnd, "boolAnd")
+    [ (AndT,    \_ _ m s -> expandAlias (maybeTermAnd m s), "expandAlias")
```
ttuegel: Why doesn't `expandAlias` use its `s` argument directly? Why is it hard-coded to `maybeTermAnd`?

kore/src/Kore/Step/Simplification/ExpandAlias.hs
```diff
@@ -0,0 +1,69 @@
...
+        -> TermLike variable
+        -> MaybeT unifier (Pattern variable)
+       )
+    -> TermLike variable
+    -> TermLike variable
+    -> MaybeT unifier (Pattern variable)
+expandAlias recurse t1 t2 = do
+    case (expandSingleAlias t1, expandSingleAlias t2) of
+        (Nothing, Nothing) -> nothing
+        (t1', t2') -> recurse (maybe t1 id t1') (maybe t2 id t2')
```
ttuegel:
```suggestion
        (t1', t2') -> recurse (fromMaybe t1 t1') (fromMaybe t2 t2')
```

kore/src/Kore/Step/Simplification/ExpandAlias.hs
```diff
@@ -0,0 +1,69 @@
...
+
+-- TODO(Vladimir): Check aliases such that the intersection of alias variables
+-- with the *bounded* variables in the rhs is empty (because we can't currently
+-- handle things like \mu.
+substituteWorker
+    :: SubstitutionVariable variable
+    => Alias (TermLike Variable)
+    -> [TermLike variable]
+    -> TermLike variable
+substituteWorker Alias { aliasLeft, aliasRight } children =
```
ttuegel: I think we should enforce the condition that the alias definition and the pattern have the same number of arguments:

```suggestion
substituteWorker Alias { aliasLeft, aliasRight } children =
    assert (length aliasLeft == length children) $
```

I know this is checked by the pattern verifier, I just feel more comfortable enforcing the pre-conditions.

kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -933,8 +939,63 @@ test_andTermsSimplification =
...
+    => SentenceAlias (TermLike Variable)
+    -> TermLike variable
+    -> TermLike variable
+applyAlias' alias arg = applyAlias alias [] [arg]
+
+assertExpectTop :: ([Pattern Variable], Maybe [Pattern Variable]) -> IO ()
+assertExpectTop =
+    assertEqualWithExplanation "" ([expectTop_], Just [expectTop_])
+  where
+    expectTop_ =
```
ttuegel: Why not [`Kore.Internal.Pattern.top`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Internal/Pattern.hs#L162)?

kore/src/Kore/Step/Simplification/AndTerms.hs
vladciobanu: It's because `AndTerm` imports this file, so this file can't also import `AndTerm` to call `maybeTermAnd` itself, so we partially apply it and pass it down.

kore/src/Kore/Step/Simplification/ExpandAlias.hs
vladciobanu: I actually considered it, added them but then figured to check if they are needed, and indeed they are also checked in the pattern verifier so I removed it. I guess it makes sense to keep them here as well, it's not costing us anything.

PR #949, opened 2019-09-11
Recognize prioritized form of rules

andreiburdusa:
---

###### Reviewer checklist


kore/test/Test/Kore/Step.hs
```diff
@@ -189,6 +189,7 @@ rewritesTo :: TestPattern -> TestPattern -> RewriteRule Variable
 rewritesTo left right =
     RewriteRule $ RulePattern
```
ttuegel: Instead of adding the `antiLeft` field, we should refactor these tests to use `Kore.Step.Rule.rulePattern`.

PR #955, opened 2019-09-12
Refactor Kore.Strategies

ana-pantilie:
This PR should provide a more general common interface for types which represent proof goals. An instance of `Kore.Strategies.Goal.Goal` is a goal provable using a `transitionRule` over a `strategy`. The rule, primitive transitions (`Prim`) and proof state types should be determined by the goal type. 

Main changes:
- the `Goal` instance for `OnePathRule` was redefined
- `Kore.Strategies.OnePath.Verification` was moved up one level because it is not specific to one-path
- `Kore.Strategies.OnePath.Actions` was removed, most of the functions had to be moved back to `Kore.Strategies.Goal`
- all one-path and all-path tests were moved to Test.Kore.Strategies

---

###### Reviewer checklist


kore/src/Kore/Strategies/ProofState.hs
```diff
@@ -6,21 +6,39 @@ module Kore.Strategies.ProofState
...
+    | TriviallyValid
+    | DerivePar [rule]
+    | DeriveSeq [rule]
+    deriving (Show)
+
 {- | The state of the all-path reachability proof strategy for @goal@.
  -}
-data ProofState goal
-    = Goal goal
+data ProofState a
```
traiansf: why did `goal` became `a`?

scripts/check.sh
```diff
@@ -16,7 +16,7 @@ fi
 
 stack install stylish-haskell
 changed=()
-for file in $(git diff --name-only $UPSTREAM_BRANCH)
+for file in $(git diff --name-only --diff-filter=d $UPSTREAM_BRANCH)
```
traiansf: why is this?

kore/src/Kore/Strategies/Verification.hs
```diff
@@ -1,17 +1,16 @@
 {-|
-Module      : Kore.OnePath.Verification
-Description : One-path verification
+Module      : Kore.Strategies.Verification
```
ttuegel: You can remove the `Module` and `Discription` lines.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -40,94 +101,136 @@ See also: 'Strategy.pickFinal', 'extractUnproven'
...
+            )
+      where
+        rewrites = rules
+        coinductiveRewrites =
+            OnePathRewriteRule
+            . RewriteRule
+            . getOnePathRule
+            <$> goals
+
+transitionRule0
```
ttuegel: The name of this function did not clarify its purpose for me. Could we rename this to `transitionRuleTemplate`? I would also prefer if it took a record of functions as its argument; this would be easier to use and clearer to document.

Because this depends on `ProofState.ProofState`, should it go live with that type?

scripts/check.sh
ana-pantilie: This is from a previous PR I already merged to master: https://github.com/kframework/kore/pull/972

kore/src/Kore/Strategies/ProofState.hs
ana-pantilie: I did this to keep the notation similar to the one used in `Goal` (in the `OnePathRule` instance the `goal` type only determines that we will use `ProofState.ProofState a`, for any `a`). If you think this is unnecessary/confusing I can change it back.

kore/src/Kore/Strategies/Goal.hs
ana-pantilie: > Because this depends on `ProofState.ProofState`, should it go live with that type?

Sorry, I'm not sure I understand what you mean here. Would you prefer I remove the type equality constraint in that function and use `ProofState.ProofState` directly?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -40,94 +102,195 @@ See also: 'Strategy.pickFinal', 'extractUnproven'
...
+            . getOnePathRule
+            <$> goals
+
+instance SOP.Generic (Rule (OnePathRule variable))
+
+instance SOP.HasDatatypeInfo (Rule (OnePathRule variable))
+
+instance Debug variable => Debug (Rule (OnePathRule variable))
+
+data TransitionRuleTemplate monad goal =
```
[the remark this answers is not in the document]
ana-pantilie: I made this change here: https://github.com/kframework/kore/pull/955/commits/8d74980195f3ed5f99da7adba989c1c1b205da54

It looks cleaner to me this way, but if @ttuegel has any objections I'll revert this.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -237,26 +351,250 @@ onePathFollowupStep claims axioms =
...
-    derivePar = OnePath.derivePar
-
-    deriveSeq = OnePath.deriveSeq
-
-instance SOP.Generic (Rule (OnePathRule variable))
-
-instance SOP.HasDatatypeInfo (Rule (OnePathRule variable))
-
-instance Debug variable => Debug (Rule (OnePathRule variable))
+instance (SimplifierVariable variable) => TransitionRule (OnePathRule variable) where
```
virgil-serbanuta: line > 80 chars.

kore/src/Kore/Strategies/Goal.hs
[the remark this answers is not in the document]
ttuegel: > This record, and the way it is used seems to me like it could be replaced by a class.

I'm confused: It was a class, and un-making that class was the motivation for this pull request, because that class wasn't reusable.

kore/src/Kore/Strategies/Goal.hs
ttuegel: I just meant to move it into the same module as `ProofState`, but it's not important.

kore/src/Kore/Strategies/Goal.hs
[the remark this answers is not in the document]
traiansf: It's a class which is reusable for at least OnePath and AllPath

The alternative was a dictionary of functions which was also only used for OnePath and AllPath.

Issue #944, opened 2019-09-10
Track pattern origin for debugging


ttuegel:
Add a `Creator` pattern attribute,

```.hs
newtype Creator = Creator { getCreator :: Maybe CallStack }
```

and add it to [`Kore.Attribute.Pattern.Pattern`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Attribute/Pattern.hs#L38). In the `TermLike` smart constructors, use [`callStack`](https://hackage.haskell.org/package/base-4.12.0.0/docs/GHC-Stack.html#v:callStack) to populate this field. The smart constructors should pop the top entry off the call stack because the smart constructor itself will always be at the top. Important: Check that this does not significantly impact performance! Add to the `Unparse` instance of `TermLike` a comment with the top entry of the `CallStack`.

PR #956, opened 2019-09-12
Track pattern origins

vladciobanu:
---

###### Reviewer checklist


vladciobanu:
## PR Summary

This PR fixes #944 by adding a `Created` attribute for `Patterns`. Interesting points:
- there is an always true `Eq` instance for it; we need `Eq` for `Pattern`s, but `CallStack` has no such instance (nor does the inner `SrcLoc`)
- we initialize the attribute with `Nothing` via its `Synthetic` instance
- we update the value via the `TermLike` smart constructors; `updateCallStack` being the key

`updateCallStack`:
- `coerced` is a lens that goes over newtypes (in this case, the `TermLike`)
- `_extract` is a helper that goes over the `Cofree` and targets its head `a` (it is very similar to [_extract](https://hackage.haskell.org/package/free-5.1.2/docs/Control-Comonad-Cofree.html#v:_extract) which already exists, but it's for the wrong `Cofree` type
- `field` is the `generic-lens` product combinator for the `Created` field in `Attribute`
- we need to pop the callstack twice: once for `updateCallStack` and once for the smart constructor

The rest of the changes are:
- reformatting on constraints to make them more consistent across smart constructors
- in order to keep things consistent, the `_` ctors (e.g. `mkTop_`) end up calling `updateCallStack` twice: once through `mkTop` and once in `mkTop_` itself. That's only needed because we want to keep the callstack consistent across all constructors. Arguably, we could have a different lens that just pops the existing callstack, but it did not seem to me that calling `getCallStack` has a detectable runtime cost.


Below are the results of running `GHCRTS='-s' ./scripts/kevm-integration.sh` on `master` (before) and this PR (after):

[after.txt](https://github.com/kframework/kore/files/3608896/after.txt)
[before.txt](https://github.com/kframework/kore/files/3608897/before.txt)

vladciobanu:
I think it's fine either waay, I'm not going to merge it until I have both approvals anyway (either explicit or in a comment like yours) :)

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1030,18 +1036,36 @@ getRigidSort pattern' =
...
-        )
+    :: GHC.HasCallStack
+    => Ord variable
+    => SortedVariable variable
+    => Unparse variable
     => TermLike variable
     -> TermLike variable
     -> TermLike variable
-mkAnd = makeSortsAgree mkAndWorker
+mkAnd t1 = updateCallStack . makeSortsAgree mkAndWorker t1
```
traiansf: consider `mkAnd t1 t2`

vladciobanu:
I also seem to have forgotten to update the `Unparse` instance for `TermLike`, so I will also do that before merging.

vladciobanu:
Do not merge until #960 is merged, since I moved the refactor in the last commit to that PR.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -458,11 +469,31 @@ instance NFData variable => NFData (TermLike variable) where
...
             $ mapVariables toVariable term
+        showCreated (name, loc) =
+            Pretty.hsep
+                [ "// Created by"
+                , Pretty.angles $ Pretty.pretty name
+                , "at"
+                , Pretty.pretty $ GHC.prettySrcLoc loc
+                ]
+
+        getCallStackHead =
```
ttuegel: Let's add a `Pretty` instance for `Created` instead.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -458,11 +469,31 @@ instance NFData variable => NFData (TermLike variable) where
...
+                    [ attrs Lens.^? getCallStackHead
+                    , Just $ unparse pat
+                    ]
       where
         freshVarTerm =
             externalizeFreshVariables
             $ mapVariables toVariable term
+        showCreated (name, loc) =
+            Pretty.hsep
+                [ "// Created by"
```
ttuegel: I think we should prefer to use the `/* C-style */` comments because the pretty-printer can split and combine lines, which might break this comment.

PR #958, opened 2019-09-12
Clarify the onepath strategy

virgil-serbanuta:
Right now we are removing the destination from the remainder and simplifying the result when deriving rules. That's a bit strange from a strategy/goal point of view, so this PR moves that at the strategy level.

To do that, it splits the `Goal` goal into `Goal` and `GoalRewritten`, so that we can easily ignore the latter until the end of a step.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Strategies/Goal.hs
+++ b/kore/src/Kore/Strategies/Goal.hs
@@ -57,8 +57,10 @@ proven = Foldable.null . unprovenNodes
 data Prim rule
     = CheckProven
     -- ^ End execution on this branch if the state is 'Proven'.
-    | CheckGoalRem
-    -- ^ End execution on this branch if the state is 'GoalRem'.
+    | CheckGoalRemainder
+    -- ^ End execution on this branch if the state is 'GoalRemainder'.
+    | ResetGoal
+    -- ^ Mark all goals rewritten previously as new goals.
     | Simplify
     | RemoveDestination
     | TriviallyValid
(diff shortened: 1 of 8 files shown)


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -106,25 +108,40 @@ transitionRule
...
       | isTriviallyValid g = return Proven
-    transitionRuleWorker TriviallyValid (GoalRem g)
+    transitionRuleWorker TriviallyValid (GoalRemainder g)
+      | isTriviallyValid g = return Proven
+    transitionRuleWorker TriviallyValid (GoalRewritten g)
       | isTriviallyValid g = return Proven
 
-    transitionRuleWorker (DerivePar rules) (GoalRem g) =
+    transitionRuleWorker (DerivePar rules) (Goal g) =
+        -- TODO (virgil): Wrap the results in GoalRemainder/GoalRewritten here.
```
vladciobanu: Is this `TODO` (and the ones below) something you forgot to do now, or does it require more (upcoming) changes in subsequent PR's?

virgil-serbanuta: More changes in the future. I don't know if it will still make sense after Ana's changes. I added some explanations.

kore/test/Test/Kore/OnePath/Step.hs
```diff
@@ -199,148 +199,147 @@ test_onePathStrategy =
...
-         -- Normal axiom: constr10(x) => constr11(x)
-         -- Start pattern: constr10(x)
-         -- Expected:
-         --   (f(b) and x=b)
-         --   or (f(c) and x=c)
-         --   or (h(x) and x!=a and x!=b and x!=c )
-         actual <-
-             runOnePathSteps
+    , testCase "Differentiated axioms" $ do
+        -- Target: constr11(a)
```
ttuegel: This test was hard for me to understand because the `Target` and the `Start pattern` are separated. Maybe this could say,

```.hs
-- Goal: constr10(x) => constr11(a)
```

virgil-serbanuta: https://github.com/kframework/kore/pull/979

PR #959, opened 2019-09-12
Function memoization

### LINKED ISSUE #953: Function memoization

ttuegel:
Fixes #953.

The function cache is implemented in `Kore.Step.Function.Memo`. An "object-oriented" approach is used to minimize the scope of changes; except for some changes to inject a "forgetful" cache into the tests, the only changes are to initialize the cache in `evalSimplifier` and to check and update the cache in `evaluateApplication`. The tests in `Test.Kore.Step.Function.Memo` illustrate the ease with which we can test such a component in isolation.

The timeout on Jenkins is reduced in anticipation of improved performance.

I am really not attached to the names in `Kore.Step.Function.Memo` if they seem too generic to anyone, but I fully intend that module to be imported qualified always.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Function/Evaluator.hs
+++ b/kore/src/Kore/Step/Function/Evaluator.hs
@@ -166,11 +166,21 @@ evaluateApplication
 
     recordOrPattern orPattern key
       | [result] <- OrPattern.toPatterns orPattern
+      , Just term <- asConcrete (Pattern.term result)
+      -- If the pattern and predicate are concrete, then we expect the predicate
+      -- to be fully-evaluated, so it must be Top. It may not be fully-evaluated
+      -- due to some uninterpreted function or an unsolved unification problem;
+      -- these are not errors, but they are unusual enough that we don't want to
+      -- deal with them here.
       , isTop (Pattern.predicate result)
+      -- We already checked that childrenPredicate has no substitutions, so we
+      -- don't expect the result to have any substitutions either. As with the
+      -- predicate, it might be possible to have a substitution in some cases,
+      -- but they are unusual enough that we don't need to deal with them here.
+      , isTop (Pattern.substitution result)
       = do
         Memo.Self { record } <- askMemo
-        let term = Pattern.term result
-        Foldable.for_ (asConcrete term) (record key)
+        record key term
       | otherwise
       = return ()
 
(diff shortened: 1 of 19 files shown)


kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -79,50 +93,76 @@ evaluateApplication
     -> Application Symbol (TermLike variable)
     -- ^ The pattern to be evaluated
     -> simplifier (OrPattern variable)
-evaluateApplication configurationPredicate childrenPredicate application = do
+evaluateApplication
```
vladciobanu: It's obviously not this PR's job to fix this, but I'm not a fan of functions as big/complex as this one. Maybe we could create a "refactor" label and make a small issue pointing to individual areas/functions everytime we find such a thing, and we can grab a few of those each sprint?

kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -79,50 +93,76 @@ evaluateApplication
...
-            error
-                (   "Attempting to evaluate unimplemented hooked operation "
-                ++  hook ++ ".\nSymbol: " ++ getIdForError symbolId
-                )
-          | otherwise ->
-            return unchanged
-        Just evaluatedPattSimplifier -> evaluatedPattSimplifier
+    recordOrPattern orPattern key
+      | [result] <- OrPattern.toPatterns orPattern
+      , isTop (Pattern.predicate result)
```
virgil-serbanuta: 1. Why aren't you also testing the substitution? If it's not needed, it would be nice to assert that.
2. Why do you need the predicate to be top? Are you expecting concrete terms to have no attached predicate, so you're only handling that case?

kore/src/Kore/Step/Function/Evaluator.hs
ttuegel: I prefer to refactor when we are already working on the affected code, so I think `TODO`s are the most effective reminder. 

As far as refactoring goes: in this case, I would factor out a memo-less "function evaluator" object in the same style as `Kore.Step.Function.Memo.Self` and add a function

```.hs
memoize :: Evaluator.Self state -> Memo.Self state -> Evaluator.Self state
```

which applies memoization to an evaluator object.

PR #960, opened 2019-09-13
Refactor 'UnsupportedPatterns' (needed for Created attribute)

vladciobanu:
See #956 and https://runtimeverification.slack.com/archives/C9B9QBF46/p1568380474021100 

---

###### Reviewer checklist

### DIFF
--- a/kore/test/Test/Kore/Comparators.hs
+++ b/kore/test/Test/Kore/Comparators.hs
@@ -95,6 +95,11 @@ import Kore.Strategies.ProofState
 import Kore.Syntax as Syntax
 import Kore.Syntax.Sentence as Syntax
 import Kore.Unification.Error
+    ( SubstitutionError (..)
+    , UnificationError (UnsupportedPatterns)
+    , UnificationOrSubstitutionError (..)
+    )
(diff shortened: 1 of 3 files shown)


kore/test/Test/Kore/Comparators.hs
```diff
@@ -990,13 +990,18 @@ instance EqualWithExplanation SymbolOrAlias where
     compareWithExplanation = structCompareWithExplanation
     printWithExplanation = show
 
-instance SumEqualWithExplanation UnificationError where
-    sumConstructorPair (UnsupportedPatterns a1) (UnsupportedPatterns a2) =
-        SumConstructorSameWithArguments
-        $ EqWrap "UnsupportedPatterns" a1 a2
+instance StructEqualWithExplanation UnificationError where
+    structConstructorName _ = "UnsupportedPatterns"
+    structFieldsWithNames
```
virgil-serbanuta: We usually match the argument with something like `expected@(UnsupportedPatterns _ _ _)` and use getters below, otherwise we seem to forget to also update this file when adding fields to the struct. See the `SymbolOrAlias` instance above.

PR #961, opened 2019-09-13
Combining a sequence of rules into one

virgil-serbanuta:
Given a list of rules
```
L1 -> R1
L2 -> R2
...
Ln -> Rn
```
 computes a predicate P such that applying the above rules in succession is the same as applying `(L1 and P) => Rn`.

Also fixes a free-variables bug for rules.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Rule/Combine.hs
+++ b/kore/src/Kore/Step/Rule/Combine.hs
@@ -7,11 +7,21 @@ module Kore.Step.Rule.Combine
     ( mergeRulesPredicate
     ) where
 
-import Data.Set (Set)
-import qualified Data.Set as Set
-import qualified Data.Map as Map
 import qualified Data.List as List
+import Data.Set
+    ( Set
+    )
+import qualified Data.Set as Set
 
+import Kore.Attribute.Pattern.FreeVariables
+    ( FreeVariables (FreeVariables)
+    )
+import Kore.Internal.TermLike
+    ( mkAnd
+    )
+import Kore.Internal.Variable
+    ( InternalVariable
+    )
 import Kore.Predicate.Predicate
     ( makeCeilPredicate
     , makeMultipleAndPredicate
@@ -20,15 +30,20 @@ import qualified Kore.Predicate.Predicate as Syntax
     ( Predicate
     )
 import Kore.Step.Rule
-    ( RewriteRule (RewriteRule), RulePattern (RulePattern), refreshRulePattern
+    ( RewriteRule (RewriteRule)
+    , RulePattern (RulePattern)
+    , refreshRulePattern
     )
-import qualified Kore.Step.Rule as Rule (freeVariables)
+import qualified Kore.Step.Rule as Rule
+    ( freeVariables
+    )
 import qualified Kore.Step.Rule as Rule.DoNotUse
-import Kore.Internal.TermLike (mkAnd)
-import Kore.Internal.Variable (InternalVariable)
-import Kore.Attribute.Pattern.FreeVariables (FreeVariables(FreeVariables))
-import Kore.Variables.UnifiedVariable (UnifiedVariable)
-import Kore.Substitute (SubstitutionVariable)
+import Kore.Substitute
+    ( SubstitutionVariable
+    )
+import Kore.Variables.UnifiedVariable
+    ( UnifiedVariable
+    )
 
 {-
 Given a list of rules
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Step/Rule.hs
```diff
@@ -568,15 +568,17 @@ refreshRulePattern (FreeVariables.getFreeVariables -> avoid) rule1 =
...
             rule1
                 { left = left'
                 , right = right'
                 , requires = requires'
+                , ensures = ensures'
                 }
     in (rename, rule2)
   where
-    RulePattern { left, right, requires } = rule1
+    RulePattern { left, right, requires, ensures } = rule1
```
ttuegel: Could you please split the bugfix into its own pull request and add a test?

kore/src/Kore/Step/Rule/Combine.hs
```diff
@@ -0,0 +1,104 @@
...
+    -> RewriteRule variable
+    -> (Set (UnifiedVariable variable), [RewriteRule variable])
+renameRuleVariable
+    (usedVariables, processedRules)
+    (RewriteRule rulePattern)
+  = (newUsedVariables, RewriteRule newRulePattern : processedRules)
+  where
+    newUsedVariables =
+        usedVariables
+        `Set.union` Map.keysSet renaming
```
ttuegel: I think it's redundant to include `ruleVariables` with the variables from `renaming`: the keys of `renaming` are free variables of `rulePattern` which are already in `usedVariables` and the values of `renaming` are free variables in `newRulePattern`. I think it is sufficient to write

```.hs
usedVariables `Set.union` getFreeVariables (Rule.freeVariables newRulePattern)
```

because `newUsedVariables` should be the set of free variables of all the rules in list.

kore/test/Test/Kore/Step/Step.hs
```diff
@@ -529,17 +529,15 @@ test_applyRewriteRule_ =
...
-                        Mock.x
-                        (mkAnd
-                            (mkElemVar Mock.x)
-                            (mkEquals_
-                                (Mock.functional11 (mkElemVar Mock.x))
-                                (Mock.functional10 (mkElemVar Mock.x))
-                            )
-                        )
-                    )
+                [ OrPattern.fromPatterns
```
ttuegel: What happened here?

kore/src/Kore/Step/Rule/Combine.hs
```diff
@@ -0,0 +1,104 @@
...
+mergeRulePair
+    ( RewriteRule RulePattern {right = right1, ensures = ensures1}
+    , RewriteRule RulePattern {left = left2, requires = requires2}
+    )
+  =
+    makeMultipleAndPredicate
+        [ makeCeilPredicate (mkAnd right1 left2)
+        , ensures1
+        , requires2
+        ]
```
ehildenb: Given:

```
l1 => r1 requires r1 ensures e1
l2 => r2 requires r2 ensures e2
```

This seems to build:

```
AND [ \ceil(\and(r1, l2)) , e1 , r2 ]
```

I'm surprised that `l1` and `r2` don't show up.

I guess I would expect something more like:

```
l1 => r2 requires AND [ \ceil(\and(r1, l2)) , e1 , r2 ]
```

traiansf: l1 and rn will be added last; the intermediary ones would become r1s and l2s.

virgil-serbanuta: This only builds the predicate, I thought it is better to wait until we actually simplify the predicate to build the rule.

ehildenb: Ah ok. In that case I guess I would suggest a different name than `mergeRulePair`, but I'm not very familiar with your style guides/naming schemes, so up to you.

Thanks for clarifying!

ehildenb:
I don't feel qualified to approve or request specific changes, but it looks "morally" like the feature I've requested, so LGTM.

kore/test/Test/Kore/Step/Step.hs
virgil-serbanuta: Caused by the `ensures` fix. This entire thing is confusing because the ensures clause uses the same `x` as the configuration, and it wasn't being refreshed before this PR.

If we change the configuration to be `y` instead of `x`, then the actual result is the one in this PR (see the first commit in https://github.com/kframework/kore/pull/967 ).

My best guess about what was happening is: the `requires` variable wasn't refreshed, and had the same name as the configuration variable, we detected that we had a rule variable which didn't disappear, so we quantified it. But it should have disappeared, because we matched on the (renamed) x in the rule.

kore/src/Kore/Step/Rule.hs
virgil-serbanuta: https://github.com/kframework/kore/pull/967

kore/src/Kore/Step/Rule/Combine.hs
virgil-serbanuta: Done, but I still need `ruleVariables`.

virgil-serbanuta:
Jenkins: test this please.

PR #974, opened 2019-09-17
Additional fixes for #956: Track Pattern Origins

vladciobanu:
This PR addresses this post-merge https://github.com/kframework/kore/pull/956#pullrequestreview-288648890

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Pattern/Created.hs
+++ b/kore/src/Kore/Attribute/Pattern/Created.hs
@@ -49,7 +49,7 @@ instance Debug Created
 
 instance Pretty Created where
     pretty (Created maybeCallStack) =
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Attribute/Pattern/Created.hs
```diff
@@ -42,5 +47,20 @@ instance Hashable Created where
 
 instance Debug Created
 
+instance Pretty Created where
+    pretty (Created maybeCallStack) =
+        maybe "/**/" go getCallStackAtHead
```
ttuegel: Why not `""` if there is no call stack? It would make the output less noisy.

Issue #937, opened 2019-09-09
Implement all-path strategy


ttuegel:
All the pieces of the all-path prover seem to be ready, but they need to be assembled into the proof strategy.

PR #975, opened 2019-09-17
Add strategy for all-path

ana-pantilie:
- Depends on https://github.com/kframework/kore/pull/955
- Fixes https://github.com/kframework/kore/issues/937

Main changes:
- implemented a `Kore.Strategies.Goal.Goal` instance for `AllPathRule`
- added `Test.Kore.Strategies.AllPath.Verification`
- implemented tests are either translated from `src/main/k/in-progress/all-path` or are present in `Test.Kore.Strategies.OnePath.Verification` as well

---

###### Reviewer checklist


kore/src/Kore/Step/Rule.hs
```diff
@@ -421,6 +428,48 @@ wEF sort = Alias
...
 
+allPathRuleToPattern
+    :: Ord variable
+    => SortedVariable variable
+    => Unparse variable
+    => AllPathRule variable
+    -> TermLike variable
+allPathRuleToPattern
+    ( AllPathRule
+        (RulePattern left _ right requires ensures _)
```
ttuegel: I think we should have some assertion that `antiLeft` is absent.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -382,6 +432,50 @@ onePathFollowupStep claims axioms =
         , TriviallyValid
         ]
 
+allPathFirstStep
+    :: [Rule (AllPathRule variable)]
+    -> Strategy (Prim (AllPathRule variable))
+allPathFirstStep axioms =
```
ttuegel: You should have a look at #991 if you haven't already.

PR #976, opened 2019-09-17
Simplification with remainders are skipped instead of failing

traiansf:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -23,6 +23,7 @@ import Data.Maybe
     )
 import qualified Data.Text as Text
 import qualified Data.Text.Prettyprint.Doc as Pretty
+import qualified Data.Text.Prettyprint.Doc.Render.Text as Pretty
 
 import qualified Kore.Attribute.Symbol as Attribute
 import qualified Kore.Internal.MultiOr as MultiOr
@@ -50,6 +51,8 @@ import qualified Kore.Step.Simplification.Simplify as AttemptedAxiomResults
 import Kore.Unparser
     ( unparse
     )
+
+import qualified Kore.Logger as Logger
 
 {-|Describes whether simplifiers are allowed to return multiple results or not.
 -}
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -234,18 +237,18 @@ applyFirstSimplifierThatWorks
...
                     && not (acceptsMultipleResults multipleResults)
-                    )
+                then
+                    do
                     -- It's not obvious that we should accept simplifications
                     -- that change only a part of the configuration, since
                     -- that will probably make things more complicated.
                     --
                     -- Until we have a clear example that this can actually
                     -- happen, we throw an error.
```
virgil-serbanuta: Out of date documentation.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -234,18 +237,18 @@ applyFirstSimplifierThatWorks
...
-                    ((error . show . Pretty.vsep)
-                        [ "Unexpected simplification result with remainder:"
+                    Logger.logDebug . Pretty.renderStrict . Pretty.layoutCompact . Pretty.vsep
```
virgil-serbanuta: Is this under 80 chars?

traiansf: Nope, it wasn't. thanks!

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
traiansf: thanks!

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -234,35 +237,34 @@ applyFirstSimplifierThatWorks
                         ++ show applicationResult
                         )
                     )
-                Monad.when
-                    (not (OrPattern.isFalse orRemainders)
+                if  not (OrPattern.isFalse orRemainders)
                     && not (acceptsMultipleResults multipleResults)
```
ttuegel: When the simplification rule applies, both checks are conditioned on `not (acceptsMultipleResults multipleResults)`. I think this part of the code would be clearer if we moved that condition to a guard.

Should we also relax the restriction on returning multiple results? That error will probably be triggered the first time someone writes a simplification rule with a disjunction in the `requires`.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -234,35 +237,34 @@ applyFirstSimplifierThatWorks
...
-        AttemptedAxiom.NotApplicable ->
-            applyFirstSimplifierThatWorks
-                evaluators
-                multipleResults
-                substitutionSimplifier
-                simplifier
-                axiomIdToSimplifier
-                patt
-                predicate
+                then
```
virgil-serbanuta: I'd put the "do" after "then", in which case the "do" contents will be properly indented.

Also, I suggest that this log statement be triggered at most once per axiom id or something like that, otherwise it may be too noisy.

traiansf: That would be interesting feature request for the logger, something like `logOnce` which would take a `Hashable` as an identifier. I'll add a TODO here

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
traiansf: > When the simplification rule applies, both checks are conditioned on `not (acceptsMultipleResults multipleResults)`. I think this part of the code would be clearer if we moved that condition to a guard.

I thought of this, but wasn't sure. Thanks for mentioning it!

 
> Should we also relax the restriction on returning multiple results? That error will probably be triggered the first time someone writes a simplification rule with a disjunction in the `requires`.

So far, the policy has been to await for the first time when it happens, and then think about how to deal with it in a clearer context.

PR #977, opened 2019-09-17
Fix a bug in the condition evaluator

ttuegel:
This fixes a bug introduced into `Kore.Step.Condition.Evaluator` by #913 and adds a test in the form of a KEVM proof. The test cannot actually be proven at this time, but we can check that it makes progress.

---

###### Reviewer checklist


kore/src/Kore/Step/Condition/Evaluator.hs
```diff
@@ -33,7 +39,7 @@ simplify
     -- to predicates. Even better, delete this one and use Predicate.simplify.
 simplify predicate = do
     simplifiedPredicates <-
-        Branch.gather . simplifyPredicate
+        Branch.gather . Predicate.simplify 0
```
virgil-serbanuta: Ok, so why is "Predicate.simplify" better?

ttuegel: `Predicate.simplify` is recursive with a depth bound. `simplifyPredicate` calls `Predicate.simplify` and propagates the local bound. If `Kore.Step.Condition.Evaluator.simplify` is called inside `Predicate.simplify`, then the depth limit may prevent `simplifyPredicate` from simplifying any predicates at all! This starts over at depth 0. This is actually a revert of an older change which unintentionally changed this behavior.

PR #980, opened 2019-09-17
Overhaul logging (Part 1)

### LINKED ISSUE #884: Overhaul logging
An overhauled logging interface will improve performance and enable greater flexibility:

```.hs
-- | Log entries
class Typeable entry => Entry entry where
    toEntry :: entry -> SomeEntry
    toEntry = SomeEntry

    fromEntry :: SomeEntry -> Maybe entry
    fromEntry (SomeEntry entry) = Data.Typeable.cast entry

    severity :: entry -> Severity

    -- | Display the log entry nicely for the user
    displayEntry :: entry -> Doc ann

    -- | Is the log entry in the named scope?
    inScope :: String -> entry -> Bool  -- derive generically, based on constructor names

-- | Encapsulates any log entry.
data SomeEntry where
    SomeEntry :: Entry entry => entry -> SomeEntry

-- | Warning emitted when a hook is missing.
-- Using distinct types for each log entry allows the LogAction 
-- to choose how it dispatches entries. For example, the LogAction 
-- may choose not to emit the same missing-hook warning multiple times.
-- Storing data instead of text lets us push rendering into a separate thread
-- when logging is enabled. The data is also stored lazily for this reason.
-- Storing data instead of text also separates concerns between the code
-- that generates the warning and the code that displays it. Besides cleanliness,
-- this would let us change log formats quickly in the future.
data WarnMissingHook =
    WarnMissingHook { hook :: Text, symbol :: Symbol }
    deriving (Eq, Typeable)

-- | Emit the missing-hook warning.
-- See the MonadLog interface below.
warnMissingHook :: MonadLog m => Text -> Symbol -> m ()

instance Entry WarnMissingHook where
    ...

-- | Wrapper for entries in the scope of function evaluation.
data ScopeFunctionEvaluation where
    ScopeFunctionEvaluation :: Entry entry => entry -> ScopeFunctionEvaluation
    deriving (Typeable)

-- | Interface for monads that support logging.
class MonadLog m where
    log :: Entry entry => entry -> m ()
    scope :: (Entry e1, Entry e2) => (e1 -> e2) -> (m a -> m a)

-- | Implementation of a logging monad transformer.
newtype LoggerT m a =
    LoggerT { getLoggerT :: ReaderT (LogAction m SomeEntry) m a }

instance Monad m => MonadLog (LoggerT m) where
    ...
```

This interface follows the standard `Exception` interface.

Besides this change, we should push log-writing into a different thread.

See also: #872 

---

- [x] Interface
- [ ] Threading

vladciobanu:
Fixes #884 

Note that I have to abandon this for a few weeks. If someone wants to pick it up, please see #884 and the comments I added here.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Logger.hs
+++ b/kore/src/Kore/Logger.hs
@@ -259,39 +259,41 @@ class Typeable entry => Entry entry where
 
     severity :: entry -> Severity
 
-    -- Would this work as a 'Pretty' constraint on 'entry'?
+    -- TODO: Would this work as a 'Pretty' constraint on 'entry'?
     displayEntry :: entry -> Pretty.Doc ann
 
+    -- TODO: This is different from the original idea, please double check it makes sense
     inScope :: Set String -> entry -> Bool
 
 data SomeEntry where
     SomeEntry :: Entry entry => entry -> SomeEntry
 
-
 instance Entry LogMessage where
     severity LogMessage { severity } = severity
-
+    inScope _ _ = True
 
 class Monad m => MonadLog m where
     logM :: Entry entry => entry -> m ()
     scope :: Entry e1 => Entry e2 => (e1 -> e2) -> m a -> m a
-
-instance Monad m => MonadLog (LoggerT m) where
-    logM _ = pure ()
-    scope _ = id
 
 newtype LoggerT m a =
     LoggerT { getLoggerT :: ReaderT (LogAction m SomeEntry) m a }
     deriving (Functor, Applicative, Monad)
     deriving (MonadIO)
 
-instance MonadTrans LoggerT where
-    lift = LoggerT . Monad.Trans.lift
-    {-# INLINE lift #-}
+instance Monad m => MonadLog (LoggerT m) where
+    logM _ = pure ()
+    scope _ = id
 
-instance Monad m => WithLog LogMessage (LoggerT m) where
+instance MonadLog m => WithLog LogMessage (LoggerT m) where
+    -- TODO: Stuck here. I can't see how to write this instance with `MonadLog`
+    -- but maybe I'm not seeing it
     askLogAction = fmap (contramap toEntry) $ ask
     {-# INLINE askLogAction #-}
 
     localLogAction locally = LoggerT . local locally . getLoggerT
     {-# INLINE localLogAction #-}
+
+instance MonadTrans LoggerT where
+    lift = LoggerT . Monad.Trans.lift
+    {-# INLINE lift #-}
--- a/kore/src/Kore/Logger.hs
+++ b/kore/src/Kore/Logger.hs
@@ -45,11 +45,24 @@ import Control.Monad.Trans.Reader
 import Data.Functor.Contravariant
     ( contramap
     )
+import Data.Maybe
+    ( fromJust
+    )
+import Data.Set
+    ( Set
+    )
 import Data.String
     ( IsString
     )
 import Data.Text
     ( Text
+    )
+import qualified Data.Text.Prettyprint.Doc as Pretty
+import Data.Typeable
+    ( Typeable
+    )
+import qualified Data.Typeable as Data.Typeable
+    ( cast
     )
 import GHC.Stack
     ( CallStack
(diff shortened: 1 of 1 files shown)


kore/src/Kore/Logger.hs
```diff
@@ -240,18 +295,82 @@ withLogScope newScope = localLogAction (contramap appendScope)
...
+                logActionReader . LogAction
+                    $ \entry ->
+                        case fromEntry entry of
+                            Nothing -> logAction entry
+                            Just entry' -> logAction $ toEntry $ f entry'
 
 instance Monad m => WithLog LogMessage (LoggerT m) where
-    askLogAction = LoggerT . fmap liftLogAction $ ask
+    askLogAction :: LoggerT m (LogAction (LoggerT m) LogMessage)
+    askLogAction = LoggerT . ReaderT $ pure . go
```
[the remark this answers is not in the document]
vladciobanu: This won't work since we can't go from `LogAction m msg -> LogAction m msg` to `msg -> msg` (needed in `logScope`). I think we can do it the other way around: provide default implementations of `MonadLog` for anything with a `WithLog` instance. Would that do us any good?

kore/src/Kore/Logger.hs
```diff
@@ -240,18 +295,82 @@ withLogScope newScope = localLogAction (contramap appendScope)
...
+        go
+            :: (LogAction n LogMessage -> LogAction n LogMessage)
+            -> (LogAction n SomeEntry -> LogAction n SomeEntry)
+        go =
+            Profunctor.dimap
+                (contramap toEntry)
+                (contramap (fromJust . fromEntry))
     {-# INLINE localLogAction #-}
+
+instance MonadTrans LoggerT where
```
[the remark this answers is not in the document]
vladciobanu: It's not, and I think the reason is `m` is not "free" in the transformer because it appears in the `r` part of the `ReaderT` (as `LogAction m`).

kore/src/Kore/Logger.hs
```diff
@@ -240,18 +295,82 @@ withLogScope newScope = localLogAction (contramap appendScope)
...
+    localLogAction locally (LoggerT (ReaderT la)) =
+        LoggerT . ReaderT $ la . go locally
+      where
+        go
+            :: (LogAction n LogMessage -> LogAction n LogMessage)
+            -> (LogAction n SomeEntry -> LogAction n SomeEntry)
+        go =
+            Profunctor.dimap
+                (contramap toEntry)
+                (contramap (fromJust . fromEntry))
```
ana-pantilie: Can this `fromJust` ever fail?

kore/src/Kore/Logger/Output.hs
```diff
@@ -124,14 +108,27 @@ data KoreLogOptions = KoreLogOptions
     }
 
 -- | Internal type used to add timestamps to a 'LogMessage'.
-data LogMessageWithTimestamp = LogMessageWithTimestamp LogMessage LocalTime
+data WithTimestamp = WithTimestamp SomeEntry LocalTime
+
+instance Pretty.Pretty WithTimestamp where
+    pretty (WithTimestamp entity time) =
+        Pretty.brackets formattedTime <> Pretty.pretty entity
```
ana-pantilie: Shouldn't `entity` be `entry`?

kore/src/Kore/Logger.hs
vladciobanu: I don't think so, we're going from `LogMessage` to `LogMessage`, so it should always work.

ttuegel: No, that can definitely fail. I don't think you can write this using `Profunctor`. If it helps, you can think of the helper as having the signature
```.hs
go
    :: (LogAction n LogMessage -> LogAction n LogMessage)
    -> LogAction n SomeEntry
    -> LogAction n SomeEntry
```
In other words, you can match on the `LogAction n SomeEntry` and choose to apply the first argument, or not. I think you need a `Choice` to do it the profunctor way, but I'd just use ordinary functions here.

PR #984, opened 2019-09-18
Fix a issue in REPL related to user-defined sorts and lemmas

yzhang90:
…red in SMT

---

###### Reviewer checklist

### DIFF
--- a/kore/app/repl/Main.hs
+++ b/kore/app/repl/Main.hs
@@ -218,7 +218,9 @@ mainWithOptions
                 lift
                 $ SMT.runSMT smtConfig (swappableLogger mLogger)
                    $ do
-                        give (MetadataTools.build indexedModule) (declareSMTLemmas indexedModule)
+                        give
+                            (MetadataTools.build indexedModule)
+                            (declareSMTLemmas indexedModule)
                         proveWithRepl
                             indexedModule
                             specDefIndexedModule
--- a/kore/app/repl/Main.hs
+++ b/kore/app/repl/Main.hs
@@ -14,6 +14,7 @@ import Control.Monad.Trans.Reader
     ( runReaderT
     )
 import Data.Maybe
+import Data.Reflection
 import Data.Semigroup
     ( (<>)
     )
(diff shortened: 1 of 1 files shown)


yzhang90:
@ttuegel @virgil-serbanuta I made a fix to REPL so that user-defined sorts and lemmas are declared in the SMT queries. However, i'm not sure how to write a test.

kore/app/repl/Main.hs
```diff
@@ -212,13 +217,15 @@ mainWithOptions
...
                 $ SMT.runSMT smtConfig (swappableLogger mLogger)
-                   $ proveWithRepl
-                        indexedModule
-                        specDefIndexedModule
-                        mLogger
-                        replScript
-                        replMode
-                        outputFile
+                   $ do
+                        give (MetadataTools.build indexedModule) (declareSMTLemmas indexedModule)
```
virgil-serbanuta: Lines should usually be < 80 chars.

ana-pantilie:
@yzhang90 
Yes, the `scripts` directory from the link @virgil-serbanuta added contains `repl` integration tests. You can find more about `repl` scripts here https://github.com/kframework/kore/wiki/Kore-REPL#repl-scripts

We check the `repl` output after running each script against its corresponding `.golden` output file. The `repl` outputs a new conjuction of claims which indicate the state of the proof after executing the script. If the proof was finished the output is `#True`. You can look at the `tiny` semantics directory for how to add this sort of integration test. Each spec should have its own directory and a `scripts` directory which tests proofs of that spec.

So, if you think that you can test this by checking the final repl output (the set of claims) after running some repl commands on a spec and semantics of your choice you can add this sort of test. Otherwise, we also have some unit tests for the parser and the interpreter in https://github.com/kframework/kore/tree/master/kore/test/Test/Kore/Repl, if you think this sort of test would be more suitable in this case.

I hope this helps.

kore/app/repl/Main.hs
yzhang90: fixed

yzhang90:
@virgil-serbanuta I added a TODO in the code. I'm busy with the sum-to-n proof and we can get back to the test later when I have time.

PR #990, opened 2019-09-19
simplify one path strategy

yzhang90:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Strategies/Goal.hs
+++ b/kore/src/Kore/Strategies/Goal.hs
@@ -171,7 +171,6 @@ allPathStrategy claims axioms =
             , RemoveDestination
             , TriviallyValid
             , DerivePar axioms
-            , RemoveDestination
             , Simplify
             , TriviallyValid
             , ResetGoal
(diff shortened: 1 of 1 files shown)


yzhang90:
I think two following removeDestination are redundant. They will only work on the remainder (because removeDestinnation will not work on GoalRewritten).  However, if the remainder can be removed from destination, it is already removed in the first removeDestination

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -225,11 +225,9 @@ onePathFollowupStep claims axioms =
         , Simplify
         , TriviallyValid
         , DeriveSeq claims
-        , RemoveDestination
         , Simplify
         , TriviallyValid
         , DeriveSeq axioms
-        , RemoveDestination
```
virgil-serbanuta: I think that this change makes sense, but I still wonder why we added the `RemoveDestination` thing to `deriveSeq` in the first place (my previous PR moved it from `deriveSeq` here). Anyway, could you also remove one of the `RemoveDestination` calls from the first step above?

yzhang90: done

PR #993, opened 2019-09-20
K-EQUAL may be undefined

ttuegel:
---

###### Reviewer checklist

### DIFF
--- a/src/main/k/working/k-equal/specs/iszero-ge-spec.k
+++ b/src/main/k/working/k-equal/specs/iszero-ge-spec.k
@@ -1,0 +1,9 @@
+module VERIFICATION
+  imports TEST-K-EQUAL
+endmodule
+
+module ISZERO-GE-SPEC
+  imports VERIFICATION
+
+  rule <k> #iszero(I) => #done(1) </k> requires I >=Int 0
+endmodule
--- a/src/main/k/working/k-equal/specs/iszero-gt-spec.k
+++ b/src/main/k/working/k-equal/specs/iszero-gt-spec.k
@@ -1,0 +1,9 @@
+module VERIFICATION
+  imports TEST-K-EQUAL
+endmodule
+
+module ISZERO-GT-SPEC
+  imports VERIFICATION
+
+  rule <k> #iszero(I) => #done(1) </k> requires I >Int 0
+endmodule
--- a/src/main/k/working/k-equal/test-k-equal.k
+++ b/src/main/k/working/k-equal/test-k-equal.k
@@ -1,0 +1,18 @@
+require "domains.k"
+
+module TEST-K-EQUAL
+    imports BOOL
+    imports INT
+    imports K-EQUAL
+
+    configuration <T> <k> $PGM:Pgm </k> </T>
+
+    syntax Pgm ::= #done( Int )
+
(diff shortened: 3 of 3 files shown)


kore/src/Kore/Internal/Pattern.hs
```diff
@@ -213,3 +215,19 @@ toPredicate = Conditional.toPredicate
...
 splitTerm = Conditional.splitTerm
+
+{- | Replace the 'term' of a 'Pattern' with the argument.
+
+It is an error to use this function where it would erase a partial term,
+i.e. the 'term' must be 'Top_' or 'Bottom_', and if it is 'Bottom_' it must have
+a false 'predicate'.
+
+ -}
+replaceTerm :: TermLike variable -> Pattern variable -> Pattern variable
```
virgil-serbanuta: This is some sort of `and` between the term and the pattern, right?

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -131,26 +133,26 @@ evalKEq true _ _ _ (valid :< app) =
...
-                )
+    evalEq termLike1 termLike2 = do
+        let pattern1 = Pattern.fromTermLike termLike1
+            pattern2 = Pattern.fromTermLike termLike2
+
+        defined1 <- Ceil.makeEvaluate Predicate.topTODO pattern1
+        defined2 <- Ceil.makeEvaluate Predicate.topTODO pattern2
+        defined <- And.simplifyEvaluated defined1 defined2
+
+        equalTerms <- Equals.makeEvaluate pattern1 pattern2 Predicate.topTODO
```
virgil-serbanuta: makeEvaluateTermsToPredicate, maybe? If that works, you might be able to remove the replaceTerm stuff, maybe also the fromTermLike stuff above.

kore/src/Kore/Step/Simplification/Pattern.hs
```diff
@@ -58,13 +58,9 @@ simplify
...
 simplify pattern' = do
-    simplifiedTerm <-
-        simplifyConditionalTermToOr predicate term
-    orPatterns <-
-        Branch.gather
-        $ traverse
-            (Pattern.mergeWithPredicate predicate)
-            simplifiedTerm
-    return (MultiOr.mergeAll orPatterns)
+    orSimplifiedTerms <- simplifyConditionalTermToOr predicate term
```
virgil-serbanuta: I assume that this fixes the `or` evaluation bug you mentioned in the PR description. So, what was the bug? I guess I could figure it out if I keep staring at the change, but it's easier to just ask.l

ttuegel: `traverse` distributes over `Or` like `And`, but we have an `Or` of `Or`; we want to concatenate, which `scatter` does.

The longer version is that `Branch` is hard to use correctly, and we should consider removing it.

kore/src/Kore/Builtin/KEqual.hs
ttuegel: I'm going to split this part of the pull request off into its own change set because it's not related to this particular bug, but something else I found while looking for this one.

Issue #982, opened 2019-09-18
[KORE-LOG] dump the internal variable -> SMT variable mapping in the log


yzhang90:
When checking the SMT queries, I always confused by the variables that are used in the SMT queries(e.g., `<0>` `<1>` ...). It will be useful if we dump the mapping between internal variable name and SMT variable name in the log.

PR #995, opened 2019-09-20
log mappings between internal variables and SMT variables

yzhang90:
Fixes: #982

---

###### Reviewer checklist


yzhang90:
I changed `SMT.MonadSMT` to `MonadSimplify` in the function signature.

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -195,4 +204,10 @@ translateUninterpreted t pat =
         n <- Counter.increment
         var <- SMT.declare ("<" <> Text.pack (show n) <> ">") t
         State.modify' (Map.insert pat (var, t))
+        Monad.Trans.lift . Monad.Trans.lift . Monad.Trans.lift $
```
ttuegel: Why were all these `lift`s necessary?

yzhang90: I' not quite sure... the program won't compile until I added three `lift`s...I looked at the definition of Translator:
```
type Translator m p = MaybeT (StateT (VarContext p) (CounterT m))
``` and Monad `m` is transformed three times. Maybe this is the reason?

PR #999, opened 2019-09-23
Check that arguments to K-EQUAL.eq are defined

ttuegel:
`evalKEq` does not check that its arguments are defined. That check _does_ occur elsewhere in the application simplifier, but... it makes me uncomfortable to leave it this way. If we really think it's unnecessary, we should at least leave the test in place.

I also added an `Unparse` instance for `Kore.Internal.Predicate.Predicate`.

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Not.hs
```diff
@@ -184,3 +209,14 @@ mkMultiAndPattern
...
 mkMultiAndPattern patterns =
     Foldable.foldrM And.makeEvaluate Pattern.top patterns
+
+{- | Conjoin and simplify a 'MultiAnd' of 'Predicate'.
+ -}
+mkMultiAndPredicate
+    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    => MultiAnd (Predicate variable)
+    -> BranchT simplifier (Predicate variable)
+mkMultiAndPredicate predicates =
```
virgil-serbanuta: Why not `MultiAnd.toPredicate`?

ttuegel: That's `Kore.Predicate.Predicate.Predicate` and this is `Kore.Internal.Predicate.Predicate`. Perhaps the latter needs a better name.

PR #1008, opened 2019-09-24
Diff

ttuegel:
The (derivable) class `Diff` displays the difference between values. This is intended to replace `Test.Kore.Comparators` with something that can be generically derived. I will remove `Test.Kore.Comparators` in my next pull request.

The generic `Debug` and `Diff` classes are extracted from module `Kore.Debug` into module `Debug` because they do not actually have anything to do with Kore. I cannot promise that the `diff` format is optimal, but because these instances are derived, it is very easy to tweak it later.

---

###### Reviewer checklist


kore/test/Test/Debug.hs
```diff
@@ -208,3 +228,25 @@ test_Debug =
...
+    , test (N (SA A), N (SB B))    $ Just "N {- was: (SA A) -} (SB B)"
+    , test (Rn { unRn = SA A }, Rn { unRn = SB B })
+        $ Just "Rn { unRn = {- was: SA A -} SB B }"
+    , test ("A" :: String, "B")    $ Just "{- was: \"A\" -} \"B\""
+    , test ('A', 'B')              $ Just "{- was: 'A' -} 'B'"
+    , test (0 :: Integer, 1)       $ Just "{- was: 0 -} 1"
+    , test (True, False)           $ Just "{- was: True -} False"
+    , test ([True], [])            $ Just "{- was: [ True ] -} []"
+    , test ([], [True])            $ Just "{- was: [] -} [ True ]"
+    , test ([True], [True, False]) $ Just "_ : {- was: [] -} [ False ]"
```
virgil-serbanuta: Could you add the following tests? Or, at least, the ones that make sense.
* `test ([True, True], [True, False, True])`
* `test (Struct {field1=True, field2=True, field3=True}, Struct {field1=False, field2=False, field3=True})`
* `test (Struct {field1=True, field2=True, field3=True}, Struct {field1=True, field2=False, field3=False})`
* A test with a constructor operator like `:<`

kore/src/Debug.hs
```diff
@@ -0,0 +1,581 @@
...
+instance (Debug a, Diff a) => Diff (Sup a)
+
+instance (Debug a, Diff a) => Diff (Identity a)
+
+instance
+    (Debug a, Debug (f b), Diff a, Diff (f b)) => Diff (CofreeF f a b) where
+    diffPrec x y =
+        -- Cannot have orphan instances of Generic and HasDatatypeInfo.
+        -- Use a fake instance instead.
+        diffPrecSOP datatypeInfoCofreeF (x, fromCofreeF x) (y, fromCofreeF y)
```
virgil-serbanuta: How hard would it be to diff the annotation last?

kore/src/Debug.hs
```diff
@@ -0,0 +1,581 @@
...
+> instance Generics.SOP.Generic DataType
+>
+> instance Generics.SOP.HasDatatypeInfo DataType
+>
+> instance Debug DataType
+>
+> instance Diff DataType
+
+ -}
+class Diff a where
```
virgil-serbanuta: Not sure if it should be part of this PR, but I think that Id needs a custom `Diff` instance, similar to the custom `Eq` instance. I wonder if we have other custom `Eq` instances.

kore/src/Debug.hs
ttuegel: You mean, print the difference in the annotations only if the patterns are the same?

kore/src/Debug.hs
ttuegel: I got around this by giving `AstLocation` a `Diff` instance which ignores all differences, because I think for debugging we would never care about that. But, I could give a custom instance for `Id` instead.

kore/src/Debug.hs
virgil-serbanuta: Yes, that's what I meant. I wasn't sure exactly how this works until I saw the additional tests. I'm not asking you to include it in this PR, but it may turn out to be useful.

Issue #1022, opened 2019-09-26
Simplification axioms over builtin symbols are not applied


ttuegel:
The backend does not recognize an `AxiomIdentifier` for patterns with an internal representations of Map, Set, or List.

ttuegel:
Even after #1023, `\ceil` of builtins does not apply user-defined simplification axioms because there is no step to do that in the `Ceil` simplifier.

PR #1023, opened 2019-09-26
Match AxiomIdentifier based on external syntax

ttuegel:
When `TermLike` patterns are `internalized`, we would not match `AxiomIdentifier`s for internal representations of Map, List, or Set. With this change, we always return an `AxiomIdentifier` based on the external pattern syntax, i.e. ordinary symbols.

Fixes: #1022 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Axiom/Identifier.hs
+++ b/kore/src/Kore/Step/Axiom/Identifier.hs
@@ -27,6 +27,7 @@ module Kore.Step.Axiom.Identifier
 import Control.Applicative
     ( Alternative (..)
     )
+import qualified Data.Foldable as Foldable
 import qualified Data.Functor.Foldable as Recursive
 import Data.Text.Prettyprint.Doc
     ( Pretty (..)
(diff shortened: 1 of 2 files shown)


ttuegel:
This pull request is larger than I wanted because I had to rearrange some modules to avoid cycles (separating builtin simplifiers from the functions for manipulating their structure). I will make another pull request with all the refactoring separately.

kore/src/Kore/Step/Axiom/Identifier.hs
```diff
@@ -85,24 +91,25 @@ Returns 'Nothing' if the 'TermLike' does not conform to one of the structures we
 recognize.
 
  -}
-matchAxiomIdentifier :: TermLike variable -> Maybe AxiomIdentifier
-matchAxiomIdentifier = Recursive.fold matchWorker
+matchAxiomIdentifier
+    :: InternalVariable variable
+    => TermLike variable
+    -> Maybe AxiomIdentifier
+matchAxiomIdentifier = Recursive.fold matchWorker . Builtin.externalize
```
virgil-serbanuta: Is `externalize` evaluated lazily?

ttuegel: Yes, it is defined by `Recursive.unfold`, so it lazily unfolds the `TermLike`.

ttuegel:
Needs #1025.

PR #1026, opened 2019-09-27
Check for function-like rule and configuration

andreiburdusa:
---

###### Reviewer checklist


ttuegel:
I just pushed fixes for all of the test failures except the `\ceil` ones. What did we decide to do about `\ceil` simplification rules again?

traiansf:
@ttuegel please take a look.

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -60,11 +61,13 @@ import Kore.Variables.Fresh
 evaluateAxioms
     :: forall variable simplifier
     .  (SimplifierVariable variable, MonadSimplify simplifier)
-    => [EqualityRule Variable]
+    => StepContext
```
ttuegel: I'm skeptical about this; `evaluateAxioms` can only be used in a function or simplification context because its arguments are `EqualityRule`s.

kore/src/Kore/Step/Rule.hs
```diff
@@ -777,3 +780,48 @@ substitute subst rulePattern' =
...
+  | AllPathContext
+  deriving (Show, Eq)
+
+
+{- | Check if the term and the 'left' fields of the rules are function-like
+
+-}
+
+checkFunctionLike
+    ::  (Debug variable
```
ttuegel: I suggest `(InternalVariable variable, InternalVariable variable') => ` here instead. Does it need `FreshVariable`?

kore/src/Kore/Step/Rule.hs
```diff
@@ -777,3 +780,48 @@ substitute subst rulePattern' =
...
+        , FreshVariable variable
+        , SortedVariable variable
+        , Show variable
+        , Unparse variable
+        , Debug variable'
+        , FreshVariable variable'
+        , SortedVariable variable'
+        , Show variable'
+        , Unparse variable')
+    => [RulePattern variable']
```
ttuegel: There is nothing here which indicates that this should be the list of _applied_ rules, yet that seems to be critical.

kore/src/Kore/Step/Rule.hs
```diff
@@ -777,3 +780,48 @@ substitute subst rulePattern' =
...
+    -> TermLike variable
+    -> Either String ()
+checkFunctionLike [] _ = pure ()
+checkFunctionLike rules term
+  | TermLike.isFunctionPattern term =
```
ttuegel: Why is it OK to omit this check if we did not apply any rules? There would still be a remainder pattern, right?

kore/src/Kore/Step/Step.hs
```diff
@@ -397,19 +397,42 @@ finalizeRulesParallel
...
-    remainders' <- Monad.Unify.gather $ applyRemainder initial remainder
-    return Step.Results
-        { results = Seq.fromList results
-        , remainders =
-            OrPattern.fromPatterns
-            $ Pattern.mapVariables Target.unwrapVariable <$> remainders'
-        }
+finalizeRulesParallel initial unifiedRules =
+    case Rule.checkFunctionLike (term <$> unifiedRules) (term initial) of
+        Left err -> error $ "Rules:\n" <> show (unparseToString . TermLike.mapVariables Target.unwrapVariable . Rule.left . term <$> unifiedRules) <> err
```
ttuegel: Because this case only throws an error, it would be better if the `Right` case did `return ()`  and the rest of the function followed at the previous indentation level.

kore/src/Kore/Step/Step.hs
```diff
@@ -397,19 +397,42 @@ finalizeRulesParallel
...
+    ::  forall unifier variable
+    .   ( SimplifierVariable variable
+        , MonadUnify unifier
+        , Log.WithLog Log.LogMessage unifier
+        )
+    => Rule.StepContext
+    -> Pattern (Target variable)
+    -> [UnifiedRule (Target variable)]
+    -> unifier (Results variable)
+finalizeRulesSequenceWithCheck Rule.SimplificationContext initial unifiedRules =
```
ttuegel: I don't think we should assume that the simplification context will always apply rules in sequence.

kore/src/Kore/Step/Step.hs
```diff
@@ -565,8 +589,11 @@ applyRewriteRulesParallel
     -> Pattern variable
     -- ^ Configuration being rewritten
     -> unifier (Results variable)
-applyRewriteRulesParallel unificationProcedure rewriteRules =
-    applyRulesParallel unificationProcedure (getRewriteRule <$> rewriteRules)
+applyRewriteRulesParallel unificationProcedure rewriteRules initialConfig =
+    applyRulesParallel
+        unificationProcedure
+        (getRewriteRule <$> rewriteRules)
+        initialConfig
```
ttuegel: Why do we need to specify `initialConfig` explicitly now?

kore/src/Kore/Step/Step.hs
```diff
@@ -397,19 +397,42 @@ finalizeRulesParallel
...
-        remainder = Predicate.fromPredicate (Remainder.remainder' unifications)
-    remainders' <- Monad.Unify.gather $ applyRemainder initial remainder
-    return Step.Results
-        { results = Seq.fromList results
-        , remainders =
-            OrPattern.fromPatterns
-            $ Pattern.mapVariables Target.unwrapVariable <$> remainders'
-        }
+finalizeRulesParallel initial unifiedRules =
+    case Rule.checkFunctionLike (term <$> unifiedRules) (term initial) of
```
ttuegel: The list of unified rules is available in the `Results`, so I don't see why this check shouldn't happen in the caller.

kore/src/Kore/Step/Axiom/Evaluate.hs
traiansf: True, but we still need to distinguish between the two cases (Function evaluation vs. simplification).

kore/src/Kore/Step/Rule.hs
traiansf: I think that if we don't apply any rules then there is no change to the original pattern, so, in principle, the remainder should be itself.

kore/src/Kore/Step/Step.hs
```diff
@@ -397,19 +397,42 @@ finalizeRulesParallel
...
+        )
+    => Rule.StepContext
+    -> Pattern (Target variable)
+    -> [UnifiedRule (Target variable)]
+    -> unifier (Results variable)
+finalizeRulesSequenceWithCheck Rule.SimplificationContext initial unifiedRules =
+    finalizeRulesSequence initial unifiedRules
+finalizeRulesSequenceWithCheck _ initial unifiedRules =
+    case Rule.checkFunctionLike (term <$> unifiedRules) (term initial) of
+        Left err -> error $ "Rules:\n" <> show (unparseToString . TermLike.mapVariables Target.unwrapVariable . Rule.left . term <$> unifiedRules) <> err
```
virgil-serbanuta: Line length

kore/src/Kore/Step/Axiom/Evaluate.hs
virgil-serbanuta: Yes, but why do we distinguish by using a context, as opposed to saying what we want? Why is it better to tell the lower levels "You are doing OnePath stuff right now, so do whatever's right for that" as opposed to "You should only accept function-like stuff when applying rules"?

If you do the second, then the higher-level code knows about high-level concepts like OnePath, function evaluation, All Path and similar stuff, while the lower level code knows about lower level concepts like "we can only handle function-like stuff".

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -34,45 +43,58 @@ import Kore.Step.Remainder
...
+-- ^ TODO (andrei.burdusa): remove this
+
+evaluateAxioms'
+    :: forall variable simplifier
+    .  (SimplifierVariable variable, MonadSimplify simplifier)
+    => StepContext
+    -> [EqualityRule Variable]
+    -> TermLike variable
+    -> Syntax.Predicate variable
+    -> simplifier (Result.Results (Step.UnifiedRule (Target variable)) (Pattern variable))
```
ttuegel:
```suggestion
    -> simplifier (Step.Results variable)
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -34,45 +43,58 @@ import Kore.Step.Remainder
...
+evaluateAxioms'
+    context
     definitionRules
     patt
     predicate
   | any ruleIsConcrete definitionRules
   , not (TermLike.isConcrete patt)
-  = return AttemptedAxiom.NotApplicable
+  = return $ Result.Results mempty mempty
```
ttuegel:
```suggestion
  = return mempty
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -95,23 +117,31 @@ evaluateAxioms
         keepResultUnchanged = id
         introduceDefinedness = flip Pattern.andCondition
         markRemainderEvaluated = fmap TermLike.mkEvaluated
+        rules = Result.appliedRule <$> Result.results result
 
     simplifiedResults <-
-        OrPattern.simplifyPredicatesWithSmt
-           predicate (Step.gatherResults result)
+        traverse (OrPattern.simplifyPredicatesWithSmt predicate)
```
ttuegel: I suggest using a `Traversal` here, so that you don't need to zip the results back up with the rules later:
```suggestion
        Lens.traverseOf (field @"results" . Lens.mapped . field @"result") 
            (OrPattern.simplifyPredicatesWithSmt predicate)
            result
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -95,23 +117,31 @@ evaluateAxioms
...
+            ((fmap Result.result . Result.results) result)
 
     simplifiedRemainders <-
```
ttuegel: I know you didn't change this part of the code, but the same comment applies as just above: 
```suggestion
    simplifiedRemainders <- Lens.traverseOf (field @"remainders") (OrPattern.simplifyPredicatesWithSmt predicate) simplifiedResults
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -34,45 +43,58 @@ import Kore.Step.Remainder
...
     definitionRules
     patt
     predicate
   | any ruleIsConcrete definitionRules
   , not (TermLike.isConcrete patt)
-  = return AttemptedAxiom.NotApplicable
+  = return $ Result.Results mempty mempty
   | otherwise
-  = AttemptedAxiom.maybeNotApplicable $ do
+  = maybeResults $ do
```
ttuegel: We don't need this, or the `guard` below, anymore.

```suggestion
  = do
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -124,8 +154,11 @@ evaluateAxioms
     rejectNarrowing (Result.results -> results) =
         (Monad.guard . not) (Foldable.any Step.isNarrowingResult results)
 
-    applyRules initial rules =
-        Monad.Unify.maybeUnifierT
+    applyRules
+        (Step.toConfigurationVariables -> initial)
+        rules
+      = Monad.Unify.maybeUnifierT
+        $ Step.checkFunctionLikeResults context initial
```
ttuegel: I still want you to move this check into `Kore.Step.Axiom.EvaluationStrategy`.

kore/src/Kore/Step/Step.hs
```diff
@@ -400,9 +401,12 @@ finalizeRulesParallel
 finalizeRulesParallel initial unifiedRules = do
     let initialCondition = Conditional.withoutTerm initial
     results <-
-        Foldable.fold <$> traverse (finalizeRule initialCondition) unifiedRules
-    let unifications = MultiOr.make (Conditional.withoutTerm <$> unifiedRules)
-        remainder = Predicate.fromPredicate (Remainder.remainder' unifications)
+        Foldable.fold
```
ttuegel: Why did this change?

kore/src/Kore/Step/Step.hs
```diff
@@ -535,17 +585,16 @@ applyRulesParallel
...
     -> unifier (Results variable)
 applyRulesParallel
     unificationProcedure
     -- Wrap the rule and configuration so that unification prefers to substitute
     -- axiom variables.
     (map toAxiomVariables -> rules)
-    (toConfigurationVariables -> initial)
-  =
-    unifyRules unificationProcedure initial rules
+    initial
```
ttuegel: Why did we remove `toConfigurationVariables` here?

kore/src/Kore/Step/Step.hs
```diff
@@ -580,19 +637,18 @@ applyRulesSequence
...
-    -> Pattern variable
+    -> Pattern (Target variable)
     -- ^ Configuration being rewritten
     -> [RulePattern variable]
     -- ^ Rewrite rules
     -> unifier (Results variable)
 applyRulesSequence
     unificationProcedure
-    -- Wrap the rule and configuration so that unification prefers to substitute
+    initial
```
ttuegel: Why did we remove `toConfigurationVariables` here?

kore/test/Test/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -54,7 +55,10 @@ test_evaluateAxioms =
     , doesn'tApply "F(x) => G(x) [concrete] doesn't apply to F(x)"
         [axiom_ (f x) (g x) & concreteEqualityRule]
         (f x, makeTruePredicate)
-    , applies "F(x) => G(x) [concrete] apply to F(a)"
+    , doesn'tApply "F(x) => G(x) doesn't apply to F(top)"
+        [axiom_ (f x) (g x)]
+        (f mkTop_, makeTruePredicate)
+     , applies "F(x) => G(x) [concrete] apply to F(a)"
```
ttuegel:
```suggestion
    , applies "F(x) => G(x) [concrete] apply to F(a)"
```

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -321,6 +322,19 @@ test_simplificationIntegration =
...
 
+simplificationRulePattern
+    :: InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+    -> RulePattern variable
+simplificationRulePattern left right =
+    patt
+        { attributes = (attributes patt)
+            { Attribute.simplification = Simplification True}
```
ttuegel:
```suggestion
            { Attribute.simplification = Simplification True }
```

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -321,6 +322,19 @@ test_simplificationIntegration =
         assertEqual "" expect actual
     ]
 
+simplificationRulePattern
+    :: InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+    -> RulePattern variable
+simplificationRulePattern left right =
+    patt
```
ttuegel: I suggest using a lens here:
```suggestion
    patt & Lens.set (field @"attributes" . field @"simplification") (Simplification True)
```

kore/test/Test/Kore/Step/Step.hs
```diff
@@ -224,20 +245,23 @@ test_applyRewriteRule_ =
     [ testCase "apply identity axiom" $ do
         let expect = Right [ OrPattern.fromPatterns [initial] ]
             initial = pure (mkElemVar Mock.x)
-        actual <- applyRewriteRule_ initial axiomId
+        actual <- applyRewriteRule_ applyRewriteRulesParallel initial axiomId
```
ttuegel: Why did we make this change? We don't seem to call `applyRewriteRule_` with any argument other than `applyRewriteRulesParallel`.

src/main/kore/prelude.kore
```diff
@@ -33,7 +33,7 @@ module INJ
     \equals{S3,R}(
       inj{S2,S3}(inj{S1,S2}(T:S1)),
       inj{S1,S3}(T:S1))
-  []
+  [simplification{}()]
```
ttuegel: Why was this necessary?

src/main/k/working/tests/fresh-vars/fresh-vars.k
```diff
@@ -8,6 +8,6 @@ module FRESH-VARS
     requires N >Int 0
   rule foo(0) => .Seq
 
-  syntax X ::= "foo"
-  rule <k> foo => foo(3) </k>
+  syntax KItem ::= "foo"
```
ttuegel: Why was this necessary?

kore/src/Kore/Step/Step.hs
andreiburdusa: Honestly, I'm not sure. It wasn't exceeding 80 characters. I'll change it back

kore/src/Kore/Step/Step.hs
andreiburdusa: I think Traian made this change and the one from `applyRulesSequence`. @traiansf , could you help us? I couldn't follow you at this part

andreiburdusa: But I think that was moved to `applyRewriteRulesParallel`
https://github.com/kframework/kore/pull/1026/files/c5216f71d35ce543a37d6eba5c3d8384e546522c#diff-595068f6f4dcaabbeb534249f4fd4c12R620

kore/test/Test/Kore/Step/Step.hs
andreiburdusa: It's used with `applyRewriteRulesSequence` at line 347, but I will change it to make it look cleaner

src/main/k/working/imp/merge/2.merge-output
```diff
@@ -0,0 +1,77 @@
+\rewrites{SortGeneratedTopCell{}}(
```
ttuegel: This shouldn't be checked in. Could we remove it, and add a `.gitignore` rule for it?

src/main/kore/prelude.kore
andreiburdusa: It was not. I removed it

kore/src/Kore/Unification/Unify.hs
```diff
@@ -169,3 +173,8 @@ runUnifierT = runExceptT . BranchT.gather . getUnifierT
  -}
 maybeUnifierT :: MonadSimplify m => UnifierT m a -> MaybeT m [a]
 maybeUnifierT = hushT . BranchT.gather . getUnifierT
+
+{- | Run a 'Unifier', returning empty list upon error.
+ -}
+listUnifierT :: MonadSimplify m => UnifierT m a -> m [a]
```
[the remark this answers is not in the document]
andreiburdusa: I will remove the `T` from the name since it doesn't make sense. But why is it not safe? `fromRight` takes `[]` as default value

src/main/k/working/tests/fresh-vars/fresh-vars.k
andreiburdusa: this was useful for debugging. I changed it back

kore/src/Kore/Step/Step.hs
```diff
@@ -409,6 +410,24 @@ finalizeRulesParallel initial unifiedRules = do
             $ Pattern.mapVariables Target.unwrapVariable <$> remainders'
         }
 
+checkFunctionLikeResults
```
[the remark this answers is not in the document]
andreiburdusa: I splitted `checkFunctionLikeResults` in two other functions, namely `assertFunctionLikeResults` and `recoveryFunctionLikeResults`. Maybe it's more intuitive now

PR #1028, opened 2019-09-27
Match abstract Map and Set

ttuegel:
@yzhang90 requested to be able to match certain combinations of abstract elements in Maps and Sets for function and simplification rules.

Needs: #1023

---

###### Reviewer checklist


ttuegel:
I'm abandoning this pull request. The tests I added at the end reveal that this implementation is subtly wrong, and that the behavior requested by the user is probably too naive to actually work. We do not need this at this time, so I am not going to fix it.

PR #1031, opened 2019-09-27
Consistent substitution normalization

ttuegel:
The matcher keeps the substitution normalized by construction, but several unification cases were inconsistent about normalizing the substitution. Unification isn't _finished_ until the substitution is normalized, so this doesn't make much sense.

This produced a lot of test failures, so it was a good exercise for tweaking the `Debug.Diff` output (see #1030).

In the future, we may consider changing how substitution normalization works. Presently, it does two different things: it normalizes the substitution itself (de-duplication, back-substitution, and occurs check) and it runs the predicate simplifier over the result. This is awkward: the predicate simplifier also calls substitution normalization. With this mutual recursion, it is hard to change either without introducing a loop. This also causes one of the big cyclic dependencies. We could consider instead keeping the unification `Predicate` always partially normalized (do de-duplication and back-substitution on-the-fly, these are pure operations we can do in the `Semigroup` instance) with a single check at the end of unification which performs the occurs check and re-unifies any predicates generated by de-duplication.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Axiom/Matcher.hs
+++ b/kore/src/Kore/Step/Axiom/Matcher.hs
@@ -178,7 +178,10 @@ matchIncremental termLike1 termLike2 =
                 $ MultiAnd.toPredicate predicate
             substitution' =
                 Predicate.fromSubstitution
-                $ Substitution.fromMap substitution
+                -- unsafeWrap is safe because the matcher guarantees the
+                -- substitution is normalized.
+                $ Substitution.unsafeWrap
+                $ Map.toList substitution
             solution = predicate' <> substitution'
         return solution
 
--- a/kore/src/Kore/Unification/Substitution.hs
+++ b/kore/src/Kore/Unification/Substitution.hs
@@ -101,13 +101,9 @@ data Substitution variable
         !(Map (UnifiedVariable variable) (TermLike variable))
     deriving GHC.Generic
 
--- | 'Eq' does not differentiate normalized and denormalized 'Substitution's.
-instance Ord variable => Eq (Substitution variable) where
-    (==) = Function.on (==) unwrap
+deriving instance Ord variable => Eq (Substitution variable)
 
--- | 'Ord' does not differentiate normalized and denormalized 'Substitution's.
-instance Ord variable => Ord (Substitution variable) where
-    compare = Function.on compare unwrap
+deriving instance Ord variable => Ord (Substitution variable)
 
 instance SOP.Generic (Substitution variable)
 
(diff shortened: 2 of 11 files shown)


kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -178,7 +178,10 @@ matchIncremental termLike1 termLike2 =
                 $ MultiAnd.toPredicate predicate
             substitution' =
                 Predicate.fromSubstitution
-                $ Substitution.fromMap substitution
+                -- unsafeWrap is safe because the matcher guarantees the
+                -- substitution is normalized.
+                $ Substitution.unsafeWrap
```
virgil-serbanuta: I actually liked that we had only one call to `unsafeWrap` in our non-testing code. Having the new checks in `unsafeWrap` certainly helps, but I'm going to make an argument that we still shouldn't use it. If, after reading it, you still think it's better this way, you can keep it as-is.

It does not seem obvious to me that "the matcher guarantees the substitution is normalized", and that it will always be so. Is there a strong argument for that?

I mean, maybe now it correctly matches `f(x, x)` to `f(g(a), g(b))` to get `x=g(a) and g(a)=g(b)`, and maybe it correctly matches `f(x, x)` and `f(y, C(y))`, where `C` is a constructor, to get `bottom`, and so on. However, it's not obvious that it does that (maybe it should be obvious, but I'm just missing something).

Even if it is obvious that it's already normalized, we may probably break this assumption at some point (intentionally or not), so why not go through the standard normalization? If it's about performance, we may consider adding a quick check in `normalizeSubstitution` to see whether everything is already normalized before doing anything else.

kore/src/Kore/Unification/Substitution.hs
```diff
@@ -209,7 +198,21 @@ unsafeWrap
     :: Ord variable
     => [(UnifiedVariable variable, TermLike variable)]
     -> Substitution variable
-unsafeWrap = NormalizedSubstitution . Map.fromList
+unsafeWrap =
+    NormalizedSubstitution . List.foldl' insertNormalized Map.empty
+  where
+    insertNormalized subst (var, termLike) =
```
virgil-serbanuta: Nice, we should have done this a long time ago...

kore/src/Kore/Step/Axiom/Matcher.hs
ttuegel: > It does not seem obvious to me that "the matcher guarantees the substitution is normalized", and that it will always be so. Is there a strong argument for that?

Yes. [`substitute`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Axiom/Matcher.hs#L445) performs an occurs-check (so there are no cyclic dependencies) and performs back-substitution (so no substitution depends on any other). For the sake of generating _matching_ solutions, it's also important that back-substitution is performed immediately: there is no possibility of generating duplicated substitutions (multiple substitutions for the same variable). We can't use the existing normalization for substitutions because it uses unification to de-duplicate substitutions.

If we ever changed this about the matcher, then we _must_ change this line too, and I would want it to throw an error otherwise.

virgil-serbanuta: Ok, I see, thanks.

ttuegel:
It turns out to be impossible to make the unification cases individually return normalized substitutions: it isn't safe to call the predicate simplifier, and it isn't possible to separate predicate simplification from substitution normalization. See #1041. We can salvage the other changes in this PR though.

PR #1039, opened 2019-09-30
Rename merge

virgil-serbanuta:
Fixes https://github.com/runtimeverification/polkadot-verification/issues/38

Some spurious diffs will go away after merging 
https://github.com/kframework/kore/pull/1038

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Rule/Combine.hs
+++ b/kore/src/Kore/Step/Rule/Combine.hs
@@ -179,9 +179,8 @@ mergeRules
     => NonEmpty (RewriteRule variable)
     -> simplifier [RewriteRule variable]
 mergeRules (a :| []) = return [a]
-mergeRules rules =
-    mergeDisjointVarRules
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Step/Rule/Combine.hs
```diff
@@ -176,7 +179,17 @@ mergeRules
     => NonEmpty (RewriteRule variable)
     -> simplifier [RewriteRule variable]
 mergeRules (a :| []) = return [a]
-mergeRules rules = BranchT.gather $ do
+mergeRules rules =
```
ttuegel: This is up to you, but you could use a view pattern here, get rid of the helper (`mergeDisjointVarRules`), and that would remove the possibility of anyone accidentally using the helper wrong in the future.

```suggestion
mergeRules (renameRulesVariables . Foldable.toList -> rules) =
```

PR #1040, opened 2019-09-30
Better error message for simplification with multiple results

yzhang90:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
@@ -236,21 +236,18 @@ applyFirstSimplifierThatWorks
               -- configuration.
               -- However, right now, we shouldn't be able to get more
               -- than one result, so we throw an error.
-              error
-                  (  "Unexpected simplification result with more "
-                  ++ "than one configuration: \n"
-                  ++ (show . Pretty.vsep $
-                        [ Pretty.indent 2 "input:"
-                        , Pretty.indent 4 (unparse patt)
-                        , Pretty.indent 2 "results:"
-                        , (Pretty.indent 4 . Pretty.vsep)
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -239,10 +239,21 @@ applyFirstSimplifierThatWorks
               error
                   (  "Unexpected simplification result with more "
                   ++ "than one configuration: "
-                  ++ show applicationResult
+                  ++ (show . Pretty.vsep $
+                        [ "Simplification result with remainder:"
```
ttuegel: Please change this to use the error message above, i.e. 
```suggestion
                        [ "Unexpected simplification result with more than one configuration:"
```

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -238,11 +238,21 @@ applyFirstSimplifierThatWorks
               -- than one result, so we throw an error.
               error
                   (  "Unexpected simplification result with more "
-                  ++ "than one configuration: "
-                  ++ show applicationResult
+                  ++ "than one configuration: \n"
+                  ++ (show . Pretty.vsep $
```
yzhang90: should i use `Pretty.renderStrict . Pretty.layoutCompact` before `Pretty.vsep` like the code below?

ttuegel: No, `show` should be fine.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
yzhang90: done

PR #1041, opened 2019-09-30
Refactor substitution normalization

ttuegel:
This pull request would refactor substitution normalization to remove a cyclic module dependency between the simplifier and the unifier. I also cleaned up and documented the implementation.

---

###### Reviewer checklist


kore/src/Kore/Unification/UnifierImpl.hs
```diff
@@ -71,123 +71,160 @@ simplifyAnds
...
+        :: Pattern variable
+        -> TermLike variable
+        -> unifier (Pattern variable)
+    simplifyAnds' intermediate pat =
+        case Cofree.tailF (Recursive.project pat) of
+            AndF And { andFirst = lhs, andSecond = rhs } ->
+                foldM simplifyAnds' intermediate [lhs, rhs]
+            _ -> do
+                result <- termUnification (Pattern.term intermediate) pat
+                normalizeExcept (result <* intermediate)
```
virgil-serbanuta: splitting intermediate into its term and predicate (`Pattern.splitTerm`) and using `andCondition` (or `<>` if it works here) instead of `<*` would make it more clear that we're not dropping the `intermediate`'s term by mistake here.

kore/src/Kore/Unification/UnifierImpl.hs
```diff
@@ -71,123 +71,160 @@ simplifyAnds
...
+        .  Ord variable1
+        => Map variable1 (NonEmpty term)
+        -> (variable1, term)
+        -> Map variable1 (NonEmpty term)
+    insertSubstitution multiMap (variable, termLike) =
+        let push = (termLike :|) . maybe [] Foldable.toList
+        in Map.alter (Just . push) variable multiMap
+
+    worker conditional@Conditional { predicate, substitution }
+      | all isSingleton substitutions || Substitution.isNormalized substitution
```
virgil-serbanuta: Doesn't `isNormalized` imply `all isSingleton`?

kore/test/Test/Kore/Step/Substitution.hs
```diff
@@ -130,8 +127,8 @@ test_mergeAndNormalizeSubstitutions =
         $ do
             let expect = Left $ UnificationError $ unsupportedPatterns
                     "Unknown unification case."
-                    (mkElemVar Mock.y)
                     (Mock.constr10 (mkElemVar Mock.y))
+                    (mkElemVar Mock.y)
```
virgil-serbanuta: Why this change and the similar one at line 222?

ttuegel: `simplifyAnds` was sensitive to the order of its arguments, even though that order has no semantic significance. The way I refactored normalization had the effect of changing the argument order. To avoid this problem in the future, I made `simplifyAnds` sort its arguments before unifying; this is the order with the arguments sorted.

kore/src/Kore/Unification/UnifierImpl.hs
ttuegel: `isNormalized` does not check the substitution itself, it only checks for the `NormalizedSubstitution` constructor. This is the short-circuit for when we try to normalize a substitution which is already normalized. (To think of it, that check should be first to short-circuit properly.)

PR #1047, opened 2019-10-01
Simplify the LHS of claims

### LINKED ISSUE #1013: Evaluate functions on the left-hand side of claims
See also: #1000

virgil-serbanuta:
Fixes #1013 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -76,6 +76,9 @@ import qualified Kore.IndexedModule.MetadataToolsBuilder as MetadataTools
 import Kore.IndexedModule.Resolvers
     ( resolveSymbol
     )
+import qualified Kore.Internal.MultiAnd as MultiAnd
+    ( extractPatterns
+    )
 import qualified Kore.Internal.MultiOr as MultiOr
 import Kore.Internal.Pattern
     ( Pattern
@@ -516,7 +519,8 @@ data InitializedProver =
 
 -- | Collect various rules and simplifiers in preparation to execute.
 initializeProver
-    :: MonadSimplify simplifier
+    :: forall simplifier a
+    .  MonadSimplify simplifier
     => VerifiedModule StepperAttributes Attribute.Axiom
     -> VerifiedModule StepperAttributes Attribute.Axiom
     -> (InitializedProver -> simplifier a)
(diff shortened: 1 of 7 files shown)


virgil-serbanuta:
@ttuegel Note that there is no sum-to-n test. It works, but takes too much time (around 15 min on my computer), so I removed it from the PR. I could rollback that commit and increase the timeout if needed.

kore/src/Kore/Exec.hs
```diff
@@ -274,15 +274,23 @@ prove
...
+    evalSimplifier definitionModule
+    $ initialize definitionModule
+    $ \initialized -> do
         tools <- Simplifier.askMetadataTools
         let Initialized { rewriteRules } = initialized
             specClaims =
                 map
                     (Bifunctor.second $ expandOnePathSingleConstructors tools)
                     (extractOnePathClaims specModule)
+            mapMSecond :: Monad m => (b -> m [c]) -> (a, b) -> m [(a, c)]
```
ttuegel: Do I understand correctly: This takes a function, which simplifies the left-hand side of a rule into a disjunction, and returns all of the rules that produces, given a single left- and right-hand side?

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -0,0 +1,74 @@
...
+    ( MonadSimplify
+    , SimplifierVariable
+    )
+
+simplifyOnePathRuleLhs
+    ::  ( MonadSimplify simplifier
+        , SimplifierVariable variable
+        )
+    => OnePathRule variable
+    -> simplifier [OnePathRule variable]
```
ttuegel: Would it make sense to return a `MultiAnd` here instead of a list? I'm thinking that because a disjunction on the left-hand side obligates us to prove all the claims, and also because there might be duplicated claims in this list.

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -0,0 +1,74 @@
...
+simplifyRuleLhs
+    ::  ( MonadSimplify simplifier
+        , SimplifierVariable variable
+        )
+    => RulePattern variable
+    -> simplifier [RulePattern variable]
+simplifyRuleLhs rule@(RulePattern _ _ _ _ _ _) = do
+    simplified <- Pattern.simplifyAndRemoveTopExists
+        (left `Conditional.withCondition` Predicate.fromPredicate requires)
+    BranchT.gather $ do
```
ttuegel: This `scatter`/`gather` combination is just a `map`.

kore/src/Kore/Step/Rule/Simplify.hs
virgil-serbanuta: MultiAnd requires a `TopBottom` instance. I did that, but I'm not sure that the result is better.

kore/src/Kore/Step/Rule/Simplify.hs
virgil-serbanuta: Fixed.

PR #1052, opened 2019-10-01
Assert that children are simplified

ttuegel:
This pull request would change the term simplifier to throw an error if any of its results is not marked `Simplified`. A term is simplified if we have exhausted our simplification routines on it, that is: if we do not know how to simplify it further. The only way to proceed with such terms is after their modification by a substitution.

---

###### Reviewer checklist


kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -0,0 +1,154 @@
...
+instance Synthetic Simplified (Ceil sort) where
+    synthetic = notSimplified
+    {-# INLINE synthetic #-}
+
+instance Synthetic Simplified (Floor sort) where
+    synthetic = notSimplified
+    {-# INLINE synthetic #-}
+
+instance Synthetic Simplified (DomainValue sort) where
+    synthetic = notSimplified
```
virgil-serbanuta: Why doesn't it return the child's simplified value?

kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -732,7 +732,8 @@ unifyEqualsNormalized
     renormalized <- normalize1 normalizedTerm
 
     let unifierTerm :: TermLike variable
-        unifierTerm = asInternal tools sort1 renormalized
+        unifierTerm =
+            TermLike.markSimplified $ asInternal tools sort1 renormalized
```
virgil-serbanuta: Shouldn't this also test that the keys/values/opaque stuff are simplified? Either as an assertion (e.g. `TermLike.markSimplified` could assert that all direct children are simplified), or as an actual computation input. Similar question for some of the other uses of `markSimplified`.

kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -0,0 +1,154 @@
...
+
+instance Synthetic Simplified (Next sort) where
+    synthetic = notSimplified
+    {-# INLINE synthetic #-}
+
+instance Synthetic Simplified (Rewrites sort) where
+    synthetic = notSimplified
+    {-# INLINE synthetic #-}
+
+instance Synthetic Simplified (Builtin sort) where
```
virgil-serbanuta: Shouldn't this be always true for ints, bools and strings, and shouldn't this be easy to compute for map, lists and sets?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -290,6 +293,10 @@ instance Synthetic syn Evaluated where
     synthetic = getEvaluated
     {-# INLINE synthetic #-}
 
+instance {-# OVERLAPS #-} Synthetic Pattern.Simplified Evaluated where
```
virgil-serbanuta: Do we still need `Evaluated` if we have the `Simplified` attribute?

kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -39,7 +39,7 @@ import Kore.Internal.OrPattern
     )
 import qualified Kore.Internal.OrPattern as OrPattern
 import Kore.Internal.Pattern as Pattern
-import Kore.Internal.TermLike
+import Kore.Internal.TermLike as TermLike
```
virgil-serbanuta: Would this be better with separate imports for the qualified and non-qualified stuff?

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -319,10 +312,11 @@ quantifyPattern variable original@Conditional { term, predicate, substitution }
     , "variable=" ++ unparseToString variable
     , "patt=" ++ unparseToString original
     ]
-  | quantifyTerm = mkExists variable <$> original
+  | quantifyTerm = markSimplified . mkExists variable <$> original
   | quantifyPredicate =
     Conditional.withCondition term
-    $ Predicate.fromPredicate
+    -- TODO (thomas.tuegel): This could probably be simplified further.
```
virgil-serbanuta: Could you give some hints about what kind of simplification you had in mind?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -153,8 +160,10 @@ simplifyInternal term predicate = simplifyInternalWorker term
     simplifyChildren = traverse simplifyInternalWorker
 
     simplifyInternalWorker termLike =
-        tracer termLike $
-        let doNotSimplify = return (OrPattern.fromTermLike termLike)
+        assertSimplifiedResults $ tracer termLike $
```
virgil-serbanuta: Shouldn't this start with checking whether `termLike` is already simplified?

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -56,7 +57,7 @@ result. The result is re-simplified once.
 
 -}
 simplify
-    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    :: (GHC.HasCallStack, SimplifierVariable variable, MonadSimplify simplifier)
```
ana-pantilie: This line is long. It's exactly at our 80 chars limit, and I think we can easily split it up, if you think that's ok.

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -129,12 +130,10 @@ See also: 'simplify'
 
 -}
 simplifyPartial
-    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    :: (GHC.HasCallStack, SimplifierVariable variable, MonadSimplify simplifier)
```
ana-pantilie: The same as above.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -111,7 +118,7 @@ import qualified Kore.Step.Simplification.Variable as Variable
 {-|'simplify' simplifies a `TermLike`, returning a 'Pattern'.
 -}
 simplify
-    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    :: (GHC.HasCallStack, SimplifierVariable variable, MonadSimplify simplifier)
```
ana-pantilie: Same as above.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -123,7 +130,7 @@ simplify patt predicate = do
 'OrPattern'.
 -}
 simplifyToOr
-    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    :: (GHC.HasCallStack, SimplifierVariable variable, MonadSimplify simplifier)
```
ana-pantilie: Same as above.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -136,7 +143,7 @@ simplifyToOr term predicate =
 
 simplifyInternal
     :: forall variable simplifier
-    .  (SimplifierVariable variable, MonadSimplify simplifier)
+    .  (GHC.HasCallStack, SimplifierVariable variable, MonadSimplify simplifier)
```
ana-pantilie: Same as above.

kore/src/Kore/Step/Simplification/TermLike.hs
ttuegel: I want to do that in another pull request, after I am satisfied that the simplifier is _actually_ idempotent (probably adding property tests).

kore/src/Kore/Step/Simplification/Exists.hs
ttuegel: Yes, I only meant that we are constructing an `\and` without invoking the corresponding simplifier. I will make the comment more specific.

kore/src/Kore/Builtin/AssociativeCommutative.hs
ttuegel: No, we should not be inspecting children here. It is not necessary because the simplifier recurses through the entire pattern, so if the children were left unsimplified, an error would already have been thrown. It would also cause a lot of spurious test failures because all the unit tests send un-simplified patterns.

kore/src/Kore/Attribute/Pattern/Simplified.hs
ttuegel: For Int, Bool, and String: yes, these are always simplified. It cannot be computed for Map, List, or Set: being "simplified" isn't a property of the pattern, it tells the history of the pattern, that is, "this pattern has been run through the simplified since the last time a substitution was applied."

kore/src/Kore/Builtin/AssociativeCommutative.hs
virgil-serbanuta: This is also called outside of the simplifier, i.e. by the unifier, which should, in principle, also sends simplified stuff, but that might not be always so. I mean, do we actually simplify the left hand term of rules? They are probably simplified since we internalize stuff, and usually there's nothing to simplify there, but I don't think that would be true if we had a rule with a function symbol in the left.

So, if my theory above is valid, then it's not always obvious that we are using `TermLike.markSimplified` properly, so maybe it would be better to do the check and find a reasonable way to fix those tests. If it's not valid, then I'll just approve the PR.

kore/src/Kore/Attribute/Pattern/Simplified.hs
virgil-serbanuta: Ok, should this instance return "alwaysSimplified" for those cases?

kore/test/Test/Kore/Step/Simplification/Ceil.hs
```diff
@@ -375,27 +373,39 @@ test_ceilSimplification =
...
-            Conditional
-                { term = Mock.builtinMap [(fOfA, fOfB), (gOfA, gOfB)]
-                , predicate = makeTruePredicate
-                , substitution = mempty
-                }
-        assertEqual "ceil(map)" expected actual
+    , testGroup "Builtin.Map"
+        [ testCase "concrete partial keys" $ do
+            -- maps assume that their keys are relatively functional, so
+            -- ceil({a->b, c->d}) = ceil(b) and ceil(d)
```
virgil-serbanuta: I guess that this is not directly related to this PR, but, since this instance of this comment showed up here, I'm going to mention this:

https://github.com/kframework/kore/issues/1077

I.e. this test is probably somewhat broken (as is the original that was used as a base for this), and using constructors or domain values as keys below may be better.

kore/src/Kore/Attribute/Pattern/Simplified.hs
ttuegel: Yes, I think it should.

kore/src/Kore/Builtin/AssociativeCommutative.hs
virgil-serbanuta: Discussed on Slack, this `markSimplified` should go away.

ttuegel: Sorry, I think I misunderstood what you were saying. `unifyEqualsNormalized` is the "canonical" simplifier for Map and Set, so this should use `markSimplified`, but you are right that it should _only_ do that if the children are simplified.

Issue #947, opened 2019-09-11
Syntax documentation is out of date


ttuegel:
_The Semantics of K_ is out-of-date in a number of ways, particularly the sections describing meta-Kore and the syntax of Kore. It is probably not useful to update the sections related to meta-Kore, which has gone away. On the other hand, it is misleading to have a document which is up-to-date in some ways and outdated in others. We should have a separate, up-to-date, and smaller document which gives the syntax of Kore, so that the teams working on other backends do not have to treat our parser as a reference implementation.

PR #1054, opened 2019-10-02
Kore syntax

traiansf:
this PR does 3 things:
* removing CharLiteral
* improved Id parsing
* adding a document containing the Syntax of Kore
---

###### Reviewer checklist


traiansf:
Fixes #947

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+--------------
+
+### Comments
+
+- C-style comments (non-nested)
+
+### Strings
+
+```
+StringLiteral ::= "\"" {Char}* "\""
```
virgil-serbanuta: It's fairly obvious, but, if it's not hard, I'd add a link to something describing the syntax used here, e.g. that things between quotation marks are literals, curly braces are used to identify things to which "repeaters" apply (e.g. `+*?`), and so on.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+    | "\u" {HexDigit}4
+    | "\U" {HexDigit}8
+
+HexDigit ::= ['0'..'9', 'A'..'F', 'a'..'f']
+```
+
+### KeyWords
+
+```
+Keywords = "module" | "endmodule" | "sort" | "symbol" | "alias" | "axiom"
```
virgil-serbanuta: "claim", "where", "hooked-sort", "hooked-symbol"?

kore/src/Kore/Parser/Lexeme.hs
```diff
@@ -217,36 +229,44 @@ idCharSet :: CharSet
...
-            idToken <- objectIdRawParser KeywordsPermitted
-            return (c:idToken)
+    start <- Parser.char '@'
+    end <- genericKoreIdRawParser KeywordsPermitted
+    return (start:end)
+
+{-|'elemVarIdRawParser' extracts the string representing a
+@set-variable-identifier@. Does not consume whitespace.
+
+Always starts with @@@
```
virgil-serbanuta: Should't @ be escaped, see, e.g., line 121 (documentation for `variableIdParser`)?

src/test/resources/test-exception-3.kore
```diff
@@ -1,8 +0,0 @@
-[]
-
-module TEST-EXCEPTION-3
-
-  axiom{S}
-    'X:S []
```
virgil-serbanuta: Do we have an equivalent test with strings?

kore/src/Kore/Parser/Lexeme.hs
```diff
@@ -217,36 +229,44 @@ idCharSet :: CharSet
...
-objectIdRawParser :: IdKeywordParsing -> Parser String
-objectIdRawParser keywordParsing = do
-    backslash <- Parser.string "\\" <|> pure ""
-    name <- objectNonslashIdRawParser keywordParsing
-    pure (backslash ++ name)
-
-objectNonslashIdRawParser :: IdKeywordParsing -> Parser String
-objectNonslashIdRawParser =
+symbolIdRawParser :: Parser String
+symbolIdRawParser = do
```
ttuegel: Why not,

```.hs
symbolIdRawParser =
    ((:) <$> char '\\' <*> genericKoreIdRawParser KeywordsPermitted)
    <|> genericKoreIdRawParser KeywordsForbidden
```

(Forgiving my formatting.)

kore/src/Kore/Parser/Lexeme.hs
```diff
@@ -97,35 +100,37 @@ sourcePosToFileLocation
...
+    name <- lexeme idRawParser
     return Id
         { getId = Text.pack name
         , idLocation = AstLocationFile pos
         }
 
+variableIdRawParser :: Parser String
+variableIdRawParser = elemVarIdRawParser <|> setVarIdRawParser
+
+{-|'variableIdParser' parses either an @element-variable-identifier@, or a
```
ttuegel: The documentation here and below should be updated to reflect the new syntax document.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Strings
+
+```
+StringLiteral ::= "\"" {Char}* "\""
+
+Char ::= EscapeChar || AsciiChar || PrintableChar
+
+AsciiChar ::= first 128 ASCII characters
+PrintableChar ::= printable according to the C++ iswprint function
```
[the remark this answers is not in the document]
traiansf: Actually their definition is pretty strange, because it depends on the current `locale` settings: I was quite surprise we do support these and I'm pretty certain that maybe we shouldn't.

docs/kore-syntax.md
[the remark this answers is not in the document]
traiansf: You are perfectly right; Virgil also mentioned that. What is interesting is that our parsed only accepts the ones I listed :-)
I'll correct both.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+
+An identifier cannot be a keyword
+
+Context-Free Syntax
+-------------------
+
+```
+Definition ::= Attributes {Module}+
+
+Attributes ::= "[" {Pattern ","}* "]"
```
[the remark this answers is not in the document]
traiansf: I kind of agree with you; however, our parser currently accepts any pattern.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+
+Module ::= "module" ModuleNameId {Sentences}* "endmodule" Attributes
+
+Sentence
+  ::= ImportStatement
```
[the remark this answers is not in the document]
traiansf: I'm not 100% sure; if it's an internal document, OK. If we mean it as external, we might want to simplify certain names.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+    | Axiom
+    | Claim
+
+ImportStatement ::= "import" ModuleNameId Attributes
+
+SortDefinition ::= "sort" SortId SortParameters Attributes
+HookedSortDefinition ::= "hooked-sort" SortId SortParameters Attributes
+
+SymbolDefinition ::=
+    "symbol" SymbolId SortParameters "(" { Sort ","}* "}" ":" Sort Attributes
```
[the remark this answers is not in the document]
traiansf: Well, in the SDF https://en.wikipedia.org/wiki/Syntax_Definition_Formalism ,  `{ Sort ","}*` is understood as not having a final comma and this notation was created especially to avoid using extra productions when defining lists.

Maybe we should clarify our syntax definition. And we could indeed define lists, if you think it would make things easier to read for newcomers.

kore/src/Kore/Parser/Lexeme.hs
traiansf: Maybe we could rewrite more of our parsing infrastructure this way;  This was a way to clearly show that you only have 1 char lookahead. In your definition, this is harder to grasp.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+- C-style comments (non-nested)
+
+### Strings
+
+```
+StringLiteral ::= "\"" {Char}* "\""
+
+Char ::= EscapeChar || AsciiChar || PrintableChar
+
+AsciiChar ::= first 128 ASCII characters
```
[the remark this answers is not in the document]
traiansf: I just stated what our parser accepts. maybe that needs to change, too.

kore/src/Kore/Parser/Lexeme.hs
traiansf: probably.

docs/kore-syntax.md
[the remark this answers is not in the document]
virgil-serbanuta: Does our parser really accept """ as a valid string? I thought it will take the first two quotation marks as being the string and produce an error at the third.

docs/kore-syntax.md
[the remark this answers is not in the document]
ttuegel: > Actually their definition is pretty strange, because it depends on the current `locale` settings

That shouldn't be the case, are you sure? The Unicode character classes are fixed. If your system is using a non-Unicode encoding, the locale settings will influence how characters are decoded from your system encoding to Unicode, but we only accept ASCII anyway. If this is really influenced by locale settings, then we need to fix the implementation.

kore/src/Kore/Parser/Lexeme.hs
ttuegel: That's true. It's a trade-off: the current implementation shows look-ahead very clearly versus the `Alternative` definition can be made to look (almost exactly) like the syntax document. We should keep it as-is to be consistent.

docs/kore-syntax.md
traiansf: I've changed it to valid BNF and added a link to wikipedia

src/test/resources/test-exception-3.kore
traiansf: good catch, we didn't.

docs/kore-syntax.md
```diff
@@ -0,0 +1,143 @@
...
+[Semantics of K](docs/semantics-of-k.pdf)
+
+Any change in the KORE Parser infrastructure should be reflected here.
+
+Lexical Syntax
+--------------
+
+### Comments
+
+- C-style comments (non-nested)
```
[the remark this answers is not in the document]
traiansf:
```.hs
skipWhitespace ::  Parser ()
skipWhitespace =
    L.space Parser.space1 lineComment blockComment
  where
    lineComment = L.skipLineComment "//"
    blockComment = L.skipBlockComment "/*" "*/"
```

So, we have line comment, and block comments are not nested as there is another function which would do that: https://hackage.haskell.org/package/megaparsec-7.0.5/docs/Text-Megaparsec-Char-Lexer.html#v:skipBlockCommentNested

docs/kore-syntax.md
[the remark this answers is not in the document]
traiansf: I based my comment on this: https://en.cppreference.com/w/c/string/wide/iswprint
and on the fact that `Data.Char.isPrint` delegates to `iswprint`: https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Unicode.html#isPrint

docs/kore-syntax.md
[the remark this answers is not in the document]
traiansf: I've changed everything to BNF and defined enumerations manually.

docs/kore-syntax.md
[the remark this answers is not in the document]
ttuegel: It does not, it delegates to `u_iswprint` which is defined somewhere in the RTS. I'm still looking it up at the moment, but I doubt it delegates to `iswprint`, which has the wrong signature on most platforms.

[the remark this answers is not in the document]
ttuegel: `u_iswprint` is defined as this [predicate](https://gitlab.haskell.org/ghc/ghc/blob/master/libraries/base/cbits/WCsubst.c#L5054) on Unicode character classes.

kore/src/Kore/Parser/Lexeme.hs
```diff
@@ -97,35 +100,37 @@ sourcePosToFileLocation
...
-@set-variable-identifier@.
-
-The @set-variable-@ version always starts with @\@@, while the @object-@
-one does not.
+{-| Generic identifier parser. Takes a parser for the string of the identifier
+ and returns an 'Id' annotated with position.
 -}
-idParser :: Parser Id
-idParser = do
+genericIdParser :: Parser String -> Parser Id
```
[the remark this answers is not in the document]
traiansf: True. I've renamed it.

docs/kore-syntax.md
```diff
@@ -0,0 +1,300 @@
...
+Context-Free Syntax
+-------------------
+
+The definition below uses standard [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form).
+
+Nevertheless, we use following notational convention for enumerations.
+- enumeration symbols are plurals of the symbols they enumerate over.
+  For example, `<attributes>` enumerates over `<attribute>`
+- enumeration uses comma as a separator; enumeration symbols using
+  whitespace as a separator are suffices with `_`, e.g., `<sentences>`
```
andreiburdusa: I can't find the `_` in the example

docs/kore-syntax.md
```diff
@@ -0,0 +1,300 @@
...
+
+String literals are defined in the [Lexical Syntax](#lexical-syntax) section
+below.
+
+### Enumerations
+
+```
+<modules_+>
+  ::= <module>
+    | <module> <modules+>
```
andreiburdusa: Shouldn't this be ```| <module> <modules_+>```

docs/kore-syntax.md
```diff
@@ -0,0 +1,300 @@
...
+
+<sentences_>
+  ::= ""
+    | <sentence> <sentences>
```
andreiburdusa: Shouldn't this be ```| <sentence> <sentences_>``` ?

kore/test/Test/Kore/Parser/Lexeme.hs
```diff
@@ -323,105 +320,6 @@ stringLiteralParserTests =
             ]
         ]
 
-charLiteralParserTests :: [TestTree]
```
ttuegel: Where did these go?

src/test/resources/test-exception-3.kore.golden
```diff
@@ -1,6 +1,6 @@
...
+../src/test/resources//test-exception-3.kore:6:12:
   |
-6 |     'X:S []
-  |       ^
-unexpected ':'
-expecting '''
+6 |     "X:S []
+  |            ^
+unexpected newline
+expecting '"' or printable ASCII character
```
ttuegel: Should we improve this error message? It could also be an escape sequence.

kore/test/Test/Kore/Parser/Lexeme.hs
traiansf: I just removed them, as we don't have any `CharLiteral` type; however, we have plenty of `StringLiteral` tests so we should be covered

src/test/resources/test-exception-3.kore.golden
traiansf: well, it cannot be an escape sequence anymore, because it doesn't start with `\`.  The message changed because I used peekChar to look one char ahead to chose between escape chars and printable chars.

ttuegel: It could be the start of an escape sequence, i.e. it could be `\` itself.

traiansf: that's what I'm saying, that there is a check for `\` before we get here, so it couldn't be  the start of an escape sequence.

ttuegel: A `\` is permitted there by the syntax.

traiansf: I see. however, these messages are generated automatically from the "labels" we're adding, and because now we're not attempting `\` anymore in this context, it got "automagically" removed. Not sure how to put it back.

ttuegel: That makes sense. I wouldn't worry about such a corner case right now.

PR #1059, opened 2019-10-03
add debug flag to kbmc

yzhang90:
---

###### Reviewer checklist


Jenkinsfile
```diff
@@ -72,26 +72,6 @@ pipeline {
         '''
       }
     }
-    stage('Integration: KEVM') {
```
ttuegel: Please re-enable these tests, but otherwise this is fine with me!

yzhang90: sure. I just want to measure how much overhead adding debug flag would incur.

PR #1061, opened 2019-10-03
Improve unparsing And

ttuegel:
---

###### Reviewer checklist


ttuegel:
> It would be nice to have a test for this, both because tests are always good, and because we'd actually see what this change does, as opposed to inferring it. Anyway, I think I understood it and I'm fine with merging it with or without a test.

Done!

PR #1063, opened 2019-10-04
Incomplete unification in intermediate results

### LINKED ISSUE #1044: Allow incomplete intermediate unification results
We would like to solve the following unification problem:

```
\in(
  <key>KEY</key>
  <map>
      KEY |-> VALUE
      MAP
  </map>,

  <key>KEY'</key>
  <map>
      KEY' |-> VALUE'
      MAP'
  </map>
)
```

There is a unique solution:

```
KEY' = KEY
VALUE' = VALUE
MAP' = MAP
```

assured by matching the `<key>` cell.

Presently, this unification is not implemented because unifying the `<map>` cells would initially branch between the accepted solution and another which has `KEY |-> VALUE` in `MAP'`. We would like to initially return a unification predicate when the Map unification problem is solvable in principle.

See also: https://runtimeverification.slack.com/archives/CC360GUTG/p1569870788227000

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Builtin/AssociativeCommutative.hs
+++ b/kore/src/Kore/Builtin/AssociativeCommutative.hs
@@ -1267,7 +1267,8 @@ unifyEqualsElementLists
             first
             second
         Just remainderTerm -> case opaque of
-            Var_ _ -> do
+            Var_ _
+              | TermLike.isFunctionPattern remainderTerm -> do
                 opaqueUnifier <- unifyEqualsChildren opaque remainderTerm
                 let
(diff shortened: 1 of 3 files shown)


ana-pantilie:
> This seems reasonable to me, but I don't understand how it fixes the problem we encountered. thinking

@ttuegel From what I understood by reading https://github.com/kframework/kore/issues/1044, this PR should only illustrate what @virgil-serbanuta is talking about here https://github.com/kframework/kore/issues/1044#issuecomment-539547578, so I don't think this should fix anything?

PR #1064, opened 2019-10-04
[DRAFT]remove simplification step

yzhang90:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Strategies/Goal.hs
+++ b/kore/src/Kore/Strategies/Goal.hs
@@ -402,8 +402,6 @@ onePathFirstStep axioms =
         , Simplify
         , TriviallyValid
         , ResetGoal
-        , Simplify
-        , TriviallyValid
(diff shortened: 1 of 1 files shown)


yzhang90:
Here is the reasoning:
`Simplify` on line 402 simplifies the GoalRemainder.
`ResetGoal` on line 404 change GoalRewritten to Goal
and `Simplify` on line 405 tries to simplify the GoalRewritten in the previous step.
**If in `DeriveSeq` on line 401, we already simpilfy the GoalRewritten**(is this true?), there is no need to simplify it  again in 405.

Testing on sum-to-n spec shows that it reduces the proving time from 11mis to 9 mins.

virgil-serbanuta:
I think DeriveSeq does not simplify the result.

yzhang90:
ok. I'll close the PR.

Issue #988, opened 2019-09-19
Integration testing with upstream alias changes


ttuegel:
The frontend changes to use aliases as the left-hand side of rules are at kframework/k#767; we need to run the backend's integration tests with those changes.

yzhang90:
please use https://github.com/kframework/k/tree/priority-group this branch

PR #1076, opened 2019-10-09
Expand aliases when parsing rules

ana-pantilie:
See https://github.com/kframework/kore/issues/988#issuecomment-539571208
Fixes https://github.com/kframework/kore/issues/988

Note: I didn't move `substituteWorker` to `Kore.Internal.Alias` because that would have resulted in a cyclic import issue between `Kore.internal.Alias` and `Kore.Internal.TermLike`

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Rule.hs
+++ b/kore/src/Kore/Step/Rule.hs
@@ -111,7 +111,7 @@ import Kore.Sort
     , SortVariable (SortVariable)
     )
 import Kore.Step.Simplification.ExpandAlias
-    ( substituteWorker
+    ( substituteInAlias
     )
 import Kore.Substitute
     ( SubstitutionVariable
@@ -565,7 +565,7 @@ patternToAxiomPattern
     -> TermLike Variable
     -> Either (Error AxiomPatternError) (QualifiedAxiomPattern Variable)
 patternToAxiomPattern attributes pat
-  | isJust . getPriority . Attribute.priority $ attributes = do
+  | isJust . getPriority . Attribute.priority $ attributes =
     case pat of
         Rewrites_ _
             (And_ _ (Not_ _ antiLeft) (And_ _ requires lhs))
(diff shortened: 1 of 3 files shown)


ttuegel:
> I didn't move `substituteWorker` to `Kore.Internal.Alias` because that would have resulted in a cyclic import issue between `Kore.internal.Alias` and `Kore.Internal.TermLike`

Oh, yeah, I should have realized that. :smile:

kore/src/Kore/Step/Rule.hs
```diff
@@ -591,6 +594,13 @@ patternToAxiomPattern attributes pat
                 , ensures = Predicate.wrapPredicate ensures
                 , attributes
                 }
+        Rewrites_ _ (ApplyAlias_ alias params) rhs ->
```
ttuegel: Could you add a unit test for this?

kore/src/Kore/Step/Simplification/ExpandAlias.hs
```diff
@@ -66,18 +67,18 @@ expandSingleAlias
...
     \case
-        ApplyAlias_ alias children -> pure $ substituteWorker alias children
+        ApplyAlias_ alias children -> pure $ substituteInAlias alias children
         _ -> Nothing
 
 -- TODO(Vladimir): Check aliases such that the intersection of alias variables
 -- with the *bounded* variables in the rhs is empty (because we can't currently
 -- handle things like \mu.
-substituteWorker
+substituteInAlias
```
vladciobanu: Would `expandAlias` be a good name for this?

vladciobanu: You're right. I think `expandAlias` would be a better name for `substituteInAlias`, but I can't think of a good name for the current `expandAlias`, since it's a bit weird with its recurse parameter. I'll think about it later, but you can disregard this comment. I can do the renaming myself, later.

Issue #1081, opened 2019-10-09
Property tests for simplification and the Simplified attribute


virgil-serbanuta:
Part of #1070

ttuegel:
I'm waiting for @ehildenb to get back to me about the pattern generator he prototyped, but basically we want to use [Hedgehog.Gen.recursive](https://www.stackage.org/haddock/lts-14.8/hedgehog-1.0.1/Hedgehog-Gen.html#v:recursive) to build a `TermLike` from two lists of (respectively) non-recursive and recursive generators. The function will look something like,

```.hs
genTermLike :: Fragment -> Gen (TermLike Variable)

-- | Describes a fragment of a definition.
data Fragment =
    Fragment
        { genSort :: Gen Sort
        , genValue :: Sort -> [Gen Sort]
            -- ^ Lists of generators for sorts with builtin values.
        , genSymbol :: Sort -> Gen Internal.Symbol
            -- ^ Generate a symbol returning the Sort.
            -- Should use 'Hedgehog.Gen.sized' to control number of arguments.
        , ...
        -- eventually need more fields here, probably
        }
```

virgil-serbanuta:
Depends on #1133 #1134 #1135 #1136

PR #1082, opened 2019-10-09
Pattern generators for simplification

virgil-serbanuta:
This does not use Gen.recursive, as #1081 expects, because we need to pass the sort down and `recursive` does not allow us to do that.

https://github.com/kframework/kore/pull/1109 , which is based on this, implements actual tests, but that PR can't be merged yet, see its comment.

---

###### Reviewer checklist

### DIFF
--- a/kore/test/Test/ConsistentKore.hs
+++ b/kore/test/Test/ConsistentKore.hs
@@ -1,0 +1,821 @@
+module Test.ConsistentKore
+    ( CollectionSorts (..)
+    , Context (..)
+    , termLikeGen
+    ) where
+
+import qualified Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+import Control.Comonad.Trans.Cofree
+    ( CofreeF ((:<))
+    )
+import Control.Monad.Reader
+    ( ReaderT
+    )
+import qualified Control.Monad.Reader as Reader
+import qualified Data.Functor.Foldable as Recursive
+import Data.Maybe
+    ( catMaybes
+    )
+import qualified Data.Set as Set
+import Data.Text
+    ( Text
+    )
+import qualified Data.Text as Text
+
+import qualified Kore.Attribute.Symbol as Attribute
+    ( Symbol
+    )
+import qualified Kore.Builtin.Bool.Bool as BuiltinBool
+    ( asBuiltin
+    )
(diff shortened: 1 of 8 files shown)


kore/test/Test/ConsistentKore.hs
```diff
@@ -0,0 +1,849 @@
...
+rewritesGenerator :: Gen TermGenerator
+rewritesGenerator = binaryOperatorGenerator mkRewrites
+
+topGenerator :: Gen TermGenerator
+topGenerator = nullaryFreeSortOperatorGenerator mkTop
+
+evaluatedGenerator :: Gen TermGenerator
+evaluatedGenerator = unaryOperatorGenerator mkEvaluated
+
+maybeStringLiteralGenerator :: Gen (Maybe TermGenerator)
```
ttuegel: I think it's a little awkward returning the `TermGenerator` inside `Gen` when we don't do any generation to prepare the `TermGenerator`. I would say the same thing about the builtin generators below.

kore/test/Test/ConsistentKore.hs
```diff
@@ -0,0 +1,849 @@
...
+        :: (Sort -> Gen (TermLike Variable))
+        -> Gen (Domain.Builtin (TermLike Concrete) (TermLike Variable))
+    }
+
+data CollectionSorts = CollectionSorts
+    { collectionSort :: !Sort
+    , elementSort :: !Sort
+    }
+
+data Context = Context
```
ttuegel: I think it would make sense to split this up: there are some fields which seem only to be used to prepare the `TermGenerator`s, and some which are only used during the actual generation (such as `allowedElementVariables`). That is, there seems to be a stage distinction here.

If you add a field like `currentSort :: Sort` alongside `allowedElementVariables`, then you can use `Gen.recursive` and use `Reader.local` to set the sort of the child pattern. However, I don't think you need to do this, I'm just remarking on how (I think) `Gen.recursive` is intended to be used.. The way it's written now seems perfectly natural to me.

kore/test/Test/ConsistentKore.hs
virgil-serbanuta: Fixed. Some of these functions got an extra argument, but that's ok, I guess.

kore/test/Test/ConsistentKore.hs
virgil-serbanuta: I split this, and didn't use `recursive`.

Issue #1069, opened 2019-10-07
Do not discard entire substitution upon simplifiable cycle


ttuegel:
Consider a substitution with a simplifiable cycle, such as

```
x = andBool(x, true) ∧ y = a()
```

Although applying the substitution for `x` would not terminate, we could apply the substitution for `y`. Presently, we convert this entire substitution to a predicate and apply none of it. This blocks further simplification because we cannot apply the rest of the substitution.

See also: #998, #1002, #1052 

---

- [x] Factor a function out of `normalizeSubstitution` which returns the normalized and denormalized parts of the substitution. (#1085)
- [x] Duplicate the predicate simplifier predicate simplifier into the unifier. (#1113)
- [x] Ensure that the unifier predicate simplifier does not call the predicate simplifier predicate simplifier, and vice versa. (#1113)
- [ ] Modify the predicate simplifier predicate simplifier to call the extracted function which returns the normalized and denormalized parts of the substitution, and only promote the denormalized part to the predicate. (#1118)

ttuegel:
This is also related to some proofs sending too many queries to Z3.

PR #1085, opened 2019-10-09
Refactor substitution normalization to allow partial normalization

ttuegel:
See also: #1069 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Unification/SubstitutionNormalization.hs
+++ b/kore/src/Kore/Unification/SubstitutionNormalization.hs
@@ -82,9 +82,6 @@ normalizeSubstitution
     -> ExceptT SubstitutionError m (Predicate variable)
 normalizeSubstitution = normalizeSubstitution'
 
-data Normalization variable term =
-    Normalization { normalized, denormalized :: !(Map variable term) }
-
 normalizeSubstitution'
     :: forall m variable
     .  (MonadSimplify m, SimplifierVariable variable)
--- a/kore/src/Kore/Unification/SubstitutionNormalization.hs
+++ b/kore/src/Kore/Unification/SubstitutionNormalization.hs
@@ -80,15 +80,15 @@ normalizeSubstitution
     => Map (UnifiedVariable variable) (TermLike variable)
     -> ExceptT SubstitutionError m (Predicate variable)
 normalizeSubstitution substitution = do
-    normalized <- liftEither $ normalizeSubstitution' substitution
+    normalized <- liftEither $ normalize substitution
     return $ fromMaybe Predicate.bottom normalized
 
-normalizeSubstitution'
+normalize
     :: forall variable
     .  SimplifierVariable variable
     => Map (UnifiedVariable variable) (TermLike variable)
     -> Either SubstitutionError (Maybe (Predicate variable))
-normalizeSubstitution' (dropTrivialSubstitutions -> substitution) =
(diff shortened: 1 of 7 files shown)


kore/src/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -78,130 +73,190 @@ x = f(x) or something equivalent).
...
+    ->  Maybe (Normalization variable)
+normalize (dropTrivialSubstitutions -> substitution) =
+    case topologicalSort allDependencies of
+        Right order -> sorted order
+        Left _ ->
+            case topologicalSort nonSimplifiableDependencies of
+                Right variables ->
+                    -- Removing the non-simplifiable dependencies removed the
+                    -- cycle, therefore the cycle is simplifiable.
+                    simplifiableCycle variables
```
virgil-serbanuta: Does this work properly with multiple cycles (a->b, b->a, c->d, d->c)? Or with cycles one within another (a->b, b->c, c->a, c->b)? Are there tests for that?

ttuegel: It works as well as it ever did, but there are not many tests for cycles. I will refresh the tests.

virgil-serbanuta: I'm not sure how to reply to "works as well as it ever did".

My point was that code that worked before when a cycle was treated as "a generic error in the substitution, so we need to dump the entire thing in the predicate" might or might not be good enough if you're splitting the substitution, i.e. when you're treating the cycle as an error with some very specific details.

I.e. if you have two cycles, does the topological sort algorithm return only one of them? If yes, would the current code consider the remaining cycle as "normalized" or "denormalized"?

I don't remember how the topological sorting algorithm works, and the current code may work properly, but I thought I'd ask. I would have checked this more carefully, but I wasn't one of the assigned reviewers, so I enjoyed the lack of responsibility and just sent a thought that crossed my head when reading the code. Sorry if I bothered you for nothing.

BTW, I didn't see any (new) tests for the issues mentioned above, not sure if that's because it's too late and I missed them, or they weren't needed, or they were needed, but the current code didn't work and required too many fixes, or whatever.

ttuegel: > My point was that code that worked before when a cycle was treated as "a generic error in the substitution, so we need to dump the entire thing in the predicate" might or might not be good enough if you're splitting the substitution, i.e. when you're treating the cycle as an error with some very specific details.

This doesn't change the behavior of substitution simplification: _any_ cycle is still a generic error, and we dump the entire substitution into the predicate. This is only a refactoring so that there is a function in the substitution normalizer that could possibly recover. In the original code, we throw a `Left` as soon as there is a problem, and there's practically no chance of recovering. In the refactored code, there is a wrapper that throws `Left` and an inner part which splits the substitution, but the wrapper throws `Left` if the substitution is split at all. The wrapper is the only part that is actually used during simplification, so the behavior of substitution normalization is the same as always. (That's what I meant by "it works as well as it ever did": I didn't change how this code works, it's only moving things around. I'm sorry, I realize now that it sounded snide.)

I haven't changed the substitution-splitting behavior yet because I need to separate the predicate simplifier used for unification from the one used for ordinary simplification. The unification simplifier should regard a denormalized substitution as an error instead of splitting on it, but we can't do that yet.

Issue #1070, opened 2019-10-07
Do not apply simplifier to fully-simplified patterns


ttuegel:
The simplifier should be idempotent, so there is no need to re-apply it to fully-simplified patterns.

See also: #1052

---

- ~Property tests for simplification (@virgil-serbanuta, #1081 )~
  - [x] Generator for well-formed patterns in a fragment of a definition
- [x] Skip simplification for patterns which are marked `Simplified` (@ana-pantilie)

PR #1086, opened 2019-10-10
Skip simplification if pattern is already simplified

ana-pantilie:
See https://github.com/kframework/kore/issues/1070

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Pattern.hs
```diff
@@ -57,10 +58,13 @@ simplify
         )
     => Pattern variable
     -> simplifier (OrPattern variable)
-simplify pattern' = do
-    orSimplifiedTerms <- simplifyConditionalTermToOr predicate term
-    fmap OrPattern.fromPatterns . Branch.gather $ do
-        simplifiedTerm <- Branch.scatter orSimplifiedTerms
-        Pattern.mergeWithPredicate predicate simplifiedTerm
+simplify pattern' =
+    if isSimplified pattern'
```
ttuegel: I think this check should go in `Kore.Step.Simplification.TermLike.simplifyInternal`. The reason is: this will re-simplify the _entire_ configuration if any part is unsimplified, but if we put the check in `simplifyInternal`, it will avoid re-simplifying _parts_ of the configuration that are unchanged.

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -94,12 +94,17 @@ simplifyPartial
...
     =>  Syntax.Predicate variable
     ->  BranchT simplifier (Predicate variable)
-simplifyPartial predicate = do
-    patternOr <-
-        Monad.Trans.lift $ simplifyTerm $ Syntax.unwrapPredicate predicate
-    -- Despite using Monad.Trans.lift above, we do not need to explicitly check
-    -- for \bottom because patternOr is an OrPattern.
-    scatter (eraseTerm <$> patternOr)
+simplifyPartial predicate =
+     if Syntax.isSimplified predicate
```
ttuegel: Is this check necessary? Does `simplifyTerm` have the same check?

kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -141,15 +141,21 @@ evaluateApplication
...
     termLike = synthesize (ApplySymbolF application)
     unevaluated =
         OrPattern.fromPattern
         $ Pattern.withCondition
-            (TermLike.markSimplified termLike)
+            (markSimplifiedIfChildren termLike)
             childrenPredicate
 
+    markSimplifiedIfChildren =
+        if all TermLike.isSimplified children
```
ttuegel: `Application` is `Foldable`, so you don't need to explicitly unpack the `children` here:
```suggestion
        if all TermLike.isSimplified application
```

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -155,10 +156,10 @@ simplifyToOr
     =>  Predicate variable
     ->  TermLike variable
     ->  simplifier (OrPattern variable)
-simplifyToOr term predicate =
+simplifyToOr predicate term =
```
ttuegel: Why reverse the argument order of `simplifyToOr` and not `simplifyInternal`?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -187,65 +189,84 @@ simplifyInternal term predicate = simplifyInternalWorker term
...
-            OrF orF -> Or.simplify <$> simplifyChildren orF
-            RewritesF rewritesF ->
-                Rewrites.simplify <$> simplifyChildren rewritesF
-            TopF topF -> Top.simplify <$> simplifyChildren topF
-            --
-            StringLiteralF stringLiteralF ->
-                return $ StringLiteral.simplify (getConst stringLiteralF)
-            VariableF variableF ->
-                return $ Variable.simplify (getConst variableF)
+        if TermLike.isSimplified termLike
```
ttuegel: I would use a guard here instead of `if`. I think it's easier to read for big blocks, and it prevents the indentation from creeping off to the right.

kore/src/Kore/Step/Simplification/TermLike.hs
ana-pantilie: I didn't change the type, I think the arguments had their names reversed and this fixes that.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -187,65 +189,84 @@ simplifyInternal term predicate = simplifyInternalWorker term
...
+           && not (Syntax.Predicate.isPredicate termLike)
```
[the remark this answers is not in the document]
ana-pantilie: I tried this in one of the previous commits and some of the integration tests were failing. We decided in one of the meetings to only check for terms now, and see what we can do for predicates in a later PR.

PR #1091, opened 2019-10-10
Refactor predicate simplifier for unification

ttuegel:
This refactors the predicate simplifier so we can use a different simplifier during unification.

~~Needs: #1085~~

---

###### Reviewer checklist


kore/src/Kore/Internal/Conditional.hs
```diff
@@ -391,3 +392,6 @@ isNormalized :: Ord variable => Conditional variable term -> Bool
 isNormalized Conditional { predicate, substitution } =
     Substitution.isNormalized substitution
     && Predicate.isFreeOf predicate (Substitution.variables substitution)
+
+hasSubstitution :: Conditional variable term -> Bool
+hasSubstitution Conditional { substitution } = Substitution.null substitution
```
virgil-serbanuta: isn't this the implementation for `doesNotHaveSubstitution`?

kore/test/Test/Kore/Step/SMT/Evaluator.hs
```diff
@@ -156,3 +172,185 @@ evaluate
...
+p = vBool (testId "p")
+q = vBool (testId "q")
+
+assertRefuted :: HasCallStack => Syntax.Predicate Variable -> Assertion
+assertRefuted prop = do
+    let expect = Just False
+    actual <- Test.runSimplifier testEnv $ SMT.Evaluator.decidePredicate prop
+    assertEqual "" expect actual
+
+unit_1 :: Assertion
```
vladciobanu: Not a huge deal, but it would be nice giving these tests names.

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -27,102 +25,57 @@ import Kore.Internal.Pattern
...
+create :: MonadSimplify simplifier => PredicateSimplifier simplifier
+create = PredicateSimplifier simplify
 
 {- | Simplify a 'Predicate'.
 
 @simplify@ applies the substitution to the predicate and simplifies the
-result. The result is re-simplified once.
+result. The result is re-simplified until it stabilizes.
+
+@simplify@ accepts any @term@, but the 'Conditional' 'term' is not simplified.
```
vladciobanu: Would renaming the type variable `term` to `a` enforce this comment?

Am I understanding it right that this function couldn't simplify its `term` even if it wanted, because it can't figure out `term ~ TermLike`?

ttuegel: That's right; `term` is `forall`-quantified, so it can't be unified with any particular type. I did it this way because it's very common that we want to simplify the predicate of some configuration and it's needlessly awkward to always take the configuration apart, simplify the predicate, and reassemble it—at every call site.

PR #1108, opened 2019-10-15
Report errors in the execution graph

### LINKED ISSUE #1098: Report errors in the execution graph
We should catch and report errors in the execution graph as a [`ProofState`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Strategies/ProofState.hs#L39). That is, one possible proof state is "encountered an error rewriting the parent configuration." This will make the REPL much more usable.

vladciobanu:
Fixes #1098 
---

###### Reviewer checklist


vladciobanu:
Wrong approach here, will re-open a different thread.

PR #1110, opened 2019-10-15
Fix markSimplified usage

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Internal/Pattern.hs
+++ b/kore/src/Kore/Internal/Pattern.hs
@@ -22,7 +22,6 @@ module Kore.Internal.Pattern
     , Kore.Internal.Pattern.freeVariables
     , Kore.Internal.Pattern.freeElementVariables
     , isSimplified
-    , markSimplified
     -- * Re-exports
     , Conditional (..)
     , Conditional.andCondition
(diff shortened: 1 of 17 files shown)


kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -253,8 +282,10 @@ makeEvaluateBoundRight variable freeSubstitution normalized = do
     TopBottom.guardAgainstBottom simplifiedPattern
     return simplifiedPattern
   where
+    simplifiedPattern :: Pattern variable
     simplifiedPattern =
-        Conditional.andCondition
+        Pattern.markSimplified
```
ttuegel: Does this eliminate the need to `markSimplified` in `quantifyPattern` below? How do we know all parts of the pattern are really simplified?

virgil-serbanuta: You're right, we don't, I'll fix that.

kore/src/Kore/Internal/Pattern.hs
```diff
@@ -84,6 +95,11 @@ isSimplified :: Pattern variable -> Bool
 isSimplified (splitTerm -> (t, p)) =
     TermLike.isSimplified t && Predicate.isSimplified p
 
+markSimplified :: Pattern variable -> Pattern variable
```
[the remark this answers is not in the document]
virgil-serbanuta: Removed

kore/src/Kore/Step/Simplification/Exists.hs
virgil-serbanuta: Fixed.

PR #1111, opened 2019-10-15
Report REPL errors

vladciobanu:
I started off with the `Either` idea, and after some amount of work I realised this might get us pretty much all the way there (the only downside being it just stops when it hits the first error, whereas if we stored errors in `Either Error ProofState` we could accumulate them in the graph). Is that worth the more complex approach?

I could use some feedback @ttuegel 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Repl.hs
+++ b/kore/src/Kore/Repl.hs
@@ -25,6 +25,7 @@ import Control.Monad
 import Control.Monad.Catch
     ( MonadCatch
     , catch
+    , catchAll
     )
 import Control.Monad.IO.Class
     ( MonadIO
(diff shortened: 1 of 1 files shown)


ttuegel:
> the only downside being it just stops when it hits the first error, whereas if we stored errors in `Either Error ProofState` we could accumulate them in the graph

If I understand correctly, this means that the user cannot continue down another branch of execution when they encounter an error on any branch, is that right? I suppose that's better than the status quo: at least the user can inspect the states leading up to the error, but part of what we had in mind here is to allow users to work around a broken branch (while we fix it, hopefully), to continue other branches of the proof.

vladciobanu:
No, what happens is we catch all exceptions (then print and discard them) as they happen during each step of execution. What this means is the error is not persisted in the graph, but rather just displayed and discarded. However, the graph is restored to the value it had before the exception occurred, so the user can either continue on another branch, or indeed reproduce the exception as many times as he wants to.

PR #1113, opened 2019-10-16
Kore.Step.Simplification.SubstitutionSimplifier

ttuegel:
- Add `Kore.Step.Simplification.SubstitutionSimplifier`.
- Change the return type of `Kore.Unification.UnifierImpl.deduplicateSubstitution`. Previously, the post-condition that the substitution is deduplicated was not checked. Now, the type is changed to make such a check unnecessary.
- `Kore.Internal.Predicate.fromNormalization`: Convert a `Kore.Unification.Substitution.Normalization` (the result of normalizing a substitution) into a `Predicate` by move the denormalized part into the predicate while preserving the normalized part as a `Substitution`.
- Several new tests.

Note: This still does not change the behavior of the predicate simplifier. The next pull request in this sequence should be the grand finale.

See also: #1069.

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -0,0 +1,147 @@
...
+simplification =
+    SubstitutionSimplifier worker
+  where
+    worker substitution = maybeUnwrapSubstitution $ do
+        deduplicated <-
+            -- TODO (thomas.tuegel): If substitution de-duplication fails with a
+            -- unification error, this will still discard the entire
+            -- substitution into the predicate. Fortunately, that seems to be
+            -- rare enough to discount for now.
+            Unifier.deduplicateSubstitution substitution
```
virgil-serbanuta: So, maybe this does not change existing behaviour (I didn't check that), but I have to ask this:

In simplification we use matching. I thought that if we were simplifying the matching substitution, we were supposed to also use matching, or we shouldn't simplify it at all. However, here we're calling `deduplicateSubstitution`, which, I think, is calling `simplifyAnds`, which is using unification. So, did I make a mistake in my assumptions?

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -0,0 +1,147 @@
...
+            & Unifier.maybeUnifierT
+        OrPredicate.fromPredicates <$> traverse normalize1 deduplicated
+      where
+        maybeUnwrapSubstitution =
+            let unwrapped =
+                    OrPredicate.fromPredicate
+                    . Predicate.fromPredicate
+                    . Syntax.Predicate.markSimplified
+                    -- TODO (thomas.tuegel): Promoting the entire substitution
```
virgil-serbanuta: I thought that promoting cyclic stuff happens in `Predicate.fromNormalization`, not here, am I missing something? Wouldn't `unwrapped` only apply to errors from deduplicateSubstitution?

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
ttuegel: > I thought that if we were simplifying the matching substitution, we were supposed to also use matching, or we shouldn't simplify it at all.

We never simplify the substitution during matching. Matching constructs the substitution so that it is always normalized; if it's ever non-normalized, matching fails. This would be used to during predicate simplification after applying (rewrite or equality) rules, after the rule is fully applied, or on the remainders. Unification is being used to simplify `\and` patterns here, so I think it's appropriate.

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
ttuegel: Yes, promoting cycles happens in `fromNormalization`, this only applies if there is a unification error (`UnsupportedPatterns`) while simplifying `\and`s. I have only seen that happen in contrived tests. I think it is unlikely to happen in practice; if the patterns are unsupported for unification, how did they get into a substitution?

virgil-serbanuta: Ok, thanks for the answer. That also clarifies that this TODO and the one above are talking about the same thing.

Would it make sense then to wrap only the `deduplicateSubstitution` call in `maybeUnwrapSubstitution`, and not the entire `do`?

kore/src/Kore/Unification/UnifierImpl.hs
```diff
@@ -90,103 +90,55 @@ simplifyAnds (NonEmpty.sort -> patterns) = do
...
-            Map.toAscList
-            . Foldable.foldl' insertSubstitution Map.empty
-            . map sortRenamedVariable
-            $ Substitution.unwrap substitution
+        simplified <- collectConditions <$> traverse simplifyAnds substitutions
+        let -- Substitutions de-duplicated by simplification.
+            substitutions' = toMultiMap $ Conditional.term simplified
+            -- New conditions produced by simplification.
+            Conditional { predicate = predicate' } = simplified
+            predicate'' = Syntax.Predicate.makeAndPredicate predicate predicate'
```
ana-pantilie: This line is kind of long (80 characters). I think it can be easily split up. The same for the line at 119.

kore/test/Test/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -0,0 +1,231 @@
...
+            [ Normalization [(z, a)] [] ]
+        ]
+    , testGroup "set-variable-only cycle"
+        [ test "length 1, alone"
+            [(xs, mkVar xs)]
+            [ mempty ]
+        -- UnificationError: UnsupportedPatterns
+        -- , test "length 1, beside related substitution"
+        --     [(xs, mkVar xs), (ys, mkVar xs)]
+        --     [([(ys, mkVar xs)], [])]
```
ana-pantilie: Should this remain commented?

kore/test/Test/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -0,0 +1,231 @@
...
+            [ Normalization [(z, c)] [(x, f (mkVar x)), (y, g (mkVar y))] ]
+        ]
+    , testGroup "set variable simplifiable cycle"
+        [ test "length 1, alone"
+            [(xs, f (mkVar xs))]
+            [ Normalization [] [(xs, f (mkVar xs))] ]
+        -- UnificationError: UnsupportedPatterns
+        -- , test "length 1, beside related substitution"
+        --     [(xs, f (mkVar xs)), (ys, mkVar xs)]
+        --     [([(ys, mkVar xs)], [(xs, f (mkVar xs))])]
```
ana-pantilie: Same as above.

kore/test/Test/Kore/Unification/SubstitutionNormalization.hs
```diff
@@ -125,6 +125,10 @@ test_normalize =
             [(x, mkAnd (mkVar y) a), (y, mkAnd (mkVar x) b)]
             []
             [(x, mkAnd (mkVar y) a), (y, mkAnd (mkVar x) b)]
+        , test "two cycles"
+            [(x, f (mkVar x)), (y, g (mkVar y)), (z, c)]
+            [(z, c)]
+            [(x, f (mkVar x)), (y, g (mkVar y))]
```
ana-pantilie: It's out of the scope of this PR, but I think the tests from this file are a little harder to read than our other tests where we make it explicit for each test which are the expected and actual values. For the tests here we also have 2 expected values and it's harder to keep track of what is what. This is just an observation, though.

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
ttuegel: > Would it make sense then to wrap only the `deduplicateSubstitution` call in `maybeUnwrapSubstitution`, and not the entire `do`?

It would, but in my very next pull request, I'm actually going to remove it here entirely.

kore/test/Test/Kore/Step/Simplification/SubstitutionSimplifier.hs
ttuegel: Yes; this test has an error due to a missing unification case. I do not know if we will ever add that unification case, but if we do then the test should be uncommented. I will add a TODO item for that.

PR #1114, opened 2019-10-16
Reject predicates which contain a contradiction

### LINKED ISSUE #1107: Reject predicates that contain contradiction
We should reject unification results that contain both `p` and `not(p)`. This is simply `unsat`.

vladciobanu:
Fixes #1107

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/And.hs
+++ b/kore/src/Kore/Step/Simplification/And.hs
@@ -190,8 +190,11 @@ makeEvaluateNonBool
         merged = Conditional.andCondition terms initialConditions
     normalized <- Substitution.normalize merged
     return
-        (applyAndIdempotenceAndFindContradictions <$> normalized)
-            { predicate =
+        normalized
+            { term =
+                applyAndIdempotenceAndFindContradictions
+                    (Conditional.term normalized)
+            , predicate =
                 applyAndIdempotenceAndFindContradictions
                     <$> Conditional.predicate normalized
             }
--- a/kore/src/Kore/Step/Simplification/And.hs
+++ b/kore/src/Kore/Step/Simplification/And.hs
@@ -189,50 +189,66 @@ makeEvaluateNonBool
         initialConditions = firstCondition <> secondCondition
         merged = Conditional.andCondition terms initialConditions
     normalized <- Substitution.normalize merged
-    return (applyAndIdempotence <$> normalized)
-        { predicate = applyAndIdempotence <$> Conditional.predicate normalized }
+    return
+        (applyAndIdempotenceAndFindContradictions <$> normalized)
+            { predicate =
+                applyAndIdempotenceAndFindContradictions
+                    <$> Conditional.predicate normalized
+            }
 
-applyAndIdempotence
+applyAndIdempotenceAndFindContradictions
     :: InternalVariable variable
     => TermLike variable
     -> TermLike variable
-applyAndIdempotence patt =
-    if contradictions
-        then mkBottom (termLikeSort patt)
-        else
+applyAndIdempotenceAndFindContradictions patt =
+    if noContradictions
+        then
             foldl1' mkAndSimplified
                 . Set.toList
-                . uncurry Set.union
-                . fmap (Set.mapMonotonic mkNot)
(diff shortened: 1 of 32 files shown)


kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -176,19 +189,66 @@ makeEvaluateNonBool
...
     -> TermLike variable
-applyAndIdempotence patt =
-    foldl1' mkAndSimplified (nub (children patt))
+applyAndIdempotenceAndFindContradictions patt =
+    if noContradictions
+        then
+            foldl1' mkAndSimplified
+                . Set.toList
+                . Set.union terms
+                . Set.mapMonotonic mkNot
```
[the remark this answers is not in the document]
ttuegel: On second thought, I think I would prefer if you left the `Not`s on top of the `negatedTerms` and checked for `Set.disjoint (Set.map mkNot terms) negatedTerms` below. The reason is memory allocation: unwrapping and re-wrapping the `negatedTerms` basically re-allocates them, discarding the original and keeping the copy. Keeping the `Not`s and temporarily wrapping the `terms` allows the temporaries to be garbage collected immediately. I might be overthinking this, but `And` simplification is used pretty heavily.

kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -176,19 +189,66 @@ makeEvaluateNonBool
...
+                $ negatedTerms
+        else mkBottom (termLikeSort patt)
+
   where
-    children (And_ _ p1 p2) = children p1 ++ children p2
-    children p = [p]
+    (terms, negatedTerms) = splitIntoTermsAndNegations patt
+    nullPredicate = ((||) `on` Set.null) terms negatedTerms
+    noContradictions =
+        Set.null (Set.intersection terms negatedTerms) || nullPredicate
```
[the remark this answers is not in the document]
vladciobanu: Whoops. This was initially `contradictions`, so I needed to make sure they aren't also empty. You are right!

kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -176,19 +189,50 @@ makeEvaluateNonBool
         initialConditions = firstCondition <> secondCondition
         merged = Conditional.andCondition terms initialConditions
     normalized <- Substitution.normalize merged
-    return (applyAndIdempotence <$> normalized)
-        { predicate = applyAndIdempotence <$> Conditional.predicate normalized }
-
-applyAndIdempotence
+    return
+        (applyAndIdempotenceAndFindContradictions <$> normalized)
```
virgil-serbanuta: So, I guess that normalized is a `Pattern`, and `applyAndIdempotenceAndFindContradictions` acts on its term. If so, I think that it would be more readable to update the `term` between curly braces below, as you do with the `predicate`, since you would use the same syntax for updating both fields.

I may be biased though, I never liked the functor instance for `Conditional` that much, so do whatever makes sense.

kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -176,19 +189,50 @@ makeEvaluateNonBool
...
+            { predicate =
+                applyAndIdempotenceAndFindContradictions
+                    <$> Conditional.predicate normalized
+            }
+
+applyAndIdempotenceAndFindContradictions
```
virgil-serbanuta: Could you add tests for the "find contradictions" part?

kore/src/Kore/Step/Simplification/And.hs
vladciobanu: The result would be:
```
    return
        normalized
            { term =
                applyAndIdempotenceAndFindContradictions
                    (Conditional.term normalized)
            , predicate =
                applyAndIdempotenceAndFindContradictions
                    <$> Conditional.predicate normalized
            }
```

I think I like this version better, thanks.

kore/src/Kore/Step/Simplification/And.hs
vladciobanu: Done!

PR #1116, opened 2019-10-17
Fix forall predicates

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Internal/Pattern.hs
+++ b/kore/src/Kore/Internal/Pattern.hs
@@ -22,6 +22,7 @@ module Kore.Internal.Pattern
     , Kore.Internal.Pattern.freeVariables
     , Kore.Internal.Pattern.freeElementVariables
     , isSimplified
+    , markSimplified
     -- * Re-exports
     , Conditional (..)
     , Conditional.andCondition
@@ -93,6 +94,11 @@ fromPredicateSorted sort = (<$) (mkTop sort)
 isSimplified :: Pattern variable -> Bool
 isSimplified (splitTerm -> (t, p)) =
     TermLike.isSimplified t && Predicate.isSimplified p
+
+markSimplified :: Pattern variable -> Pattern variable
+markSimplified (splitTerm -> (term, predicate)) =
+    (TermLike.markSimplified term)
+    `Conditional.withCondition` (Predicate.markSimplified predicate)
 
 freeVariables
     :: Ord variable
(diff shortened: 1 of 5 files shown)


kore/src/Kore/Internal/Pattern.hs
```diff
@@ -94,6 +95,11 @@ isSimplified :: Pattern variable -> Bool
 isSimplified (splitTerm -> (t, p)) =
     TermLike.isSimplified t && Predicate.isSimplified p
 
+markSimplified :: Pattern variable -> Pattern variable
```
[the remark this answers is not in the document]
virgil-serbanuta: Oops, yes.

kore/src/Kore/Step/Simplification/Forall.hs
```diff
@@ -98,8 +118,25 @@ makeEvaluate
 makeEvaluate variable patt
   | Pattern.isTop patt    = Pattern.top
   | Pattern.isBottom patt = Pattern.bottom
+  | not variableInTerm && not variableInPredicate = patt
+  | predicateIsBoolean =
+    TermLike.markSimplified (mkForall variable term)
```
ana-pantilie: Should we check that the `term` is simplified here? Or do we only call this when we're sure it was simplified before.

kore/src/Kore/Step/Simplification/Forall.hs
```diff
@@ -98,8 +118,25 @@ makeEvaluate
...
+    `Conditional.withCondition` predicate
+  | termIsBoolean =
+    term
+    `Conditional.withCondition` Predicate.markSimplified
+        (Predicate.fromPredicate
+            (makeForallPredicate variable (Predicate.toPredicate predicate))
```
ana-pantilie: Same as above for the `predicate`.

kore/src/Kore/Step/Simplification/Forall.hs
virgil-serbanuta: I ask @ttuegel the same question in another PR :)

I think we just assume that the input to all these functions (simplify, makeEvaluate, and so on) is already simplified.

kore/src/Kore/Step/Simplification/Forall.hs
virgil-serbanuta: Same answer

virgil-serbanuta:
> I left some questions, also I think that the docs in `kore/src/Kore/Step/Simplification/Forall.hs` from line 65 to 78 might need to be updated.

Thanks! Fixed.

kore/src/Kore/Step/Simplification/Forall.hs
ana-pantilie: Ah, right. Found the conversation.

PR #1128, opened 2019-10-21
Fix term sorts in predicates

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/test/Test/Kore/Step/Simplification/Not.hs
+++ b/kore/test/Test/Kore/Step/Simplification/Not.hs
@@ -3,12 +3,20 @@ module Test.Kore.Step.Simplification.Not
     ) where
 
 import Test.Tasty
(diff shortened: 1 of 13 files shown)


kore/test/Test/Kore/Step/Simplification/Not.hs
```diff
@@ -35,13 +43,17 @@ import Kore.Variables.UnifiedVariable
 
 import qualified Test.Kore.Step.MockSymbols as Mock
 import Test.Kore.Step.Simplification
+import Test.Tasty.HUnit.Ext
```
ttuegel: `Test.Tasty.HUnit.Ext` re-exports `Test.Tasty.HUnit`, with the intent that we would no longer directly import the latter.

Issue #1148, opened 2019-10-23
Term not fully simplified (2)


virgil-serbanuta:
For the following term we don't properly set the "simplified" bit. Creating a test in `kore/test/Test/Kore/Step/Simplification/Integration.hs` will almost surely reproduce the issue.

```
\and{listSort{}}(
    concatList{}(
        /* evaluated: */ /* evaluated: */ /* evaluated: */ \top{listSort{}}(),
        /* evaluated: */ /* evaluated: */ /* evaluated: */ \top{listSort{}}()
    ),
    \and{listSort{}}(
        \not{listSort{}}(\not{listSort{}}(unitList{}())),
        \mu{}(
            s:listSort{},
            \or{listSort{}}(
                \implies{listSort{}}(s:listSort{}, \top{listSort{}}()),
                /* evaluated: */ \bottom{listSort{}}()
            )
        )
    )
)
```

virgil-serbanuta:
This is the failing example:
```
\and{listSort{}}(
    concatList{}(
        /* evaluated: */
        \implies{listSort{}}(
            \implies{listSort{}}(\bottom{listSort{}}(), \top{listSort{}}()),
            \in{testSort0{}, listSort{}}(cfSort0{}(), cgSort0{}())
        ),
        \implies{listSort{}}(
            \and{listSort{}}(
                \mu{}(m:listSort{}, \bottom{listSort{}}()),
                \bottom{listSort{}}()
            ),
            \implies{listSort{}}(
                unitList{}(),
                \nu{}(Ue:listSort{}, unitList{}())
            )
        )
    ),
    unitList{}()
)
```

PR #1152, opened 2019-10-24
Fix the simplified bit for list unification failures

virgil-serbanuta:
Fixes https://github.com/kframework/kore/issues/1148

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Builtin/Builtin.hs
+++ b/kore/src/Kore/Builtin/Builtin.hs
@@ -147,15 +147,22 @@ import Kore.Internal.Pattern
 import Kore.Internal.Pattern as Pattern
     ( fromTermLike
     , top
+    , withCondition
+    )
+import Kore.Internal.Predicate as Predicate
+    ( topTODO
     )
 import Kore.Internal.TermLike as TermLike
 import Kore.Predicate.Predicate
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -983,31 +996,30 @@ getAttemptedAxiom attempt =
...
-unifyEqualsUnsolved SimplificationType.And a b =
-    let
-        unified = mkAnd a b
-        predicate = makeCeilPredicate unified
-    in
-        return (pure unified) { predicate }
+    -> unifier (Pattern variable)
+unifyEqualsUnsolved SimplificationType.And a b = do
+    let unified = TermLike.markSimplified $ mkAnd a b
+    orPredicate <- Ceil.makeEvaluateTerm Predicate.topTODO unified
```
vladciobanu: Weren't we done with `topTODO`? Is this a new one? If memory serves, this was only supposed to be used for a limited amount of time and removed later.

ttuegel: It is intended to be temporary, but I think we never decided what to do about passing predicates in during unification. So, I guess this is alright.

PR #1157, opened 2019-10-24
BYTES hooks

### LINKED ISSUE #936: Missing hooks: BYTES
The KEVM semantics have been updated to use byte arrays for program storage. This improves concrete execution performance because byte arrays have constant-time access, but symbolic execution performance is severely degraded because byte arrays are implemented as lists with linear-time access. (Previously, maps with log-time access were used for storage in concrete and symbolic backends.)

We will not need any of the functions for converting between `Bytes` and `Int`, or the `Endianness` or `Signedness` sorts.

See also: [`domains.k`](https://github.com/kframework/k/blob/9bdeac205c4f4d3cb938e539fa25826123305869/k-distribution/include/builtin/domains.k#L662-L687)

vladciobanu:
Fixes #936 
---

###### Reviewer checklist


kore/src/Kore/Domain/Builtin.hs
```diff
@@ -657,6 +704,7 @@ data Builtin key child
     | BuiltinInt !InternalInt
     | BuiltinBool !InternalBool
     | BuiltinString !InternalString
+    | BuiltinBytes !InternalBytes
```
ttuegel: Instead of adding this to `Builtin`, could we add an `InternalBytesF` constructor to `TermLikeF`? The `Builtin` type is a historical accident, and it's difficult to work with because it mashes many different things together.

kore/Makefile
```diff
@@ -41,9 +41,8 @@ ghcid-repl:
 	$(stack) exec -- ghcid -c "stack ghci $(package) --test --ghci-options='-fobject-code -fno-warn-unused-do-bind' --main-is $(package):kore-repl"
 
 
-# TODO(Vladimir): remove 'hyperlink-source' when we upgrade from lts-12.10 (8.4.3)
 hoogle-gen:
-	$(stack) haddock; $(stack) hoogle generate --rebuild --no-haddock-hyperlink-source  -- --local
+	$(stack) haddock; $(stack) hoogle generate --rebuild -- --local
```
ttuegel: Could you open a separate pull request for this?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -2209,6 +2243,11 @@ pattern App_ applicationSymbolOrAlias applicationChildren <-
 pattern Bottom_ bottomSort <-
     (Recursive.project -> _ :< BottomF Bottom { bottomSort })
 
+pattern Bytes_ bytesSort string2BytesSymbol bytesValue <-
```
ttuegel: I would prefer to use `InternalBytes` everywhere.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -277,6 +283,16 @@ simplifyInternal term predicate =
                 --
                 StringLiteralF stringLiteralF ->
                     return $ StringLiteral.simplify (getConst stringLiteralF)
+                InternalBytesF (Const
+                    Bytes.InternalBytes { bytesValue, string2BytesSymbol }) ->
+                    Application.simplify predicate =<< simplifyChildren
```
ttuegel: This should get handed off to an internal simplifier instead.

kore/src/Kore/Syntax/Bytes.hs
```diff
@@ -0,0 +1,95 @@
+{-|
+Copyright   : (c) Runtime Verification, 2019
+License     : NCSA
+-}
+
+module Kore.Syntax.Bytes
```
ttuegel: This should be `Kore.Internal.InternalBytes` because it is an internal representation and we usually have the module name match the type name.

kore/src/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,371 @@
...
+    evalBytes2String0 resultSort = \arguments ->
+        Builtin.getAttemptedAxiom $ do
+            let _bytes =
+                    case arguments of
+                        [_bytes] -> _bytes
+                        _ -> Builtin.wrongArity bytes2StringKey
+            (_, bytestring) <- expectBuiltinBytes bytes2StringKey _bytes
+            Builtin.appliedFunction
+                . String.asPattern resultSort
+                . E.decodeUtf8
```
ttuegel: We may need to ask @dwightguth, but I think that a raw 8-bit encoding is expected, not UTF-8.

kore/src/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,371 @@
...
+                        , applicationSymbolOrAlias
+                        } -> (applicationSymbolOrAlias, _string)
+                    _ -> Builtin.wrongArity string2BytesKey
+            Attribute.Pattern { patternSort } = attrs
+
+        _string <- String.expectBuiltinString string2BytesKey _string
+
+        Builtin.appliedFunction
+            . asPattern patternSort symbol
+            . E.encodeUtf8
```
ttuegel: See above.

kore/src/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,371 @@
...
+    evalUpdate0 resultSort = \arguments ->
+        Builtin.getAttemptedAxiom $ do
+            let (_bytes, _index, _value) =
+                    case arguments of
+                        [_bytes, _index, _value] -> (_bytes, _index, _value)
+                        _ -> Builtin.wrongArity updateKey
+            (symbol, _bytes) <- expectBuiltinBytes updateKey _bytes
+            _index <- fromInteger <$> Int.expectBuiltinInt updateKey _index
+            _value <- fromInteger <$> Int.expectBuiltinInt updateKey _value
+            if _index < 0 || _index > (BS.length _bytes - 1)
```
ttuegel: For K Lists, negative indices are measured from the end of the list. Are we sure that's not also the case for byte arrays?

kore/test/Test/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,522 @@
...
+import Test.SMT
+
+genString :: Gen Text
+genString = Gen.text (Range.linear 0 256) Gen.alphaNum
+
+genString' :: Int -> Gen Text
+genString' i = Gen.text (Range.linear 0 i) Gen.alphaNum
+
+test_bytes2string :: TestTree
+test_bytes2string =
```
ttuegel: I don't think that `test_bytes2string` and `test_string2bytes` are useful tests individually, especially because they are tested to be inverses of each other below.

kore/test/Test/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,522 @@
...
+        let expect = asPattern $ E.encodeUtf8 str
+        actual <- evaluateT
+            $ mkApplySymbol
+                string2bytesBytesSymbol
+                (Test.String.asInternal <$> [str])
+        (===) expect actual
+
+test_update :: [TestTree]
+test_update =
+    [ testPropertyWithSolver "∀ b v. update b 0 v = v" $ do
```
ttuegel:
```suggestion
    [ testPropertyWithSolver "∀ v. update v 0 v = v" $ do
```

kore/src/Kore/Builtin/Bytes.hs
vladciobanu: I was looking at the `K` implementation, which reads:
```
  syntax Bytes ::= Bytes "[" Int "<-" Int "]" [function]
  rule BS [ N <- M ] => substrBytes(BS, 0, N) +Bytes M : substrBytes(BS, N +Int 1, lengthBytes(BS))

  syntax Bytes ::= substrBytes(Bytes, Int, Int) [function]
  rule substrBytes(BS, 0, 0) => nilBytes
  rule substrBytes(B : BS, N, M) => substrBytes(BS, N -Int 1, M -Int 1) requires N >Int 0
  rule substrBytes(B : BS, 0, M) => B : substrBytes(BS, 0, M -Int 1) requires M >Int 0
```
Unless I'm reading it wrong, it requires positive indices.

kore/src/Kore/Step/Simplification/TermLike.hs
vladciobanu: Does this mean just moving the code and calling `return $ Bytes.simplify`?

kore/src/Kore/Internal/TermLike.hs
vladciobanu: Does that include module names like `Kore/Builtin/InternalBytes` instead of `Bytes`?

ttuegel: Yes. I think it's better to have the module name match the type name.

kore/src/Kore/Step/Simplification/TermLike.hs
ttuegel: Well, I don't think we need to do anything here. The value is already fully evaluated. We don't call the application simplifier for any other builtins. I think we should just return the value we were given.

kore/src/Kore/Builtin/Bytes.hs
ttuegel: Sounds good!

kore/src/Kore/Builtin/Bytes.hs
```diff
@@ -0,0 +1,371 @@
...
+    , symbolVerifiers
+    , builtinFunctions
+    , asInternal
+    , internalize
+    , asTermLike
+    , asPattern
+      -- * Keys
+    , bytes2StringKey
+    , string2BytesKey
+    , updateKey
```
virgil-serbanuta: Could you add the new builtins to `docs/hooks.md`?

kore/src/Kore/Builtin/Bytes/Bytes.hs
```diff
@@ -0,0 +1,169 @@
...
+      literal = E.decodeUtf8 bytesValue
+
+internalize
+    :: InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+internalize =
+    \case
+        TermLike.App_ symbol args
+          | isSymbolString2Bytes symbol, [TermLike.StringLiteral_ str] <- args ->
```
virgil-serbanuta: >80 chars

vladciobanu:
This should be ready for re-review. I am least sure about whether this is right: https://github.com/kframework/kore/pull/1157/files#diff-c3eb9f1a5b644763fbe9bcdf5d2fe6ddR19

Issue #1071, opened 2019-10-07
Remove trivial branches when simplifying claims at startup


yzhang90:
I'm debugging the `https://github.com/kframework/wasm-semantics/blob/memory-spec/tests/proofs/memory-spec.k` under memory-spec branch in wasm-semantics.
I found that in `0449856d84f82eea8df31f8510ed5ed1066cdb36`, after simplifying the LHS, it split the original claim into 2. One of them is
```
  ( #Ceil ( ADDR modInt 4294967296 )
    #And
      {
        EA
      #Equals
        ADDR modInt 4294967296
      }
  #Implies
    {
      true
    #Equals
      EA +Int #numBytes ( ITYPE ) <=Int SIZE *Int 65536 andBool #isByteMap ( ByteMap <| BM0 |> )
    } )
#And
  #Ceil ( ADDR modInt 4294967296 )
#And
  #Not ( #Ceil ( ADDR modInt 4294967296 )
  #And
    #Not ( #Ceil ( ADDR modInt 4294967296 )
    #And
      {
        EA
      #Equals
        ADDR modInt 4294967296
      } ) )
#And
  #Not ( #Ceil ( ADDR modInt 4294967296 )
  #And
    {
      EA
    #Equals
      ADDR modInt 4294967296
    } )
```

- [x] `#Ceil (... modInt ...)` is simplifiable.
- [ ] We should simplify the predicate to remove trivial branches of the proof.

yzhang90:
We saw similar not(a) #and not(not(a)) in #1060

ttuegel:
Duplicate of #1103.

ana-pantilie:
I think 1. is fixed on the latest `kore` master, could you please double check?

ttuegel:
Yes, the `Ceil(... modInt ...)` problem is definitely fixed.

PR #1158, opened 2019-10-25
Simplify predicate with SMT at start-up

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1071

---

###### Reviewer checklist


kore/src/Kore/Exec.hs
```diff
@@ -575,11 +575,11 @@ initializeProver definitionModule specModule within =
             specClaims =
                 map (Bifunctor.second fromMaybeChanged) changedSpecClaims
 
+        assertSomeClaims specClaims
```
ttuegel: Could you add a small comment why this check comes first? It seems like the sort of thing which is obvious now, but will quickly be forgotten.

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -79,10 +83,12 @@ simplifyRuleLhs rule@(RulePattern _ _ _ _ _ _) = do
         Conditional.withCondition
             left
             $ Predicate.fromPredicate
-                ( makeAndPredicate
-                    requires
-                    (makeCeilPredicate left)
-                )
+                lhsPredicate
+    lhsPredicate =
```
ttuegel: Could we move `lhsPredicate` and `definedLhs` into a `let` block above? I think this bit would be easier to understand if they were closer to where they are used.

PR #1162, opened 2019-10-25
Generalized Exec.prove to accept any Claim

traiansf:
Preliminary refactoring towards allowing all-path proofs along one-path ones.

---

###### Reviewer checklist


kore/src/Kore/Exec.hs
```diff
@@ -282,20 +284,31 @@ search verifiedModule strategy termLike searchPattern searchConfig =
...
     -- ^ The main module
     -> VerifiedModule StepperAttributes Attribute.Axiom
     -- ^ The spec module
+    -> (VerifiedModule StepperAttributes Attribute.Axiom
+        -> [(Attribute.Axiom, claim)])
+    -- ^ claims extractor
     -> smt (Either (TermLike Variable) ())
-prove limit definitionModule specModule =
-    evalProver definitionModule specModule
+prove limit definitionModule specModule claimExtractor =
```
ttuegel: Could we push the `claimExtractor` out into the `Claim` class?

kore/src/Kore/Step/Rule/Expand.hs
```diff
@@ -79,12 +82,14 @@ import Kore.Variables.UnifiedVariable
...
-expandOnePathSingleConstructors metadataTools (OnePathRule rule) =
-    OnePathRule (expandSingleConstructors metadataTools rule)
+expandClaimSingleConstructors
+    :: Coercible (RulePattern Variable) claim
+    => Coercible claim (RulePattern Variable)
+    => SmtMetadataTools attributes
+    -> claim
+    -> claim
+expandClaimSingleConstructors metadataTools claim =
+    coerce (expandSingleConstructors metadataTools (coerce claim))
```
ttuegel: When we have a function applied over a coercion like this, it makes me think this would be better in the class. Could we lift this to the class? Or going even further, make this a strategy step, so that we expand any single-constructor variables at the start of each proof step? Or going perhaps to far: make this a step in the variable simplifier? :sunglasses: 

That last is probably too far because of the quantification problem, but I think we could generalize this a bit.

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -51,15 +55,17 @@ import Kore.Step.Simplification.Simplify
...
+        , Coercible rule (RulePattern variable)
         )
-    => OnePathRule variable
-    -> simplifier (MultiAnd (OnePathRule variable))
-simplifyOnePathRuleLhs (OnePathRule rule) = do
-    simplifiedList <- simplifyRuleLhs rule
-    return (fmap OnePathRule simplifiedList)
+    => rule
+    -> simplifier (MultiAnd rule)
+simplifyGoalRuleLhs rule = do
```
ttuegel: This is another one I would probably prefer to see lifted into the class.

kore/src/Kore/Exec.hs
traiansf: Thanks! I did that (kind of). However, I now need a `Proxy` for the `prove` and `proveWithRepl` functions as they have no `claim`-related arguments anymore. Vlad suggested we `AllowAmbiguousTypes` for the `Exec` module and use `TypeApplications`  instead of `Proxy` at call-time.  What do you think?

ttuegel: > Vlad suggested we AllowAmbiguousTypes for the Exec module and use TypeApplications instead of Proxy at call-time. What do you think?

`AllowAmbiguousTypes` is a bad idea because it _requires_ using `TypeApplications`, but it doesn't tell you _where_ they are required. Using `Proxy` documents exactly where the type annotation is required.

Why are `prove` and `proveWithRepl` polymorphic now?

vladciobanu: I would argue that `prove @claim ...` is more readable because it's explicit in what happens: we're passing a type to resolve ambiguity. I also suggested we make sure `claim` is the first type argument to make it easy (rather than `prove @_ @claim`).

I find it odd that `prove` does not take any `claim` itself (which would fix all of these problems). Do we have some types which hide `claims` inside existentials or such?

traiansf: it's because we've made the `claimExtractor` part of the `Claim` so now both the extraction and the processing is done in the methods below.

vladciobanu: Ah so previously we were essentially picking the `claim` type by passing in a concrete `claimExtractor`, but now that we moved that to a type class, the callers of `prove` are no longer picking a `claim` instance so it becomes ambiguous. Is this correct?

kore/src/Kore/Step/Rule/Simplify.hs
traiansf: I lifted both the above, but each to its own class and included them both into the `Claim` type.  Please take another look.

kore/app/exec/Main.hs
```diff
@@ -439,7 +449,7 @@ koreProve execOptions proveOptions = do
                     Bounded.Failed final -> return (failure final)
             else
                 either failure (const success)
-                <$> prove stepLimit mainModule specModule
+                <$> prove stepLimit mainModule specModule (Proxy :: Proxy (OnePathRule Variable))
```
ttuegel: Although I don't like requiring `TypeApplication` in the interface of a function, I think it's good to make using `Proxy` less verbose:

```suggestion
                <$> prove stepLimit mainModule specModule (Proxy @(OnePathRule Variable))
```

kore/src/Kore/Step/Rule.hs
```diff
@@ -555,15 +555,30 @@ wAF sort = Alias
     , aliasRight = mkTop sort
     }
 
+pathRuleToPattern
```
ttuegel: Because we renamed the type above:

```suggestion
reachabilityRuleToPattern
```

Issue #1153, opened 2019-10-24
Incomplete simplification error


virgil-serbanuta:
For the term below we get an "Incomplete simplification! Expected all patterns to be fully simplified." error. Creating a test in kore/test/Test/Kore/Step/Simplification/Integration.hs will almost surely reproduce the issue.

```
\nu{}(
    gt:#String{},
    \equals{boolSort{}, #String{}}(
        \in{testSort1{}, boolSort{}}(
            \implies{testSort1{}}(
                \rewrites{testSort1{}}(\top{testSort1{}}(), cfSort1{}()),
                \equals{otherSort{}, testSort1{}}(
                    \top{otherSort{}}(),
                    \top{otherSort{}}()
                )
            ),
            \nu{}(g:testSort1{}, \or{testSort1{}}(aSort1{}(), g:testSort1{}))
        ),
        \top{boolSort{}}()
    )
)
```

virgil-serbanuta:
This bug works something like this:

We have an `equals (OrPattern [predicate1 predicate2]) top`. `OrPattern [predicate1 predicate2]` gets transformed to a pattern with a top term and the obvious `or` as the predicate (this pattern will be called `or-predicate` in the following). Since the initial `equals` becomes an `equals` between predicates, it is simplified as an `iff or-predicate top`. The `iff` sees that the second term is `top`, so it simplifies to `or-predicate`.

The problem is that the  `(or predicate1 predicate2) -> or-predicate` transformation does not mark the `or-pattern` as being simplified (there are some arguments that this is actually reasonable, we could start a discussion), while the `iff` code simply returns the (non-simplified) argument.

PR #1163, opened 2019-10-25
Fix equals simplification

virgil-serbanuta:
Fixes https://github.com/kframework/kore/issues/1153

See the issue for details about what this attempts to solve.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/Equals.hs
+++ b/kore/src/Kore/Step/Simplification/Equals.hs
@@ -59,6 +59,7 @@ import {-# SOURCE #-} qualified Kore.Step.Simplification.Ceil as Ceil
     )
 import qualified Kore.Step.Simplification.Iff as Iff
     ( makeEvaluate
+    , simplifyEvaluated
     )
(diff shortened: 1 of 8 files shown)


kore/src/Kore/Step/Simplification/Equals.hs
```diff
@@ -181,18 +182,21 @@ simplifyEvaluated predicate first second
...
                 makeEvaluateFunctionalOr predicate firstP secondPatterns
         (_, [secondP])
             | isFunctionConditional secondP ->
                 makeEvaluateFunctionalOr predicate secondP firstPatterns
-        _ ->
-            makeEvaluate
-                (OrPattern.toPattern first)
-                (OrPattern.toPattern second)
-                predicate
+        _ -> if OrPattern.isPredicate first && OrPattern.isPredicate second
```
ttuegel: I think a guard here would be more consistent with the cases above than using `if`, but I don't feel strongly.

virgil-serbanuta: Fixed.

Issue #1077, opened 2019-10-09
Check that (concrete) map keys are distinct.


virgil-serbanuta:
We check whether concrete map keys are not identical patterns, but that's not enough if we have unevaluated symbols.

As an example, say we have two total function symbols from int to int, `f` and `g`, that do not have any definition. Then, when taking the ceil of `{f(1) -> 1, g(1) -> 1}` we get `top`, but the map might still be bottom if `f(1) == g(1)`.

Also see the "ceil with map domain value" unit test.

ttuegel:
I don't know what your thoughts are about this, but instead of allowing any concrete keys, we could allow only types of patterns that we know are fully evaluated—Int, Bool, String—but this might cause problems with semantics that use Maps from identifiers to values.

virgil-serbanuta:
We can also allow anything that has a constructor or inj at the top.

PR #1169, opened 2019-10-28
Allow only non-simplifiable concrete keys in maps and sets

ana-pantilie:
~Needs https://github.com/kframework/kore/pull/1182~
~Needs https://github.com/kframework/kore/pull/1227~
Fixes https://github.com/kframework/kore/issues/1077

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -313,6 +319,34 @@ makeEvaluateNormalizedAc
             ++ elementsWithVariablesDistinct
     And.simplifyEvaluatedMultiPredicate (MultiAnd.make allConditions)
   where
+    assertCanBeEqCheck
+        :: TermLike variable
+        -> simplifier (TermLike variable)
+    assertCanBeEqCheck key =
+        case key of
+            App_ symbol _ ->
+                if not (Symbol.isInjective symbol)
```
ttuegel: `Symbol.isInjective` is redundant because all constructors are injective. Did you mean to check if it's a sort injection?

There is a complication here: This only checks the head of the pattern, but it could have a constructor at the top and unevaluated functions as arguments (for example). I think we may need to follow [this suggestion](https://github.com/kframework/kore/blob/master/kore/src/Kore/Attribute/Symbol.hs#L171) by @virgil-serbanuta and introduce a `NonSimplifiable` pattern attribute to track this.

kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -313,6 +319,34 @@ makeEvaluateNormalizedAc
...
+                      \ which can be tested for equality."
+                    , Pretty.indent 2 "key:"
+                    , Pretty.indent 4 (unparse key)
+                    ]
+                else return key
+            BuiltinBool_ _   -> return key
+            BuiltinInt_ _    -> return key
+            BuiltinString_ _ -> return key
+            StringLiteral_ _ -> return key
+            DV_ _ _          -> return key
```
ttuegel: We need to check that the child of the domain value is not simplifiable, too.

kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -313,6 +319,34 @@ makeEvaluateNormalizedAc
...
   where
+    assertCanBeEqCheck
+        :: TermLike variable
+        -> simplifier (TermLike variable)
+    assertCanBeEqCheck key =
+        case key of
+            App_ symbol _ ->
+                if not (Symbol.isInjective symbol)
+                   && not (Symbol.isConstructor symbol)
+                then (error . show . Pretty.vsep)
```
ttuegel: Let's factor out the error so that we don't have to write the message twice.

kore/src/Kore/Step/Simplification/Ceil.hs
virgil-serbanuta: I know that I said differently in the past (sorry for the confusion), but:
1. I think that having constructors at the top is not enough, since the keys can be, say, `C(f())` and `C(g())`, where `C` is a constructor, and `f` and `g` are uninterpreted symbols, and then there's no way of comparing them.
2. I don't think all injections work, only sort injections do, since you can't compare random distinct injections.

3. @ttuegel I think sort injection do not work unless they are simplified, since we can transform `inj(inj)` into a single `inj`. Should we assume that? If yes, should we check it?

ttuegel: > I think sort injection do not work unless they are simplified, since we can transform `inj(inj)` into a single `inj`. Should we assume that? If yes, should we check it?

We should not assume that sort injections are non-simplifiable, but we should inspect the child to see if it is a sort injection. We did this in the past, but I don't know what happened.

kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -420,11 +421,14 @@ Return 'Nothing' if there are any duplicate keys.
...
  -}
 updateConcreteElements
-    :: Ord key
-    => Map key value
-    -> [(key, value)]
-    -> Maybe (Map key value)
-updateConcreteElements = Foldable.foldrM (uncurry insertMissing)
+    :: Map (TermLike Concrete) value
+    -> [(TermLike Concrete, value)]
+    -> Maybe (Map (TermLike Concrete) value)
```
ana-pantilie: @virgil-serbanuta I added the check in this function as well (I hope that makes sense). Judging by the name of this function and its context it doesn't seem like we ever intend it to be used outside of the case where we have a `TermLike Concrete` as the key type. Do you think this seems reasonable or do we actually need a more general type here?

kore/src/Kore/Builtin/Map.hs
```diff
@@ -329,9 +330,10 @@ evalUpdate =
                         _ -> Builtin.wrongArity Map.updateKey
             _key <- hoistMaybe $ Builtin.toKey _key
             _map <- expectConcreteBuiltinMap Map.updateKey _map
-            returnConcreteMap
-                resultSort
-                (Map.insert _key (Domain.MapValue value) _map)
+            TermLike.assertNonSimplifiableKeys (_key : Map.keys _map)
```
ttuegel: You could also do this here:

```.hs
TermLike.assertNonSimplifiableKeys (_key : Map.keys _map) $ return ()
returnConcreteMap ...
```

kore/src/Kore/Builtin/AssociativeCommutative.hs
virgil-serbanuta: I'm not sure which kind of generalization you mean. To pick one of them, this function was meant only for concrete elements. We also have `updateAbstractElements` for updating the others. Checking this here makes sense.

ana-pantilie: Alright. I was just asking since I changed `key` to `TermLike Concrete` to make this work. Thanks!

virgil-serbanuta: Oh, I see! Well, I'm not going to complain if it compiles...

PR #1172, opened 2019-10-29
Iterate simplification

virgil-serbanuta:
We should follow this with removing the various redundant `markSimplified` calls.

---

###### Reviewer checklist


kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1014,78 +1015,112 @@ forceSort forcedSort = Recursive.apo forceSortWorker
...
+        [ "Could not force pattern to sort "
+        , Pretty.squotes (unparse forcedSort)
+        , ", instead it has sort "
+        , Pretty.squotes (unparse (termLikeSort original))
+        , ":"
+        ]
+    , Pretty.indent 4 (unparse original)
+    ]
+
+forceSortWorkerPredicate
```
traiansf: I was wondering whether this is a good name.

kore/src/Kore/Predicate/Predicate.hs
```diff
@@ -474,23 +508,95 @@ makePredicate = Recursive.elgot makePredicateBottomUp makePredicateTopDown
...
+        wasSimplified =
+            Attribute.Simplified.isSimplified
+                (Attribute.Pattern.simplified attrs)
+
+    updateHasChanged
+        :: HasChanged
+        -> TermLike variable
+        -> Predicate variable
+        -> HasChanged
+    updateHasChanged before term predicate =
```
ttuegel: Is this tracking whether the term's sort was changed building the predicate?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -201,101 +225,80 @@ simplifyInternal term predicate =
...
-                OrF orF -> Or.simplify <$> simplifyChildren orF
-                RewritesF rewritesF ->
-                    Rewrites.simplify <$> simplifyChildren rewritesF
-                TopF topF -> Top.simplify <$> simplifyChildren topF
-                --
-                StringLiteralF stringLiteralF ->
-                    return $ StringLiteral.simplify (getConst stringLiteralF)
-                VariableF variableF ->
-                    return $ Variable.simplify (getConst variableF)
+          = assertTermNotPredicate $ tracer termLike $ do
```
ttuegel:
```suggestion
        = assertTermNotPredicate $ tracer termLike $ do
```

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -319,6 +322,77 @@ simplifyInternal term predicate =
                     , "Expected all predicates to be removed from the term."
                     ]
 
+    descendAndSimplify :: TermLike variable -> simplifier (OrPattern variable)
```
ttuegel: This is the old worker function, right?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -201,101 +225,80 @@ simplifyInternal term predicate =
...
-                --
-                StringLiteralF stringLiteralF ->
-                    return $ StringLiteral.simplify (getConst stringLiteralF)
-                VariableF variableF ->
-                    return $ Variable.simplify (getConst variableF)
+          = assertTermNotPredicate $ tracer termLike $ do
+            termOr <- descendAndSimplify termLike
+            termPredicateList <- BranchT.gather $ do
+                termOrElement <- BranchT.scatter termOr
+                simplified <- simplifyCondition termOrElement
```
ttuegel: How is this going to work with `simplifyCondition` calling back into the term simplifier?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -201,101 +225,80 @@ simplifyInternal term predicate =
...
+                return (applyTermSubstitution simplified)
+
+            case termPredicateList of
+                [] -> return OrPattern.bottom
+                [result] ->
```
ttuegel: Why is a single result handled differently from many results, below?

kore/test/Test/Kore/Predicate/Predicate.hs
```diff
@@ -344,9 +345,64 @@ test_predicate =
...
+        (toBool termSimplification)
+        (TermLike.isSimplified term)
+    assertEqual "Predicate simplification"
+        (Right (toBool predicateSimplification))
+        (Predicate.isSimplified <$> eitherPredicate)
+  where
+    toBool IsSimplified = True
+    toBool NotSimplified = False
+
+
```
ttuegel:
```suggestion
```

kore/src/Kore/Internal/TermLike.hs
virgil-serbanuta: Better now?

kore/src/Kore/Step/Simplification/TermLike.hs
virgil-serbanuta: Not that well, it seems that I missed some stuff when moving code around. Thank you for pointing it out. I rewrote it, in the new version it should work like this (siw = simplifyInternalWorker, sc = simplifyCondition):
```
1. siw t1 computes t2 and p2, then calls sc p2
2. sc p2 calls siw p2
3. siw p2 computes t3 and p3. t3 should be top. If p2 == p3, i.e. this was fully simplified, mark it as such and continue. Otherwise call sc p2, go to step 2
```

kore/src/Kore/Predicate/Predicate.hs
ttuegel: What does the `HasChanged` value mean, that this is updating?

Issue #1174, opened 2019-10-29
Add NonSimplifiable attribute for patterns


ana-pantilie:
This is needed for https://github.com/kframework/kore/issues/1077.

A pattern is `NonSimplifiable` if it's a:
1. `BuiltinBool` or `BuiltinInt` or `BuiltinString`
2. a constructor applied over a `NonSimplifiable` pattern
3. a sort injection applied over a `NonSimplifiable` pattern `pat`, where `pat` does not also have a sort injection at the top

See https://runtimeverification.slack.com/archives/CC360GUTG/p1572364797433900?thread_ts=1572364637.433400&cid=CC360GUTG

PR #1182, opened 2019-10-30
Kore.Attribute.Pattern.NonSimplifiable

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1174

---

###### Reviewer checklist


kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
```diff
@@ -0,0 +1,191 @@
...
+import Kore.Syntax
+import Kore.Syntax.Application
+    ( Application (..)
+    )
+import Kore.Variables.UnifiedVariable
+    ( UnifiedVariable
+    )
+
+{- | A pattern is `NonSimplifiable` if:
+    1. it's a `BuiltinBool`, `BuiltinInt` or a `BuiltinString`
```
virgil-serbanuta: @ana-pantilie You asked me about this before. I'm sorry I only realized this now:

@ana-pantilie @ttuegel Should we also include domain values of unrecognized types (i.e. not int, bool, string) here? I think (but I may be wrong) that imp uses such domain values for IDs, and I think it uses them as storage keys.

Also, strictly speaking, we could probably include sets, lists and maps when all their elements are non-simplifiable, but it's probably not worth the trouble.

kore/test/Test/Kore/Step/MockSymbols.hs
```diff
@@ -50,6 +50,7 @@ import qualified Kore.Attribute.Sort.Unit as Attribute
 import qualified Kore.Attribute.Symbol as Attribute
 import qualified Kore.Builtin.Bool as Builtin.Bool
 import qualified Kore.Builtin.Int as Builtin.Int
+import Kore.Builtin.String as Builtin.String
```
virgil-serbanuta: Why not make this a qualified-only import? Chances are you won't need the `sort -> sort'` change below anymore.

kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
ana-pantilie: @virgil-serbanuta Yes, we use domain values as keys as well. Sorry, I marked this pull request as ready before testing the integration tests with the other PR as well. Why do you think we should only include domain values of unrecognized types?

virgil-serbanuta: We could think about that, but the only kind of recognizable domain values that I saw had a string child, and we usually check that it's a valid string. And we simplify those to BuiltinBool/Int/String.

I did try to find hypothetical examples of bool/int/string domain values that do not have strings as children, and which are unsimplifiable at the same time, and I'm not sure that makes sense (i.e. I would expect that any non-string child would be simplifiable to a string), but I'm not fully sure yet.

ttuegel: > I think (but I may be wrong) that imp uses such domain values for IDs, and I think it uses them as storage keys.

This is correct.

> Also, strictly speaking, we could probably include sets, lists and maps when all their elements are non-simplifiable, but it's probably not worth the trouble.

Agreed.

kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
```diff
@@ -0,0 +1,191 @@
...
+import qualified Kore.Internal.Symbol as Symbol
+import Kore.Syntax
+import Kore.Syntax.Application
+    ( Application (..)
+    )
+import Kore.Variables.UnifiedVariable
+    ( UnifiedVariable
+    )
+
+{- | A pattern is `NonSimplifiable` if:
```
ttuegel: Haddock uses apostrophe instead of backtick:

```suggestion
{- | A pattern is 'NonSimplifiable' if:
```

kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
```diff
@@ -0,0 +1,201 @@
...
+    synthetic = const (NonSimplifiable . Just $ BuiltinHead)
+    {-# INLINE synthetic #-}
+
+instance Synthetic NonSimplifiable (Top sort) where
+    synthetic = const (NonSimplifiable Nothing)
+    {-# INLINE synthetic #-}
+
+data NonSimplifiableHead = ConstructorLikeHead
+                         | SortInjectionHead
+                         | BuiltinHead
```
virgil-serbanuta: Why is `BuiltinHead` distinct from `ConstructorLikeHead`?

kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
ttuegel: > Why do you think we should only include domain values of unrecognized types and not all domain values with a non-simplifiable child?

During the daily meeting we checked and found that for all the recognized types of domain value, the pattern verifier checks that they have a literal string argument and decodes them to their builtin representation; it is an error otherwise. So, it should be safe to assume that any `\dv` remaining during execution is in an unrecognized sort.

kore/src/Kore/Attribute/Pattern/NonSimplifiable.hs
ana-pantilie: Yeah, I guess this distinction didn't turn out to be necessary. I'll remove one of them.

PR #1184, opened 2019-10-30
Improved recognition for priority rules

traiansf:
---

###### Reviewer checklist


traiansf:
@yzhang90 this currently breaks for rules with top priority because they are not encoded the same way as the other priority rules.

traiansf:
I've refactored rule recognition per @virgil-serbanuta 's suggestion, and now we can parse all rules generated by @yzhang90 's code.
I've also made failing to recognize a rule a hard error and added exceptions for functional, subsort and constructor axioms.

kore/src/Kore/Step/Rule.hs
```diff
@@ -579,24 +621,45 @@ patternToAxiomPattern
...
+patternToAxiomPattern attributes pat =
     case pat of
+        Rewrites_ _ (ApplyAlias_ alias params) rhs ->
+            case substituteInAlias alias params of
+                And_ _ requires lhs ->
+                    patternToAxiomPattern
+                        attributes
+                        (mkRewrites (mkAnd requires lhs) rhs)
+                _ ->
+                    error
```
ana-pantilie: This looks like it will print the error message and the pattern on the same line. I think it would be nicer if we used `Pretty` here to also indent the pattern so the error is more readable like we do in other places in the code.

kore/src/Kore/Step/Rule.hs
```diff
@@ -671,7 +727,16 @@ patternToAxiomPattern attributes pat
...
                     , attributes
                     }
-        _ -> koreFail "Unsupported pattern type in axiom"
+        _
+            | (isDeclaredFunctional . Attribute.functional $ attributes)
+            || (isConstructor . Attribute.constructor $ attributes)
+            || (not . null . getSubsorts . Attribute.subsorts $ attributes)
+            -> koreFail "Patterns of this type do not represent rules"
+            | otherwise ->
+                error
```
ana-pantilie: The same as above.

kore/src/Kore/Attribute/Axiom.hs
```diff
@@ -97,6 +99,10 @@ data Axiom =
     , constructor :: !Constructor
     -- ^ Shows that this is one of the constructor axioms
     -- (e.g. no confusion, no junk)
+    , functional :: !Functional
```
ana-pantilie: What does this attribute mean for axioms? Also, does this use `Kore.Attribute.Functional`? I think we only use that for symbols so the documentation should need updating if we were to use it with axioms as well.

traiansf: it is copied by the frontend to axioms stating the functionality of a symbol.

ana-pantilie: Oh, I see. Then disregard my comment. Thanks for the explanation!

kore/src/Kore/Attribute/Subsort.hs
```diff
@@ -6,23 +6,41 @@ License     : NCSA
...
 
 {- | The @subsort@ attribute. -}
 data Subsort = Subsort
     { subsort :: Sort
     , supersort :: Sort
     }
-    deriving (Eq, Generic, Ord, Show)
+    deriving (Eq, GHC.Generic, Ord, Show)
+
+instance SOP.Generic Subsort
```
ttuegel: The style guide specifies one blank line between top-level declarations. (Should we update the guide?)

src/main/k/in-progress/tests/priority/tests/1.output.golden
```diff
@@ -0,0 +1,3 @@
+<k>
+  #Exists _0 . _0 : #Exists _0 . _0 : #Exists _0 . _0 : .Seq
```
ttuegel: Is this the right output?

kore/src/Kore/Attribute/Axiom.hs
traiansf: It's actually a very good comment. I've added a note to the documentation of the functional attribute stating that it can be used for axioms, too.

kore/src/Kore/Attribute/Subsort.hs
traiansf: I don't mind leaving a blank line. It's just that I don't know the guide by heart and I saw some similar instances not separated...
Maybe it's good to have them separated; it's harder to confuse them for a single block.

src/main/k/in-progress/tests/priority/tests/1.output.golden
traiansf: no. I just forgot to update it. thanks for mentioning it.

PR #1188, opened 2019-10-30
Add LIST.update hook

ttuegel:
Needs:

- [x] Implementation
- [ ] Unit tests
- [ ] Integration tests
- [x] Documentation in `hooks.md`

Fixes: #1186 

---

###### Reviewer checklist


kore/src/Kore/Builtin/List.hs
```diff
@@ -254,11 +251,36 @@ evalGet =
...
+evalUpdate =
+    Builtin.functionEvaluator evalUpdate0
+  where
+    evalUpdate0 :: Builtin.FunctionImplementation
+    evalUpdate0 resultSort [_list, _ix, value] =
+        Builtin.getAttemptedAxiom $ do
+            let emptyList = do
+                    _list <- expectBuiltinList updateKey _list
+                    if Seq.null _list
+                        then Builtin.appliedFunction Pattern.bottom
```
virgil-serbanuta: Why does this return bottom for an empty list, while returning empty if the index is out of bounds below?

kore/test/Test/Kore/Builtin/List.hs
```diff
@@ -96,6 +99,39 @@ test_getLastElement =
...
+        ix <- forAll genSeqIndex
+        let len = fromIntegral $ length values
+            patValues = asTermLike $ Test.Int.asInternal <$> values
+            patUpdated = mkApplySymbol updateListSymbol
+                [ patValues
+                , Test.Int.asInternal ix
+                , value
+                ]
+        if (-len) <= ix && ix < len then do
+            let patGet = mkApplySymbol getListSymbol
```
ttuegel: Let's use `getList` here, and the same for `updateList` above.

kore/src/Kore/Builtin/List.hs
```diff
@@ -254,11 +251,30 @@ evalGet =
...
+    evalUpdate0 :: Builtin.FunctionImplementation
+    evalUpdate0 resultSort [_list, _ix, value] =
+        Builtin.getAttemptedAxiom $ do
+            _list <- expectBuiltinList getKey _list
+            _ix <- fromInteger <$> Int.expectBuiltinInt getKey _ix
+            let len = Seq.length _list
+                ix
+                  | _ix < 0 =
+                    -- negative indices count from end of list
+                    _ix + len
```
vladciobanu: What if it wraps around twice? Should it be
```
_ix `mod` len
```
And I think you can remove the special case since this would work for both negative and positive numbers alike.

PR #1189, opened 2019-10-30
Replace the missing hook warning with a helpful error

ttuegel:
@vladciobanu I thought we could get away with a warning here, but the backend makes a spectacle of itself if a hook is missing. Sorry for pushing you to do that. At least it helped refine the logging interface. :frowning:

---

###### Reviewer checklist


ttuegel:
> sorry I didn't test it better but it seemed to work reasonably on trivial cases.

It's not your fault, it had nothing to do with the warning itself: When faced with the missing `KRYPTO.sha256` hook, the backend tried to continue executing symbolically as far as it could. In the Beacon Chain spec, that hook is used as an argument of a recursive call... without a concrete string, the backend can't decide when to stop recursing.

PR #1200, opened 2019-11-01
decidePredicate: Error if predicate is not decided

ttuegel:
Throw an error if the SMT solver returns `Unknown` because we probably won't continue correctly.

---

###### Reviewer checklist


ttuegel:
This does not seem to be something we can promote to an error, but at least the warning is shown now.

PR #1201, opened 2019-11-01
Merge rules in batches

virgil-serbanuta:
To merge in batch mode:
```
kore-exec definition.kore --merge-rules rule-list --module WASM-WITH-K-TERM --merge-batch-size 2
```

To merge without batches, just omit the flag:
```
kore-exec definition.kore --merge-rules rule-list --module WASM-WITH-K-TERM
```
---

###### Reviewer checklist

### DIFF
--- a/kore/app/share/GlobalMain.hs
+++ b/kore/app/share/GlobalMain.hs
@@ -74,6 +74,7 @@ import GHC.Stack
 import Options.Applicative
     ( InfoMod
     , Parser
+    , ReadM
     , argument
     , disabled
     , execParser
@@ -85,6 +86,7 @@ import Options.Applicative
     , info
     , internal
     , long
+    , maybeReader
     , metavar
     , option
     , readerError
@@ -132,6 +134,9 @@ import Kore.Syntax.Definition
 import qualified Kore.Verified as Verified
 import qualified Paths_kore as MetaData
     ( version
+    )
+import Text.Read
+    ( readMaybe
     )
 
 type Main = LoggerT IO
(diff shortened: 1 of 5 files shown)


kore/app/share/GlobalMain.hs
```diff
@@ -200,17 +205,34 @@ data KoreMergeOptions =
...
+        <> long "merge-rules"
+        <> help
+            "List of rules to merge."
+        )
+    <*> option
+        maybeIntReader
+        (  metavar "MERGE_BATCH_SIZE"
+        <> long "merge-batch-size"
+        <> help
+            "The size of a merge batch."
```
ehildenb: How do we say "all the rules"? Is it `--merge-batch-size 0`?

virgil-serbanuta: Wow, that was fast! If you don't mention --merge-batch-size it should merge all rules at once, but I didn't test it yet

virgil-serbanuta: Hm, my message below does not seem to be attached to this one. Oh well...

This merged the rule list you last sent me in ~6 seconds. I'm running it like this:
```
kore-exec definition.kore --merge-rules rule-list --module WASM-WITH-K-TERM --merge-batch-size=2
```
It produces two rather large rules, but they seem reasonable at first sight (I mostly looked at their "shape", not at their contents).

virgil-serbanuta: `--merge-batch-size 0` merges all rules at once.

ehildenb:
It looks fine to me, but should have @ttuegel 's approval I think.

PR #1209, opened 2019-11-05
functionLike checks for simplification

traiansf:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -499,6 +499,9 @@ recoveryFunctionLikeResults initial unifier = do
             , "doesn't match configuration"
             , Pretty.indent 4 $ unparse phi_t
             ]
+        -- what we would like to check above is that phi_p -> phi_t = alpha_t,
+        -- but that's hard to do for non-functional patterns,
+        -- so we check for (syntactic) equality instead.
         unifier
     fullyOverrideSort' sort term
       | sort == termLikeSort term = term
--- a/kore/src/Kore/Step/Step.hs
+++ b/kore/src/Kore/Step/Step.hs
@@ -79,6 +79,8 @@ import Kore.Logger
 import qualified Kore.Logger as Log
 import Kore.Predicate.Predicate as Predicate
 import qualified Kore.Step.Remainder as Remainder
+import qualified Kore.Step.Result as Result
+import qualified Kore.Step.Result as Results
 import qualified Kore.Step.Result as Step
 import Kore.Step.Rule
     ( RewriteRule (..)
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Step/Step.hs
```diff
@@ -438,7 +439,61 @@ recoveryFunctionLikeResults
...
+    results <- Step.results <$> unifier
+    let unifiedRules = Step.appliedRule <$> results
+    case checkFunctionLike unifiedRules (term initial) of
+        Right _  -> unifier
+        Left err -> moreChecksAfterError err
+  where
+    moreChecksAfterError err = do
+        rules <- fmap Step.appliedRule . Step.results <$> unifier
+        let
+            rule = case Seq.length rules of
```
virgil-serbanuta: s/rule/appliedRule/?

kore/src/Kore/Step/Step.hs
```diff
@@ -438,7 +439,61 @@ recoveryFunctionLikeResults
     => Pattern (Target variable)
     -> unifier (Results variable)
     -> unifier (Results variable)
-recoveryFunctionLikeResults _ = id
+recoveryFunctionLikeResults initial unifier = do
```
virgil-serbanuta: Could you add a pointer to the document explaining this recovery process?

Also, tests?

kore/src/Kore/Step/Step.hs
```diff
@@ -438,7 +448,61 @@ recoveryFunctionLikeResults
...
+        Monad.when (res1 /= Just False) $ error $ show $ Pretty.vsep
+            [ "Couldn't recover simplification coverage error: "
+            , Pretty.indent 4 (Pretty.pretty err)
+            , "Configuration condition "
+            , Pretty.indent 4 $ unparse phi_p
+            , "doesn't imply rule condition "
+            , Pretty.indent 4 $ unparse alpha_p
+            ]
+        let alpha_t_sorted = fullyOverrideSort' (termLikeSort phi_t) alpha_t
+        Monad.when (phi_t /= alpha_t_sorted) $ error $ show $ Pretty.vsep
```
virgil-serbanuta: Could you add a comment here saying that what we would like to check is that `\implies(phi_t  alpha_t_sorted)`, but that's hard to do for non-functional patterns, so we check for (syntactic) equality instead?

BTW, the design document says "This could be checked with an SMT", but I'm not sure how to do that.

traiansf: > BTW, the design document says "This could be checked with an SMT", but I'm not sure how to do that.

With a very smart SMT which would know all about equality in Matching Logic :-)

kore/src/Kore/Step/Step.hs
```diff
@@ -438,7 +448,64 @@ recoveryFunctionLikeResults
...
+    results <- Results.results <$> unifier
+    let appliedRules = Result.appliedRule <$> results
+    case checkFunctionLike appliedRules (Conditional.term initial) of
+        Right _  -> unifier
+        Left err -> moreChecksAfterError appliedRules err
+  where
+    moreChecksAfterError appliedRules err = do
+        let
+            appliedRule = case Seq.length appliedRules of
+                1 -> Seq.index appliedRules 0
```
ttuegel: We can match on `appliedRules` using pattern synonyms instead of applying a partial function:

```suggestion
                rule Seq.:< Seq.EmptyL -> rule
```

PR #1212, opened 2019-11-05
Rewrite the merge rule batch size flag

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -466,14 +466,21 @@ koreMerge execOptions mergeOptions = do
     definition <- loadDefinitions [definitionFileName]
     let KoreExecOptions {mainModuleName} = execOptions
     mainModule <- loadModule mainModuleName definition
+    let KoreExecOptions { patternFileName } = execOptions
+    maybeInitial <- case patternFileName of
+        Nothing -> return Nothing
+        Just _ -> do
+            initial <- loadPattern mainModule patternFileName
+            return (Just initial)
     let KoreMergeOptions {rulesFileName} = mergeOptions
     ruleIds <- lift $ loadRuleIds rulesFileName
     let KoreMergeOptions {maybeBatchSize} = mergeOptions
     eitherMergedRule <- execute execOptions mainModule $
         case maybeBatchSize of
             Just batchSize ->
-                mergeRulesConsecutiveBatches batchSize mainModule ruleIds
-            Nothing -> mergeAllRules mainModule ruleIds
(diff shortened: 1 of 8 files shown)


ehildenb:
Hmmmm, this looks fine to me, but now that I see all you're doing is building a new ID rule `INIT => INIT`, and then adding it to the front of the list of rules to merge, I wonder if it makes more sense for me to do this myself in the `pyk` library rather than using the backend? Using the frontend, I could add side-conditions, for example, to provide constraints over the initial term.

ehildenb:
But the change to make `merge-batch-size` an `optional` instead of using `0` to mean "all of them" seems good on its own.

kore/app/exec/Main.hs
```diff
@@ -466,14 +466,21 @@ koreMerge execOptions mergeOptions = do
     definition <- loadDefinitions [definitionFileName]
     let KoreExecOptions {mainModuleName} = execOptions
     mainModule <- loadModule mainModuleName definition
+    let KoreExecOptions { patternFileName } = execOptions
+    maybeInitial <- case patternFileName of
```
ttuegel: Why are we unpacking this `Maybe` but passing it untouched into `loadPattern`?

kore/src/Kore/Exec.hs
```diff
@@ -448,7 +458,19 @@ mergeRules ruleMerger verifiedModule ruleNames =
                 rules <- extractRules rewriteRules ruleNames
                 case rules of
                     [] -> Left "Empty rule list."
-                    (r : rs) -> Right (r :| rs)
+                    (r : rs) -> case maybeInitialConfiguration of
+                        Nothing -> Right (r :| rs)
+                        Just cfg -> Right (identityRule cfg :| r : rs)
+
+            identityRule :: TermLike Variable -> RewriteRule Variable
```
ttuegel: The `identityRule` is because you can only insert rules, not configurations?

virgil-serbanuta: As discussed on Slack, I'm rolling this back.

To answer your question, I could have changed the merging algorithms to take an optional configuration, or to take a mixed list of configurations and rules, but then I realized that, while we are starting with a list of rules of the type `a=>b` where `a` is being rewritten in one step to `b`, when the algorithm is running we are merging rules where the rewriting is being done in any number `n>0` of steps. So then the obvious extension was to also include rules that rewrite in `0` steps, a.k.a. (partial) configurations, which makes `identityRule` the natural representation here.

Issue #1187, opened 2019-10-30
Clean up logging options


ttuegel:
These are user interface bugs:

- [ ] `--log none` makes no sense; the user should not be able to suppress error reporting.
- [ ] The default `--log-level` should be `warning`.
- [ ] `--log stdout` should be `--log stderr`.
- [ ] Fix the `--help` to explain how to use `--log` to specifying a filename.

PR #1213, opened 2019-11-06
Start cleaning up logging options

andreiburdusa:
---
Referencing #1187 

- remove `--log none`
- set default `--log-level` to `warning`
- change `--log stdout` in `--log stderr`

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Logger/Output.hs
+++ b/kore/src/Kore/Logger/Output.hs
@@ -172,7 +172,8 @@ parseKoreLogOptions =
             "warning"  -> pure Warning
             "error"    -> pure Error
             "critical" -> pure Critical
-            _          -> pure Warning
+            ""         -> pure Warning
+            _          -> Nothing
     parseScope =
         option
             parseCommaSeparatedScopes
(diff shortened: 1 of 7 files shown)


kore/src/Kore/Repl/Parser.hs
```diff
@@ -275,9 +275,8 @@ logScope =
             <* optional (literal ",")
 
 logType :: Parser LogType
-logType = noLogging <|> logStdOut <|> logFile
+logType = logStdOut <|> logFile
```
andreiburdusa: I did't remove `NoLogging` constructor from `data LogType` because I thought it might be used in other cases beside command line arguments, but only removed the possibility for it to be specified by user. I don't know if this is correct

kore/src/Kore/Logger/Output.hs
```diff
@@ -177,7 +172,7 @@ parseKoreLogOptions =
             "warning"  -> pure Warning
             "error"    -> pure Error
             "critical" -> pure Critical
-            _          -> Nothing
+            _          -> pure Warning
```
ttuegel: If the user supplies `--log-level`, but we don't recognize the string, it should be a parse failure.

```suggestion
            _          -> Nothing
```

kore/src/Kore/Repl/Parser.hs
ttuegel: That's fine, but it should be `stderr` again instead of `stdout`.

PR #1218, opened 2019-11-06
Simplify or simplification results

virgil-serbanuta:
---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Bottom.hs
```diff
@@ -11,14 +11,19 @@ module Kore.Step.Simplification.Bottom
...
-import qualified Kore.Internal.OrPattern as OrPattern
 import Kore.Sort
 import Kore.Syntax.Bottom
 
 {-| simplifies a Bottom pattern, which means returning an always-false or.
 -}
-simplify :: Ord variable => Bottom Sort child -> OrPattern variable
-simplify Bottom {} = OrPattern.bottom
+simplify :: InternalVariable variable => Bottom Sort child -> Pattern variable
+simplify Bottom {} = Pattern.bottom
```
ttuegel: We could preserve the sort here now.

kore/src/Kore/Step/Simplification/Application.hs
```diff
@@ -71,7 +71,7 @@ simplify predicate application = do
         (symbolConstructor symbol)
         (length childrenCrossProduct)
         (length result)
-    return result
+    return (OrPattern.toPattern result)
```
ttuegel: If this `OrPattern` contains multiple branches, won't this have to be unpacked from a `Pattern` to an `OrPattern` again later?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -369,64 +369,83 @@ simplifyInternal term predicate = do
             --
             AndF andF ->
                 And.simplify =<< simplifyChildren andF
-            ApplySymbolF applySymbolF ->
-                Application.simplify predicate
+            ApplySymbolF applySymbolF -> OrPattern.fromPattern <$>
```
ttuegel: Why are we unpacking the `Pattern` we just packed up?

kore/src/Kore/Step/Simplification/Application.hs
virgil-serbanuta: Yes, but it will be unpacked byt the `or` simplifier, which unpacks it "the right way". Also see the `TermLike.hs` answer below.

kore/src/Kore/Step/Simplification/Bottom.hs
virgil-serbanuta: If we're going to preserve sorts, I'd like to do that in a different PR, because it's not enough to change this.

As an example, when computing `ceil(bottom)` we just remove the `ceil` and return the bottom. However, that `ceil` might have changed the sorts, so, normally, we'd need to change the predicate's sort when simplifying `ceil`s. Likely we also need to change `floor` simplification. I don't think we need to do anything about it, but there's a small chance that `Equal` simplification has the same problem.

So: not a big problem, but I'd prefer to separate these two issues (sort consistency and `or` simplification).

kore/src/Kore/Step/Simplification/TermLike.hs
virgil-serbanuta: Strictly speaking, this is not unpacking, it's wrapping: this is an `OrPattern` packed in a `Pattern` wrapped in an `OrPattern`. I am doing the `OrPattern -> Pattern` packing so that the `or` simplifier has its chance to simplify the `or`, and I am doing the `Pattern -> OrPattern` wrapping because `OrPattern` is still the standard data structure in `TermLike.simplify`.

This seems somewhat ridiculous, so I'll try to get rid of it (either by returning `OrPattern`s or by making `Pattern` the standard thing in `TermLike.simplify`), but I think that the `Pattern -> OrPattern` wrapping is not that bad for performance.

virgil-serbanuta:
> I don't get a good feeling about packing `OrPattern` into `Pattern`. In the past, repeated packing/unpacking has been a performance problem; not just a constant factor, but an asymptotic slowdown. I can't make a specific argument for that here without measuring, though.

Me neither. I think that some of these would go away as I finish my PR, while others will not matter as much, but, as you said, I don't actually know how much these will affect performance.

> 
> It's also not clear to my why this helps the idempotence problem, particularly because the packed `Pattern` is immediately unpacked into an `OrPattern` again by the `TermLike` simplifier.

I provided more details through the in-line answers, but I also need to answer the "why" question:
This helps the idempotence problem because the `or` patterns are simplified in an uniform way, so when doing an `OrPattern->TermLike` transformation as we do for, say, `not` simplification, we won't get a different result'. Right now that happens because the `OrPattern` is not a simplified `or`, it's just something returned by a random simplifier, while in the resimplification call we would actually simplify the `or` under the `not`.

ttuegel:
I'm doing a spot-check of performance of this branch vs. master on my local machine, just to make sure Jenkins is giving us an accurate picture. I am suspicious of the results when changes to `elapsed_ns` are uncorrelated with changes to `allocated_bytes`.

ttuegel:
I checked the `sum-to-n-spec.k` on my machine with no other load, and unfortunately it is still significantly worse. :slightly_frowning_face:

Issue #1207, opened 2019-11-05
Incorrect set variable de-duplication


ttuegel:
During substitution normalization, we de-duplicate element variable substitutions in this way:

```
x = t₁ ∧ ... ∧ x = tₙ
x = (t₁ ∧ ... ∧ tₙ)
```

Set variables are de-duplicated in the same way, which is not correct. Set variables should be de-duplicated this way:

```
@x = t₁ ∧ ... ∧ @x = tₙ
@x = t₁ ∧ t₁ = t₂ ∧ ... ∧ t₁ = tₙ
```

These transformations are equivalent for element variables.

See also: https://runtimeverification.slack.com/archives/CC360GUTG/p1572914597064200

ttuegel:
During the iteration planning meeting, we discussed this issue, but could not come to a resolution. We don't actually need to implement set variable deduplication right now because it is not required by any semantics. We should implement an error message in case it is required.

In [`Kore.Step.Simplification.SubstitutionSimplifier.deduplicateSubstitution`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs#L198), the [second argument](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs#L215) is the set of substitutions. The keys of that `Map` are the substitution variable, and the `NonEmpty` list of values are all the assignments for that variable. We should add an assertion that if any key is a `SetVar`, then its list of assignments `isSingleton` (`isSingleton = isJust . getSingleton`).

ttuegel:
Fixed in #1224.

PR #1224, opened 2019-11-08
Add check in deduplicateSubstitution

andreiburdusa:
---
Referencing #1207 

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
+++ b/kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
@@ -45,6 +45,9 @@ import Data.Map.Strict
     ( Map
     )
 import qualified Data.Map.Strict as Map
+import Data.Maybe
+    ( isJust
+    )
 import qualified GHC.Generics as GHC
 
 import Branch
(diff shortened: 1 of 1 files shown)


kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -208,8 +212,23 @@ deduplicateSubstitution
...
+      | Map.null problematicSubmap = checkedMap
+      | otherwise = (error . unlines)
+        [ "found SetVar key:"
+        , show key
+        , "with non-singleton list of assignments as value:"
+        , show value
+        ]
+        where
+            problematicSubmap =
+                Map.filterWithKey
```
ttuegel: We should use `Map.foldMapWithKey` here, something like

```.hs
getAny $ Map.foldMapWithKey (\k v -> Data.Monoid.Any $ isSetVar k && isNotSingleton v) ...
```

This way, we don't construct any intermediate `Map`.

andreiburdusa: Oh, ok. I thought we need to save some example of (key, value) pair that fails the check and display it in the error message.

ttuegel: Hm, that's a good point, that would be useful in the error message. Can you revert the "remove map filter" change, but update the error message to display all the bad substitutions? Sorry. :sweat_smile:

PR #1225, opened 2019-11-08
Add extra warnings

ttuegel:
We should sync up about a standard set of warnings to enable.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Subsort.hs
+++ b/kore/src/Kore/Attribute/Subsort.hs
@@ -1,10 +1,14 @@
 {-|
-Module      : Data.Attribute.Subsort
-Description : Representation and parser for subsort attribute
 Copyright   : (c) Runtime Verification, 2018
 License     : NCSA
 -}
-module Kore.Attribute.Subsort where
+module Kore.Attribute.Subsort
+    ( Subsort (..)
+    , Subsorts (..)
+    , subsortId
+    , subsortSymbol
+    , subsortAttribute
+    ) where
 
 import Control.DeepSeq
     ( NFData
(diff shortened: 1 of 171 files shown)


kore/src/Kore/Attribute/Subsort.hs
```diff
@@ -1,10 +1,14 @@
 {-|
-Module      : Data.Attribute.Subsort
-Description : Representation and parser for subsort attribute
 Copyright   : (c) Runtime Verification, 2018
 License     : NCSA
 -}
-module Kore.Attribute.Subsort where
+module Kore.Attribute.Subsort
```
virgil-serbanuta: This seems like a common theme, so I'll ask even if the PR is not ready: Why would one switch to exporting everything or importing everything?

vladciobanu: This is switching to exporting a specific list, rather than everything. I think Tom is fixing this (due to adding `-Wmissing-export-lists`). This helps with not accidentally exporting a helper function when doing a refactoring, for example

ttuegel: `-Wmissing-export-lists` requires every module to have an export list. The most useful thing about export lists, in my opinion, is that the compiler will warn you when you have unused code. This may not be a big problem, but I did find a few unused functions this way.

ttuegel:
> I might have to go back on `redundant-constraints` because I heard persistent/db libraries might not work very well with this warning.

I think that is because of their reliance on `TemplateHaskell` (generating code with lots of boilerplate constraints). The generated code can (should) be separated from most other code, and you can set `-Wno-redundant-constraints` on a module-by-module basis.

> `-Wcpp-undef` but we're not using CPP macros so it probably doesn't matter

Maybe we need `-Wcpp`! :wink: 

> `-Wmissing-local-signatures` I like adding local signatures but I wouldn't force it

I'm unclear about what this does. The manual says:

> GHC will warn you about any polymorphic local bindings

What does polymorphic mean here, i.e., polymorphic with respect to what? Is this a weakened version of the monomorphism restriction?

> `-Wmissing-import-lists` I would almost say it's nice having this, but it would annoy most people

I think that import lists provide relatively little value when importing from the local package.

Issue #906, opened 2019-08-29
Interface to choose one-path or all-path proofs


ttuegel:
Needs: #937 

See also: #714

---

- [x] Module and rule attributes in frontend
- [x] Update frontend for backend integration tests
- [x] Allow all-path to be used in the backend ( initial work in PR #1162 )
  - [x] `Goal` instance for `ReachabilityRule` (https://github.com/kframework/kore/pull/1206)
  - [x] `ReachabilityRule` as `Claim` + Generalize `verifyClaimStep` to `ReachabilityRule` (https://github.com/kframework/kore/pull/1232)
- [ ] Update integration tests to use new interface (https://github.com/kframework/kore/pull/1251)

traiansf:
PR at https://github.com/kframework/k/pull/842

PR #1232, opened 2019-11-11
ReachabilityRule as Claim

ana-pantilie:
Checkpoint in https://github.com/kframework/kore/issues/906

- `kore-exec` and `kore-repl` now use `ReachabilityRule Variable`
- Replaces the `Coercible` constaints in `Claim` with `ToRulePattern` and `FromRulePattern` on `goal` and `Rule goal`
- Adds `goalToRule` and `ruleToGoal` as methods in `Goal`.
- `Rule (ReachabilityRule Variable)` is now a newtype over `RewriteRule Variable`.
- `Goal.strategy` now returns a `Stream` to make a general version of `verifyClaimStep` typesafe when extracting the first and followup steps from the `Strategy` list/stream. This needed a new package dependency, will revert if you think it's a bit overkill.
- Added `Test.Kore.Strategies.Reachability.Verification`.

---

###### Reviewer checklist

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -120,6 +120,7 @@ import Kore.Profiler.Data
 import Kore.Step
 import Kore.Step.Rule
     ( OnePathRule (..)
+    , ReachabilityRule (..)
     , RewriteRule (..)
     )
 import Kore.Step.Search
@@ -449,7 +450,7 @@ koreProve execOptions proveOptions = do
                     Bounded.Failed final -> return (failure final)
             else
                 either failure (const success)
-                <$> prove stepLimit mainModule specModule (Proxy @(OnePathRule Variable))
+                <$> prove stepLimit mainModule specModule (Proxy @(ReachabilityRule Variable))
     lift $ renderResult execOptions (unparse final)
     return exitCode
   where
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -22,9 +22,6 @@ import qualified Data.Foldable as Foldable
 import Data.List
     ( intercalate
     )
-import Data.Proxy
-    ( Proxy (..)
-    )
 import Data.Reflection
 import Data.Semigroup
     ( (<>)
@@ -118,9 +115,6 @@ import Kore.Profiler.Data
     ( MonadProfiler
     )
 import Kore.Step
-import Kore.Step.Rule
-    ( ReachabilityRule (..)
-    )
 import Kore.Step.Search
     ( SearchType (..)
     )
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -119,18 +119,13 @@ import Kore.Profiler.Data
     )
 import Kore.Step
 import Kore.Step.Rule
-    ( OnePathRule (..)
-    , ReachabilityRule (..)
-    , RewriteRule (..)
+    ( ReachabilityRule (..)
     )
 import Kore.Step.Search
     ( SearchType (..)
     )
 import qualified Kore.Step.Search as Search
 import Kore.Step.SMT.Lemma
-import Kore.Strategies.Goal
-    ( Rule (OnePathRewriteRule)
-    )
 import Kore.Syntax.Definition
     ( ModuleName (..)
     )
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -445,7 +445,7 @@ koreProve execOptions proveOptions = do
                     Bounded.Failed final -> return (failure final)
             else
                 either failure (const success)
-                <$> prove stepLimit mainModule specModule (Proxy @(ReachabilityRule Variable))
+                <$> prove stepLimit mainModule specModule
     lift $ renderResult execOptions (unparse final)
     return exitCode
   where
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -40,9 +40,6 @@ import qualified Data.Bifunctor as Bifunctor
     ( first
     , second
     )
-import Data.Coerce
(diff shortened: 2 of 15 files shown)


kore/src/Kore/Exec.hs
```diff
@@ -526,25 +521,26 @@ assertSomeClaims claims =
         ++  "on the representation of claims."
 
 makeClaim
-    :: Coercible (RulePattern Variable) claim
+    :: Goal.FromRulePattern claim
+    => Goal.ToRulePattern claim
     => (Attribute.Axiom, claim) -> claim
 makeClaim (attributes, rule) =
```
ttuegel:
```suggestion
makeClaim (attributes, Goal.toRulePattern -> rule) =
```

kore/src/Kore/Exec.hs
```diff
@@ -648,13 +644,16 @@ initializeProver definitionModule specModule within =
...
         simplifiedSpecClaims <-
             mapM (mapMSecond simplifyToList) specClaims
         specAxioms <- Profiler.initialization "simplifyRuleOnSecond"
             $ traverse simplifyRuleOnSecond (concat simplifiedSpecClaims)
-        let
-            axioms = coerce <$> rewriteRules
-            claims = fmap makeClaim specAxioms
+        let claims = fmap makeClaim specAxioms
+            axioms =
+                Goal.fromRulePattern (Goal.goalToRule firstClaim)
```
ttuegel: Why are we inserting the first claim into the list of axioms?

kore/src/Kore/Repl.hs
```diff
@@ -193,10 +190,12 @@ runRepl axioms' claims' logger replScript replMode outputFile = do
...
-    addIndexesToClaims len cls =
-        fmap
-            (coerce . addIndex)
-            (zip (fmap coerce cls) [len..])
+    addIndexesToClaims len claims'' =
+        let toAxiomAndBack (claim', index) =
+                ruleToGoal
+                    claim'
+                    $ addIndex (goalToRule claim', index)
+        in fmap toAxiomAndBack (zip claims'' [len..])
```
ttuegel:
```suggestion
        in zipWith toAxiomAndBack claims'' [len..]
```

kore/src/Kore/Step/Rule.hs
```diff
@@ -916,3 +922,45 @@ applySubstitution substitution rule =
...
+
+instance ToRulePattern (ImplicationRule Variable)
+
+instance ToRulePattern (ReachabilityRule Variable) where
+    toRulePattern (OnePath rule) = toRulePattern rule
+    toRulePattern (AllPath rule) = toRulePattern rule
+
+-- | We need to know the context when transforming a 'RulePattern Variable'
+-- into a 'rule', hence the first 'rule' argument. In general it can be ignored
+-- when the 'rule' and the 'RulePattern Variable' are representationally equal.
```
ttuegel: Can the first argument be a `Proxy` then?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -164,7 +188,7 @@ class Goal goal where
         :: goal
         -> [goal]
         -> [Rule goal]
-        -> [Strategy (Prim goal)]
+        -> Stream (Strategy (Prim goal))
```
ttuegel: Why this change?

kore/app/exec/Main.hs
```diff
@@ -449,7 +445,7 @@ koreProve execOptions proveOptions = do
                     Bounded.Failed final -> return (failure final)
             else
                 either failure (const success)
-                <$> prove stepLimit mainModule specModule (Proxy @(OnePathRule Variable))
+                <$> prove stepLimit mainModule specModule (Proxy @(ReachabilityRule Variable))
```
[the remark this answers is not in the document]
ana-pantilie: Doing this will mostly revert https://github.com/kframework/kore/pull/1162, just want to make sure that's what we want.

kore/src/Kore/Exec.hs
ana-pantilie: This doesn't insert the claim, we just need to get the axiom type here which is determined by the claim type. If we revert the generalization of `Exec.prove` I think this awkward way of getting the axiom type at start-up won't be needed anymore.

kore/src/Kore/Strategies/Goal.hs
ana-pantilie: This was for generalizing `verifyClaimStep`. I think there were 2 ways to solve this:
  1. assume `strategy` returns an infinite list and then `verifyClaimStep` can extract steps from it
  2.  add `strategyFirstStep` and `strategySecondStep` to the class but that would assume all claim types have at maximum 2 different steps (but it's very possible that we wouldn't want something more general anyway)

I opted for 1 since it seemed more general and easier to refactor. I can do 2 if that makes more sense, or maybe there is a better way.

kore/app/exec/Main.hs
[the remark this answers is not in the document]
ttuegel: I think that's OK, the goal of #1162 was to facilitate this change now. I don't think we can generalize `prove` beyond reachability logic, and that has only `OnePath` and `AllPath` claims.

kore/src/Kore/Strategies/Goal.hs
ttuegel: I agree, returning an infinite stream as the strategy makes sense!

kore/src/Kore/Step/Rule.hs
ana-pantilie: I don't think it can, or I just don't know how to do this with a `Proxy`. I need a case on that first argument, if it's `OnePathRule` apply the `OnePath` constructor and if it's an `AllPathRule` apply the `AllPath` constructor.

PR #1234, opened 2019-11-11
LTS Haskell 14.14

ttuegel:
---

###### Reviewer checklist


ttuegel:
> Is that change on the shell script only there to make the process faster by skipping tests/benchmarks?

That, and I forgot to have it build the benchmarks before.

PR #1237, opened 2019-11-11
Collect RTS statistics

ttuegel:
- Enable the `-T` GHC RTS flag by default for `kore-exec` and `kore-test` so that RTS statistics are always available. This does not seem to have significant overhead.
- Add a flag `--rts-statistics ⟨filename⟩` to `kore-exec` which collects the RTS statistics at the end of execution and writes them to the named file (JSON format).
- Use `--rts-statistics` and `KORE_EXEC_OPTS` to collect statistics on Jenkins, instead of using the `time` command. This gives better performance indicators because it does not include frontend time or memory usage (the latter is especially significant).

---

###### Reviewer checklist

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -381,25 +381,29 @@ main = do
     Foldable.forM_ (localOptions options) mainWithOptions
 
 mainWithOptions :: KoreExecOptions -> IO ()
-mainWithOptions execOptions@KoreExecOptions { koreLogOptions } = do
-    exitCode <- runLoggerT koreLogOptions
-        $ case () of
-            ()
-              | Just proveOptions <- koreProveOptions execOptions ->
-                koreProve execOptions proveOptions
-
-              | Just searchOptions <- koreSearchOptions execOptions ->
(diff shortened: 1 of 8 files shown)


kore/app/exec/Main.hs
```diff
@@ -370,20 +381,25 @@ main = do
...
 
 mainWithOptions :: KoreExecOptions -> IO ()
-mainWithOptions execOptions@KoreExecOptions { koreLogOptions } =
-    (=<<) exitWith $ runLoggerT koreLogOptions $ case () of
-    ()
-      | Just proveOptions <- koreProveOptions execOptions ->
-        koreProve execOptions proveOptions
+mainWithOptions execOptions@KoreExecOptions { koreLogOptions } = do
+    exitCode <- runLoggerT koreLogOptions
+        $ case () of
```
vladciobanu: I know it was here before, but instead of the case unit trick to get guards, can't we use a function in a where clause instead?

ttuegel: Done!

PR #1240, opened 2019-11-13
Handle IOException when we can't find 'z3'

### LINKED ISSUE #1131: Produce a helpful error message when Z3 is missing
Presently, the error produced when Z3 is missing is just an internal exception:

```
kore-exec: z3: runInteractiveProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

We should catch this error and re-throw a more useful exception.

vladciobanu:
Fixes #1131 

Does it make sense to check the `IOException` and make sure it's this specific error? I'm not sure it can fail in any other way. Maybe if the user doesn't have permission to run `z3`...? But I'm not sure how that would happen.

I also removed `newSolver` as an alternate way to run SMT queries since it was not used anywhere, thus making `runSMT` the only point of failure -- unless we use `SimpleSMT` directly.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/SMT.hs
+++ b/kore/src/SMT.hs
@@ -444,15 +444,16 @@ stopSolver mvar = do
 runSMT :: Config -> Logger -> SMT a -> IO a
 runSMT config logger SmtT { runSmtT } =
     Exception.bracket
-        (newSolver config logger `Exception.catch` formatError)
+        (newSolver config logger `Exception.catch` showZ3NotFound)
         stopSolver
         (runReaderT runSmtT)
   where
-    formatError :: Exception.IOException -> IO a
-    formatError _ =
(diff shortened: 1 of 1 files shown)


ttuegel:
> Does it make sense to check the `IOException` and make sure it's this specific error? I'm not sure it can fail in any other way. Maybe if the user doesn't have permission to run `z3`...? But I'm not sure how that would happen.

I think it's fine not to check for the exact error, but then we should print the exception with `displayException` in case it's _not_ what we expect here. Otherwise, I fear the day will come when someone reports a bug: "The backend says it can't find Z3, but I have it installed!"

Let's have it print the exception first, followed by something to the effect of: "We couldn't start Z3 due to the exception above. Is Z3 installed?"

PR #1248, opened 2019-11-15
WarnUnhookedSymbol

### LINKED ISSUE #1185: Reject sentences with `hook` attributes that are not `hooked-`
It is illegal to have a `hook` attribute on a sentence that is not `hooked-` and the backend will ignore the attribute, to the detriment and confusion of the user.

vladciobanu:
Fixes #1185 also see https://runtimeverification.slack.com/archives/CC360GUTG/p1573726962193100

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Function/Evaluator.hs
+++ b/kore/src/Kore/Step/Function/Evaluator.hs
@@ -28,9 +28,6 @@ import Control.Exception
 import qualified Control.Monad.Trans as Trans
 import qualified Data.Foldable as Foldable
 import Data.Function
-import Data.Functor
-    ( ($>)
-    )
 import qualified Data.Map as Map
 import Data.Text
     ( Text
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -121,6 +127,8 @@ evaluateApplication
           -- have a separate evaluator for startup.
           , (not . null) axiomIdToEvaluator
           = criticalMissingHook symbol hook
+          | (not . null) axiomIdToEvaluator
+          = warnUnhookedSymbol symbol $> unevaluated
```
ttuegel: This isn't going to do what we want. This simplifier is invoked whenever the function is left unevaluated. For a hook, that means the hook is missing, but for a user-defined symbol there could be many reasons we can't evaluate it yet. I think we should move both of these checks [here](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Function/Evaluator.hs#L262) so that we can directly check if an evaluator is available. It would probably be a good idea to extract that function, too.

PR #1254, opened 2019-11-15
Overloaded constructors and domain values

### LINKED ISSUE #1255: Cannot resolve inequality of overloaded constructors and domain values.
An overloaded constructor does not resolve to a domain value injected from a lower sort.

Perhaps related: because we do not apply the actual overloading axioms, sort injections may not be split correctly.

See also: #1254

ttuegel:
Fixes #1255 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/Overloading.hs
+++ b/kore/src/Kore/Step/Simplification/Overloading.hs
@@ -16,6 +16,9 @@ import Control.Exception
     ( assert
     )
 import qualified Control.Monad.Trans as Monad.Trans
+import Data.Text.Prettyprint.Doc
+    ( Doc
+    )
 import qualified GHC.Stack as GHC
 import Prelude hiding
     ( concat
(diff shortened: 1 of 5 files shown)


src/main/k/working/tests/overloads/domain-values/prove/domain-values-spec.k
```diff
@@ -0,0 +1,9 @@
+module VERIFICATION
+  imports DOMAIN-VALUES
+endmodule
+
+module DOMAIN-VALUES-SPEC
+  imports VERIFICATION
+  rule <k> 1 => end </k>
+  rule <k> (ctor a) => end </k>
```
traiansf: are you sure here `ctor a` is parsed as `Sub` and not directly as `Sup`?

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -83,19 +86,37 @@ overloadedAndEqualsOverloading
...
         empty
+  where
+    notUnifiableType :: TermLike variable -> Maybe (Doc ())
+    notUnifiableType (App_ appHead _)
+      | MetadataTools.isConstructorOrOverloaded tools appHead
+      = Just "different injected ctor"
+    notUnifiableType (DV_ _ _) = Just "injected domain value"
+    notUnifiableType (BuiltinBool_ _) = Just "injected builtin bool"
+    notUnifiableType (BuiltinInt_ _) = Just "injected builtin int"
+    notUnifiableType (BuiltinList_ _) = Just "injected builtin list"
```
traiansf: If one overloads the list constructor, couldn't they unify against a builtin list?

traiansf: actually, I think I know the answer to the question above: builtin list is List{K} and cannot be overloaded, right?

virgil-serbanuta: I was asking myself the same question, and I had roughly the same answer. But since we both doubted this, maybe it's safer to remove these and to add them only when needed.

src/main/k/working/tests/overloads/domain-values/prove/domain-values-spec.k
virgil-serbanuta: Better now?

PR #1271, opened 2019-11-19
Error for functions with ensures

### LINKED ISSUE #1193: Haskell:  Rewrite a term to a constrained symbolic variable
Whats the correct way to create a symbmolic Int bounded above and below, using the Haskell backend?I'm trying this:

```
module TEST
    imports INT                                                                 
    syntax Interval ::= "[" Int "," Int "]"                                     
                                                                                
    rule [ I , J ] => ?X:Int                                                    
      ensures ?X >=Int I                                                        
      andBool ?X <=Int J                                                        
endmodule 
```

But I see `kore-exec: Quantifying both the term and the predicate probably means that there's
an error somewhere else.`I tried requires as well

### LINKED ISSUE #1243: Functions with ensures
A function rule's `ensures` clause is currently, and incorrectly, ignored.

See also: #1193

andreiburdusa:
---
Fixes #1243 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Rule.hs
+++ b/kore/src/Kore/Step/Rule.hs
@@ -655,15 +655,20 @@ patternToAxiomPattern attributes pat =
                 }
         -- function axioms: general
         Implies_ _ requires (And_ _ (Equals_ _ _ lhs rhs) _ensures) ->
-            -- TODO (traiansf): ensure that _ensures is \top
-            pure $ FunctionAxiomPattern $ EqualityRule RulePattern
-                { left = lhs
-                , antiLeft = Nothing
-                , right = rhs
-                , requires = Predicate.wrapPredicate requires
-                , ensures = Predicate.makeTruePredicate
-                , attributes
-                }
+            case _ensures of
+                TermLike.Top_ _ ->
+                    pure $ FunctionAxiomPattern $ EqualityRule RulePattern
+                        { left = lhs
+                        , antiLeft = Nothing
+                        , right = rhs
+                        , requires = Predicate.wrapPredicate requires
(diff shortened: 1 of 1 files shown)


kore/src/Kore/Step/Rule.hs
```diff
@@ -655,15 +655,20 @@ patternToAxiomPattern attributes pat =
...
+                    pure $ FunctionAxiomPattern $ EqualityRule RulePattern
+                        { left = lhs
+                        , antiLeft = Nothing
+                        , right = rhs
+                        , requires = Predicate.wrapPredicate requires
+                        , ensures = Predicate.makeTruePredicate
+                        , attributes
+                        }
+                _ -> koreFail $ "Found ensures clause:\n"
+                        <> show _ensures
```
ttuegel: We used the initial underscore to prevent the compiler from complaining that `ensures` is not used, but now it is, so we can give this its proper name.

```suggestion
                        <> show ensures
```

kore/src/Kore/Step/Rule.hs
```diff
@@ -655,15 +655,20 @@ patternToAxiomPattern attributes pat =
...
+                TermLike.Top_ _ ->
+                    pure $ FunctionAxiomPattern $ EqualityRule RulePattern
+                        { left = lhs
+                        , antiLeft = Nothing
+                        , right = rhs
+                        , requires = Predicate.wrapPredicate requires
+                        , ensures = Predicate.makeTruePredicate
+                        , attributes
+                        }
+                _ -> koreFail $ "Found ensures clause:\n"
```
ttuegel: Let's use `unlines` instead of the escaped newlines.

PR #1273, opened 2019-11-20
Repl: can only try to apply claims of the same type

ana-pantilie:
Needs https://github.com/kframework/kore/pull/1232

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Repl/Interpreter.hs
+++ b/kore/src/Kore/Repl/Interpreter.hs
@@ -829,23 +829,22 @@ tryAxiomClaimWorker mode ref = do
         case eitherAxiomClaim of
             Left _ -> True
             Right claimToBeApplied ->
-                case castToReachabilityRule claimToBeApplied of
-                    Just reachClaimToBeApplied ->
-                        case castToReachabilityRule currentClaim of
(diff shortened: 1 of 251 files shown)


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -807,16 +809,44 @@ tryAxiomClaimWorker mode ref = do
...
-           case mode of
-               Never ->
-                   showUnificationFailure axiomOrClaim node
-               IfPossible ->
-                   tryForceAxiomOrClaim axiomOrClaim node
+        Nothing ->
+            putStrLn' "Could not find axiom or claim."
+        Just axiomOrClaim -> do
+            claim <- Lens.use (field @"claim")
+            if isReachabilityRule claim && not (equalClaimTypes axiomOrClaim claim)
```
vladciobanu: This could be `either (const True) (equalClaimTypes claim)`, and that would make `equalClaimTypes :: claim -> claim -> Bool` which makes more sense to me.

PS: `equalClaimTypes` could be implemented as
```
isSameType `on` castToReachability
```

PR #1284, opened 2019-11-22
Decode function axiom ensures clause

ttuegel:
See also: kframework/k#923

---

###### Reviewer checklist


ttuegel:
> Is this what I was supposed to do instead of #1271 ?

No, the plan was always to leave the error in place until someone reported a problem. That just happened much sooner than expected. :sweat_smile: It looks like there is some simplification rule (which we already rely on) from the frontend that uses an `ensures` clause.

PR #1285, opened 2019-11-22
Fix predicate sorts

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Internal/Conditional.hs
+++ b/kore/src/Kore/Internal/Conditional.hs
@@ -149,14 +149,14 @@ instance InternalVariable variable => Applicative (Conditional variable) where
     (<*>) predicated1 predicated2 =
         Conditional
             { term = f a
-            , predicate = Predicate.makeAndPredicate predicate1 p2
+            , predicate = Predicate.makeAndPredicate predicate1 sortedPredicate2
             , substitution = substitution1 <> substitution2
             }
       where
         Conditional f predicate1 substitution1 = predicated1
         Conditional a predicate2 substitution2 = predicated2
         sort = Predicate.predicateSort predicate1
-        p2 = Predicate.coerceSort sort predicate2
+        sortedPredicate2 = Predicate.coerceSort sort predicate2
 
 {- | 'Conditional' is equivalent to the 'Control.Comonad.Env.Env' comonad.
 
--- a/kore/src/Kore/Internal/Conditional.hs
+++ b/kore/src/Kore/Internal/Conditional.hs
@@ -130,7 +130,7 @@ instance
     mempty =
         Conditional
             { term = mempty
-            , predicate = Predicate.makeTruePredicate
+            , predicate = Predicate.makeTruePredicate_
             , substitution = mempty
             }
     {-# INLINE mempty #-}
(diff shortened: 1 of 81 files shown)


kore/test/Test/Kore/Step/Simplification/Equals.hs
```diff
@@ -805,9 +808,13 @@ test_equalsSimplification_TermLike =
             in
                 testCase "[a] `concat` x /\\ [a, b] "
                     (assertTermEquals
-                        (Condition.fromSingleSubstitution
-                            (ElemVar x, Mock.builtinList [Mock.b])
-                        )
+                        Conditional
+                            { term = ()
+                            -- TODO(virgil): This sort should be listSort.
```
ttuegel: Does this need to be fixed?

kore/src/Kore/Internal/Pattern.hs
```diff
@@ -218,9 +225,78 @@ fromTermLike term
   | otherwise =
     Conditional
         { term
-        , predicate = Predicate.makeTruePredicate
+        , predicate = Predicate.makeTruePredicate (termLikeSort term)
         , substitution = mempty
         }
 
+fromTermLikeUnsorted
```
ana-pantilie: `Unsorted` and `_` at the end of a function name mean the same thing, right?

virgil-serbanuta:
> Just confirming my own understanding: The new `Predicate` functions with `_` suffixes are predicates with `_PREDICATE` sort, right?

Yes, it follows the same convention as the mk* functions for terms.

kore/src/Kore/Internal/Pattern.hs
virgil-serbanuta: Yes, though I find the "Unsorted" thing to be somewhat more readable. I used "_" for the predicate stuff since it had a clear relation to the mk* term constructors, but I wasn't sure that here it's similarly clear. If you think I should use underscores, I will.

kore/src/Kore/Internal/Conditional.hs
```diff
@@ -142,19 +142,21 @@ instance InternalVariable variable => Applicative (Conditional variable) where
...
             { term = f a
-            , predicate = Predicate.makeAndPredicate predicate1 predicate2
+            , predicate = Predicate.makeAndPredicate predicate1 p2
             , substitution = substitution1 <> substitution2
             }
       where
         Conditional f predicate1 substitution1 = predicated1
         Conditional a predicate2 substitution2 = predicated2
+        sort = Predicate.predicateSort predicate1
+        p2 = Predicate.coerceSort sort predicate2
```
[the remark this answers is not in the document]
virgil-serbanuta: Done.

kore/test/Test/Kore/Step/Simplification/Equals.hs
virgil-serbanuta: Maybe, but fixing it properly would require fixing this in the entire simplifier code, as opposed to fixing it in `Simplifier/TermLike.hs`, as it's done now. This may be a large change in itself, so I'm hesitant to do it now. How about doing it in the next PR if you think it's worth having consistent output from the individual simplifiers?

kore/src/Kore/Internal/Pattern.hs
ana-pantilie: I understand, makes sense in this context. I was just reviewing the code and was confused why the convention changed, but as you say, we don't have such a convention for `Pattern`.

virgil-serbanuta: It turned out that, after fixing all the broken tests, this was used only once in the entire project, so I removed it.

kore/test/Test/Kore/Step/Simplification/Equals.hs
ttuegel: Sounds good!

PR #1293, opened 2019-11-26
Remove destination with configuration splitting

ana-pantilie:
https://github.com/kframework/kore/pull/1300 should be merged before this.

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -156,41 +167,73 @@ simplifyEvaluated variable simplified
...
-            matched <- Monad.Trans.lift $ matchesToVariableSubstitution
-                variable
-                normalized { Conditional.substitution = boundSubstitution }
-            if matched
-                then return normalized
-                    { Conditional.predicate = Predicate.makeTruePredicate_
-                    , Conditional.substitution = freeSubstitution
-                    }
-                else makeEvaluateBoundRight
+makeEvaluate variables original = do
```
ttuegel: Is the primary change here that we are simplifying multiple quantifiers at once? (The list of variables.)

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -731,22 +741,29 @@ allPathFollowupStep claims axioms =
...
-    pure $ makeRuleFromPatterns goal result destination
+    removal <- removalPredicate (destination right') configuration
+    let result = Conditional.andPredicate configuration removal
+    pure $ makeRuleFromPatterns goal result (Pattern.fromTermLike right')
   where
     configuration = getConfiguration goal
     configFreeVars = Pattern.freeVariables configuration
 
-    RulePattern { right, ensures } = toRulePattern goal
+    RulePattern { right } = toRulePattern goal
```
ttuegel: Should we check that `ensures` is `\top`?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -921,39 +938,88 @@ deriveSeq rules goal = errorBracket $ do
...
+                                Exists.makeEvaluate extraElemVariables remainderPattern
+                            return
+                                . Predicate.makeNotPredicate
+                                . Condition.toPredicate
+                                . Conditional.withoutTerm
+                                . OrPattern.toPattern
+                                $ evaluatedRemainder
+                _ -> error . show . Pretty.vsep $
+                    [ "Unifying the terms of the configuration and the\
+                      \ destination has unexpectedly produced more than one\
```
ttuegel: Indentation should be a multiple of four.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -921,39 +938,88 @@ deriveSeq rules goal = errorBracket $ do
...
+    destination
+    configuration
+  | isFunctionPattern configTerm
+  , isFunctionPattern destTerm
+  = do
+    unifiedConfigs <- simplifyTerm (mkAnd configTerm destTerm)
+    if OrPattern.isFalse unifiedConfigs
+        then return Predicate.makeTruePredicate_
+        else
+            case Foldable.toList unifiedConfigs of
```
ttuegel:
```suggestion
            case OrPattern.toPatterns unifiedConfigs of
```

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -921,39 +938,88 @@ deriveSeq rules goal = errorBracket $ do
...
+                [substPattern] ->
+                    let extraNonElemVariables =
+                            remainderNonElemVariables configuration substPattern
+                        extraElemVariables =
+                            remainderElementVariables configuration substPattern
+                    in if not (null extraNonElemVariables)
+                        then
+                            error . show . Pretty.vsep $
+                                "Cannot quantify non-element variables: "
+                                : fmap (Pretty.indent 4 . unparse) extraNonElemVariables
```
ttuegel: Let's factor this branch out into a function which returns `extraElemVariables` or throws an error if there are non-element variables. Then, we only need to write the "happy path" here.

kore/test/Test/Kore/Strategies/AllPath/Verification.hs
```diff
@@ -298,7 +303,15 @@ simpleClaim
     -> TermLike Variable
     -> AllPathRule Variable
 simpleClaim left right =
-    AllPathRule . getRewriteRule $ simpleRewrite left right
+    AllPathRule
+    $ RulePattern
```
ttuegel: Record brackets bind stronger than application:

```suggestion
    RulePattern
```

kore/test/Test/Kore/Strategies/OnePath/Verification.hs
```diff
@@ -345,7 +344,15 @@ simpleClaim
     -> TermLike Variable
     -> OnePathRule Variable
 simpleClaim left right =
-    OnePathRule . getRewriteRule $ simpleRewrite left right
+    OnePathRule
+    $ RulePattern
```
ttuegel:
```suggestion
    RulePattern
```

kore/src/Kore/Step/Simplification/Exists.hs
ana-pantilie: Yes, that should be it. I updated the documentation here as @traiansf suggested.

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -156,41 +167,76 @@ simplifyEvaluated variable simplified
...
+    let sortedVariables = sortBy substVariablesFirst variables
```
virgil-serbanuta: Shouldn't we resimplify the predicate and then resort the variables after substituting the ones that we can? Alternately, we could call `simplifyCondition` before sorting, which may extract some substitutions that are "pendind" inside the predicate, e.g. a postponed map lookup (I don't remember if we still have those).

ana-pantilie: @virgil-serbanuta Does it matter if we call `simplifyCondition` before or after sorting the variables? We don't use the variables when simplifying the condition, and the (newly-formed) condition always gets simplified at the beginning of `makeEvaluateWorker`. Have I misunderstood what you're saying?

virgil-serbanuta: My question is something like this: can the `simplifyCondition original'` call in `makeEvaluateWorker` add more variables to the substitution? If yes, should that change the sorting order?

As an example, if our `original` condition is `C(x)=f(y) /\ [y=5]`, where `C` is a constructor, `f` is a function and `f(5)=C(A)`, then should we simplify that to `[x=A, y=5]` before sorting, or we should consider `x` as not being a substituted variable?

Now, the above will not be a valid example after https://github.com/kframework/kore/pull/1320 , but you can ask the same about `C(A)=f(x,y) and [y=5]` where `f(x, 5)=C(x)`.

ana-pantilie: Oh, I see. Thank you for the explanation! I'll try to fix this in a follow-up PR.

Issue #1296, opened 2019-11-26
Hold symbol and alias arguments in arrays instead of lists


ttuegel:
A linked list uses `3n+1` pointers to hold `n` elements, whereas an array uses `n+1` pointers. This improves memory use slightly.

PR #1297, opened 2019-11-26
Hold symbol arguments in arrays instead of lists

ttuegel:
Using arrays reduces overall memory use and time spent in the garbage
collector.

A `newtype` wrapper `Arguments` is added to handle pattern matching. Most of the changes below are related to adding the wrapper. The interesting part is `Kore.Syntax.Arguments`.

Fixes: #1296

---

###### Reviewer checklist


ttuegel:
This improves concrete execution slightly, but it makes symbolic execution slightly worse in some cases. (In all cases, we are looking at <5% difference either way.) I'm not going to merge this because there is no benefit to the added complexity.

PR #1300, opened 2019-11-28
Existentials

traiansf:
Fixes: #1193

---

###### Reviewer checklist


traiansf:
Note to reviewers: it might be helpful do review this commit by commit, as, except for one commit, all others are refactorings.

kore/src/Kore/Step/Rule.hs
```diff
@@ -821,6 +825,13 @@ axiomPatternToPattern
   =
     TermLike.mkImplies left right
 
+errorEnsures :: InternalVariable variable => Predicate variable -> a
```
virgil-serbanuta: How about a TODO to use a different representation for the RulePattern so that this error is not needed anymore?

kore/test/Test/Kore/Step/Step.hs
```diff
@@ -41,19 +41,19 @@ import Kore.Internal.TermLike
 import qualified Kore.Step.Result as Result
     ( mergeResults
     )
+import Kore.Step.RewriteStep hiding
```
virgil-serbanuta: Why `hiding` instead of importing what you need?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -685,6 +686,29 @@ refreshVariables
...
+    -> TermLike variable
+freeTopExists
+    (FreeVariables.getFreeVariables -> avoid)
+    term
+  =
+    substitute subst termWithoutExists
+  where
+    (termWithoutExists, exists) = removeTopExists term Set.empty
+    originalFreeVariables = FreeVariables.getFreeVariables (freeVariables term)
+    rename = Fresh.refreshVariables (avoid <> originalFreeVariables) exists
```
virgil-serbanuta: Having a type for `rename` would make the `<$>` below easier to follow.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -739,8 +740,13 @@ removeDestination goal = errorBracket $ do
...
     pure $ makeRuleFromPatterns goal result destination
   where
-    destination = getDestination goal
     configuration = getConfiguration goal
+    configFreeVars = Pattern.freeVariables configuration
+
+    RulePattern { right, ensures } = toRulePattern goal
+    right' = topExistsToImplicitForall configFreeVars right
+    destination =
+        Pattern.withCondition right' (Conditional.fromPredicate ensures)
```
ana-pantilie: I think I already asked you, so sorry for asking again, but why do we keep the ensures here? Also this applies to the function `getDestination` as well (it's at the bottom of this file and we use it in several places in this module and in the `Verification` module as well I think).

ana-pantilie:
I think keeping the `ensures` in the type is error-prone, or at least I find it kind of confusing. If I understand this PR correctly, now if I want to get the _real_ `ensures` clause I actually need to do some further processing of the RHS of a rule, and this isn't obvious only by looking at the types and the functions we currently use to operate on `RulePattern`s.

traiansf:
I agree.  In my opinion the best option would be to have different types for Equational rules (with ensures in the record) than Rewrite rules/claims (without ensures in the record, but embedded in the RHS).

traiansf:
> I'm still not happy about the lack of tests for the new/duplicated code. However, I don't know @ttuegel 's plans, so maybe new tests do not make sense. If you're not in a hurry, I would wait for his feedback.

@virgil-serbanuta I've added tests.

PR #1308, opened 2019-12-02
Internal representation of sort injections

ttuegel:
An internal representation of sort injections to modularize simplification and
evaluation of the same.

~After: #1307~

The primary purpose of this refactoring is to make the bug fix in #1316 small.

---

###### Reviewer checklist


kore/src/Kore/Internal/Inj.hs
```diff
@@ -0,0 +1,143 @@
...
+        { injConstructor :: !Id
+        , injFrom        :: !Sort
+        , injTo          :: !Sort
+        , injAttributes  :: !Attribute.Symbol
+        , injChild       :: !a
+        }
+    deriving (GHC.Generic, Show)
+    deriving (Functor, Foldable, Traversable)
+
+instance Eq a => Eq (Inj a) where
```
virgil-serbanuta: Below I'll assume that you're not comparing attributes in order to make things faster (if not, then what's the actual reason?).

Is it reasonable to exclude all attributes from equality testing? The synthetic ones and the ones determined by the symbol definition should be fine, should we add a comment to `Attribute.Symbol` that we should pay attention to that? I.e. that we should not add one which "mutable" in a way similar to the `simplified` bit?

As a general solution, maybe we should also compare attributes, but we should have a special comparison function which ignores the part that's uniquely determined (i.e. for `Attribute.Symbol` it should always return that the operands are equal). Whenever we change the attribute, we would also be forced to change the comparison function, so we would make a decision about that change.

kore/src/Kore/Internal/Inj.hs
```diff
@@ -0,0 +1,143 @@
...
+    (==) a b =
```
virgil-serbanuta: Here and below, why not match on `(Inj _ _ _ _ _)` in the same way as you do for `toSymbol`?

kore/src/Kore/Step/Simplification/InjSimplifier.hs
```diff
@@ -0,0 +1,135 @@
...
+      | otherwise = injFrom `isSubsortOf'` injTo
+
+    evaluateInj
+        :: forall variable
+        .  GHC.HasCallStack
+        => InternalVariable variable
+        => Inj (TermLike variable)
+        -> TermLike variable
+    evaluateInj inj
+      -- _| not (isOrderedInj inj) = unorderedInj inj
```
virgil-serbanuta: Why these comments?

kore/src/Kore/Step/Simplification/Inj.hs
```diff
@@ -0,0 +1,51 @@
...
+    ( InjSimplifier (..)
+    )
+import Kore.Step.Simplification.Simplify as Simplifier
+
+{- |'simplify' simplifies an 'Inj' of 'OrPattern'.
+
+-}
+simplify
+    :: (SimplifierVariable variable, MonadSimplify simplifier)
+    => Condition variable
```
virgil-serbanuta: I guess it's fine if you want to leave this here, but we have simplifiers that do not take a `Condition` (e.g. `Top.simplify`), so you won't have a uniform interface for all simplifiers anyway, and removing it would mean that this function takes as input only what's actually needed.

kore/src/Kore/Step/Simplification/Inj.hs
```diff
@@ -0,0 +1,51 @@
+{- |
```
virgil-serbanuta: It would be nice to have unit tests for this.

kore/src/Kore/Builtin/Inj.hs
```diff
@@ -0,0 +1,87 @@
+{- |
```
virgil-serbanuta: It would be nice to have unit tests for this.

kore/src/Kore/Internal/Inj.hs
ttuegel: > As a general solution, maybe we should also compare attributes, but we should have a special comparison function which ignores the part that's uniquely determined

I considered having the `Eq` instance `assert` that the attributes are the same if the symbol names are the same. I only didn't do that because the `Eq Symbol` instance doesn't do that.

Issue #1270, opened 2019-11-19
RemoveDestination should try unification first


ttuegel:
@yzhang90 points out that the `RemoveDestination` predicate has the form `¬⌈t₁ ∧ t₂⌉ ∨ ¬(P₁ ∧ P₂)` where `tᵢ` are the terms and `Pᵢ` are the path conditions, respectively. If the unification of terms `⌈t₁ ∧ t₂⌉` is `\bottom` then we know removal predicate is `\top` without evaluating the path conditions. This is important because computing `⌈t₁ ∧ t₂⌉` is usually much faster than computing `¬(P₁ ∧ P₂)`, and the unification is _almost always_ `\bottom`. At present, we construct a single predicate from both terms and conditions and simplify the entire thing. Instead, we should check the unification of terms first, and proceed with checking the conditions only if the unification succeeds.

This is similar to the boundary pattern check (#1262) that the Java backend has: avoid checking the final implication until you are sure rewriting is done. We believe that this is more general and that by making this smaller change, we may avoid needing to implement boundary pattern matching. If the unification of terms succeeds, we should check implication of conditions, and we should _not_ continue rewriting.

See also: #1262

---

- [x] Check unification of terms before implication of path conditions. (https://github.com/kframework/kore/pull/1293)
- [ ] Once the terms unify, do not continue rewriting. If the condition does not imply the goal, the proof is "stuck". (https://github.com/kframework/kore/pull/1311)

PR #1311, opened 2019-12-03
Decide in removeDestination if proof is stuck

ana-pantilie:
Needs https://github.com/kframework/kore/pull/1293

Fixes https://github.com/kframework/kore/issues/1270

- Adds a new `ProofState` and a new `Prim` constructor
- `removeDestination` now handles state transitions directly since doing so in `transitionRuleTemplate` would now make it depend on the representation of goals
- removes `Simplify` after `RemoveDestination` since `removeDestination` now does simplification as well
- The modified/added tests in `Kore.Strategies.OnePath.Step` should illustrate the desired behavior of this check

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -743,21 +747,55 @@ allPathFollowupStep claims axioms =
...
-        let result = Conditional.andPredicate configuration removal
-        pure $ makeRuleFromPatterns goal result (Pattern.fromTermLike right')
+        if isTop removal
+            then return . stateConstructor $ goal
+            else do
+                simplifiedRemoval <-
+                    SMT.Evaluator.filterMultiOr
+                    =<< simplifyAndRemoveTopExists
+                        (Conditional.andPredicate configuration removal)
+                if not (null simplifiedRemoval)
```
ttuegel: Is `simplifiedRemoval` an `OrPattern`? It would be better to use `isBottom` instead of `null`.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -743,21 +747,55 @@ allPathFollowupStep claims axioms =
...
+                    then return . GoalStuck $ goal
+                    else
```
ttuegel: What does this branch return, when `simplifiedRemoval` is `\bottom` (that is, when the right-hand side is implied by the current left-hand side)?

ana-pantilie: The `else` branch returns a `goal` with `bottom` on the LHS.

ttuegel: Why not just `Proven`?

ana-pantilie: Well we have the `TriviallyValid` step right after, and we do the same in `simplify`, although I guess what you're suggesting make sense here since we're defining the transitions directly in `removeDestination` instead of `transitionRuleTemplate`. I'll change it, thanks!

PR #1312, opened 2019-12-03
Configurable proof search order

### LINKED ISSUE #1264: Configurable proof search order
By default, proof search is breadth-first, but we also support a depth-first mode for the bounded model checker. We would like a command-line option to select the search order for reachability proofs, too; this is useful for debugging.

---

- [ ] `--search-order=breadth-first` (no behavior change, just add the option)
- [ ] `--search-order=depth-first`

andreiburdusa:
---
Fixes #1264 
###### Reviewer checklist


kore/test/Test/Kore/Strategies/Common.hs
```diff
@@ -46,6 +49,7 @@ runVerification stepLimit axioms claims =
     runSimplifier mockEnv
     $ runExceptT
     $ Verification.verify
+        BreadthFirst
```
andreiburdusa: I'm not sure if I should add a `GraphSearchOrder` parameter to `runVerification` or live `BreadthFirst` here

andreiburdusa:
It looks like the search order option was already parsed (`graph-search`)
https://github.com/kframework/kore/blob/master/kore/app/share/GlobalMain.hs#L157
but it was only used in the case of bmc
https://github.com/kframework/kore/blob/master/kore/app/exec/Main.hs#L446

kore/test/Test/Kore/Strategies/Common.hs
ttuegel: No, I think this is fine. The test was written for `BreadthFirst` originally.

PR #1319, opened 2019-12-04
Limit number of concurrent branches

### LINKED ISSUE #1265: Limit number of concurrent branches
A useful debugging option is to limit the number of concurrent branches in a proof. Basically, we are limiting the size of the frontier set. This is extremely important when debugging loop invariants.

---

- [ ] `--breadth LIMIT` option (opposite `--depth`)

andreiburdusa:
---
Fixes #1265 
###### Reviewer checklist


kore/app/exec/Main.hs
```diff
@@ -247,7 +247,8 @@ data KoreExecOptions = KoreExecOptions
     , smtTimeOut          :: !SMT.TimeOut
     , smtPrelude          :: !(Maybe FilePath)
     , smtSolver           :: !Solver
-    , stepLimit           :: !(Limit Natural)
+    , breadthLimit          :: !(Limit Natural)
```
ttuegel:
```suggestion
    , breadthLimit        :: !(Limit Natural)
```

kore/app/exec/Main.hs
```diff
@@ -329,6 +332,12 @@ parseKoreExecOptions =
             , ("any-heating-cooling", heatingCooling anyRewrite)
             , ("all-heating-cooling", heatingCooling allRewrites)
             ]
+    breadth =
+        option auto
+            (  metavar "BREADTH"
+            <> long "breadth"
+            <> help "Execute up to BREADTH steps."
```
ttuegel:
```suggestion
            <> help "Allow up to BREADTH parallel execution branches."
```

kore/src/Kore/Exec.hs
```diff
@@ -220,20 +221,21 @@ execGetExitCode
...
     -> smt ExitCode
-execGetExitCode indexedModule strategy' finalTerm =
+execGetExitCode breadthlimit indexedModule strategy' finalTerm =
     case resolveInternalSymbol indexedModule $ noLocationId "LblgetExitCode" of
         Nothing -> return ExitSuccess
         Just mkExitCodeSymbol -> do
             exitCodePattern <-
                 -- TODO (thomas.tuegel): Run in original execution context.
-                exec indexedModule strategy'
+                exec breadthlimit indexedModule strategy'
```
ttuegel: The `breadthLimit` here should be hardcoded to `1`. Execution must not branch when we are choosing the exit code.

kore/src/Kore/Step/Strategy.hs
```diff
@@ -414,6 +421,8 @@ constructExecutionGraph transit instrs0 searchOrder0 config0 = do
       | []              <- instrs = unfoldWorker rest searchOrder
       | instr : instrs' <- instrs
       = Profile.executionQueueLength (Seq.length rest) $ do
+        when ((not . withinLimit depthLimit . fromIntegral . Seq.length) rest)
```
ttuegel: Let's factor out this error into a helper function. We should also `unparse` the branches for the error message.

kore/test/Test/Kore/Step/Strategy.hs
```diff
@@ -232,8 +232,8 @@ test_Or =
         ]
```
ttuegel:
```suggestion
```

kore/src/Kore/Step/Strategy.hs
```diff
@@ -368,6 +376,11 @@ emptyExecutionGraph config =
 -}
 data GraphSearchOrder = BreadthFirst | DepthFirst deriving Eq
 
+data LimitExceeded = LimitExceeded | Other
```
ttuegel: What is `Other` for? It would also be nice if the exception type captured the execution queue when it is thrown.

andreiburdusa: I know I did an ugly thing, but I did this in order to distinguish between the failing of the function and the failing of the test caused by the success of the function. I mostly pushed this to save my work. It was the only way I was able to make it work in almost 6 hours. I'll try to come with a better solution

andreiburdusa: I think I already have a better idea :))

ttuegel: It looks like you figured this out below, but you can `catch` or `handle` the `LimitExceeded` exception and use `assertFailure` if the exception is not thrown.

kore/src/Kore/Step.hs
```diff
@@ -66,6 +66,7 @@ import Kore.Unparser
 {- | A strategy primitive: a rewrite rule or builtin simplification step.
  -}
 data Prim rewrite = Simplify | Rewrite !rewrite
+        deriving (Show)
```
ttuegel:
```suggestion
    deriving (Show)
```

kore/test/Test/Kore/Exec.hs
```diff
@@ -113,6 +124,78 @@ test_search =
...
+
+    assertion searchType =
+        shouldExeedBreadthLimit searchType `catch`
+            \(_ :: LimitExceeded (Strategy (Prim Rewrite))) -> pure ()
+
+    shouldExeedBreadthLimit :: SearchType -> IO ()
+    shouldExeedBreadthLimit searchType = do
+        a <- actual searchType
+        when (a == expected searchType)
+            $ assertFailure "Did not exceed bredth limit"
```
ana-pantilie:
```suggestion
            $ assertFailure "Did not exceed breadth limit"
```

Issue #1256, opened 2019-11-15
Missing unification cases no longer produce errors


ttuegel:
At one time, it was an error if a particular unification case was not recognized when applying rules. Now it seems that we promote unknown cases to equalities in a (likely) futile attempt to continue.

See also: #1255

virgil-serbanuta:
Technically speaking, we still produce errors when we don't know how to unify things. However, we have a fallback case for unifying functional stuff, so, in practice, we think we know how to all things that we actually see in our input files.

Producing an error for this case does not work out of the box, since we have rules that (roughly) work like this:
```if true then A else B => A```
which we may try to apply to
```if (X <Int Y) then something else somethingelse```
and this will crash at unification time, because we don't know how to handle `true` vs `X <Int Y`.
To see an example of this, for now you can run the `src/main/k/working/smc` test with this PR: https://github.com/kframework/kore/pull/1302 , though the PR may change or disappear in the future.

Possible solutions:
1. Maybe it's not that bad to get an error after unification, saying that the axiom does not cover all the variables - searching the predicate for an `equals` containing the said variable is almost equivalent to getting an error about an unknown unification case. We could even have some heuristics in our error code doing that for the user.
2. Whitelist various cases in which it's ok to return equalities, e.g. when the term coming from the rule is a concrete one. This does not solve all issues (e.g. `C(X)` in the rule vs `Y` in the term would still be accepted, though that might be solvable). Also, it's not a big issue, but whoever does this should also pay attention to argument order swapping in the unifier, e.g. here: https://github.com/kframework/kore/blob/master/kore/src/Kore/Builtin/AssociativeCommutative.hs#L820 . Or we could blacklist stuff, of course.
3. Make the semantics to not use, say, domain values in rules. Instead, we should use variables, and we should use the domain value in the `requires` clause. That could be done automatically when compiling the `k` rules to `kore`.

I like 1 the most, then 3 if we do it at compilation time, then 2, then 3 if we do it manually. All of them should work, and all have some unknown elements that could surprise us (though 1 can't really go wrong, it can only be less useful than people hope).

virgil-serbanuta:
@ttuegel I'll unassign this bug since it's Thanksgiving, which means that I can't ask the people for which this was annoying to see if item 1 would be good enough for them.

ttuegel:
> We could even have some heuristics in our error code doing that for the user.

We could probably identify subterms of the conditions involving the missing variables.

[...]

PR #1320, opened 2019-12-04
functionAnd: Restrict to function symbols

ttuegel:
Because we do not evaluate functions during unification, we sometimes have to
emit an Equals predicate between two patterns that we will evaluate later. This
suppresses many errors surrounding missing unification cases. We will restrict
this fallback to the case of And between a function symbol and a
non-simplifiable pattern. This allows us to handle cases like `f(x) ∧ 1` without
suppressing errors for cases like `f(x) ∧ f(y)`.

Fixes: #1256

This turns the `wrc20-spec.k` error into:

```
kore-exec: Not implemented error:
    Unsupported patterns: Unknown unification case.
    first = 
        Lbl'Unds'ModuleInstCellMap'Unds'{}(
            LblModuleInstCellMapItem{}(
                Lbl'-LT-'modIdx'-GT-'{}(VarCUR0:SortInt{}),
                Lbl'-LT-'moduleInst'-GT-'{}(
                    Lbl'-LT-'modIdx'-GT-'{}(VarCUR0:SortInt{}),
                    Var'Unds'101:SortExportsCell{},
                    Lbl'-LT-'typeIds'-GT-'{}(VarTYPEIDS:SortMap{}),
                    Lbl'-LT-'types'-GT-'{}(VarTYPES0:SortMap{}),
                    Var'Unds'111:SortNextTypeIdxCell{},
                    Lbl'-LT-'funcIds'-GT-'{}(VarIDS:SortMap{}),
                    Lbl'-LT-'funcAddrs'-GT-'{}(VarADDRS:SortMap{}),
                    Lbl'-LT-'nextFuncIdx'-GT-'{}(VarNEXTIDX:SortInt{}),
                    Var'Unds'121:SortTabIdsCell{},
                    Var'Unds'131:SortTabAddrsCell{},
                    Var'Unds'141:SortMemIdsCell{},
                    Var'Unds'151:SortMemAddrsCell{},
                    Var'Unds'161:SortGlobIdsCell{},
                    Var'Unds'171:SortGlobalAddrsCell{},
                    Var'Unds'181:SortNextGlobIdxCell{}
                )
            ),
            VarDotVar6:SortModuleInstCellMap{}
        )
    second = 
        /* builtin: */
        Lbl'Unds'ModuleInstCellMap'Unds'{}(
            /* element: */ LblModuleInstCellMapItem{}(
                Lbl'-LT-'modIdx'-GT-'{}(VarCUR:SortInt{}),
                Lbl'-LT-'moduleInst'-GT-'{}(
                    Lbl'-LT-'modIdx'-GT-'{}(VarCUR:SortInt{}),
                    Lbl'-LT-'exports'-GT-'{}(Var'Unds'160:SortMap{}),
                    Lbl'-LT-'typeIds'-GT-'{}(Var'Unds'170:SortMap{}),
                    Lbl'-LT-'types'-GT-'{}(VarTYPES:SortMap{}),
                    Lbl'-LT-'nextTypeIdx'-GT-'{}(VarNEXTTYPEIDX:SortInt{}),
                    Lbl'-LT-'funcIds'-GT-'{}(Var'Unds'1:SortMap{}),
                    Lbl'-LT-'funcAddrs'-GT-'{}(Var'Unds'3:SortMap{}),
                    Lbl'-LT-'nextFuncIdx'-GT-'{}(VarNEXTFUNCIDX:SortInt{}),
                    Lbl'-LT-'tabIds'-GT-'{}(Var'Unds'180:SortMap{}),
                    Lbl'-LT-'tabAddrs'-GT-'{}(Var'Unds'190:SortMap{}),
                    Lbl'-LT-'memIds'-GT-'{}(Var'Unds'200:SortMap{}),
                    Lbl'-LT-'memAddrs'-GT-'{}(
                        /* builtin: */
                        /* concrete element: */ Lbl'UndsPipe'-'-GT-Unds'{}(
                            inj{SortInt{}, SortKItem{}}(
                                /* builtin: */ \dv{SortInt{}}("0")
                            ),
                            inj{SortInt{}, SortKItem{}}(VarMEMADDR:SortInt{})
                        )
                    ),
                    Lbl'-LT-'globIds'-GT-'{}(Var'Unds'210:SortMap{}),
                    Lbl'-LT-'globalAddrs'-GT-'{}(Var'Unds'220:SortMap{}),
                    Lbl'-LT-'nextGlobIdx'-GT-'{}(Var'Unds'230:SortInt{})
                )
            ),
            /* opaque child: */ VarDotVar7:SortModuleInstCellMap{}
        )
while applying a \rewrite axiom to the pattern:
...
```

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -70,7 +70,7 @@ import Kore.Internal.Predicate
     )
 import qualified Kore.Internal.Predicate as Predicate
 import qualified Kore.Internal.Symbol as Symbol
-import Kore.Internal.TermLike
+import Kore.Internal.TermLike as TermLike
 import qualified Kore.Logger as Logger
 import Kore.Step.Simplification.ExpandAlias
     ( expandAlias
@@ -831,15 +831,29 @@ functionAnd
     -> TermLike variable
     -> Maybe (Pattern variable)
 functionAnd first second
-  | isFunctionPattern first, isFunctionPattern second =
-    return Conditional
-        { term = first  -- different for Equals
+  | not (isFunctionPattern first) || not (isFunctionPattern second) = empty
+  | App_ symbol _ <- first
+  , Symbol.isDeclaredFunction symbol
+  , TermLike.isNonSimplifiable second
+  , isFunctionPattern first, isFunctionPattern second
+  =
+    return (Pattern.withCondition second condition)
+  | App_ symbol _ <- second
+  , Symbol.isDeclaredFunction symbol
+  , TermLike.isNonSimplifiable first
+  =
+    return (Pattern.withCondition first condition)
+  | App_ symbol1 _ <- first
+  , App_ symbol2 _ <- second
+  , Symbol.isDeclaredFunction symbol1
+  , Symbol.isDeclaredFunction symbol2
+  =
+    return (Pattern.withCondition first condition)
+  | otherwise = empty
+  where
+    condition =
         -- Ceil predicate not needed since first being
         -- bottom will make the entire term bottom. However,
         -- one must be careful to not just drop the term.
-        , predicate =
-            Predicate.markSimplified
-            $ makeEqualsPredicate_ first second
-        , substitution = mempty
-        }
-  | otherwise = empty
+        Condition.fromPredicate . Predicate.markSimplified
+        $ makeEqualsPredicate_ first second
--- a/kore/test/Test/Kore/Unification/Unifier.hs
+++ b/kore/test/Test/Kore/Unification/Unifier.hs
@@ -190,8 +190,9 @@ andSimplifySuccess
     -> Assertion
 andSimplifySuccess term1 term2 results = do
     let expect = map unificationResult results
-    Right subst' <-
-        runSMT
+    subst' <-
+        fmap (either (error . show) id)
+        $ runSMT
         $ runSimplifier testEnv
         $ Monad.Unify.runUnifierT
         $ simplifyAnds (unificationProblem term1 term2 :| [])


kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -831,15 +831,23 @@ functionAnd
...
+  | not (isFunctionPattern first) || not (isFunctionPattern second) = empty
+  | App_ symbol _ <- first
+  , Symbol.isDeclaredFunction symbol
+  , TermLike.isNonSimplifiable second
+  , isFunctionPattern first, isFunctionPattern second
+  =
+    return (Pattern.withCondition second condition)
+  | App_ symbol _ <- second
+  , Symbol.isDeclaredFunction symbol
+  , TermLike.isNonSimplifiable first
```
virgil-serbanuta: why is there an asymmetry between this case and the above? Above you are checking that the second is non-simplifiable and both the first and the second are non-simplifiable, here you are checking that the first is non-simplifiable and (implicitly) that at least one of the two is functional.

Also, both check that the top symbol for the perhaps-simplifiable term is a function. This is also mentioned in the PR description, but why would one check only the top symbol and not the entire term?

I mean, I know that this is a heuristic, but I'd like to understand it.

Something else that's confusing: I didn't yet figure out why this works for the example you provide in the PR description. I would have thought that `Lbl'Unds'ModuleInstCellMap'Unds'`, the top symbol in the first term, is a function symbol (or should be one, i.e. we should fix it if it isn't), both terms are functional, and the second one is non-simplifiable (since it's a builtin). If so, then it should match the second guard in this function, so we should not get an error.

virgil-serbanuta: Ok, I think I found the answer to the last question - variables are not non-simplifiable.

ttuegel:
We rely on this unification case to handle, for example, `(A, A |-> B) ∧ (X, X |-> V _)`. So, I don't think we can do this.

Issue #1317, opened 2019-12-03
Show TypeRep in log entry


yzhang90:
I found that in the log, the scope information is not shown anymore, I randomly picked some of the entry:
```
[2019-12-03 16:22:17.477785115][Debug] : [Debug] : (assert (and <0> (= (and (<= 0 <1> ) (< <1> 1461501637330902918203684832716283019655932542976 ) ) true ) ) ) [CallStack (from HasCallStack):
  debug, called at src/SMT/SimpleSMT.hs:316:5 in kore-0.0.1.0-HZLca9freouSbuMEw9GYn:SMT.SimpleSMT]
```
the second (or third) column used to display scope information. Now it only shows `Debug`.

---

- [ ] Print `Severity` only once; remove it from [here](https://github.com/kframework/kore/blob/master/kore/src/Kore/Logger/Output.hs#L275)
- [ ] In place of the duplicate `Severity`, display the [`typeOf`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Type-Reflection.html#v:typeOf) the entry, in lieu of the scope.
- [ ] `DebugSolverSend`
- [ ] `DebugSolverRecv`

PR #1325, opened 2019-12-05
Show TypeRep in log entry

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1317 + removes `Logger.WithScope`

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Logger/DebugAppliedRule.hs
+++ b/kore/src/Kore/Logger/DebugAppliedRule.hs
@@ -174,8 +174,4 @@ filterDebugAppliedRule debugAppliedRuleOptions baseLogAction logAction =
 
 matchDebugAppliedRule :: SomeEntry -> Maybe DebugAppliedRule
 matchDebugAppliedRule entry =
-    fromEntry entry <|> throughScope
-  where
-    throughScope = do
-        entry' <- fromEntry entry
-        matchDebugAppliedRule entry'
+    fromEntry entry
--- a/kore/src/Kore/Logger/DebugAppliedRule.hs
+++ b/kore/src/Kore/Logger/DebugAppliedRule.hs
@@ -177,5 +177,5 @@ matchDebugAppliedRule entry =
     fromEntry entry <|> throughScope
   where
     throughScope = do
-        WithScope { entry = entry' } <- fromEntry entry
+        entry' <- fromEntry entry
         matchDebugAppliedRule entry'
--- a/kore/src/Kore/Logger/Output.hs
+++ b/kore/src/Kore/Logger/Output.hs
@@ -85,6 +85,7 @@ import Options.Applicative
     )
 import qualified Text.Megaparsec as Parser
 import qualified Text.Megaparsec.Char as Parser
+import qualified Type.Reflection as Reflection
 
 import Kore.Logger
 import Kore.Logger.DebugAppliedRule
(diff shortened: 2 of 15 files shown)


kore/src/Kore/Logger/Output.hs
```diff
@@ -270,11 +271,9 @@ swappableLogger mvar =
     worker a logAction = Colog.unLogAction logAction a
 
 defaultLogPretty :: SomeEntry -> Pretty.Doc ann
-defaultLogPretty entry =
+defaultLogPretty (SomeEntry entry) =
     Pretty.hsep
-        [ Pretty.brackets (Pretty.pretty severity)
+        [ Pretty.brackets (Pretty.viaShow . Reflection.typeOf $ entry)
```
vladciobanu: Do you have a sample of how this looks like, now?

ana-pantilie: Sure.
When we have an `Entry` type for the log:
`[2019-12-06 15:46:26.950404633][DebugSolverSend] : (push 1 )`
When something was logged as a `LogMessage`:
`[2019-12-06 15:01:33.170087234][LogMessage] : [Debug] : Evaluator functionVariableAndEquals does not apply. []`

ana-pantilie: Oops, just noticed that the `Severity` is missing from the output in one of the cases. I'll debug this.

ana-pantilie: @vladciobanu Fixed the above in a new commit and edited the example.

vladciobanu: Ah, so it's literally the type name. That's reasonable. I think we need to do something about the repl: either just remove everything related to filtering by scope or replace the scope filter with a type filter.

ana-pantilie: I agree about the type filter. I was planning to do that in a separate PR, since filtering is a different feature and filtering by scope is already unusable because we don't print any scope information to the user as of the latest master. I can move these changes to this PR if you think it would make more sense, though.

kore/src/Kore/Logger/DebugAppliedRule.hs
```diff
@@ -177,5 +177,5 @@ matchDebugAppliedRule entry =
     fromEntry entry <|> throughScope
   where
     throughScope = do
```
ttuegel: The `throughScope` branch is no longer necessary.

kore/src/SMT/SimpleSMT.hs
```diff
@@ -281,48 +285,37 @@ newSolver exe opts logger = do
...
-warn :: GHC.HasCallStack => Solver -> [Logger.Scope] -> Text -> IO ()
-warn = logWith Logger.Warning
+warn :: GHC.HasCallStack => Solver -> Text -> IO ()
+warn = logMessageWith Logger.Warning
 
 send :: Solver -> SExpr -> IO ()
-send solver@Solver { hIn } command' = do
-    debug solver ["send"] (buildText command')
+send Solver { hIn, logger } command' = do
+    logDebugSolverSendWith logger (buildText command')
```
ttuegel: I would prefer if `logDebugSolverSendWith` took an `SExpr` as its second argument. We do not need to render the `SExpr` to text unless the log entry is displayed.

PR #1322, opened 2019-12-04
Rename NonSimplifiable to ConstructorLike

ttuegel:
I hope this will end a regrettable source of confusion.

I also removed three unused modules.

Fixes: #1309 

---

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`

PR #1333, opened 2019-12-09
Evaluate concrete functions only on non-simplifiable arguments

### LINKED ISSUE #1276: Only apply concrete rules to constructor-like arguments
When the user gives a rule the `concrete` attribute, they mean that the arguments are concrete and non-simplifiable. The backend currently interprets "concrete" to mean "no variables", but for the purpose of applying rules (at least) we should take a more restrictive interpretation.

virgil-serbanuta:
Fixes #1276 

---

###### Reviewer checklist


kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -107,6 +111,9 @@ evaluateAxioms definitionRules termLike predicate
     return simplified
 
   where
+    termLikeF = case Recursive.project termLike of
```
traiansf: Why not
```
    termLikeF = Cofree.tailF (Recursive.project termLike)
```

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -68,7 +72,7 @@ evaluateAxioms
     -> simplifier (Step.Results variable)
 evaluateAxioms definitionRules termLike predicate
   | any ruleIsConcrete definitionRules
-  , not (TermLike.isConcrete termLike)
+  , not (all TermLike.isNonSimplifiable termLikeF)
```
traiansf: This looks a little strange. If I understand correctly, since this is supposed to be a Function application, and since that is `Foldable`, what `all` will do would be to iterate over the arguments. Is that correct?  If so, although quite clever, it does not seem straight-forward to read, since `TermLikeF` being `Foldable` is not something we currently use on a regular basis in the code.  Therefore it might be useful to accompany this change by a comment.

kore/src/Kore/Step/Axiom/Evaluate.hs
virgil-serbanuta: Done.

kore/src/Kore/Step/Axiom/Evaluate.hs
virgil-serbanuta: Better now?

traiansf: better by why `isConstructorLike` instead of `isNonSimplifiable`?

virgil-serbanuta: Because @ttuegel just renamed it ;)

virgil-serbanuta: https://github.com/kframework/kore/pull/1322

PR #1336, opened 2019-12-10
Refactored right and ensures to RHS

### LINKED ISSUE #1305: Representation for rewrite rules with existentially-quantified variables on the right-hand-side
- [ ] Refactoring `RulePattern` to allow existential quantification (see PR #1336)
- [ ] Creating a separate structure dedicated to equational rules

traiansf:
This is the first part of addressing  #1305 , refactoring `RulePattern` such that all types of rules can have existentially quantified variables.

However, I'd like to get this merged because it touches a lot of code.  Second part would be creating a structure dedicated for equality rules.

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1064,27 +1045,23 @@ getDestination
...
 makeRuleFromPatterns
     :: forall rule
     .  FromRulePattern rule
     => rule
     -> Pattern Variable
-    -> Pattern Variable
+    -> RHS Variable
     -> rule
-makeRuleFromPatterns ruleType configuration destination =
+makeRuleFromPatterns ruleType configuration rhs =
```
ana-pantilie: Maybe this should get renamed to something like `makeRuleFromComponents`, although I'm not sure that is the best name. Do you or @ttuegel have any other ideas?

src/main/k/working/imp/merge/1.merge-golden
```diff
@@ -25,27 +25,20 @@
...
-                    Lbl'Unds'Map'Unds'{}(
-                        /* element: */ Lbl'UndsPipe'-'-GT-Unds'{}(
-                            /* Inj: */ inj{SortId{}, SortKItem{}}(
-                                VarX:SortId{}
-                            ),
-                            VarI:SortKItem{}
-                        ),
-                        /* opaque child: */ VarDotVar3:SortMap{}
-                    )
+    Lbl'-LT-'generatedTop'-GT-'{}(
```
ana-pantilie: Why has the output from these two tests changed?

traiansf: Because it had to be changed previously when I added exists for the first time. now it reverted to its initial form.

kore/src/Kore/Step/Rule.hs
```diff
@@ -488,29 +557,48 @@ fromSentenceAxiom sentenceAxiom = do
             (Syntax.sentenceAxiomAttributes sentenceAxiom)
     patternToAxiomPattern attributes (Syntax.sentenceAxiomPattern sentenceAxiom)
 
+rhsToPattern
```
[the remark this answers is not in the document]
traiansf: I can do that, although the  naming was chosen to complement patterntoAxiom and axiomToPattern which also act on TermLike. If we are to rename them, maye we should remove those other two as well, to be consistent.

kore/src/Kore/Strategies/Goal.hs
traiansf: I've renamed it as `configurationDestinationToRule` to match the naming of the "destructors" above

kore/src/Kore/Step/Rule.hs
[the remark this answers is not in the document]
traiansf: I've done that. I've also gone forward and renamed `patternToAxiomPattern` and `axiomPatternToPattern` following the same convention.

kore/src/Kore/Step/Rule.hs
```diff
@@ -486,31 +555,50 @@ fromSentenceAxiom sentenceAxiom = do
...
         (Attribute.Parser.liftParser . Attribute.Parser.parseAttributes)
             (Syntax.sentenceAxiomAttributes sentenceAxiom)
-    patternToAxiomPattern attributes (Syntax.sentenceAxiomPattern sentenceAxiom)
+    termToAxiomPattern attributes (Syntax.sentenceAxiomPattern sentenceAxiom)
+
+rhsToTerm
+    :: InternalVariable variable
+    => RHS variable
+    -> TermLike.TermLike variable
+rhsToTerm (RHS existentials right ensures) =
```
virgil-serbanuta: Using curly braces around the fields seems safer.

kore/src/Kore/Step/Rule.hs
```diff
@@ -919,71 +984,100 @@ refreshRulePattern
...
-mapVariables mapping rule1@(RulePattern _ _ _ _ _ _) =
+mapVariables mapping rule1@(RulePattern _ _ _ _ _) =
     rule1
         { left = TermLike.mapVariables mapping left
         , antiLeft = fmap (TermLike.mapVariables mapping) antiLeft
-        , right = TermLike.mapVariables mapping right
         , requires = Predicate.mapVariables mapping requires
-        , ensures = Predicate.mapVariables mapping ensures
+        , rhs = RHS
+            { existentials =  fmap mapping <$> existentials
```
virgil-serbanuta: single space after equals?

virgil-serbanuta:
Could you add one or more integration steps for applying `A(x)=>exists x . B(x)` to `A(1)`?

kore/src/Kore/Step/Rule.hs
```diff
@@ -148,15 +155,72 @@ newtype AxiomPatternError = AxiomPatternError ()
 
 instance NFData AxiomPatternError
 
+{-| Defines the right-hand-side of a rewrite rule / claim
+-}
+data RHS variable = RHS
+    { existentials :: ![TermLike.ElementVariable variable]
```
ttuegel: Should this be a `Set`?

kore/test/Test/Kore/Step/EquationalStep.hs
```diff
@@ -513,29 +515,38 @@ test_applyEquationalRule_ =
         EqualityRule RulePattern
             { left = Mock.a
             , antiLeft = Nothing
-            , right = mkBottom Mock.testSort
             , requires = makeTruePredicate_
-            , ensures = makeTruePredicate_
+            , rhs = RHS
```
ttuegel: We should have a helper like `RHS.fromTermLike :: TermLike variable -> RHS variable` with `existentials = [], ensures = makeTruePredicate_`, because that pattern seems to occur often. That will make future changes easier.

kore/src/Kore/Step/Rule.hs
traiansf: I could be, but I was reluctant to make it a set because then rhsToTerm would not be the inverse of termToRHS.  Do you think we should make it a Set?

kore/test/Test/Kore/Step/EquationalStep.hs
traiansf: I've added a helper named `injectTermIntoRHS`

kore/src/Kore/Step/Rule.hs
ttuegel: > then rhsToTerm would not be the inverse of termToRHS

Why not?

traiansf: because when we fold the set to add back the existentials, they might come up in a different orde. So, although semantically equivalent, they won't be syntactically equivalent.  If you think we shouldn't worry about that, I don't mind making it a set.

ttuegel: Oh, I see. It's nice to have the two functions be inverses of each other, so let's not use a `Set` then.

PR #1338, opened 2019-12-10
SMT solver transcript

virgil-serbanuta:
---

###### Reviewer checklist


kore/src/Kore/Logger/DebugSolver.hs
```diff
@@ -67,3 +92,44 @@ logDebugSolverRecvWith
...
+    $ \entry -> case matchDebugSolverSend entry of
+        Just sendEntry -> unLogAction textLogger (messageToText sendEntry)
+        Nothing -> unLogAction mempty entry
+  where
+    messageToText :: DebugSolverSend -> Text
+    messageToText =
+        Pretty.renderStrict
+        . Pretty.layoutPretty Pretty.defaultLayoutOptions
+        . pretty
+
```
ttuegel:
```suggestion
```

kore/src/Kore/Logger/DebugSolver.hs
```diff
@@ -67,3 +92,44 @@ logDebugSolverRecvWith
...
+{- | Options (from the command-line) specifying where to create a solver
+transcript.
+
+See also: 'parseDebugSolverOptions'
+
+-}
+newtype DebugSolverOptions =
+    DebugSolverOptions
+        { logType :: KoreLogType
```
ttuegel: I'm confused about the `KoreLogType`. I thought that this would be something like `Maybe FilePath`: either the user requested the transcript in a particular file, or they did not. What am I missing?

virgil-serbanuta: It allows me to provide a uniform interface for `startLoggers` in a natural way. Also, it seemed more flexible, but we don't actually need that. 

Since I can always do a `Maybe FilePaht` to `KoreLogType` conversion before calling `startLoggers`, I'll change this to `Maybe FilePath`.

kore/src/Kore/Logger/DebugSolver.hs
virgil-serbanuta: Fixed.

kore/src/Kore/Logger/DebugSolver.hs
virgil-serbanuta: Fixed.

kore/src/Kore/Logger/Output.hs
```diff
@@ -183,13 +294,22 @@ parseKoreLogOptions =
...
         option
             (maybeReader parseTypeString)
             $ long "log"
             <> help "Name of the log file"
-    parseTypeString filename = pure $ LogFileText filename
+
+      where
+        parseTypeString filename =
+            case filename of
+                "none" -> pure LogNone
```
ttuegel: We recently removed the command-line options for `none` and `stderr` because they don't make sense. I would not like to add them back.

kore/src/Kore/Logger/Output.hs
```diff
@@ -113,25 +132,117 @@ data KoreLogOptions = KoreLogOptions
...
+        , ioMode = AppendMode
+        , handleLogger = makeMainLogger koreLogOptions
+        }
+    smtSolverLoggerStartData = LoggerStartData
+        { logType = maybe LogNone LogFileText smtSolverLogFile
+        , ioMode = WriteMode
+        , handleLogger = makeSmtSolverLogger
+        }
+
+makeSmtSolverLogger :: MonadIO m => Handle -> LogAction m SomeEntry
```
ttuegel: I think the code to set up the loggers would be simpler if we replaced `makeSmtSolverLogger` and `makeMainLogger` with

```.hs
withSmtSolverLogger :: DebugSolverOptions -> (LogAction IO SomeEntry -> IO a) -> IO a

withMainLogger :: KoreLogOptions -> (LogAction IO SomeEntry -> IO a) -> IO a
```

`withLogger` becomes:

```.hs
withLogger
    koreLogOptions@KoreLogOptions { debugSolverOptions }
    continue
  =
    withMainLogger koreLogOptions $ \mainLogger ->
    withSmtSolverLogger debugSolverOptions $ \smtSolverLogger ->
        continue (mainLogger <> smtSolverLogger)
```

(`<>` for `LogAction` sends each `SomeEntry` to every logger.) I think this is easier to understand than `startLoggers` and we don't need to change `KoreLogType` or the command-line options.

ttuegel:
Bah! I should have removed the `automerge` label until you saw my review. :slightly_frowning_face:

virgil-serbanuta:
I'll try to do a follow-up PR tomorrow.

PR #1345, opened 2019-12-13
Write result.kore if there is an error

### LINKED ISSUE #1341: Write result.kore if there is an error

andreiburdusa:
---
Fixes #1341 

###### Reviewer checklist


kore/app/exec/Main.hs
```diff
@@ -375,7 +381,11 @@ parserInfoModifiers =
 main :: IO ()
 main = do
     options <- mainGlobal parseKoreExecOptions parserInfoModifiers
-    Foldable.forM_ (localOptions options) mainWithOptions
+    Foldable.forM_ (localOptions options) $ \o ->
+        catch (mainWithOptions o) $
+            \(Goal.GoalException (e :: SomeException) p) -> do
+                renderResult o ("last configuration = " <> unparse p)
```
traiansf: I think here you don't want to print any additional text, as the KORE parser int the frontend would not be able to handle it.
Alternatively, consider making the additional text into a comment.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -782,13 +789,17 @@ removeDestinationWorker stateConstructor goal = do
 
     destination = topExistsToImplicitForall configFreeVars rhs
 
+    enrichEventualException action =
```
traiansf: Why don't you extract this as a separate function and use it below as well.

kore/app/exec/Main.hs
```diff
@@ -375,7 +381,11 @@ parserInfoModifiers =
 main :: IO ()
 main = do
     options <- mainGlobal parseKoreExecOptions parserInfoModifiers
-    Foldable.forM_ (localOptions options) mainWithOptions
+    Foldable.forM_ (localOptions options) $ \o ->
+        catch (mainWithOptions o) $
```
ttuegel: We should move the `catch` down into `mainWithOptions`

kore/app/exec/Main.hs
```diff
@@ -375,7 +381,11 @@ parserInfoModifiers =
...
+            \(Goal.GoalException (e :: SomeException) p) -> do
```
ttuegel: The type annotation here should not be necessary. Please do not use single-letter variable names.

```suggestion
            \(Goal.GoalException someException p) -> do
```

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -824,6 +833,12 @@ isTrusted =
     . RulePattern.attributes
     . toRulePattern
 
+data GoalException exception = GoalException exception (Pattern Variable)
```
ttuegel:
```suggestion
data GoalException = GoalException (Pattern Variable) SomeException
```

Please also add documentation.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -879,11 +894,9 @@ derivePar rules goal = errorBracket $ do
     configuration :: Pattern Variable
     configuration = getConfiguration goal
 
-    errorBracket action =
-        onException action
-            (formatExceptionInfo
-                ("configuration=" <> unparseToText configuration)
-            )
+    enrichEventualException action =
```
ttuegel: Please lift this to the top level as:

```.hs
mapGoalException :: MonadCatch m => Goal goal => goal -> m a -> m a
```

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -879,11 +894,9 @@ derivePar rules goal = errorBracket $ do
...
+        catch action $
```
ttuegel: If you flip the arguments of `GoalException` and remove the type parameter, this function is simply

```suggestion
        handle (throw . GoalException configuration)
```

kore/src/Kore/Strategies/Goal.hs
ttuegel: Actually, I think I would prefer naming the type `WithConfiguration` and the function `withConfiguration`. The name `GoalException` sounds like there is a problem with the `Goal`, which is not exactly what's happening.

kore/app/exec/Main.hs
```diff
@@ -380,7 +385,12 @@ main = do
 mainWithOptions :: KoreExecOptions -> IO ()
 mainWithOptions execOptions = do
     let KoreExecOptions { koreLogOptions } = execOptions
-    exitCode <- runLoggerT koreLogOptions go
+    exitCode <- catch (runLoggerT koreLogOptions go) $
+        \(Goal.WithConfiguration lastConfiguration someException) -> do
+            renderResult
+                execOptions
+                ("// Last configuration:\n" <> unparse lastConfiguration)
+            throw someException
```
ttuegel: Use `Control.Monad.Catch.throwM` instead of `Control.Exception.throw` whenever we are in a monad context.

Issue #1335, opened 2019-12-09
Potential space leakage of Kore-repl/Kore-exec


yzhang90:
k-master: 413f849a889 + kore-master: b523da7 + kevm: https://github.com/kframework/evm-semantics/tree/edsl2

I ran the following command 
```
kprove --haskell-backend-command "/home/zy/workspace/kore/.build/kore/bin/kore-repl --repl-script /home/zy/workspace/kore/dist/kast.kscript " \
    ~/workspace/evm-semantics/tests/specs/erc20/ds/balanceOf-spec.k \
    -d ~/workspace/evm-semantics/.build/defn/haskell/ \
    --def-module VERIFICATION \
    --concrete-rules EDSL.#ceil32,EDSL.keccakIntList,EVM-DATA.#signed.positive,EVM-DATA.#signed.negative,EVM-DATA.#unsigned.positive,EVM-DATA.#unsigned.negative,EVM-DATA.powmod.nonzero,EVM-DATA.powmod.zero,EVM-DATA.signextend.invalid,EVM-DATA.signextend.negative,EVM-DATA.signextend.positive,EVM-DATA.keccak,EVM-DATA.#asWord.recursive,EVM-DATA.#asByteStack,EVM-DATA.#asByteStackAux.recursive,EVM-DATA.#padToWidth,EVM-DATA.#padRightToWidth,EVM-DATA.#newAddr,EVM-DATA.#newAddrCreate2,EVM-DATA.mapWriteBytes.recursive,EVM-DATA.#range,EVM-DATA.#lookup.some,EVM-DATA.#lookup.none,EVM.#memoryUsageUpdate.some,EVM.Cgascap,EVM.Csstore.new,EVM.Csstore.old,EVM.Rsstore.new,EVM.Rsstore.old,EVM.Cextra,EVM.Cmem
``` 
followed by
```
stepf 44
```
 REPL runs for a while and eats 96% of the memory (I have a 32G memory).
However, if i type `stepf 11`  every time, the memory consumption can be maintained at a reasonable ratio. (e.g., ~10% after executing 33 steps). 
I tried `kore-exec` with `--depth 33` and monitored it with `top`. memory peak: 68.1% time:~10mins. memory consumption is always increasing.

yzhang90:
only `graph` command will reduce the memory consumption from ~80% to ~3%

ttuegel:
> only `graph` command will reduce the memory consumption from ~80% to ~3%

[...]

PR #1347, opened 2019-12-13
Better profiling experience

ttuegel:
This pull request improves the profiling experience.

You must use `cabal` to build the package, and you must have version 3.0 or later (run `cabal --version` to check). Stack does not seem to allow the user to override the profiling options passed to GHC, but this is essential for acceptable performance. You can install `cabal` and GHC with [`ghcup`](https://www.haskell.org/ghcup/).

To build:

```
# This writes the cabal.project.local file.
# The setting persists until you --disable-profiling or delete the file.
cabal configure --enable-profiling
cabal build kore
# At the end of the build, cabal will print the path of
# the kore-exec binary it built, you need to copy it into place:
cp [...]/kore-exec ./.build/kore/bin
```

With these changes, profiling overhead is about 500%, which is better than the 3000% we were seeing with Stack.

I am adding appropriate `INLINE` pragmas to get reasonable profiling reports. (Anything marked `INLINE` will not be reported automatically.)

Fixes: #1346 
See also: #1335 

---

###### Reviewer checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ test-results.xml
 .directory
 /dist-newstyle/
 /dist/
+cabal.project.local
--- a/cabal.project
+++ b/cabal.project
@@ -6,6 +6,8 @@ allow-newer:
 
 package *
   ghc-options: -fhide-source-paths
+  ghc-options: -fno-prof-auto
 
 package kore
   ghc-options: -Wall -Werror
+  ghc-options: -fprof-auto-top
--- a/kore/package.yaml
+++ b/kore/package.yaml
@@ -23,6 +23,12 @@ extra-source-files:
 # common to point users to the README.md file.
 description: Please see the [README](README.md) file.
 
+flags:
+  threaded:
+    description: Use the threaded runtime. Recommended to disable for profiling.
+    manual: true
+    default: true
+
 dependencies:
   - base >=4.7
   - aeson >=1.4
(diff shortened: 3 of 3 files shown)


kore/src/Kore/Step/Strategy.hs
```diff
@@ -291,14 +296,17 @@ insEdge (fromNode, toNode) exe@ExecutionGraph { graph } =
...
+    :: (NFData config, NFData rule)
+    => MonadState (Gr config (Seq rule)) m
     => ChildNode config rule
     -> m Graph.Node
-insChildNode configNode =
-    State.state insChildNodeWorker
+insChildNode childNode = do
+    graph <- deepseq childNode State.get
+    let (node', graph') = insChildNodeWorker graph
+    deepseq graph' (State.put graph')
```
ana-pantilie: Just to be sure I understand, this is what you were saying in the meeting about fully evaluating the graph when adding a new node, similar to calling `graph` after each step?

kore/src/Kore/ModelChecker/Bounded.hs
```diff
@@ -61,6 +61,7 @@ import Kore.Step.Strategy
     , pickFinal
     , runStrategyWithSearchOrder
     )
+import Kore.Strategies.ProofState ()
```
ana-pantilie: Why was this added?

ttuegel:
My workaround for the issue makes the KEVM tests time-out, so I must have affected time performance, too. :slightly_frowning_face:

ttuegel:
I moved all the mitigations to #1378 so that we can merge this pull request, which now only adds the profiling options.

PR #1351, opened 2019-12-16
kore-exec: Use compacting garbage collector on oldest branch

ttuegel:
This is a work-around for #1335. It doesn't solve the underlying problem, but it
should reduce memory use in the meantime.

---

###### Reviewer checklist

### DIFF
--- a/kore/package.yaml
+++ b/kore/package.yaml
@@ -142,7 +142,7 @@ executables:
     source-dirs:
       - app/exec
       - app/share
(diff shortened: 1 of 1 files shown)


yzhang90:
The fix is not that promising...The kore-repl still eats most of the memory after executing 22 steps...

yzhang90:
I still need to call graph  every 11 steps... ( it seems that 15 - 18 steps is the limit on my computer)

Issue #1277, opened 2019-11-20
Resume in-progress proofs


ttuegel:
The output of `kprove` should not be a single pattern, but rather an update spec module where

- Each claim that was proved appears with the `trusted` attribute.
- For each claim in progress,
  - For each unproven configuration, there is one claim from that configuration to the claim target.
  - There is one `trusted` claim from the original source of the claim to the disjunction of all unproven configurations.
- Each claim which has not been started appears unchanged.

The user can modify the updated spec module and resume the proof by pointing `kprove` at the new spec module. The user will mainly add new lemmas to remove infeasible branches, which is usually safe. Of course, this also allows the user to do unsound things, so the updated spec shall come with a prominent warning that they must eventually run the proof from the beginning with any modifications.

ttuegel:
Clarification from @yzhang90:

We do not need to output any un-`trusted` claims into the updated spec module. We should output only the `trusted` claims to recover the current proof state. The current proof state should be represented as a Kore module of trusted claims and applied by the backend _once_ at start-up. We do not want to apply the trusted claims as circularity rules.

yzhang90:
what you described looks good to me. One thing that I'm not sure is that whether we should trust the frontend or not. Tom thinks that the frontend may not be consistent between runs. As a result, we need to convert the generated module back to K and use the frontend to compile the whole thing again.

ttuegel:
> As a result, we need to convert the generated module back to K and use the frontend to compile the whole thing again.

As I said, I don't think we should actually do that, I think we should just invalidate the trusted claims file whenever the definition changes.

yzhang90:
If users only add some lemmas, would that be considered as definition changes? Note that all the semantic rules and lemmas will be combined into a single vdefinition.kore file.

ehildenb:
If we are going with adding `trusted` claims to the spec, then I think it needs to be done by the frontend (because there is no reason for this to be backend specific).

If we use a structured proof log, the trusted claims can remain as Kore and be backend specific, which means we're not limited to interactions that the frontend supports.

ehildenb:
Discussed with @ttuegel IRL, and it sounds like a good approach (different than what I expected). Here is a summary of my understanding:

-   Haskell backend has extra option `--save-proofs` or something.
-   If that option is turned on, all proofs are saved in a `trusted-claims.kore` file which the frontend never sees.
[...]

PR #1356, opened 2019-12-17
Dump proven claims to file

virgil-serbanuta:
This solves half of #1277 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Strategies/Verification.hs
+++ b/kore/src/Kore/Strategies/Verification.hs
@@ -9,29 +9,40 @@ This should be imported qualified.
 module Kore.Strategies.Verification
     ( Claim
     , CommonProofState
+    , StuckVerification (..)
     , verify
     , verifyClaimStep
     , toRulePattern
     ) where
 
+import qualified Control.Monad as Monad
+    ( foldM_
+    )
 import Control.Monad.Catch
     ( MonadCatch
     )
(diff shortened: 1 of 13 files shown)


kore/src/Kore/Strategies/Verification.hs
```diff
@@ -95,6 +103,22 @@ didn't manage to verify a claim within the its maximum number of steps.
 If the verification succeeds, it returns ().
 -}
 
+data StuckVerification patt claim
+    = StuckVerification
+        { stuckPattern :: !patt
```
ttuegel: Is it possible to actually return the current proof state here, or only the pattern we got stuck on?

src/main/k/working/imp/save-proofs/sum-spec.save-proofs-golden
```diff
@@ -0,0 +1,128 @@
+claim
```
ttuegel: Should this be a full Kore definition, so we can have multiple claims?

virgil-serbanuta: I was just doing this in the next PR, but I guess I could move it here.

virgil-serbanuta: Done.

PR #1357, opened 2019-12-17
Skip proving rules that are already proven

### LINKED ISSUE #1277: Resume in-progress proofs

virgil-serbanuta:
~Needs https://github.com/kframework/kore/pull/1356~

Fixes #1277 

---

###### Reviewer checklist

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -71,6 +71,9 @@ import Options.Applicative
     , value
     )
 import qualified Options.Applicative as Options
+import System.Directory
+    ( doesFileExist
+    )
 import System.Exit
     ( ExitCode (..)
     , exitWith
@@ -89,6 +92,7 @@ import Kore.Error
 import Kore.Exec
 import Kore.IndexedModule.IndexedModule
     ( VerifiedModule
+    , indexedModuleRawSentences
     )
 import qualified Kore.IndexedModule.MetadataToolsBuilder as MetadataTools
     ( build
@@ -120,6 +124,12 @@ import Kore.Profiler.Data
     ( MonadProfiler
     )
 import Kore.Step
+import Kore.Step.Rule
+    ( ReachabilityRule
+    )
+import qualified Kore.Step.Rule as Rule
+    ( toSentence
+    )
 import Kore.Step.Search
     ( SearchType (..)
     )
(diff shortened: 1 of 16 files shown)


kore/app/exec/Main.hs
```diff
@@ -470,6 +480,19 @@ koreProve execOptions proveOptions = do
     mainModule <- loadModule mainModuleName definition
     let KoreProveOptions { specMainModule } = proveOptions
     specModule <- loadModule specMainModule definition
+    let KoreProveOptions { saveProofs } = proveOptions
+    maybeAlreadyProvenModule <- case saveProofs of
```
ttuegel: Let's factor this out into `loadSavedProofs`.

kore/src/Kore/Exec.hs
```diff
@@ -700,9 +737,15 @@ evalProver
     -- ^ The main module
     -> VerifiedModule StepperAttributes Attribute.Axiom
     -- ^ The spec module
+    -> Maybe (VerifiedModule StepperAttributes Attribute.Axiom)
```
ttuegel: This module contains only the claims which were proved?

kore/src/Kore/Strategies/Verification.hs
```diff
@@ -128,14 +143,71 @@ verify
     => Show (Rule claim)
     => Limit Natural
     -> GraphSearchOrder
-    -> [claim]
-    -> [Rule claim]
-    -> [(claim, Limit Natural)]
+    -> AllClaims claim
+    -> Axioms claim
+    -> AlreadyProven
+    -> ToProve claim
```
ttuegel: Do these four arguments always go together? It might be nicer to pack them up in a record, if so.

PR #1366, opened 2019-12-18
Add design doc for ?X with ensures

traiansf:
---

###### Reviewer checklist


docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+(RAND0)  rand() -> * 0
+(RAND1)  rand() -> * 1
+(RAND2)  rand() -> * 2
+  ...    ...
+(RAND-1) rand() -> * (-1)
+(RAND-2) rand() -> * (-2)
+  ...    ...
+```
+
+It is not easy to define many of the variants of `rand()`. For example. `randPrime()` can rewrite to every _prime numbers_. Here is its axiom
```
yzhang90: not easy?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+
+There are many variants of `arb()`. For example. `arbDifferentFromList(Is)` is an unspecific function whose return value must be different than those in `Is`. Here is its axiom
+
+```
+(ARBDIFF) \forall Is:List{Is} . \exists I:Int . \neg(I \in Is) /\ arbDifferentFromList(Is) = I
+```
+
+### Frontend Syntax
+
+We do not need special frontend syntax to define `arb()`. We only need to define it in the usual way as a function (instead of a language construct).
```
yzhang90: if we treat `arb()` as uninterpreted function, the current frontend syntax is enough. However, if we want to rewrite it to some variable, the current frontend syntax doesn't support it.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+
+The random number construct `rand()` is a language construct. It is part of the syntax of the PL:
+
+```
+Stmt ::= "rand" "(" ")"
+```
+
+### Axioms
+
+The intended semantics of `rand()` is that it can rewrite to every integers. Here is its axiom
```
virgil-serbanuta: all integers? all integers at once? the set of all integers?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+
+The fresh integer construct `fresh(Is)` is a language construct and part of the PL syntax
+
+```
+Stmt ::= ... | "fresh" "(" List{Int} ")"
+```
+
+### Axioms
+
+The intended semantics of `fresh(Is)` is that it can rewrite to any integers that are not in `Is`.
```
virgil-serbanuta: any integer that is not? any set of integers that are not?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+Stmt ::= ... | "fresh" "(" List{Int} ")"
+```
+
+### Axioms
+
+The intended semantics of `fresh(Is)` is that it can rewrite to any integers that are not in `Is`.  
+
+Note that `fresh(Is)` and `randDifferentFromList(Is)` are different; the former does not need to rewrite to _every_ integers not in `Is`, while the latter requires so. For example, it is _correct_ to implement `fresh(Is)` so it always returns the smallest positive integers that are not in `Is`, but same implementation for `randDifferentFromList(Is)` is considered _incorrect_. In other words, there exist multiple correct implementations of `fresh(Is)`, some of which may be deterministic, but there only exists a unique implementation of `randDifferentFromList(Is)`. Finally, it is a _correct_ if one implements `fresh(Is)` the same as `randDifferentFromList(Is)`.
```
virgil-serbanuta: smallest positive integer that is not? the/a set of smallest positive integers?

also: "a correct"?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
+This design document describes four typical example language constructs or mathematical functions that have unspecified or nondeterministic behaviors and shows their axiomatization in Kore. We also propose the corresponding K frontend syntax to define some of these behaviors. The frontend syntax uses question mark variables `?X`, which when translated to Kore axioms become existentially quantified, and use `ensures` to specify the logical constraints that `?X` should satisfy. The `ensures` clause will be translated to the RHS in Kore axioms.
```
virgil-serbanuta: Below I've added a comment referring to something in the middle of one of these long lines, and I guess it's harder to resolve the reference than it could be. Why not 80-chars lines?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Axioms
+
+The intended semantics of `arb()` is that it is an unspecified nullary function. The exact return value of `arb()` is unspecified in the semantics but up to the implementations. However, being a mathematical function, `arb()` must return the same value in any one implementation. 
+
+The axiom of `arb()` is 
+
+```
+(ARB) \exists I:Int . arb() = I
+//    this is exactly (Function) axiom
```
virgil-serbanuta: What does this mean? "This is exactly the axiom specifying that `arb` is a function"?

traiansf: yes. it's what is generated from the `[functional]` annotation.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+Note that `arbDifferentFromList(Is)` is different from `fresh(Is)`, because
+`fresh(Is)` _transitions_ to an integer not in `Is` (could be a different one
+each time it is used), while `arbDifferentFromList(Is)` _is equal to_ a (fixed)
+integer not in `Is`.
+
+### Skolemization
+
+One possible approach to implementing the abovewould be through [Skolemization](https://en.wikipedia.org/wiki/Skolem_normal_form),
```
virgil-serbanuta: Line break before the square brace?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+Its axiom is
+
+```
+(INTERVAL) \forall M:Int . \forall N:Int . interval(M,N) = \exists X:Int . X:Int /\ X >= M /\ X <= N
+```
+
+### Frontend Syntax
+
+The symbol declaration would require a special attribute to signal the fact that
+it represent a proper ML symbol which is defined equationally. The precise
```
virgil-serbanuta: represents

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+name would have to be establishes, but let's call it here `power-function`
```
virgil-serbanuta: estabilished

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+syntax Exp ::= randList (List{Int})
+rule randList(Is) => I:Int
+  requires notBool (I inList{Int} Is)
+```
+
+Note 1: all above are not function symbols, but languages constructs.
+
+Note 2: Currently the frontend does not allow rules with universally quantified
+variables in the RHS which are not bound in the LHS.
+
```
grosu: The above looks good to me.  So basically you would use a fresh, non-? variable in the RHS. The frontend may still issue a warning, just in case the user meant the other case below.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+### Frontend Syntax
+
+We use the following K syntax to define `fresh(Is)`
+
+```
+syntax Exp ::= fresh (List{Int})
+rule fresh(Is:List{Int}) => ?I:Int
+  ensures notBool (?I inList{Int} Is)
+```
+
```
grosu: LGTM, too. So the idea is that the ?-variables are existentially quantified after the universally quantified ones, also at the top. Interestingly, you can have both ?- and non-? fresh variables in the RHS of a rule, the former being existentially quantified after the latter (so they ? variables depend, a la Skolem, on the non-? ones; it may be a good idea to have such an example, too).

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+is a symbol satisfying the `(Function)` axiom, without further constraints.
+
+### Frontend Syntax
+
+We do not need special frontend syntax to define `arb()`.
+We only need to define it in the usual way as a function
+(instead of a language construct), and provide no axioms for it
+(the `(Function)` axiom is generated automatically from the `function`
+attribute annotation.
+
```
grosu: Yeah, I agree with that.  This is just saying that `arb(...)` is a function, so no need for anything special in terms of notation.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+1. they are not allowed, since the users could themselves Skolemize the axiom
+
+    ```
+    syntax Int ::= arbDiffI(List{Int}) [function, functional]
+    syntax Int ::= arbDifferentFromList(List{Int}) [function, functional]
+    rule arbDifferentFromList(Is:List{Int}) => arbDiffI(Is)
+      ensures notBool (arbDiffI(Is) inList{Int} Is)
+    ```
+
```
grosu: Allowing both `function` and `functional` as attributes is really confusing.  We should change one of them.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+   the `ensures` clause would trigger Skolemization and the generation of
+   the rule above.
+   That is, `adbDifferentFromList` could be written in K as
+
+    ```
+    syntax Int ::= arbDifferentFormList(List{Int}) [function]
+    rule arbDifferentFromList(Is:List{Int}) => ?I:Int
+      ensures notBool (?I inList{Int} Is)
+    ```
+
```
grosu: The rule above looks totally fine to me.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
grosu: Also, the rule above looks strange to me.  `arbDiffI(Is)` seems artificial there.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
grosu: `power-function` ... hm, then all ML symbols are power-functions by default.  Since almost all the ML symbols in a definition end up being function symbols or constructors, the poor proper ML symbols lost their default status.  Not sure what name is better here: power-function, relation, proper-symbol, bare-symbol, default-symbol, etc.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+it represent a proper ML symbol which is defined equationally. The precise
+name would have to be establishes, but let's call it here `power-function`
+to signal that it's a function interpreted in the powerset of the domain.
+
+```
+syntax Int ::= interval(Int, Int) [power-function]
+rule interval(M:Int,N:Int) = ?X
+    ensures ?X >=Int M andBool ?X <=Int N
+```
+
```
grosu: I am not too excited about the above, because it breaks the convention that ?-variables are existentially quantified at the top.  Because of some attribute somewhere else, the scope of the ?-variable changes later.  I think we should be able to desugar rules locally, without global information.  How about a new kind of ?-variable, like `??X` instead of `?X`, with the meaning that its scope is the RHS of the rule and not the entire rule?

grosu: Of course, the ideal would be to simply allow people to write the existential quantifier themselves, but then the `ensures` makes no sense because it needs to go inside the existential quantifier scope.  I would personally like to write the rule as in ML:
```
rule interval(M:Int,N:Int) = \exists ?X . (?X /\ ?X >=Int M /\ ?X <=Int N)
```

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: my option for power-function was to signal that its definition is to be done through equations, since the front-end only uses the `=>` symbol for both rewrites and equalities.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,272 @@
...
+where `isPrime(_)` is a predicate that can be defined in the usual way.
+
+As another example, `randDifferentFromList(Is)` takes a list `Is` of integers
+and can rewrite in one step to any integer that is not in `Is`.
+Here is its axiom
+
+```
+(RANDLIST) \forall Is:List{Int} . \forall I:Int . \neg(I \in Is) -> randDifferentFromList(Is) -> * I
```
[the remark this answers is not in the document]
traiansf: I'd rather not.  implication, as the functional arrow, associates to the right.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,276 @@
...
+syntax Exp ::= rand ()
+rule rand() => X:Int
+
+syntax Exp ::= randPrime ()
+rule randPrime() => X:Int
+  requires isPrime(X)
+
+syntax Exp ::= randList (List{Int})
+rule randList(Is) => I:Int
+  requires notBool (I inList{Int} Is)
```
ehildenb: Why the `notBool` here? Are you trying to pick a random element _in_ the list, or _not in_ the list?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
ehildenb: I think for rules like this, let's fall back to using ML syntax directly for now:

```
axiom interval(M, N) = \exists X . X >= M /\ X <= M
```

We'll need to expose an `axiom` keyword for this, and allow the `=` symbol.

But I don't see an immediate need to support this ML-Set reference directly yet.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: All these rules are non-executable in a concrete setting, therefore, we should maybe add attributes to them to signal that. Moreover, it might be useful if we use different attribute names because this would help the compiler better identify the rules and process them accordingly.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: We could, but this again looks more like a case in which it might be clearer to have a proper ML rule.  I would not add an example for this unless it's a compelling and useful one, which I can't thing of at this time.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: I removed this and only kept the `?` syntax and promised an Skolemization pass in the Kompiler

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: I proposed a combination between @ehildenb and @grosu 's answers, with the current KORE-like syntax in K, and mentioned we won't be implementing this for now.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: There was some confusion. I decided to implement both cases and use better names.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+axiomatization in Kore.
+
+We also propose the corresponding K frontend syntax to define some of these
+behaviors.
+
+### Implementation concerns
+
+Except for the `fresh` variants (Example B), none of the other are currently
+supported by either the K frontend of the Haskell backend.
+
```
virgil-serbanuta: Remove empty line?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+integer not in `Is`.
+
+### Skolemization
+
+One possible approach to implementing the above would be through
+[Skolemization](https://en.wikipedia.org/wiki/Skolem_normal_form),
+i.e., by replacing the existential variables with fresh uninterpreted functions
+applied to the universally quantified variables.
+
+For example, the `(ARBDIFF)` rule above could be Skolemized to
```
virgil-serbanuta: There's no `ARBDIFF` above.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+is a symbol satisfying the `(Function)` axiom, without further constraints.
+
+### Frontend Syntax
+
+We do not need special frontend syntax to define `arb()`.
+We only need to define it in the usual way as a function
+(instead of a language construct), and provide no axioms for it
+(the `(Function)` axiom is generated automatically from the `function`
```
virgil-serbanuta: Parenthesis not closed.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+attribute annotation.
+
+The frontend doesn't currently provide a syntax to define the variants of
+`arb()`, such as `arbInList(Is)` in the above.
+
+We propose to implement a Skolemization compilation step triggered by a
+`function` annotation in the syntax definition, together with the use of `?`
+variables in the `rhs` of the rule and the `ensures`.
+
+That is, `adbInList` could be written in K as
```
virgil-serbanuta: s/adb/arb/

virgil-serbanuta: Why is this an example for Skolemization?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+   and can rewrite in one step to any item _not_ in `Is`.
+
+    ```
+    \forall Is:List . \forall I . \not(I \in Is) -> randNotInList(Is) -> * I
+    ```
+
+1. `randPrime()`, can rewrite to any _prime number_.
+
+    ```
+    \forall P:Int . isPrime(P) -> (randPrime() -> * P)
```
xc93: remove parentheses?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Summary
+
+The frontend syntax uses question mark variables of the form `?X` to refer to
+existential variables, and uses `ensures` to specify logical constraints on
+those variables.
+These variables are only allowed to appear in the RHS of a rule.
+
+When K rules are translated to Kore axioms, the `?` variables are existentially
+quantified at the top of the RHS and the `ensures` clause is and-ed to the RHS.
```
xc93: "In addition, the regular `non-?` variables are universally quantified at the top of the entire rule. When there are both `?`-variables and the regular variables, we quantify regular variables first, followed by `?`-variables.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,313 @@
...
+In addition, the regular non-`?` variables are universally quantified at the top
+of the entire rule.
+When both `?`-variables and regular variables are present, we quantify regular
+variables first, followed by `?`-variables.
```
ttuegel: Isn't this sentence implied by these?

> the `?` variables are existentially quantified at the top of the RHS

> the regular non-`?` variables are universally quantified at the top of the entire rule

Is this saying something else?

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,313 @@
...
+functions that have unspecified or nondeterministic behaviors and shows their
+axiomatization in Kore.
+
+We also propose the corresponding K frontend syntax to define some of these
+behaviors.
+
+### Implementation concerns
+
+Except for the `fresh` variants (Example B), none of the other are currently
+supported by either the K frontend of the Haskell backend.
```
ttuegel:
```suggestion
supported by either the K frontend or the Haskell backend.
```

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,313 @@
...
+    ```
+    \forall M:Int.\forall N:Int.\forall I:Int.
+        I >= M /\ I <= N -> randBounded(M, N) -> * I
+    ```
+
+1. `randInList(Is)` takes a list `Is` of items
+   and can rewrite in one step to any item in `Is`.
+
+    ```
+    \forall Is:List . \forall I . I \in Is -> randInList(Is) -> * I
```
ttuegel: We probably want a different symbol here, because `\in` is the Kore alias for unification:

```suggestion
    \forall Is:List . \forall I . inList(I, Is) -> randInList(Is) -> * I
```

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,313 @@
...
+In other words, there exist multiple correct implementations of `fresh(Is)`,
+some of which may be deterministic, but there only exists a unique
+implementation of `randNotInList(Is)`.
+Finally, note that `randNotInList(Is)` is a _correct_ implementation
+for `fresh(Is)`.
+
+Here is the axiom of `fresh(Is)`
+
+```
+(FRESH) \forall Is:List{Int} . \exists I:Int .
```
ttuegel: I'm confused by this, because it says above that the `?` variables will be quantified over the right-hand side only.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
```diff
@@ -0,0 +1,313 @@
...
+
+```
+(ARB') \forall Is:List{Is} . arbI(Is) \in Is /\ arbInList(Is) = arbI(Is)
+```
+where
+```
+arbI : List{Int} -> Int
+```
+
+is a symbol satisfying the `(Function)` axiom, without further constraints.
```
ttuegel: Why is it necessary to introduce an auxiliary symbol instead of using an existential quantifier as for `arb`? For example, why is this insufficient?

```
\forall Is:List{Int} . \exists I:Int . notEmpty(Is) -> arbInList(Is) = I ∧ inList(I, Is)
```

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: My bad. I understood Xiaohong's suggestion as adding text, while he obviously meant replacing.  The idea is that we sometimes can move the quantifier to the RHS, but other times, like in this case, we cannot.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: Logically, it is sufficient, but operationally I think we are better off if Skolemizing, because it's not clear how we could instantiate this axiom if we don't.

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
traiansf: I think Xiaohong meant for me to replace the fact that ? vars are quantified at the top of the RHS with them being quantified at the top, as it covers the fresh example, too. I'll reword.

traiansf:
Fixes #1305

docs/2019-12-16-Question-Mark-Variables-With-Ensures.md
ttuegel: It's not clear to me how we could instantiate `arbI(Is)` using the `(Function)` axiom, either. We have to introduce the same variable for each occurrence, but also distinguish between different arguments.

traiansf:
> I approved, but I am still somewhat unconvinced that we need Skolemization in the frontend. We can revisit that when the time comes.

Concerning `arb(Is)`, we will just treat it as an uninterpreted, simplified symbol.  That is the point of Skolemization: to introduce fresh uninterpreted symbols avoiding an existential quantification.  Being an uninterpreted symbol, it would only compare equal to itself, and you can only discharge it through the predicates it came with.

traiansf:
@ttuegel I'm not saying Skolemization is the only possible solution; it's just that in its absence, it's not clear (to me, at least) how to operationally use axioms with an exists at the top.  Skolemization allows one to fix a witness for the existence once and for all.

PR #1368, opened 2019-12-18
Existentials 2

traiansf:
- Extracted `EqualityPattern` and `EqualityRule`; refactored `Rule` module into 3 modules
- Created 'HasFreeVariables' class and instances of it for all relevant types
- Created 'HasLeftPattern' 'HasMapVariables' 'HasRequiresPredicate' 'HasRefreshPattern' classes to be able to reuse unification code for `EqualityPattern`s
---

###### Reviewer checklist


kore/src/Kore/Step/AxiomPattern.hs
```diff
@@ -0,0 +1,60 @@
...
+    ( SubstitutionVariable
+    )
+import Kore.Variables.Fresh
+    ( Renaming
+    )
+
+{-| Should be implemented for rules having a clearly identifiable
+left-hand-side pattern used for matching / unifying with the configuration.
+-}
+class HasLeftPattern rule variable where
```
ttuegel: I am generally opposed to having `Has...` classes because 1. they don't actually allow flexible behavior and 2. they constrain all instances to have the same underlying representation without giving the benefit of actually having the same data type. We should either refactor the data types so that all the rules have the same (possibly parameterized) representation _or_ we should introduce a new class whose members are every function that currently has a `HasLeftPattern` constraint. The same goes for `HasRequiresPredicate` and `HasMapVariable`.

kore/src/Kore/Logger/DebugAppliedRule.hs
```diff
@@ -61,23 +70,23 @@ The rule's 'RulePattern.requires' clause is combined with the unification
...
-type UnifiedRule variable = Conditional variable (RulePattern variable)
+type UnifiedRule = Conditional
 
 {- | A log 'Entry' when a rule is applied.
 
 We will log the applied rule and its unification or matching condition.
 
  -}
-newtype DebugAppliedRule =
+newtype DebugAppliedRule rule =
```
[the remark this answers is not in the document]
traiansf: Got it. I admit I found it hard to work with them. Well, anyway this filtering based on the top symbol is only useful for simplification, so I've revert it to only work for equations now.

kore/src/Kore/Step/AxiomPattern.hs
traiansf: I've created a `UnifyingRule` class to comprise the common unification code used by both `EqualityPattern` and `RulePattern`.  I agree with you that `HasLeftPattern` and `HasRequires` are not very good.

Nevertheless, I think that a class/classes providing 'freeVariables', 'mapVariables', 'refreshVariables' would actually be beneficial, as we already have these functions implemented for multiple types, sometimes with slight changes in the signature.
Having a class/classes to provide these might recognize the existing common functionality.
What do you think?

ttuegel: This sounds like a generalization of the `Binding` class used to implement substitution. What if we implement a `Substitute` class that has `freeVariables`, `refreshVariables`, `mapVariables`, and `traverseVariables` (which basically support capture-avoiding substitution), and also `substitute`?

traiansf: Sounds good!  Should we do it as part of this PR, or open a new one?

kore/src/Kore/Attribute/Pattern/FreeVariables.hs
```diff
@@ -83,21 +84,28 @@ mapFreeVariables
...
 {-# INLINE traverseFreeVariables #-}
 
 {- | Extracts the list of free element variables
 -}
 getFreeElementVariables :: FreeVariables variable -> [ElementVariable variable]
 getFreeElementVariables =
     mapMaybe extractElementVariable . Set.toList . getFreeVariables
+
+-- | Class for extracting the free variables of a pattern, term, rule, ...
+class HasFreeVariables pat variable where
```
ttuegel: This should aid type inference greatly:

```suggestion
class HasFreeVariables thing where
    type FreeVariableType thing
    freeVariables :: thing -> FreeVariables (FreeVariableType thing)
```

kore/src/Kore/Step/Step.hs
```diff
@@ -0,0 +1,405 @@
...
+    refreshRule
+        :: SubstitutionVariable variable
+        => FreeVariables variable  -- ^ Variables to avoid
+        -> rule variable
+        -> (Renaming variable, rule variable)
+
+    {-| Apply a given function to all variables in a rule. This is used for
+    distinguishing rule variables from configuration variables.
+    -}
+    mapVariables
```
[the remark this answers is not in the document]
traiansf: I'm planning to implement `traverseVariables` with the `Substitute` class we discussed earlier.  Then both `mapVariables` and `refreshVariables` will be moved to that class.

kore/src/Kore/Attribute/Pattern/FreeVariables.hs
traiansf: I tried to do that, but it does not seem to play nice with the instance below (`HasFreeVariables () variable`) as in that case the variable type does not depend on the pattern type. That is why I couldn't use functional dependencies in the first place.

kore/src/Kore/Step/Step.hs
```diff
@@ -0,0 +1,405 @@
...
+        :: Ord variable2
+        => (variable1 -> variable2)
+        -> rule variable1
+        -> rule variable2
+
+        -- Minimal complete definition:
```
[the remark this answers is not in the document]
traiansf: no longer needed

Issue #1353, opened 2019-12-17
Kore.Logger.entryTypeText is slow


ttuegel:
[`Kore.Logger.entryTypeText`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Logger.hs#L293) takes a surprising amount of time and does about 10% of the allocation in a typical run.

---

- [ ] Create a registry (`Map`) from textual representations to `TypeRep`.
- [ ] Use that map to validate command-line input.
- [ ] Compare `TypeRep` instead of `Text` when deciding which log entries to display.

PR #1370, opened 2019-12-18
Add registry of log entry types

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1353

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Logger/Registry.hs
+++ b/kore/src/Kore/Logger/Registry.hs
@@ -10,6 +10,7 @@ module Kore.Logger.Registry
     , registry
     , typeToText
     , textToType
+    , getEntryTypesAsText
     -- entry types
     , debugAppliedRuleType
     , debugAxiomEvaluationType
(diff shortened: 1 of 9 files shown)


kore/src/Kore/Logger/Registry.hs
```diff
@@ -0,0 +1,164 @@
...
+    ( WarnFunctionWithoutEvaluators
+    )
+import Kore.Logger.WarnSimplificationWithRemainder
+    ( WarnSimplificationWithRemainder
+    )
+
+data Registry =
+    Registry
+    { textToType :: Map Text SomeTypeRep
+    , typeToText :: Map SomeTypeRep Text
```
ttuegel: Why do we need to store `typeToText`?

ana-pantilie: We need to print the entry type information when logging (see https://github.com/ana-pantilie/kore/blob/447988ddb59f8c0ec0a2b22bdb1877d36fd5f36d/kore/src/Kore/Logger/Output.hs#L337).

kore/src/Kore/Logger/Registry.hs
```diff
@@ -0,0 +1,164 @@
...
+import Kore.Logger.WarnFunctionWithoutEvaluators
+    ( WarnFunctionWithoutEvaluators
+    )
+import Kore.Logger.WarnSimplificationWithRemainder
+    ( WarnSimplificationWithRemainder
+    )
+
+data Registry =
+    Registry
+    { textToType :: Map Text SomeTypeRep
```
ttuegel: These fields should be strict.

PR #1378, opened 2019-12-20
Workaround for heap size explosion

ttuegel:
@yzhang90 I'm moving the workaround to this branch so that I can merge the other pull request.

---

###### Reviewer checklist


kore/src/ListT.hs
```diff
@@ -83,81 +83,79 @@ newtype ListT m a =
...
 
 instance Applicative (ListT m) where
-    pure a = ListT $ \yield -> yield a
     {-# INLINE pure #-}
+    pure = \a -> ListT $ \yield -> yield a
 
-    (<*>) fs as =
-        ListT $ \yield -> foldListT fs $ \f -> foldListT as $ \a -> yield (f a)
     {-# INLINE (<*>) #-}
+    (<*>) = \fs as ->
```
virgil-serbanuta: I'm trying to learn some of this Haskell magic performance optimization, so: Why does this help?

ttuegel: GHC will only inline a [fully-applied function](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=inline#pragma-INLINE). The reason is simply that, to inline a partially-applied function, the compiler must introduce a lambda-expression, and doing that automatically might lead to a space explosion at runtime. By introducing the lambda-expression manually, this tells the compiler, "Yes, please really inline this _no matter what_."

Actually, I wasn't trying to get the compiler to inline this. When profiling, you can manually add pragmas `{-# SCC someName #-}` to the things you want to see in the profile, or the compiler can do it automatically based on some criteria. The `-fprof-auto-top` option tells GHC: I want to see every top-level definition in the profile, _except_ those marked INLINE. I was still seeing `(<*>)` in the profile, so I was trying to inline it more aggressively to hide it. (I don't think it's very interesting.) However, that didn't work because there is a [bug](https://gitlab.haskell.org/ghc/ghc/issues/17573) which causes all instance methods to get automatic SCCs, even if marked INLINE. But, I hadn't reported that bug yet! So I was still trying to inline this.

virgil-serbanuta: Thank you!

PR #1381, opened 2019-12-21
Reduce inappropriate laziness

### LINKED ISSUE #1335: Potential space leakage of Kore-repl/Kore-exec

ttuegel:
Careful reading of the profiling report yielded up to an 80% reduction in max residency. In particular, I noticed places where the call graph looked like this:

```
  parent
    child
      ...
        ...
          parent
```

That is, somewhere deep inside `child`, the program is re-entering `parent`. This is a red-flag for a laziness problem unless one of these conditions holds:

1. `parent` is passed as an argument to `child`
2. `parent` and `child` are mutually-recursive
3. `parent` passes to `child` a gradually-unfolded lazy data structure

Basically, unless you really intend laziness, this pattern indicates a problem. This pattern was unknown to me, so I'm calling it a "re-entrant call graph".

None of these conditions held here, so I applied strictness. In particular, I noticed that:

1. `removeDestinationWorker` calls `askSimplifierTermLike` (:+1:) which eventually descends back into `removeDestinationWorker` (:thinking:)
1. `runSimplifier` wraps the simplifier which eventually calls `evaluateApplication` (:+1:) which descends back into `runSimplifier` (:thinking:)

To fix `removeDestinationWorker` I lifted the body of that function so that it would run in `simplifier` instead of `TransitionT _ simplifier`. `TransitionT` is `ListT`-based, so it can be prone to these problems. My comments earlier today about removing `BranchT` were not far off: if we removed the `ListT`-based things, this could not happen. However, I am very pleased to solve the issue more tactically.

While I was working on `removeDestinationWorker`, I noticed that the `ProofGoal` data type was lazy; I made it strict to keep with our code guidelines and prevent future problems, but it is not related to this problem.

To fix `runSimplifier`, I applied bang-patterns to the `Env` (environment) so that it is fully evaluated before execution begins. After that, the call stack was still re-entrant. I observed that the re-entrant call to `runSimplifier` always appeared above `axiomPatternsToEvaluators`. Upon examining that function, I noticed that it uses `Data.Map` instead of `Data.Map.Strict`; correcting the module's imports finally resolved the re-entrant call stack.

Resolving these two bugs reduces max residency by up to 80% (measured on the `tests/specs/erc20/ds/balanceOf-spec.k` spec in `evm-semantics` with `--depth 2`). I also observed that max residency only increases by 1 MB between `--depth 2` and `--depth 4` on that proof, which satisfies me that the memory use is now appropriate.

To prevent this problem from occurring again, I have added heap limits to the `evm-semantics` and `wasm-semantics` integration tests in our repository. Changes that cause the heap limits to be exceeded will fail in CI. (@ehildenb We could also upstream the heap limits, if you prefer.)

To prevent problems like accidental use of lazy `Data.Map`, I recommend we add `hlint` to our CI checks. We can use `hlint` to forbid importing certain modules. (It allows to make individual exceptions, of course.)

Fixes #1335 

---

###### Reviewer checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -29,6 +29,9 @@
 # - functions:
 #   - {name: unsafePerformIO, within: []} # unsafePerformIO can only appear in no modules
 
+- modules:
+  - {name: [Data.Map], within: []}
+
 
 # Add custom hints for this project
 #
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,10 +80,6 @@ pipeline {
         sh '''
           ./scripts/integration-kevm.sh
         '''
-        archiveArtifacts 'kevm-add0-stats.json'
-        archiveArtifacts 'kevm-pop1-stats.json'
-        archiveArtifacts 'kevm-sum-to-10-stats.json'
-        archiveArtifacts 'kevm-sum-to-n-spec-stats.json'
       }
     }
     stage('Integration: KWASM') {
@@ -94,10 +90,6 @@ pipeline {
         sh '''
           ./scripts/integration-kwasm.sh
         '''
-        archiveArtifacts 'kwasm-simple-arithmetic-spec-stats.json'
-        archiveArtifacts 'kwasm-loops-spec-stats.json'
-        archiveArtifacts 'kwasm-memory-symbolic-type-spec-stats.json'
-        archiveArtifacts 'kwasm-locals-spec-stats.json'
       }
     }
     stage('Update K Submodules') {
--- a/kore/app/parser/Main.hs
+++ b/kore/app/parser/Main.hs
@@ -9,7 +9,7 @@ import Control.Monad.Trans
 import Control.Monad.Trans.Reader
     ( runReaderT
     )
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
 import Data.Semigroup
     ( (<>)
     )
--- a/kore/app/share/GlobalMain.hs
+++ b/kore/app/share/GlobalMain.hs
@@ -40,10 +40,10 @@ import Data.Function
 import Data.List
     ( intercalate
     )
-import Data.Map
+import Data.Map.Strict
     ( Map
     )
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
 import Data.Semigroup
     ( (<>)
     )
--- a/kore/bench/exec/Main.hs
+++ b/kore/bench/exec/Main.hs
@@ -9,7 +9,7 @@ import Data.Limit
     ( Limit
     )
 import qualified Data.Limit as Limit
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
 import Data.Maybe
     ( fromMaybe
     )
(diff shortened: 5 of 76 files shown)


ttuegel:
Strangely, it seems the heap limits need to be higher on Jenkins. :thinking: It should be alright for me to increase them because the explosion is so large when there is a problem.

virgil-serbanuta:
How would I get this kind of profiling report?

ehildenb:
Awesome! Well, let's leave the heap limits here, because I think the proof tests in KEVM/KWasm are more likely to need to change than the ones here. Good to have a test that memory usage doesn't get worse.

ttuegel:
> How would I get this kind of profiling report?

I set `GHCRTS='-p'` in the environment to generate a `kore-exec.prof` when the test runs. I analyzed the profile with [`profiteur`](https://github.com/jaspervdj/profiteur) to visually navigate the call graph. Starting at the root of the graph, I navigated down the paths with the largest cumulative allocation.

virgil-serbanuta:
> > How would I get this kind of profiling report?
> 
> I set `GHCRTS='-p'` in the environment to generate a `kore-exec.prof` when the test runs. I analyzed the profile with [`profiteur`](https://github.com/jaspervdj/profiteur) to visually navigate the call graph. Starting at the root of the graph, I navigated down the paths with the largest cumulative allocation.

Thank you!

PR #1385, opened 2019-12-24
Add log entry for ProofState (v2)

### LINKED ISSUE #1358: Log entries for proof states
`transitionRule` should emit an `Entry` for each `ProofState`.

ana-pantilie:
Fixes #1358

Also fixes:
- determining the sorts in `onePathRuleToTerm`/`allPathRuleToTerm`
- forces main thread to wait for logging to finish (introduced in https://github.com/kframework/kore/pull/1387)

---

###### Reviewer checklist


kore/test/Test/Kore/Strategies/AllPath/AllPath.hs
```diff
@@ -381,15 +381,96 @@ instance Debug (Goal.Rule Goal)
 
 instance Diff (Goal.Rule Goal)
 
+transitionRuleTemplate
```
virgil-serbanuta: Why does the function with the same name in `Goal.hs` log transition information, while this one doesn't?

ana-pantilie: @virgil-serbanuta This was my original problem, that the `Goal` type for these tests does not implement `ToRulePattern`, and since logging depends on this representation we need separate `transitionRuleTemplate`s for those `goal`s which are `ToRulePattern` and those which aren't.

virgil-serbanuta: Ok, maybe duplicating the function is not so bad, but here are some suggestions, would any of them be better? 

* implement `ToRulePattern` with an `error` if it's possible somehow to make it trigger only when actually logging.
* implement `ToRulePattern` with something trivial like `\top => \top`
* put the logging functions themselves in a class (in the Goal? a new one?) - I don't like this one that much, really, but it may work.

ana-pantilie: I think your second suggestion would be the best solution if we didn't want two `transitionRuleTemplate`. Maybe @ttuegel also has an opinion on this?

kore/src/Kore/Logger/DebugProofState.hs
```diff
@@ -0,0 +1,65 @@
...
+    )
+import Kore.Strategies.ProofState
+    ( Prim (..)
+    , ProofState (..)
+    )
+
+data DebugProofState =
+    DebugProofState
+        { configuration :: !TransitionState
+        , transition :: !(Prim (RulePattern Variable))
```
ttuegel: Why `RulePattern` instead of `ReachabilityRule`?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1077,3 +1096,33 @@ configurationDestinationToRule ruleType configuration rhs =
...
+    => Prim goal ~ ProofState.Prim (Rule goal)
+    => ProofState goal goal
+    -- ^ current configuration
+    -> Prim goal
+    -- ^ transition
+    -> log (ProofState goal goal)
+    -- ^ action to be computed
+    -> log (ProofState goal goal)
+debugProofStateBracket
+    (fmap toRulePattern -> config)
```
ttuegel: I think we should call this `proofState` instead of `config`.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1077,3 +1096,33 @@ configurationDestinationToRule ruleType configuration rhs =
...
+    (fmap toRulePattern -> transition)
+    action
+  = do
+    logM DebugProofState
+        { configuration = Before config
+        , transition
+        }
+    result <- action
+    logM DebugProofState
```
ttuegel: We never log an `After` without a `Before`, right? I think the log would be more useful if we logged the initial and final proof states together. Otherwise, I don't think there is an easy way to associate them with each other. Could we change the entry type to be `DebugProofState { initialProofState, finalProofState, transition }`?

kore/src/Kore/Logger/DebugProofState.hs
```diff
@@ -0,0 +1,53 @@
...
+    , RewriteRule (..)
+    )
+import Kore.Strategies.ProofState
+    ( Prim (..)
+    , ProofState (..)
+    )
+
+data DebugProofState =
+    DebugProofState
+        { proofstate :: !(ProofState (ReachabilityRule Variable))
```
ttuegel:
```suggestion
        { proofState :: !(ProofState (ReachabilityRule Variable))
```

kore/test/Test/Kore/Strategies/AllPath/AllPath.hs
ttuegel: My suggestion would be: instead of modifying the existing `transitionRuleTemplate` to implement logging, implement a function that adds logging to the existing `transitionRuleTemplate`. That is:

```.hs
-- I'm listing this for brevity, but it's not necessary for you to add this type synonym
type TransitionRule monad goal =
        Prim goal
    ->  ProofState goal goal
    ->  TransitionT (Rule goal) monad (ProofState goal goal)

-- existing function:
transitionRuleTemplate :: TransitionRuleTemplate monad goal -> TransitionRule monad goal

-- new function:
withDebugProofState 
    :: MonadLog monad
    => ToRulePattern goal
    => TransitionRule monad goal
    -> TransitionRule monad goal
withDebugProofState = _
-- might need a better name
```

If I understand correctly, we could then use the same `transitionRuleTemplate` in the tests as in the prover, right?

kore/src/Kore/Logger/DebugProofState.hs
ana-pantilie: It would clash with https://github.com/kframework/kore/blob/d09b5e74e0a87ba613a7ce786618f6dcfa81b801/kore/src/Kore/Strategies/ProofState.hs#L106

virgil-serbanuta: Where? I.e. are you talking about a specific clash, or about the fact that the clash could occur? Usually qualified imports for record fields (as recommended by our coding standards) or importing only what's needed are enough to solve this issue, but it may not work in all cases.

ana-pantilie: Hmm yes, you're right, sorry. I was just thinking it might be confusing to have two things named `proofState` which are both related to the same type, but after thinking about this a bit more `proofstate` might be more error-prone.

kore/src/Kore/Logger/DebugProofState.hs
```diff
@@ -0,0 +1,53 @@
+{- |
+Copyright   : (c) Runtime Verification, 2019
```
virgil-serbanuta: 2020?

kore/src/Kore/Logger/Output.hs
```diff
@@ -223,20 +229,26 @@ runLoggerT :: KoreLogOptions -> LoggerT IO a -> IO a
 runLoggerT options loggerT = do
     let runLogger = runReaderT . getLoggerT $ loggerT
     withLogger options $ \logger -> do
-        modifiedLogger <- concurrentLogger logger
-        runLogger modifiedLogger
+        (asyncThread, modifiedLogger) <- concurrentLogger logger
+        result <- runLogger modifiedLogger
+        wait asyncThread
```
ttuegel: We should use [`finally`](https://hackage.haskell.org/package/exceptions-0.10.4/docs/Control-Monad-Catch.html#v:finally) so that if `runLogger modifiedLogger` throws an exception, the logging thread gets a chance to finish writing.

Issue #1132, opened 2019-10-22
Catch and log errors


ttuegel:
Presently an error looks like this:

```
kore-exec: z3: runInteractiveProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

It is prefixed with `kore-exec:` so at least the user knows where it is coming from, but it will not appear in the log, and there is no explicit indication that this is an error. The logger should catch errors at the top-level and insert them into the log before terminating.

Obviously, this should be done only after we fix the performance problem with logging and enable it by default.

PR #1386, opened 2019-12-27
Catch and log errors

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1132

---

###### Reviewer checklist


kore/src/Kore/Logger/CriticalExecutionError.hs
```diff
@@ -0,0 +1,39 @@
...
+newtype CriticalExecutionError =
+    CriticalExecutionError
+    { executionError :: SomeException
+    }
+
+instance Pretty CriticalExecutionError where
+    pretty CriticalExecutionError { executionError } =
+        pretty . displayException $ executionError
+
+instance Entry CriticalExecutionError where
```
andreiburdusa: Could you explain me why did you add this instance?

ana-pantilie: @andreiburdusa When we want to log something new, we add a new type which implements the `Entry` class so that we can log it (see https://github.com/kframework/kore/blob/504d615fddb8a3705de36337a78e542a61710a9d/kore/src/Kore/Logger.hs#L243). The reason we make new types for different logging information (instead of just using `LogMessage`, see the functions here https://github.com/kframework/kore/blob/504d615fddb8a3705de36337a78e542a61710a9d/kore/src/Kore/Logger.hs#L148) is that we can use the types themselves to provide additional information in the log and to filter what gets printed. That's my understanding of our logger.

kore/app/exec/Main.hs
```diff
@@ -404,12 +407,13 @@ main = do
...
     let KoreExecOptions { koreLogOptions } = execOptions
-    exitCode <- catch (runLoggerT koreLogOptions go) $
+    exitCode <- runLoggerT koreLogOptions $ catch go $
         \(Goal.WithConfiguration lastConfiguration someException) -> do
-            renderResult
+            liftIO $ renderResult
                 execOptions
                 ("// Last configuration:\n" <> unparse lastConfiguration)
-            throwM someException
+            criticalExecutionError someException
```
ttuegel: This actually needs to be lifted out to a higher handler because we might have an exception before there is any configuration. The body of the `mainWithOptions` should look something like,

```.hs
    exitCode <- runLoggerT koreLogOptions
        $ handle handleSomeException
        $ handle handleWithConfiguration
        $ go
```

Issue #1364, opened 2019-12-18
Write log in a separate thread


ttuegel:
---

- [ ] A transformer, `LogAction IO a -> IO (LogAction IO a)` which creates a `TChan`, runs the original `LogAction` in another thread reading from the `TChan`, and returns a `LogAction` that puts entries into the `TChan`.

PR #1387, opened 2019-12-27
Write log in a separate thread

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1364

---

###### Reviewer checklist


kore/src/Kore/Logger/Output.hs
```diff
@@ -207,7 +217,29 @@ filterSeverity level entry =
...
+runLoggerT options loggerT = do
+    let runLogger = runReaderT . getLoggerT $ loggerT
+    withLogger options $ \logger -> do
+        modifiedLogger <- concurrentLogger logger
+        runLogger modifiedLogger
+
+concurrentLogger :: LogAction IO a -> IO (LogAction IO a)
+concurrentLogger logger = do
+    tChan <- newTChanIO
+    _ <- forkIO $ do
```
ttuegel: This will read one message from the `TChan`, but we want the other thread to run forever:
```suggestion
    _ <- forkIO $ forever $ do
```

kore/src/Kore/Logger/Output.hs
```diff
@@ -207,7 +217,29 @@ filterSeverity level entry =
...
+            val <- atomically $ readTChan tChan
+            logger Colog.<& val
+            return ()
```
ttuegel: ... following the suggestion above, we don't need to explicitly `return ()` here.
```suggestion
```

kore/src/Kore/Logger/Output.hs
```diff
@@ -207,7 +217,29 @@ filterSeverity level entry =
...
+    return $ writeTChanLogger tChan logger
+
+writeTChanLogger :: TChan a -> LogAction IO a -> LogAction IO a
+writeTChanLogger tChan logger =
```
ttuegel: Here we should *only* write to the `TChan`, instead of sending to both loggers:

```suggestion
writeTChanLogger tChan = LogAction $ \msg -> atomically $ writeTChan tChan msg
```

PR #1398, opened 2020-01-06
Fix ConstructorLike for maps

### LINKED ISSUE #1393: `.Map` is not considered as concrete
In evm-semantics, I marked the following rule as concrete:
```
    syntax Map ::= Map "[" Int ":=" ByteArray "]" [function, klabel(mapWriteBytes)]
 // -------------------------------------------------------------------------------
    rule WM[ N := WS ] => WM [ N := WS, 0, #sizeByteArray(WS) ] [concrete]
```
During the execution, the prover reported the following error:
```
Stepper evaluation errored.
WithConfiguration ...
Unification case that should be handled somewhere else:
attempting normalized unification with a 
non-element-variable opaque term or 
non-function maps could lead to infinite loops.
first=/* builtin: */ Lbl'Stop'Map{}()
second=Lbl'UndsLSqBUndsColnEqlsUndsRSqBUnds'EVM-TYPES'Unds'Map'Unds'Map'Unds'Int'Unds'ByteArray{}(
    /* builtin: */ Lbl'Stop'Map{}(),
    /* builtin: */ \dv{SortInt{}}("64"),
    \dv{SortBytes{}}(
        "0000000000000000000000000000000000000000000000000000000000000060"
    )
)
```
and I found that term `.Map [ 64 := b"0000000000000000000000000000000000000000000000000000000000000060" ]` is not evaluated in the configuration.

virgil-serbanuta:
Fixes #1393 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
+++ b/kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
@@ -7,10 +7,12 @@ License     : NCSA
 module Kore.Attribute.Pattern.ConstructorLike
     ( ConstructorLike (..)
     , ConstructorLikeHead (..)
+    , HasConstructorLike (..)
     ) where
 
 import Control.DeepSeq
 import Data.Hashable
+import qualified Data.Map.Strict as Map
 import Data.Maybe
     ( isJust
     )
@@ -64,7 +66,7 @@ Examples of patterns that are not constructor-like:
 -}
 newtype ConstructorLike =
     ConstructorLike
-        { isConstructorLike :: Maybe ConstructorLikeHead
+        { getConstructorLike :: Maybe ConstructorLikeHead
         }
     deriving (Eq, GHC.Generic, Ord, Show)
 
@@ -130,7 +132,7 @@ means constructor-like here).
  -}
 instance Synthetic ConstructorLike (DomainValue sort) where
     synthetic domainValue
-        | isJust . isConstructorLike $ child =
+        | isJust . getConstructorLike $ child =
             ConstructorLike . Just $ ConstructorLikeHead
         | otherwise =
             ConstructorLike Nothing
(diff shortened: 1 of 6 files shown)


kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
```diff
@@ -193,15 +195,42 @@ A builtin value is not technically a constructor, but it is constructor-like for
...
+            { elementsWithVariables = []
+            , concreteElements
+            , opaque = []
+            }
+              | all pairIsConstructorLike concreteElementsList
+                -> ConstructorLike . Just $ ConstructorLikeHead
+              where
+                concreteElementsList = Map.toList concreteElements
+                pairIsConstructorLike (key, value) =
+                    isConstructorLike key && isConstructorLike value
```
ttuegel: Why do we need to check the `key`? It should always be constructor-like. If we are going to check, I think it should at least be an assertion, but if we don't check, then we also don't need the `HasConstructorLike` abstraction.

virgil-serbanuta: We don't, but I thought it's safer to check it, especially since the requirements for the key changed in the past. You're right, an assertion makes more sense, so I changed it to that.

yzhang90:
it solves my problem. thanks

kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
```diff
@@ -245,3 +274,23 @@ instance Diff ConstructorLikeHead where
...
 instance Hashable ConstructorLikeHead
+
+class HasConstructorLike a where
+    extractConstructorLike :: a -> ConstructorLike
+
+    isConstructorLike :: a -> Bool
+    isConstructorLike a = case extractConstructorLike a of
+        (ConstructorLike constructorLike) -> isJust constructorLike
+
+    assertConstructorLike :: String -> a -> b -> b
```
ttuegel: Does this need to be a member of the class? Would we ever want to override it?

kore/test/Test/Kore/Attribute/Pattern/ConstructorLike.hs
```diff
@@ -3,31 +3,53 @@ module Test.Kore.Attribute.Pattern.ConstructorLike
     ) where
 
 import Test.Tasty
-import Test.Tasty.HUnit
+--import Test.Tasty.HUnit
```
ttuegel:
```suggestion
```

PR #1404, opened 2020-01-07
Wrap side conditions in their own type

virgil-serbanuta:
This PR is mostly a refactoring which attempts to make the code more readable and safer. The current code uses many distinct names for the side condition, and places it randomly in function argument lists.

While this PR does not solve the random placement, it uses the same name everywhere (hopefully a more descriptive one) and wraps the condition is a newtype, which makes it obvious, say, which argument is the side condition and which is the condition being simplified.

This made it possible to discover some issues in the various `makeAnd` implementations we use for unification.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Internal/SideCondition.hs
+++ b/kore/src/Kore/Internal/SideCondition.hs
@@ -5,7 +5,7 @@ License     : NCSA
 
 module Kore.Internal.SideCondition
     ( SideCondition  -- Constructor not exported on purpose
-    , addAssumedTrue
+    , andCondition
     , assumeTrueCondition
     , assumeTruePredicate
     , mapVariables
@@ -40,6 +40,13 @@ import Kore.Unparser
     ( Unparse (..)
     )
 
+{-| Side condition used in the evaluation context.
+
+It is not added to the result.
+
+It is usually used to remove infeasible branches, but it may also be used for
+other purposes, say, to remove redundant parts of the result predicate.
+-}
 newtype SideCondition variable =
     SideCondition
         { assumedTrue :: Condition variable }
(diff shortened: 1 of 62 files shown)


kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -0,0 +1,123 @@
...
+top :: InternalVariable variable => SideCondition variable
+top = SideCondition
+    { assumedTrue = Condition.top
+    }
+
+-- | A 'top' 'Condition' for refactoring which should eventually be removed.
+topTODO :: InternalVariable variable => SideCondition variable
+topTODO = top
+
+addAssumedTrue
```
ttuegel: I would (slightly) prefer to call this `andCondition` both because we are lifting `Condition.andCondition` through the `newtype` and because it has the logical operation in the name.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -0,0 +1,123 @@
...
+    => SideCondition variable
+    -> Condition variable
+    -> SideCondition variable
+addAssumedTrue sideCondition@SideCondition {assumedTrue} newCondition =
+    sideCondition
+        { assumedTrue = assumedTrue `Condition.andCondition` newCondition }
+
+assumeTrueCondition
+    :: InternalVariable variable => Condition variable -> SideCondition variable
+assumeTrueCondition condition =
```
ttuegel: Why not just use the constructor here?

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -0,0 +1,123 @@
...
+    ( InternalVariable
+    )
+import Kore.TopBottom
+    ( TopBottom (..)
+    )
+import Kore.Unparser
+    ( Unparse (..)
+    )
+
+newtype SideCondition variable =
```
ttuegel: Could you add a comment here to indicate something like: this is a `Condition` used in the evaluation context which is not added to the result?

virgil-serbanuta:
> This looks great, a big improvement!
> 
> Not that you need to take on this task now (or ever), but we could improve the "inconsistent argument order" problem by making every function that takes a `SideCondition` argument and returns a `MonadSimplify simplifier => simplifier something` into a `ReaderT (SideCondition variable) simplifier something`. For example, the signature of `simplifyToOr` would change as follows:
> 
> ```haskell
> simplifyToOr :: (...) => SideCondition variable -> TermLike variable -> simplifier (OrPattern variable)
> simplifyToOr :: (...) => TermLike variable -> ReaderT (SideCondition variable) simplifier (OrPattern variable)
> ```
> 
> The argument order would be consistent _and_ the side conditions would propagate automatically.

I thought about that, but it seemed like a larger change than wrapping some function arguments. I'll merge this PR as-is, and I'll try to find the time for a follow-up PR.

PR #1405, opened 2020-01-07
Regression tests

ttuegel:
The KEVM and KWASM integration tests are not proper regression tests because upstream changes can break our tests. This pull request would "freeze" the upstream dependencies.

---

###### Reviewer checklist


scripts/kollect.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
virgil-serbanuta: Could you add a comment describing what this script does?

test/regression-evm-semantics-39dd1e5151e7fb5b2047b6cacab3f414757464c7/generate.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
virgil-serbanuta: Could you add a comment describing what this script does?

PR #1409, opened 2020-01-08
No branching for functions

### LINKED ISSUE #1065: Function evaluation without configuration splitting
Here is how we evaluate functions:

```
@virgil-serbanuta : if we evaluate `sigma(f(x), g(y))`, we first evaluate `f(x)` to `F(x) /\ Pf(x)` and `g(y)` to `G(y) /\ Pg(y)`.
We will then evaluate `sigma(F(x), G(y))`. The child predicate while evaluating this is `Pf(x)` and `Pg(y)`.
so then, if `Pt(x, y)` is the top-level predicate, and we evaluate `sigma(F(x), G(y))` to `S(x, y) /\ Ps(x, y)`.
We will send `Ps(x, y)` to z3 with `Pt(x, y)` and `Pf(x)` and `Pg(y)` as a side predicate. 
We will and the main predicate and the side one before sending them to z3.
```
But the current approach doesn't fully take the advantage of passing top level predicate down and it will still split the configuration when doing function evaluation.

@ttuegel and I have the following proposal (also related to #945 ):
suppose we have two function eval rules:
```
rule f(x) => f1(x) requires x >=0
rule f(x) => f2(x) requires x<=0
```
the current config is `C(f(x)) /\ \top`.
`f(x)` can potentially have two results `f1(x) /\ x >=0` or `f2(x) /\ x <0`, when passing the top level predicate down, we check the validity of  `top -> x >=0` and `top -> x < 0`. In this example, none of the validity holds and then we still return `C(f(x)) /\ top` as the result and keep internally two possible outcomes of f(x) (using #945 ).
When semantics rules causes new constraint to be added to the configuration(config split), for example, the current configuration becomes `C(f(x)) /\ x >= 5`. Now, we know `x >= 5 -> x >=0`, and we can safely return `C(f1(x)) /\ x >= 5` as the result.

Suppose that the top level predicate is `Pt` and the side condition of the rule is `Ps`.
1. The function evaluation is different from applying semantics rules in the sense that:
For semantics rules, we should check if `Pt /\ Ps` is sat (if sat, we can apply the semantics rule);
For function rules, we should check if `Pt /\ not(Ps)` is unsat (if unsat, then Pt -> Ps and we can be sure that only this function rule can be applied and it will not cause the split of configuration)
2. Function evaluation can be localized. when evaluating `sigma(f(x), g(y))`, we first evaluate the arguments `f(x)` and `g(y)`. And when we evaluate `sigma`, we should only consider top level predicate `Pt` and we don't need to consider the child predicate generated from evaluating `f` and `g`.

virgil-serbanuta:
Fixes #1065 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Pattern.hs
+++ b/kore/src/Kore/Attribute/Pattern.hs
@@ -35,6 +35,9 @@ import Data.Hashable
     )
 import qualified Generics.SOP as SOP
 import qualified GHC.Generics as GHC
+import GHC.Stack
+    ( HasCallStack
+    )
 
 import Kore.Attribute.Pattern.ConstructorLike
 import Kore.Attribute.Pattern.Created
@@ -130,26 +133,33 @@ instance HasConstructorLike (Pattern variable) where
         Pattern {constructorLike}
       = constructorLike
 
-simplifiedAttribute :: Pattern variable -> Simplified
-simplifiedAttribute Pattern {constructorLike, simplified}
-  | Simplified.isFullySimplified simplified = simplified
-  | isConstructorLike constructorLike =
-    error "Inconsistent attributes, constructorLike implies fully simplified."
-  | otherwise = simplified
+simplifiedAttribute :: HasCallStack => Pattern variable -> Simplified
+simplifiedAttribute patt@Pattern {simplified} =
+    assertSimplifiedConsistency patt simplified
 
-isSimplified :: SideCondition.Representation -> Pattern variable -> Bool
-isSimplified sideCondition Pattern {constructorLike, simplified}
-  | Simplified.isFullySimplified simplified = True
-  | isConstructorLike constructorLike =
-    error "Inconsistent attributes, constructorLike implies fully simplified."
-  | otherwise = Simplified.isSimplified sideCondition simplified
+{- Checks whether the pattern is simplified relative to the given side
+condition.
+-}
+isSimplified
+    :: HasCallStack
+    => SideCondition.Representation -> Pattern variable -> Bool
+isSimplified sideCondition patt@Pattern {simplified} =
+    assertSimplifiedConsistency patt
+    $ Simplified.isSimplified sideCondition simplified
 
-isFullySimplified :: Pattern variable -> Bool
-isFullySimplified Pattern {constructorLike, simplified}
-  | Simplified.isFullySimplified simplified = True
-  | isConstructorLike constructorLike =
+{- Checks whether the pattern is simplified relative to any side condition.
+-}
+isFullySimplified :: HasCallStack => Pattern variable -> Bool
+isFullySimplified patt@Pattern {simplified} =
+    assertSimplifiedConsistency patt
+    $ Simplified.isFullySimplified simplified
+
+assertSimplifiedConsistency :: HasCallStack => Pattern variable -> a -> a
+assertSimplifiedConsistency Pattern {constructorLike, simplified}
+  | isConstructorLike constructorLike
+  , not (Simplified.isFullySimplified simplified) =
     error "Inconsistent attributes, constructorLike implies fully simplified."
-  | otherwise = False
+  | otherwise = id
 
 setSimplified :: Simplified -> Pattern variable -> Pattern variable
 setSimplified simplified patt = patt { simplified }
--- a/kore/src/Kore/Attribute/Pattern.hs
+++ b/kore/src/Kore/Attribute/Pattern.hs
@@ -8,10 +8,11 @@ License     : NCSA
 
 module Kore.Attribute.Pattern
     ( Pattern (Pattern, patternSort, freeVariables, functional, function, defined, created, constructorLike)
-    -- 'simplified' was intentionally left out above.
+    -- 'simplified' and 'constructorLike' were intentionally left out above.
     , mapVariables
     , traverseVariables
     , deleteFreeVariable
+    , isFullySimplified
     , isSimplified
     , setSimplified
     , simplifiedAttribute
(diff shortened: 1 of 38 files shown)


kore/src/Kore/Attribute/Pattern.hs
```diff
@@ -115,13 +132,24 @@ instance HasConstructorLike (Pattern variable) where
...
-    | isConstructorLike constructorLike = Simplified
-    | otherwise = simplified
-
-isSimplified :: Pattern variable -> Bool
-isSimplified Pattern {constructorLike, simplified}
-  | isConstructorLike constructorLike = True
-  | otherwise = isFullySimplified simplified
+  | Simplified.isFullySimplified simplified = simplified
+  | isConstructorLike constructorLike =
+    error "Inconsistent attributes, constructorLike implies fully simplified."
```
traiansf: why not reusing `isFullySimplified` here and in `isSimplified` below?

kore/src/Kore/Attribute/Pattern.hs
```diff
@@ -115,13 +132,24 @@ instance HasConstructorLike (Pattern variable) where
...
+  | otherwise = simplified
```
traiansf: since this returns simplified whenever it does not error, it seems more like the combination of an assert and a getter.

kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -35,17 +136,39 @@ pattern.
...
 simplified status is reset by any substitution under the pattern.
-
 -}
-data Simplified = Simplified | NotSimplified
+data Simplified
+    = Simplified !(Type, Condition)
+    | NotSimplified
+    deriving (Eq, GHC.Generic, Ord, Show)
+
+{- |
```
traiansf: is this still needed?

kore/src/Kore/Attribute/Pattern.hs
```diff
@@ -115,13 +132,24 @@ instance HasConstructorLike (Pattern variable) where
...
+
+isSimplified :: SideCondition.Representation -> Pattern variable -> Bool
```
traiansf: documentation?

kore/src/Kore/Attribute/Pattern.hs
```diff
@@ -115,13 +132,24 @@ instance HasConstructorLike (Pattern variable) where
...
+isSimplified sideCondition Pattern {constructorLike, simplified}
+  | Simplified.isFullySimplified simplified = True
+  | isConstructorLike constructorLike =
+    error "Inconsistent attributes, constructorLike implies fully simplified."
+  | otherwise = Simplified.isSimplified sideCondition simplified
+
+isFullySimplified :: Pattern variable -> Bool
```
traiansf: documentation

kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -60,16 +183,57 @@ instance NFData Simplified
 
 instance Hashable Simplified
 
+simplifiedTo :: HasCallStack => Simplified -> Simplified -> Simplified
```
traiansf: documentation. here is even more needed than above.

kore/src/Kore/Internal/Condition.hs
```diff
@@ -6,9 +6,9 @@ License     : NCSA
 module Kore.Internal.Condition
     ( Condition
     , isSimplified
-    , markSimplified
-    , setPredicateSimplified
     , simplifiedAttribute
+    , markPredicateSimplified
```
traiansf: maybe documentation?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -433,35 +444,98 @@ assertConstructorLikeKeys keys a
                 , Pretty.indent 2 "Simplifiable keys:"
                 ]
                 <> fmap (Pretty.indent 4 . unparse) simplifiableKeys
+    | any (not . isFullySimplified) keys =
+        let simplifiableKeys =
+                filter (not . isFullySimplified) $ Foldable.toList keys
+        in
+            (error . show . Pretty.vsep) $
+                [ "Map and Set may only contain simplified \
+                  \keys (resp. elements)."
```
traiansf: why "resp."?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -433,35 +444,98 @@ assertConstructorLikeKeys keys a
...
+  where
+    mergedSimplified =
+        foldMap (Attribute.simplifiedAttribute . extractAttributes) termLikeF
+
+checkedSimplifiedFromChildren
+    :: (GHC.HasCallStack, InternalVariable variable)
+    => TermLikeF variable (TermLike variable) -> Pattern.Simplified
+checkedSimplifiedFromChildren termLikeF =
+    case simplifiedFromChildren termLikeF of
+        Pattern.NotSimplified -> error
```
traiansf: maybe extract error expression as separate function?

kore/src/Kore/Attribute/Pattern.hs
```diff
@@ -8,10 +8,11 @@ License     : NCSA
 
 module Kore.Attribute.Pattern
     ( Pattern (Pattern, patternSort, freeVariables, functional, function, defined, created, constructorLike)
-    -- 'simplified' was intentionally left out above.
+    -- 'simplified' and 'constructorLike' were intentionally left out above.
```
ttuegel: `constructorLike` was exported, though. Why are these hidden?

kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -35,17 +147,21 @@ pattern.
 
 Most patterns are assumed un-simplified until marked otherwise, so the
 simplified status is reset by any substitution under the pattern.
-
 -}
-data Simplified = Simplified | NotSimplified
+data Simplified
+    = Simplified !(Type, Condition)
```
ttuegel: `(,)` is not strict in its arguments, so this might not do what you want.

kore/src/Kore/Internal/SideCondition/SideCondition.hs
```diff
@@ -0,0 +1,56 @@
...
+    , HasDatatypeInfo
+    )
+import qualified GHC.Generics as GHC
+
+import Kore.Debug
+    ( Debug
+    , Diff (..)
+    )
+
+data Representation =
```
ttuegel: We could also get this from

```.hs
newtype Representation = Representation { getRepresentation :: Hashed Text }
```

and most of the instances would come for free, except `Debug` and `Diff` (and we can define catch-all instances for those in `Debug`).

PR #1417, opened 2020-01-10
Fixes #1412 (removalPredicate now uses unification instead of and simplification)

### LINKED ISSUE #1412: removeDestination should use unification
Fix this TODO:

https://github.com/kframework/kore/blob/11769a0e62d62eba17b5b5a9a9a8123745aa4bf6/kore/src/Kore/Strategies/Goal.hs#L980

If unification fails (not `\bottom`, but two patterns that aren't handled) then we should abort execution, as we do when applying rules.

traiansf:
---

###### Reviewer checklist


traiansf:
Decided not to pursue this further in this iteration.

Issue #1266, opened 2019-11-18
Structured logs


ttuegel:
We need to turn Kore patterns back into K patterns for logs and errors. This is one important reason to log _structured data_: the conversion back to K patterns can be handled in the logging apparatus so that the symbolic execution engine does not become complicated with such concerns.

ttuegel:
After discussion with @ehildenb, I think it would be ideal to expose as much structured log information as possible and rely on post-processing to digest it into something acceptable to the user. Keeping structured information as long as possible in the analysis pipeline increases the number of things we can do with it, and the number of people who can create new analysis tools.

### Prototype design

The log will be a directory with one `index` file and many entry files (one per log entry). The index will be used to put the log entries in order; each line corresponds to an entry and has the format:

```
⟨integer timestamp⟩ ⟨filename.json⟩
```

The timestamp should be the number of nanoseconds since start-up. We can just format it as a fixed-width hexademical string; that's easy to parse, and 16 hex digits covers 584 years before it rolls over, so "long enough".

The entry file should be JSON-formatted. Every entry should have a tag indicating the type; a string with the Haskell type name should be fine. Each entry should also have a reference to its "parent" (if any) for context.

ttuegel:
I hid my earlier comment because @vladciobanu talked me out of re-implementing SQLite badly. We should just emit structured logs to a SQLite database using something like `persistent-sqlite`. It's as easy as using `aeson`, but would enable us to develop very rich tools on the other end. It's also fast and compact.

ttuegel:
The first task is to evaluated different libraries for interacting with the database. The most popular library is `persistent`, but it is designed for the database to be used only by the Haskell program, which does not agree with our purposes. Therefore, we should compare it with other options.

PR #1419, opened 2020-01-10
Refactor logging

ttuegel:
Several refactoring changes to prepare for #1266:

- Separate the parts of logging that depend on Kore from the parts that don't. The parts that don't go into `module Log` and the parts that do go into `module Kore.Log`.
- Separate `KoreLogOptions` (the logging configuration) from the parts that execute logging.
- Clean up some log option parsers.
- Add `Default` instances for log option types. This ensures the same default options are consistently used throughout the code. Prior to this change, the defaults for most options were defined by their parser, which doesn't seem like the right place to have that information.

---

###### Reviewer checklist


kore/src/Kore/Repl.hs
```diff
@@ -168,14 +167,14 @@ runRepl axioms' claims' logger replScript replMode outputFile = do
             , labels     = Map.empty
             , aliases    = Map.empty
             , koreLogOptions =
-                Logger.KoreLogOptions
-                    { logType = Logger.LogStdErr
+                Log.KoreLogOptions
```
ana-pantilie: I think this is equal to the default value for `KoreLogOptions`. Do we want to use `def` here instead?

PR #1422, opened 2020-01-13
Fixed no-helpers found bug

traiansf:
@andreiburdusa  and I looked earlier over an issue where a `test-...-.sh.out` rule was not firing.

Andrei had to leave but I managed to find the issue in the fact that the dependency rule `$(TEST_DIR)/test-%-*` was requiring that at least one file matching that pattern exists.

Assuming we don't always want to provide such a file, I've modified the rule to usa a `wildcard` (and had to enable [Make Secondary Expansion](https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html).

@andreiburdusa : I hope that works.

What also confused us was the default rule of `make a.out` to copy `a` to `a.out` which was being used since matching our rule didn't work...

---

###### Reviewer checklist

### DIFF
--- a/test/include.mk
+++ b/test/include.mk
@@ -125,7 +125,7 @@ PATTERN_OPTS = --pattern "$$(cat $*.k)"
 
 ### SCRIPTS
 
-test-%.sh.out: $(KORE_EXEC) $(TEST_DIR)/test-%.sh $(TEST_DIR)/test-%-*
+test-%.sh.out: $(KORE_EXEC) $(TEST_DIR)/test-%.sh
 	@echo ">>>" $(CURDIR) $(@:.out=)
 	rm -f $@
 	$(TEST_DIR)/$(@:.out=) > $@ || true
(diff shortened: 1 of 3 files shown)


test/include.mk
```diff
@@ -125,7 +125,8 @@ PATTERN_OPTS = --pattern "$$(cat $*.k)"
 
 ### SCRIPTS
 
-test-%.sh.out: $(KORE_EXEC) $(TEST_DIR)/test-%.sh $(TEST_DIR)/test-%-*
+.SECONDEXPANSION:
+test-%.sh.out: $(KORE_EXEC) $(TEST_DIR)/test-%.sh $$(wildcard $(TEST_DIR)/test-$$*-*)
```
ttuegel: `SECONDEXPANSION` is kinda tricky; I think I would prefer just to remove the default dependency on `$(TEST_DIR)/test-%-*` and leave it to specific tests to declare their dependencies in a downstream `Makefile`.

PR #1426, opened 2020-01-13
hie.yaml: Use hie-bios.sh to setup mega-component cradle

ttuegel:
All components of package `kore` are loaded into a single environment because
ghcide does not yet support multi-cradle projects. (It's not even clear to me
that multi-cradle projects are a good idea to begin with.)

---

###### Reviewer checklist


virgil-serbanuta:
What is a "cradle"? I briefly looked at https://github.com/haskell/haskell-ide-engine/blob/master/README.md which seems to use the term, but I didn't find a definition.

ttuegel:
I answered in the daily meeting, but for the sake of completeness: A cradle (described at [hie-bios](https://github.com/mpickering/hie-bios)) is the configuration for a "workspace" of `haskell-ide-engine` or `ghcide`. Other IDEs may use the word "reactor" instead of "workspace". `hie-bios` allows a single project to define multiple cradles, that is multiple configurations which will lead to multiple workspaces. This is useful if the project defines many packages, for example. It is also useful because ordinarily a workspace can target only a single component of a single package. If there are multiple components in the package (which is common) then we might want to have a cradle for each one. However, it is not well-supported yet. I think there is another problem with this (and many people agree, but not all): any dependencies that cross the boundaries between workspaces have to be built outside of the IDE, that is, if your test suite depends on your library, then you have to `stack build` the library before you can load the test suite in the IDE, and the IDE will not detect changes to the library while you are editing the tests. So, instead of using `hie-bios`'s multi-cradle support, I am providing a script (which `hie-bios` will run automatically) which generates the correct command-line options for GHC to treat the entire package (all components) as a single mega-component.

PR #1430, opened 2020-01-15
Side conditions for substitution simplification

virgil-serbanuta:
---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -120,7 +120,8 @@ newtype SubstitutionSimplifier simplifier =
         { simplifySubstitution
             :: forall variable
             .  SubstitutionVariable variable
-            => Substitution variable
+            => SideCondition variable
```
ttuegel: Checking that I understand correctly: The "core" of the work needed to use side conditions for substitution simplification is to take this argument and pass it through to the correct simplification calls inside the substitution simplifier. Is that right?

virgil-serbanuta: That's correct.

PR #1440, opened 2020-01-16
Structured logging

ttuegel:
This implements structured logging for the simple log entry types (product types). I need to do some clean-up. I think I will implement the sum types in a separate pull request, as they are somewhat more complicated.

---

###### Reviewer checklist


ttuegel:
> Could you add some documentation showing the intended use of the new SQL package?

Do you mean like how to derive instances of the classes?

ttuegel:
I'm going to address the documentation in my subsequent pull request because the interface will be much simpler once sum types are available.

virgil-serbanuta:
> > Could you add some documentation showing the intended use of the new SQL package?
> 
> Do you mean like how to derive instances of the classes?

Yes, perhaps with some explanation about when to use certain things and when to use others.

As an example: When should one use `createTableUnwrapped` and when should one use `createTableGeneric`? When should one derive `Column` instances? When should one use foreign keys?

kore/src/Kore/Log/WarnSimplificationWithRemainder.hs
```diff
@@ -46,10 +51,49 @@ data WarnSimplificationWithRemainder =
...
+
+instance Unparse Remainders where
+    unparse = unparse . toTermLike . getRemainders
+    unparse2 = unparse2 . toTermLike . getRemainders
+
+instance SQL.Column Remainders where
+    defineColumn = SQL.defineTextColumn
+    toColumn = SQL.toColumn . Pretty.renderText . Pretty.layoutOneLine . unparse
+
+toTermLike :: OrPattern Variable -> TermLike Variable
```
ana-pantilie: Why not use https://github.com/kframework/kore/blob/b6522cdd9153d35f8c7caac8d28088cbc56d9635/kore/src/Kore/Internal/OrPattern.hs#L168?

Issue #1395, opened 2020-01-03
Rule variable is not substituted correctly


yzhang90:
Kwasm: https://github.com/kframework/wasm-semantics
compile: 
```
kompile --backend haskell --directory .build/defn/haskell -I .build/defn/haskell --main-module WASM-TEST --syntax-module WASM-TEST-SYNTAX .build/defn/haskell/test.k
```
kprove:
```
kprove --haskell-backend-command "/home/zy/workspace/kore/.build/kore/bin/kore-repl --repl-script /home/zy/workspace/kore/dist/kast.kscript " ~/workspace/wasm-semantics/tests/proofs/wrc20-spec.k -d ~/workspace/wasm-semantics/.build/defn/haskell/ --def-module KWASM-LEMMAS
```
in REPL mode:
```
prove 1
stepf 2
select 2
konfig
```
in step 2 , the following rule is applied:
```
    rule <k> ( func OID:OptionalId TUSE:TypeUse LDECLS:LocalDecls INSTRS:Instrs ) => #checkTypeUse ( TUSE ) ... </k>
         <curModIdx> CUR </curModIdx>
         <moduleInst>
           <modIdx> CUR </modIdx>
           <typeIds> TYPEIDS </typeIds>
           <types>   TYPES   </types>
           <funcIds> IDS => #saveId(IDS, OID, NEXTIDX) </funcIds>
           <nextFuncIdx> NEXTIDX => NEXTIDX +Int 1 </nextFuncIdx>
           <funcAddrs> ADDRS => ADDRS [ NEXTIDX <- NEXTADDR ] </funcAddrs>
           ...
         </moduleInst>
         <nextFuncAddr> NEXTADDR => NEXTADDR +Int 1 </nextFuncAddr>
         <funcs>
           ( .Bag
          => <funcDef>
               <fAddr>    NEXTADDR                             </fAddr>
               <fCode>    INSTRS                               </fCode>
               <fType>    asFuncType  ( TYPEIDS, TYPES, TUSE ) </fType>
               <fLocal>   asLocalType ( LDECLS )               </fLocal>
               <fModInst> CUR                                  </fModInst>
             </funcDef>
           )
           ...
         </funcs>
```
and the result config changes from
```
<funcs>_5 </funcs>
```
to
```
      <funcs>
        <funcDef>
          <fAddr>
            NEXTADDR
          </fAddr>
          <fCode>
[...]

PR #1452, opened 2020-01-20
Remove map constructor for single, opaque element maps

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1395

---

###### Reviewer checklist


kore/src/Kore/Domain/Builtin.hs
```diff
@@ -236,6 +237,26 @@ emptyNormalizedAc = NormalizedAc
...
+    => Ord valueWrapper
+    => NormalizedAc key valueWrapper child
+    -> Maybe child
+asSingleOpaqueElem
+    NormalizedAc
+        { elementsWithVariables
+        , concreteElements
+        , opaque
+        }
+    | elementsWithVariables == mempty
```
ttuegel: I would suggest `null` instead of `== mempty`. You can also use a pattern guard to avoid repeating the `Nothing`:

```suggestion
    | null elementsWithVariables
    , null concreteElements 
    , [singleOpaqueElem] <- opaque
    = Just singleOpaqueElem
    | otherwise = Nothing
```

Issue #1462, opened 2020-01-23
Prover admits universally-quantified claim incorrectly


ttuegel:
Define a symbol `Det` which rewrites deterministically to `true`:

```
module TEST where
    syntax KItem ::= "Det"
    rule Det => true
endmodule
```

Then, the Haskell backend admits the following claim, which should not be proven:

```
module DET-FORALL-SPEC [all-path]
  import TEST
  // claim: Det rewrites to all Bool
  rule Det =>  X:Bool  [no-warning-i-know-what-i-am-doing-here]
endmodule
```

Tests: #1463

PR #1463, opened 2020-01-23
Tests for all-path proofs of non-deterministic semantics

ttuegel:
See also: #1462 

---

###### Reviewer checklist


ttuegel:
When #1462 is fixed, these claims will be rejected.

PR #1468, opened 2020-01-23
Stop using side conditions for substitutions

virgil-serbanuta:
Probably it's not worth the trouble, but this seems to somewhat improve the run time for `test/regression-evm-semantics-39dd1e5/test-sum-to-n.sh` by ~5%

Without this PR:
455.71user 30.96system 4:50.48elapsed 167%CPU (0avgtext+0avgdata 1243420maxresident)k
0inputs+0outputs (0major+340944minor)pagefaults 0swaps

With this PR:
434.27user 30.46system 4:36.44elapsed 168%CPU (0avgtext+0avgdata 1250492maxresident)k
0inputs+0outputs (0major+342744minor)pagefaults 0swaps

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Builtin/KEqual.hs
+++ b/kore/src/Kore/Builtin/KEqual.hs
@@ -145,7 +145,7 @@ evalKEq true (valid :< app) =
 
         defined1 <- Ceil.makeEvaluate sideCondition pattern1
         defined2 <- Ceil.makeEvaluate sideCondition pattern2
-        defined <- And.simplifyEvaluated sideCondition defined1 defined2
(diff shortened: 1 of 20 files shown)


ttuegel:
I checked using the structured logging feature, and with or without this change, we send the same number of queries to the solver and evaluate the same number of function rules. It's harder to determine if we send exactly the same conditions to the solver, but I think it's a good bet that we do. So, I want to keep investigating this.

ttuegel:
After reviewing the profile, I am reasonably satisfied that the performance changes because of many small factors, and is not due to any big change. Therefore, I'm happy to merge this.

PR #1474, opened 2020-01-27
Fix the simplified attribute when refusing to branch

virgil-serbanuta:
---

###### Reviewer checklist


kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -201,6 +208,10 @@ evaluateBuiltin
     isValue pat =
         maybe False TermLike.isConstructorLike $ asConcrete pat
 
+data NonSimplifiability
```
ttuegel: Is this similar to `ConstructorLike`?

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -211,11 +222,30 @@ applyFirstSimplifierThatWorks
...
+    applyFirstSimplifierThatWorksWorker evaluators multipleResults Always
+
+applyFirstSimplifierThatWorksWorker
+    :: forall variable simplifier
+    .  ( SimplifierVariable variable
+       , MonadSimplify simplifier
+       )
+    => [BuiltinAndAxiomSimplifier]
+    -> AcceptsMultipleResults
+    -> NonSimplifiability
```
ttuegel: What does the `NonSimplifiability` argument do here?

test/simplified/five-spec.k
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2019 K Team. All Rights Reserved.
```
ttuegel: For each of these specs, could you give a brief description of what it is testing? It's not clear to me what the failure modes are.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -256,13 +308,16 @@ applyFirstSimplifierThatWorks
                 orRemainders
             -- TODO (traiansf): this might generate too much output
             --    replace log with a logOnce when that becomes available
-            tryNextSimplifier
+            tryNextSimplifier Conditional
           | otherwise -> return applicationResult
-        AttemptedAxiom.NotApplicable -> tryNextSimplifier
+        AttemptedAxiom.NotApplicable -> tryNextSimplifier nonSimplifiability
```
ttuegel: Just checking my understanding: If a rule doesn't match, that would be an example of a `NotApplicable` result, because the `SideCondition` doesn't determine matching, so changing it would never allow the match to succeed. On the other hand, if a rule matches, but the `requires` isn't satisfied, that would be an example of `NotApplicableUntilConditionChanges` because changing the `SideCondition` could allow the rule to apply. Does that sound right?

virgil-serbanuta: Close. I added a comment.

Issue #1407, opened 2020-01-08
Unsuccessful removal of variable in `owise` rule application on concrete KEVM tests


ehildenb:
On KEVM `master`, if you run:

```
git clone 'https://github.com/kframework/evm-semantics'
cd evm-semantics
git submodule update --init --recursive
make deps -B RELEASE=true
make build-haskell -j8
make test-vm TEST_CONCRETE_BACKEND=haskell
```

the tests there pass, but generate a bunch of warnings which don't make much sense:

```
[2020-01-08 14:26:13.577821392][Warning] : [LogMessage] : Failed to decide predicate:                           
    \not{SortGeneratedTopCell{}}(                                                    
        \equals{SortString{}, SortGeneratedTopCell{}}(                                                          
LblsubstrString'LParUndsCommUndsCommUndsRParUnds'STRING'Unds'String'Unds'String'Unds'Int'Unds'Int{}(
                Var'Unds'0:SortString{},                                             
                /* builtin: */ \dv{SortInt{}}("0"),
                /* builtin: */ \dv{SortInt{}}("2")                                                          
            ),                                    
            /* builtin: */ \dv{SortString{}}("0x")                                   
        )                                                                                                       
    ) []
```

Which makes it look like it's trying to decide `substrString(_:String, 0, 2) #Equals "0x"`.

This comes from the rules (from `serialization.md`):

```
    rule #parseWord("") => 0 
    rule #parseWord(S)  => #parseHexWord(S) requires lengthString(S) >=Int 2 andBool substrString(S, 0, 2) ==String "0x" 
    rule #parseWord(S)  => String2Int(S) [owise] 
```

Because it's using underbar variable `_:String` instead of variable `S:String`, we think it's coming from the `owise` rule, where it's not able to detect that the `owise` rule does/doesn't apply immediately.

Somehow this doesn't make much sense, because the test is concrete, so it should be able to definitively say which of the rules apply.

ttuegel:
I remember discussing this with @yzhang90 and @xc93, but I don't think I ever created an issue for it.

When evaluating conditions, we should always evaluate logical predicates (such as `\in(_, _)` in the side-condition of an `owise` rule) before evaluating functions and sending the result to the solver. Evaluating logical connectives first allows for short-circuiting, which is essential in this case and others.

ttuegel:
[...]

PR #1475, opened 2020-01-27
Begin to clean up function evaluation

ttuegel:
This pull request begins to clean up function evaluation, in particular:

- Hard-coding matching instead of unification.
- Removing `MonadUnify` from the matcher so that the type of the match result is constrained (e.g. no branching).

See also: #1407 

---

###### Reviewer checklist


kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -353,22 +364,93 @@ See also: 'applyRewriteRule'
...
     initial
     -- Wrap the rules so that unification prefers to substitute
     -- axiom variables.
     (map EqualityPattern.toAxiomVariables -> rules)
-  = unifyRules unificationProcedure sideCondition initial rules
+  = unifyRules sideCondition initial rules
     >>= finalizeRulesSequence sideCondition initial
+
+-- |Unifies/matches a list a rules against a configuration. See 'unifyRule'.
+unifyRules
```
virgil-serbanuta: Here and below: why `unify` and not `match`?

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -353,22 +364,93 @@ See also: 'applyRewriteRule'
...
+    -- ^ Top level condition.
+    -> Pattern variable
+    -- ^ Initial configuration
+    -> rule variable
+    -- ^ Rule
+    -> unifier (UnifiedRule variable (rule variable))
+unifyRule sideCondition initial rule = do
+    let (initialTerm, initialCondition) = Pattern.splitTerm initial
+        mergedSideCondition =
+            sideCondition `SideCondition.andCondition` initialCondition
```
virgil-serbanuta: Not relevant for this PR, but we can probably remove this merge and get a performance improvement. Or make a special case for `andCondition` with top in which we don't update the representation.

ttuegel:
I am holding off on merging this for a moment because I noticed the integration tests taking longer than normal. I am checking locally that it's due to the load on Jenkins and not a performance regression.

ttuegel:
I found no significant change in performance locally.

PR #1479, opened 2020-01-29
refreshVariable tests

ana-pantilie:
- adds simple unit tests for the most important `FreshVariable` instances
- adds generators for these types of variables to use in future refactoring PR

---

###### Reviewer checklist


kore/test/Test/Kore/Variables/Fresh.hs
```diff
@@ -55,12 +74,213 @@ test_refreshVariable =
         in do
             assertRefreshes original metaVariableDifferentSort
             assertRefreshes metaVariableDifferentSort original
+
+    -- Target Variable
```
ttuegel: Instead of the comments, could we put the tests for each combination of types in a `testGroup` so that it is visible in the output of a failing test?

PR #1480, opened 2020-01-29
Verifying overload attributes

### LINKED ISSUE #1447: Verifying Overload attributes
We are already doing ad hoc attribute verification in [`Kore.ASTVerifier.AttributesVerifier`](https://github.com/kframework/kore/blob/master/kore/src/Kore/ASTVerifier/AttributesVerifier.hs). I suggest we implement verifiers for the specific attribute types (`verifyAxiomAttributes`,  `verifySortAttributes`, `verifySymbolAttributes`) and use them in [`Kore.ASTVerifier.SentenceVerifier`](https://github.com/kframework/kore/blob/master/kore/src/Kore/ASTVerifier/SentenceVerifier.hs) where `parseAttributes` is called.

Verifying attributes will help the consumers of axiom and sort attributes

---

- [x] Add empty `verifyAxiomAttributes`, `verifySortAttributes`, and `verifySymbolAttributes` functions to `Kore.ASTVerifier.AttributesVerifier`.
- [x] Add a `symbol` type parameter to `Overload`. All existing occurrences should be `Overload SymbolOrAlias`.
- [x] Propagate the `symbol` type parameter up to `Kore.Attribute.Axiom.Axiom`. Again the default should be `SymbolOrAlias`.
- [x] `verifySymbolAttributes` should take an `Attribute.Axiom SymbolOrAlias` and return an `Attribute.Axiom Symbol`. The type of `VerifiedModule` must make a corresponding change.
- [x] Remove the `OverloadSimplifier` code that turns a `SymbolOrAlias` into a `Symbol` because it's no longer needed.

In a subsequent task, we can do the same for sort attributes, but that is a little trickier: it requires lazy verification of symbols.

andreiburdusa:
---
Fixes #1447 
###### Reviewer checklist


kore/src/Kore/ASTVerifier/AttributesVerifier.hs
```diff
@@ -106,3 +131,46 @@ verifyNoHookAttribute attributes = do
...
+                    , symbolParams
+                    , symbolAttributes
+                    , symbolSorts
+                    }
+        return symbol
+      where
+        symbolConstructor = symbolOrAliasConstructor symbolOrAlias
+        symbolParams = symbolOrAliasParams symbolOrAlias
+
+verifySymbolAttributes :: a
```
ttuegel: Let's give this a signature like `verifyAxiomAttributes`, something like

```.hs
verifySymbolAttributes
    :: forall error a
    .  MonadError (Error VerifyError) error
    => IndexedModule Verified.Pattern Attribute.Symbol a
    -> Attribute.Symbol
    -> error Attribute.Symbol
verifySymbolAttributes _ attrs = return attrs
```

and also for `verifySortAttributes`.

kore/src/Kore/ASTVerifier/ModuleVerifier.hs
```diff
@@ -123,9 +126,15 @@ newVerifiedModule :: ParsedModule -> Verifier VerifiedModule'
 newVerifiedModule module' = do
     VerifierContext { implicitModule } <- Reader.ask
     let Module { moduleName, moduleAttributes } = module'
+        ImplicitIndexedModule indexedModule = implicitModule
+        emptyIM = emptyIndexedModule moduleName
+    indexedModule' <- traverse
```
ttuegel: Do we ever need an `implicitModule` that is _not_ verified? Maybe we could make that module verified from the beginning.

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -437,3 +450,11 @@ parseAttributes'
     -> error attrs
 parseAttributes' =
     Attribute.Parser.liftParser . Attribute.Parser.parseAttributes
+
+parseAttributesAndVerify
```
ttuegel: We should call this `parseAndVerifyAxiomAttributes`, because we will also be verifying other attributes soon.

kore/src/Kore/Attribute/Overload.hs
```diff
@@ -18,32 +18,37 @@ import qualified GHC.Generics as GHC
...
-newtype Overload =
+newtype Overload symbol =
     Overload
-        { getOverload :: Maybe (SymbolOrAlias, SymbolOrAlias) }
-    deriving (Eq, GHC.Generic, Ord, Show)
+        { getOverload :: Maybe (symbol, symbol) }
+    deriving (Eq, GHC.Generic, Ord, Show, Functor)
 
-instance SOP.Generic Overload
+instance SOP.Generic (Overload SymbolOrAlias)
```
ttuegel: All of these instances should be polymorphic, for example:

```.hs
instance SOP.Generic (Overload symbol)
```

kore/src/Kore/Attribute/Overload.hs
```diff
@@ -68,7 +73,7 @@ overloadAttribute symbol1 symbol2 =
         , attributePattern_ symbol2
         ]
 
-instance ParseAttributes Overload where
+instance ParseAttributes (Overload SymbolOrAlias) where
```
ttuegel: This will probably be the only instance for `Overload` that is _not_ polymorphic.

kore/src/Kore/Builtin.hs
```diff
@@ -173,7 +172,7 @@ evaluators builtins indexedModule =
         justAttributes (attrs, _) = attrs
 
     importHookedSymbolAttributes
-        :: (a, b, VerifiedModule StepperAttributes Attribute.Axiom)
+        ::  ( a, b, VerifiedModule StepperAttributes)
```
ttuegel:
```suggestion
        ::  (a, b, VerifiedModule StepperAttributes)
```

kore/src/Kore/IndexedModule/Resolvers.hs
```diff
@@ -12,6 +12,7 @@ module Kore.IndexedModule.Resolvers
     , getSymbolAttributes
     , resolveSort
     , resolveAlias
+    , resolveThing
```
ttuegel: Why does this need to be exported now?

kore/src/Kore/Step/EqualityPattern.hs
```diff
@@ -69,7 +72,7 @@ data EqualityPattern variable = EqualityPattern
     , left  :: !(TermLike.TermLike variable)
     , right :: !(TermLike.TermLike variable)
     , ensures :: !(Predicate variable)
-    , attributes :: !Attribute.Axiom
+    , attributes :: !(Attribute.Axiom SymbolOrAlias)
```
ttuegel: At this point, the attributes should be verified, so let's use `Attribute.Axiom Symbol` here.

kore/src/Kore/Step/RulePattern.hs
```diff
@@ -196,7 +199,7 @@ data RulePattern variable = RulePattern
     , antiLeft :: !(Maybe (TermLike.TermLike variable))
     , requires :: !(Predicate variable)
     , rhs :: !(RHS variable)
-    , attributes :: !Attribute.Axiom
+    , attributes :: !(Attribute.Axiom SymbolOrAlias)
```
ttuegel:
```suggestion
    , attributes :: !(Attribute.Axiom Symbol)
```

kore/src/Kore/IndexedModule/OverloadGraph.hs
```diff
@@ -38,11 +38,7 @@ import Kore.IndexedModule.IndexedModule
     ( VerifiedModule
     , recursiveIndexedModuleAxioms
     )
-import Kore.IndexedModule.MetadataTools as MetadataTools
-import Kore.Internal.Symbol
-import Kore.Syntax.Application
-    ( SymbolOrAlias (..)
-    )
+import Kore.Internal.Symbol as Internal.Symbol
```
ttuegel: Is the named import necessary now?

```suggestion
import Kore.Internal.Symbol
```

kore/src/Kore/Attribute/Overload.hs
```diff
@@ -87,3 +94,24 @@ instance ParseAttributes Overload where
       where
         toAttribute (symbol1, symbol2) =
             Attributes [overloadAttribute symbol1 symbol2]
+
+instance ParseAttributes (Overload Symbol) where
+    parseAttribute _                    (Overload Nothing)           =
+        return $ Overload Nothing
+
+    parseAttribute attrPattern overload@(Overload (Just symbolPair)) = do
```
andreiburdusa: This is the part I'm a little concerned about, especially because the unit tests passed even when I made the result `undefined`

ttuegel: This instance should not exist. It is not possible to parse an `Overload Symbol`. There are fields of `Symbol` missing, and I have no idea why that isn't an error.

kore/src/Kore/Attribute/Axiom.hs
```diff
@@ -153,7 +163,52 @@ instance Default Axiom where
...
+            , toAttributes . label
+            , toAttributes . sourceLocation
+            , toAttributes . constructor
+            , toAttributes . functional
+            , toAttributes . subsorts
+            , toAttributes . uniqueId
+            , toAttributes . owise
+            ]
+
+instance ParseAttributes (Axiom Symbol) where
```
ttuegel: Because it's not possible to parse an `Overload Symbol` attribute, it's also not possible to parse an `Axiom Symbol` attribute.

andreiburdusa: As I remember, after I changed `RulePattern` 's `attributes` to `Attribute.Axiom Symbol` , the compiler requested a `SQL.Column` instance for  `Axiom Symbol`. I just kept adding the required instances until I did this. I'll try to do it in a different way

ttuegel: Oh, it wants the `toAttributes` function. That's the only one you should need to implement. You can leave the other methods as `undefined`, and I will push a separate pull request which deals with that; I already have it in a branch.

andreiburdusa: but without
`parseAttribute  _ (Overload Nothing) = return $ Overload Nothing`
4 tests are failing.
Should I just use `undfined` and wait for your pull request?

ttuegel: Let's just merge #1510 quickly, then this.

PR #1481, opened 2020-01-29
scripts/check.sh: Do not rely on git fetch

ttuegel:
This removes the `git fetch` command so that (hopefully) this script can work with the Git reference cache on Jenkins. It should also make the command faster on developers' machines.

---

###### Reviewer checklist


ehildenb:
I've approved (looks like we'll be checking more files against stylish, but that's fine I think) but up to you if you want to merge it. If you go to the step which does the git checkout, you should see the flag `--reference /home/jenkins-slave/devops.git` being used in various clones to make sure the reference repo is in use.

PR #1482, opened 2020-01-29
Short-circuiting for owise rules

### LINKED ISSUE #1407: Unsuccessful removal of variable in `owise` rule application on concrete KEVM tests

ttuegel:
Fixes #1407.

---

###### Reviewer checklist


ttuegel:
This fixes the issue above, but there are other problems, see #1490.

kore/src/Kore/Internal/Predicate.hs
```diff
@@ -734,6 +738,12 @@ setSimplified
 setSimplified simplified (GenericPredicate termLike) =
     GenericPredicate (TermLike.setSimplified simplified termLike)
 
+forgetSimplified
+    :: InternalVariable variable
+    => Predicate variable -> Predicate variable
+forgetSimplified (GenericPredicate termLike) =
+    GenericPredicate (resynthesize termLike)
```
virgil-serbanuta: Why not have a `TermLike.forgetSimplified` which calls `resynthesize`?

Right now `resynthesize` does the right thing because the `Simplified` attribute is the only one that cannot be recomputed from the term, but that may change in the future, and it would be nice to have a single place to change when that happens.

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -456,3 +460,21 @@ matchRule sideCondition initial rule = do
...
+    :: forall unifier variable
+    .  SimplifierVariable variable
+    => MonadUnify unifier
+    => SideCondition variable
+    -> Condition variable
+    -> unifier (Condition variable)
+evaluateRequires sideCondition requires = do
+    _requires' <-
+        simplifyPredicate sideCondition Nothing requires
+        & withoutAxioms
```
virgil-serbanuta: Why "withoutAxioms"? Is it just attempting to take a shortcut in case this can be solved without axioms? If yes, why not call `Condition.forgetSimplified` as part of `withoutAxioms`, since that seems a safer way of handling this?

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -456,3 +460,21 @@ matchRule sideCondition initial rule = do
...
+        & Unifier.maybeUnifierT
+        & maybeT (return requires) Unifier.scatter
+    deepseq _requires' $ simplifyPredicate sideCondition Nothing
```
virgil-serbanuta: Why this `deepseq`?

ttuegel: Ah, sorry, I requested review but forgot to push my latest version! That is left over from testing. :sweat_smile:

kore/src/Kore/Step/EquationalStep.hs
ttuegel: Yes, that's a good idea, with the benefit that it doesn't reset the simplified status if the first pass fails.

kore/src/Kore/Internal/Predicate.hs
ttuegel: Done!

PR #1488, opened 2020-01-30
Remove CharSet

ttuegel:
This pull request cleans up a few things about the identifier parser that I ran across when I was adding the custom prelude.

---

###### Reviewer checklist


virgil-serbanuta:
Isn't a Set a less efficient representation?

ttuegel:
There are two places we used `CharSet`. In the `Id` parser, I think it is more efficient to avoid constructing a `Set` or `CharSet` at all. In the escape-code parser, the `Set` lookup has complexity `O(log 11)`, versus `O(1)` for the `CharSet`. The `CharSet` uses a lot more memory, but neither uses a significant amount of memory. I can measure, if we want to be sure, or I can just close the PR.

virgil-serbanuta:
> There are two places we used `CharSet`. In the `Id` parser, I think it is more efficient to avoid constructing a `Set` or `CharSet` at all. In the escape-code parser, the `Set` lookup has complexity `O(log 11)`, versus `O(1)` for the `CharSet`. The `CharSet` uses a lot more memory, but neither uses a significant amount of memory. I can measure, if we want to be sure, or I can just close the PR.

Ok, I'll let Traian decide, for me it's fine either way. If you are confident it does not affect the performance significantly, in my opinion it's fine to skip the measurement.

Issue #1399, opened 2020-01-06
refreshVariable should avoid allocation


ttuegel:
`refreshVariable` is performance critical and instances should avoid allocating whenever possible. Profiling indicates this is a significant hot spot. I suggest the following refactoring so that we can use a non-allocating implementation whenever possible:

```.hs
class SortedVariable variable where
    -- New: A Lens to get *or* set the Sort
    lensVariableSort :: Lens' variable sort

    sortedVariableSort = Lens.view lensVariableSort

class (Ord variable, SortedVariable variable) =>  FreshVariable variable where
    -- New: The greatest lower bound on variables
    -- with the same name as the given variable.
    infVariable :: variable -> variable
    -- New: The least upper bound on variables
    -- with the same name as the given variable.
    supVariable :: variable -> variable
    -- New: The least variable greater than the given variable.
    nextVariable :: variable -> variable

    -- This is a default implementation in terms of the above.
    -- The default implementation is suitable for most types
    -- and does O(1) allocation.
    refreshVariable :: Set variable -> variable -> Maybe variable
```

ttuegel:
As an example, the instance for the basic `Variable` type would be

```.hs
instance FreshVariable Variable where
    infVariable variable = variable { variableCounter = Nothing }
    supVariable variable = variable { variableCounter = Just Sup }
    nextVariable variable = {- already implemented, see below -}
    refreshVariable avoiding variable = do
        fixedLargest <- Lens.set lensVariableSort theSort <$> Set.lookupLT pivotMax avoiding
        Monad.guard (fixedLargest >= pivotMin)
        return (nextVariable fixedLargest)
      where
        pivotMax = supVariable variable
        pivotMin = infVariable variable
        theSort = Lens.view lensVariableSort variable
```

The `nextVariable` function is [already implemented](https://github.com/kframework/kore/blob/master/kore/src/Kore/Variables/Fresh.hs#L128), but it's not a member of the class.

Some pseudo-code properties:

```.hs
properties = do -- this line is to make GitHub's syntax highlighting happy
    -- generators for different variable types in Test.Kore
    variable <- ...

    -- any variable is between infVariable and supVariable:
    infVariable variable <= variable
    variable < supVariable variable

    -- nextVariable generates an increasing sequence of variables with the same "stem" (the text part of the name):
    variable < nextVariable variable
    nextVariable variable < nextVariable (nextVariable variable)
    --      etc.

    -- supVariable is strictly greater than every variable in the sequence:
[...]

PR #1496, opened 2020-01-31
mapVariables: Avoid capturing variables

ttuegel:
This pull request fixes `mapVariables` and `traverseVariables` (on `TermLike`) so that they will not capture variables. That is, they are not naive mappings. This makes them significantly more useful! Unfortunately, this pull request has grown larger than I would like, and there is not a clear way to break it up.

Summary:

- `mapVariables` and `traverseVariables` will not capture variables.
- `mapVariables` and `traverseVariables` now take a separate mapping or traversal for `ElementVariable` and `SetVariable`.
- To avoid capturing variables, `mapVariable` and `traverseVariable` now require a `FreshVariable` constraint on the result's variable type.
- Due to its proliferation, the `FreshVariable` constraint is added to `InternalVariable`.

See also: #1399 
~After: #1503~

---

###### Reviewer checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -137,6 +137,10 @@
       - Kore.Step.Step
       - SQL.SOP
 - ignore: {name: "Avoid lambda", within: [Kore.Step.Function.Memo]}
+- ignore:
+    name: "Redundant bracket"
+    within:
+      - Kore.Internal.TermLike.TermLike
 
 
 # Haskell names match K names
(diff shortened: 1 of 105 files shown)


ttuegel:
I'm currently investigating the possibility of a performance regression due to these changes.

ttuegel:
I have determined that this pull request is not responsible for any significant change in performance.

PR #1500, opened 2020-02-03
Subterm ceil

virgil-serbanuta:
Fixes 7 of the evm lemma tests, 2 are still failing.

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -361,3 +362,66 @@ makeEvaluateNormalizedAc
         }
   | Map.null concreteElements = Just $ makeEvaluateTerm sideCondition opaqueAc
 makeEvaluateNormalizedAc _  _ = Nothing
+
+makeSimplifiedCeil
```
ttuegel: Could you add a comment here explaining that this is the terminal case of the `Ceil` simplifier, and that it adds the `Ceil` of subterms, as required?

PR #1501, opened 2020-02-03
Prelude.Kore: Alternative and optional

ttuegel:
---

###### Reviewer checklist


kore/src/Branch.hs
```diff
@@ -16,7 +16,7 @@ module Branch
     , alternate
     ) where
 
-import Prelude.Kore
+import Prelude
```
andreiburdusa: Why did you replace the `Prelude.Kore` here and in ListT.hs ?

ttuegel: These two modules were originally written for the default `Prelude`, and they don't rely on anything we use in `Prelude.Kore`. I was going to have to write a lot of `hiding (...)` statements, so I decided to go back to the default `Prelude` instead.

PR #1510, opened 2020-02-04
From class

ttuegel:
This pull request adds a `From` class for (total) conversions between types. This idea (from Rust, among others) is an oft-requested feature for the standard library, but it requires `TypeApplications` to work well. Fortunately, we already use `TypeApplications`!

In particular, this splits up the `ParseAttributes` type class to make some things from #1480 less awkward.

---

###### Reviewer checklist


kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -3,6 +3,9 @@ Copyright   : (c) Runtime Verification, 2020
 License     : NCSA
 -}
 
+-- For instance Applicative:
```
virgil-serbanuta: I am curious why this extension is needed now, but this comment does not help me. In particular, I didn't find any Applicative instance in this file.

kore/src/Prelude/Kore.hs
```diff
@@ -22,6 +22,8 @@ module Prelude.Kore
     , Applicative (..)
     , Alternative (..)
     , optional
+    -- * From
+    , module From
```
virgil-serbanuta: Would this lead to code that to code that is harder to read because it's not obvious what is being converted to what?

In this particular PR this is not an issue, but I could easily imagine us writing code in which it's perfectly clear to the to whoever is writing the code what the types are, perhaps it would also be easy to figure out for the PR reviewer since he sees all the relevant pieces of code at once, but which would be fairly opaque to other people.

Of course, using type annotations solves this issue, I wonder if we'll actually use them or we'll take shortcuts, especially when the types are on the complex side (e.g. multi-line, deeply nested types).

FWIW, I think that `coerce` is similar to `from` without type annotations, and I remember spending more time figuring some of its usage than it would have taken whoever wrote the code to write something that manually unpacks and repacks the data. I'm not sure if I'm the only one that spent too much time on that.

So, I'm not saying we shouldn't do this, I just wanted to point out the possible costs to make sure we take them into account (the benefits are obvious, I guess).

kore/src/Kore/Internal/SideCondition.hs
ttuegel: Yes, it doesn't seem to be used. I wonder what it is left-over from...

kore/src/Prelude/Kore.hs
ttuegel: > FWIW, I think that `coerce` is similar to `from` without type annotations, and I remember spending more time figuring some of its usage than it would have taken whoever wrote the code to write something that manually unpacks and repacks the data. I'm not sure if I'm the only one that spent too much time on that.

Yes, there is a hazard here. I will say that, because `coerce` works only over `newtype`s, I think it's easier to get `coerce` working without `TypeApplications`. In practice, it seems pretty hard to get `from` working without them, or at least without type annotations. I would rather prefer there be a way to disable type inference for certain parameters, but type inference is hard-coded into the Haskell mindset.

The alternative in this particular case would be a `ToAttributes` class, separate from `ParseAttributes`. The problem blocking #1480 is that we need a `toAttributes` implementation for a type that doesn't have a meaningful `ParseAttributes`. We also have other class methods following a similar pattern, for example: `fromVariable` and `toVariable`.

As long as we are diligent about annotating occurrences of `from`, I think I prefer one `From` class over a proliferation of many `FromThing` and `ToThing` classes. We could add something to the style guide that says you must give an explicit annotation for at least the `from` type or the `to` type; would that address your concerns?

Issue #1505, opened 2020-02-04
Implement matching for overloaded constructors


ttuegel:
Unification is implemented for overloaded constructors, but now we need to implement matching. The existing overloaded constructor unification code should be refactored so that matching can use the same code.

See also: #1269

PR #1512, opened 2020-02-05
Overload matching

traiansf:
Fixes: #1505 

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -27,113 +27,154 @@ import Kore.Internal.Pattern
...
+        else return $ Right Nothing
+
+unifyOverloadingCommonOverload
+    :: MonadSimplify unifier
+    => SimplifierVariable variable
+    => Application Symbol (TermLike variable)
+    -> Application Symbol (TermLike variable)
+    -> Inj ()
+    -> unifier
+        (Either (Doc ()) (Maybe (Pair (TermLike variable))))
```
virgil-serbanuta: Could you make a custom UnificationResult type instead of Either+Maybe?

ttuegel: I agree, this should at least be `unifier (Either UnifyOverloadingError (Pair (TermLike variable)))`. Another option, which should minimize the need for `case` and `if _ then _ else _`, is `ExceptT UnifyOverloadingError unifier (Pair (TermLike variable))`. You could define a type synonym such as,

```.hs
type UnifyOverloadingResult a = Either UnifyOverloadingError (Pair a)
-- or:
type UnifyOverloading unifier a = ExceptT UnifyOverloadingError unifier (Pair a)
```

or a `newtype`. If you want to go the `newtype` route, please also define functions to make that type as easy to deconstruct as `either` makes `Either` easy to deconstruct.

`UnifyOverloadingError` should be a sum type so we can match on the result, at least for testing.

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -27,113 +27,154 @@ import Kore.Internal.Pattern
...
+ the first and second terms in a pair are not interchangeable.
+-}
+unifyOverloading
+    :: MonadSimplify unifier
+    => SimplifierVariable variable
+    => Pair (TermLike variable)
+    -> unifier
+        (Either (Doc ()) (Maybe (Pair (TermLike variable))))
+unifyOverloading termPair = case termPair of
+    Pair
```
ttuegel: Could you please add brief comment to each of these cases, explaining what it is checking for?

kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -969,6 +970,19 @@ test_matching_Map =
         ]
     ]
 
+test_matcherOverloading :: [TestTree]
+test_matcherOverloading =
```
ttuegel: Can we add some tests for constructors that are overloaded, but don't match? I would also like to see some unit tests of `Kore.Step.Simplification.Overloading` directly.

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
+    OverloadSimplifier { isOverloaded, isOverloading, resolveOverloading }
+        <- Simplifier.askOverloadSimplifier
+    Monad.guard (isOverloaded overloadingHead)
+    isSecondHeadConstructor <- isConstructorOrOverloaded overloadedHead
+    Monad.guard isSecondHeadConstructor
+    let overloadedTerm' =
+            resolveOverloading injProto overloadingHead overloadedChildren
+    if isOverloading overloadingHead overloadedHead
+        then return $ Pair overloadingTerm overloadedTerm'
+        else throwBottom "different injected ctor"
```
ttuegel: This is because the second symbol _is_ overloaded, but by a different symbol than the first?

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
+
+notUnifiableError
+    :: Monad unifier => TermLike variable -> UnifyOverloading unifier a
+notUnifiableError (DV_ _ _) = throwBottom "injected domain value"
```
ttuegel: All these failures are because domain values and builtins cannot be overloaded, is that right?

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
+ See <https://github.com/kframework/kore/blob/master/docs/2019-08-27-Unification-modulo-overloaded-constructors.md>
+
+ If the overloading rules are not applicable, it would return @Nothing@.
+ However, if it detects an unification failure, it would return a reason
+ for the failure which can be used as an explanation for it.
+
+ Note: this algorithm is used for both matching and unification; that is why
+ the first and second terms in a pair are not interchangeable.
+-}
+unifyOverloading
```
ttuegel: What happens if I feed the result of `unifyOverloading` back into itself? (That would probably be a good unit test.)

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -354,6 +361,16 @@ matchInj (Pair (Inj_ inj1) (Inj_ inj2)) = do
     unifyInj inj1 inj2 & either (const empty) (push . injChild)
 matchInj _ = empty
 
+matchOverload
+    :: (MatchingVariable variable, MonadSimplify unifier)
+    => Pair (TermLike variable)
+    -> MaybeT (MatcherT variable unifier) ()
+matchOverload termPair = do
+    termPair' <- Monad.Trans.lift . runExceptT $ unifyOverloading termPair
```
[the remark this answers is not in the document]
traiansf: Got it! I was actually wondering how might I write this better. Thanks!

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
+  and
+    inj{S2,injTo}(secondHead(secondChildren)) =
+        inj{S,injTo}(headUnion(inj2(secondChildren)))
+
+  Then reduces the problem to
+    inj{S,injTo}(headUnion(inj1(firstChildren)))
+   vs
+    inj{S,injTo}(headUnion(inj2(secondChildren)))
+-}
+unifyOverloadingCommonOverload
```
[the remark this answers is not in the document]
traiansf: Strictly speaking, yes; I like to think about them like this:
* for `unifyOverloadingVsOverloaded` you apply an equation to transform the injected symbol into the non-injected one
* for `unifyOverloadingCommonOverload`, you find a symbol overloading both injected symbols and apply equations to transform both symbols to the common overloading one.

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
+    )
 import Kore.Unification.Unify as Unify
+import Pair
 
-{- | Unify an overloaded constructor application pattern with a sort injection
-pattern over an (overloaded) constructor.
+{- | Describes the possible errors encountered during unification:
 
-See <https://github.com/kframework/kore/blob/master/docs/2019-08-27-Unification-modulo-overloaded-constructors.md>
+- current procedure not applicable
```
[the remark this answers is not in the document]
traiansf: > I don't understand what this means.

I've changed to "the unification problem could not be solved by the current method"

Basically, it means we need to fall-through and attempt a different procedure.

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -15,125 +19,236 @@ import Control.Error
...
         <- Simplifier.askOverloadSimplifier
     Monad.guard (isOverloaded firstHead && isOverloaded secondHead)
-    let injProto = inj { injChild = () }
     case unifyOverloadWithinBound injProto firstHead secondHead injTo of
-        Nothing -> Monad.Trans.lift
-            $ explainAndReturnBottom
-                "overloaded constructors not unifiable"
-                first
-                second
+        Nothing -> throwBottom "overloaded constructors not unifiable"
```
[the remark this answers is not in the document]
traiansf: Basically, yes.  more concretely, there is no common super-symbol fitting within the top sort fixed by the injection.

kore/src/Kore/Step/Simplification/Overloading.hs
traiansf: either that, or a non-overloaded constructor.

kore/src/Kore/Step/Simplification/Overloading.hs
traiansf: Yes.  But you should know better, as you've added this code ;-) fe47fbe340a395c020b1d7737e60167ad65c9e29

kore/src/Kore/Step/Simplification/Overloading.hs
traiansf: I think it should probably return a `NotApplicable` answer, but we can definitely try it.

PR #1513, opened 2020-02-05
test/bmc/fail-2-bmc-spec.k: Stabilize output

ttuegel:
The spec fail-2-bmc-spec.k has two failing branches; which branch is reported is
not deterministic. To prevent spurious test failures, the output is stabilized
by splitting the spec on \equals(>Int(X, 0), true), which selects one failing
branch or the other without affecting the outcome of the test.

---

###### Reviewer checklist


yzhang90:
thanks tom!

PR #1517, opened 2020-02-06
One path algorithm limitations

ana-pantilie:
---

###### Reviewer checklist


design-decisions/2020-02-06-One-Path-Deterministic.md
```diff
@@ -0,0 +1,190 @@
+One-Path Reachability implemented only for deterministic semantics
+==================================================================
+
+Background
+----------
+
+Proving an one-path reachability formula `∀X.φ → ◇w ∃y.ψ` means proving that,
```
yzhang90: `y` should be capitalized in order to match the `Y` in the following text.

design-decisions/2020-02-06-One-Path-Deterministic.md
```diff
@@ -0,0 +1,190 @@
...
+)
+\/
+(
+  d => c \/ e                     -- derivation using 3rd rule
+  /\
+  test(X) /\ x <> false => c \/ e -- remainder after applying 3rd rule
+)
+\/
+(
+  e => c \/ e                     -- derivation using 1st rule
```
yzhang90: derivation using 4th (instead of 1st) rule

design-decisions/2020-02-06-One-Path-Deterministic.md
```diff
@@ -0,0 +1,190 @@
...
+`e => c \/ e`, which holds immediately, so it can be eliminated.
+Its remainder, too, can be eliminated, as 
+` x <> true /\ x <> false` is equivalent to `⊥`.
+
+This concludes the proof, as we have found witnesses of reaching the
+destination for all instances of the input.
+
+### Note
+
+When deriving goals, care should be taken about the same rule unifing in
```
yzhang90: I think in the current implementation of one path, it will consider all the unification instances instead of picking one of them. In other words, it consider unification results as `and` instead of `or`.

traiansf: I'm not 100% sure about that; An interesting anecdote that @ana-pantilie  can confirm is that in the `select` example above, attempting to prove that `select` can rewrite into `a` succeeds, while attempting to prove that it rewrites to either `b` or `c` fails...

ana-pantilie: The following claim can be proven:
```  rule <k> select => ?X </k>
       <state> SetItem(a) SetItem(b) </state> 
    ensures ?X ==K a [one-path]
```
while if we switch `a` for `b` we get the following result:
```<generatedTop>
  <k>
    a
  </k>
  <state>
    SetItem ( a )
    SetItem ( b )
  </state>
</generatedTop>
```
@traiansf and I spent quite a while debugging this, and we also observed that in the case of `all-path` it behaves as expected, which led us to believe that our `one-path` implementation is the culprit.

ttuegel: @traiansf This is a bug, right? Shouldn't one-path implementation consider all the unifiers?

traiansf: yet, it should , but it should make a disjunction between them.

PR #1518, opened 2020-02-06
Reject claims with universally-quantified variables appearing only on the right-hand side

### LINKED ISSUE #1466: Reject claims with universally-quantified variables appearing only on the right-hand side
After: #1460
Fixes: #1462

andreiburdusa:
---
Fixes #1466 
###### Reviewer checklist


kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -377,10 +390,22 @@ verifyClaimSentence sentence =
...
   where
-    addClaim verified attrs =
-        Lens.over
-            (field @"indexedModuleClaims")
-            ((attrs, verified) :)
+    addClaim verifiedSentenceClaim@(SentenceClaim verified) attrs =
+        case fromSentenceAxiom (attrs, verified) of
+            Right (OnePathClaimPattern (OnePathRule rulePattern))
+              | not $ containsForall $ left rulePattern
+              , containsForall $ right $ rhs rulePattern ->
```
andreiburdusa: Should I also check in the `antiLeft`, `requires` or `ensures`?

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -377,10 +390,22 @@ verifyClaimSentence sentence =
...
+                    error "Found claim with universally-quantified variables \
+                        \appearing only on the right-hand side"
```
andreiburdusa: Since the issue is called "Reject claims..." I only checked the `OnePathClaimPattern` and `AllPathClaimPattern` constructors. Should I also add cases for the others, like `RewriteAxiomPattern`?

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
ttuegel: Yes; `antiLeft` and `requires` are considered part of the left-hand side and `ensures` is considered part of the right-hand side.

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
ttuegel: No, it's alright to have variables that only appear in the right-hand side of a `RewriteAxiomPattern`. We can apply those rules. We want to reject the claims because we can't prove them.

andreiburdusa: That makes sense

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -377,10 +393,26 @@ verifyClaimSentence sentence =
...
   where
-    addClaim verified attrs =
-        Lens.over
-            (field @"indexedModuleClaims")
-            ((attrs, verified) :)
+    addClaim verifiedSentenceClaim@(SentenceClaim verified) attrs =
+        case fromSentenceAxiom (attrs, verified) of
+            Right (OnePathClaimPattern (OnePathRule rulePattern))
+              | rejectRulePattern rulePattern ->
+                    error "Found claim with universally-quantified variables \
```
ttuegel: Please use `withSentenceClaimContext` and `koreFail` instead of `error`, so that the error message will refer to the correct line of the file.

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -375,12 +392,30 @@ verifyClaimSentence sentence =
...
+              | rejectRulePattern rulePattern -> True
+            _ -> False
+    rejectRulePattern
+        RulePattern { left, antiLeft, requires, rhs = RHS { right, ensures } }
+      =
+        (not . any containsForall)
+            (catMaybes [antiLeft] <> [left, unwrapPredicate requires])
+        && any containsForall [right, unwrapPredicate ensures]
+    sentenceClaimError sentenceClaim =
+        withSentenceClaimContext sentenceClaim $ koreFail "Found claim with \
```
ttuegel: I was partly mistaken: we already use `withSentenceClaimContext` above, so we don't need to repeat it here.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -445,6 +446,33 @@ isFullySimplified = Attribute.isFullySimplified . extractAttributes
 simplifiedAttribute :: TermLike variable -> Pattern.Simplified
 simplifiedAttribute = Attribute.simplifiedAttribute . extractAttributes
 
+containsForall :: TermLike variable -> Bool
+containsForall = Recursive.cata containsForall0
```
ttuegel: This is not what we need to be checking. We want to check that the free variables on the right-hand side of the claim (which are _implicitly_ universally quantified) also appear on the left-hand side of the claim. We don't need to walk the entire pattern, we just need to look at the `freeVariables`. If the right-hand side has an explicit `\forall` quantifier, I think we should assume that the user intended to write that, because by default the frontend won't allow it.

kore/src/Kore/ASTVerifier/PatternVerifier.hs
```diff
@@ -381,7 +381,7 @@ verifyApplyAlias application =
             _ ->
                 koreFail
                     "The alias was declared with an element variable, but its\
-                    \argument is not an element variable."
+                    \ argument is not an element variable."
```
andreiburdusa: It's not linked to my pr, I just noticed this missing space while debugging

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -417,15 +421,24 @@ verifyClaimSentence sentence =
...
             rhs = [right, unwrapPredicate ensures]
             freeVariablesUnion terms
                 = getFreeVariables
                     (foldMap freeVariables terms :: FreeVariables Variable)
+            existentialsInRHS = Set.fromList $ fmap ElemVar existentials
         in
-            not $ isSubsetOf (freeVariablesUnion rhs) (freeVariablesUnion lhs)
+            not 
+                $ isSubsetOf
+                    (freeVariablesUnion rhs \\ existentialsInRHS)
```
ttuegel: Please add an instance of `HasFreeVariables` for `RHS` instead, so that `freeVariables` actually works correctly.

Issue #1477, opened 2020-01-28
Suppressed "unknown sort" error


ttuegel:
I get this error in the log:

```
Debug (LogMessage): 0 |-> fTestInt{}(z:testSort{}) []
Debug (DebugSolverSend): (declare-fun <1> () |HB_testSort| )
Debug (DebugSolverRecv): (error "line 6 column 20: Invalid function declaration: unknown sort 'HB_testSort'")
```

but it isn't reported as an error. Instead, it continues until we try an assertion:

```
Debug (DebugSolverSend): (assert (and (= (< <0> 0 ) false ) (not (and (not (= (< <0> 0 ) false ) ) (= <1> <2> ) ) ) ) )
Debug (DebugSolverRecv): (error "line 9 column 77: unknown constant <1>")
```

The failure to declare the sort is probably a mistake in the test environment, but the suppression of the initial "unknown sort" error makes it quite difficult to debug!

PR #1531, opened 2020-02-10
Declare constructors first to the SMT + add SMT declarations to test env

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1477

- Will prefer to use constructor declaration instead of symbol declaration
- Adds tests with actual encoded predicates
- Fixes tests which run using the SMT:
    - initialize SMT with MockSymbols declarations
    - fix empty module test
---

###### Reviewer checklist


kore/src/Kore/Step/SMT/Representation/All.hs
```diff
@@ -46,9 +53,40 @@ sorts).
...
+removeDuplicates :: Step.AST.SmtDeclarations -> Step.AST.SmtDeclarations
+removeDuplicates Step.AST.Declarations { sorts, symbols } =
+    let constructorSymbols =
+            join $ mapMaybe getConstructorNames (Map.elems sorts)
+     in Step.AST.Declarations
+         { sorts
+         , symbols = filter (isNotDuplicate constructorSymbols) symbols
+         }
+  where
+    getConstructorNames
```
ttuegel: We should make this a left-fold returning a `Set`:

```.hs
getConstructorNames
    :: Set Text
    -> Step.AST.Sort SMT.AST.SExpr Text Text
    -> Set Text
```

so that above we need only `let constructorSymbols = foldl' getConstructorNames Set.empty (Map.elems sorts)`. A `Set` will be more efficient for filtering symbols below.

kore/src/Kore/Step/SMT/Representation/All.hs
```diff
@@ -46,9 +53,40 @@ sorts).
...
-    -> AST.SmtDeclarations
+    -> Step.AST.SmtDeclarations
 build indexedModule sortConstructors =
-    resolve (sorts `AST.mergePreferFirst` symbols)
+    removeDuplicates $ resolve (sorts `Step.AST.mergePreferFirst` symbols)
   where
     sorts = Sorts.buildRepresentations indexedModule sortConstructors
     symbols = Symbols.buildRepresentations indexedModule
+
+removeDuplicates :: Step.AST.SmtDeclarations -> Step.AST.SmtDeclarations
```
ttuegel: Could we call this `removeDuplicateConstructorDeclarations` instead? Because it's only removing constructor symbol declarations, and someone might look at it and think there could be other forms of duplication.

kore/src/Kore/Step/SMT/Representation/All.hs
```diff
@@ -46,9 +57,46 @@ sorts).
...
+    -> Step.AST.SmtDeclarations
 build indexedModule sortConstructors =
-    resolve (sorts `AST.mergePreferFirst` symbols)
+    removeDuplicateConstructorDeclarations
+    $ resolve (sorts `Step.AST.mergePreferFirst` symbols)
   where
     sorts = Sorts.buildRepresentations indexedModule sortConstructors
     symbols = Symbols.buildRepresentations indexedModule
+
+removeDuplicateConstructorDeclarations
```
virgil-serbanuta: I think it would be preferable to solve this in `kore/src/Kore/Step/SMT/Representation/Symbols.hs` when merging the different symbol types, i.e. in this code:
```
    listToDeclarations builtinDeclarations
    `AST.mergePreferFirst` listToDeclarations smtlibDeclarations
    `AST.mergePreferFirst` listToDeclarations constructorDeclarations
```
i.e. instead of `AST.mergePreferFirst` we could use something else that uses the constructor, or we could put the constructors before the smtlibDeclarations.

Also, note that, if you remove a symbol from this list, it likely means that you will not be able so send it to the SMT solver. Using the constructor entry for the symbol would solve this issue.

Could you add a test for this? A combination between `Constructors work` and `Uses smtlib name for constructor` in `kore/test/Test/Kore/Step/SMT/Symbols.hs`, but using an actual encoded predicate, would be nice.

I may approve this if the test works, even of you keep the current solution.

ana-pantilie: @virgil-serbanuta I'm sorry but I haven't been able to implement a test which would reproduce the issue this PR intends to fix. I think this is because I haven't been able to figure out a sequence of sentence declarations which would reproduce this. Also, we didn't really get to the reason why there are duplicate symbol declarations, only that they are not resolved by the current master code. 

For example, when running one of the specs where this happened, the `SmtDeclarations` in `Kore.Step.Representation.All` would end up with the following declarations:
```
( Id {getId = "SortWordStack", idLocation = AstLocationFile (FileLocation {fileName = "test-sum-to-n-vdefinition.kore", line = 153, column = 8})}
, Sort { smtFromSortArgs, declaration = SortDeclarationDataType (DataTypeDeclaration {name = "|HB_SortWordStack|", typeArguments = [], constructors = [Constructor {name = "_dotWS", arguments = []},Constructor {name = "_WS_", arguments = [ConstructorArgument {name = "|HB_Lbl'UndsColnUndsUnds'EVM-TYPES'Unds'WordStack'Unds'Int'Unds'WordStack1|", argType = Atom "Int"},ConstructorArgument {name = "|HB_Lbl'UndsColnUndsUnds'EVM-TYPES'Unds'WordStack'Unds'Int'Unds'WordStack2|", argType = Atom "|HB_SortWordStack|"}]}]})}
)

( Id {getId = "Lbl'Stop'WordStack'Unds'EVM-TYPES'Unds'WordStack", idLocation = AstLocationFile (FileLocation {fileName = "test-sum-to-n-vdefinition.kore", line = 492, column = 10})}
, Symbol { smtFromSortArgs, declaration = SymbolDeclaredDirectly (FunctionDeclaration {name = "_dotWS", inputSorts = [], resultSort = Atom "|HB_SortWordStack|"})}
)
```
The first entry declares a sort with a constructor named `_dotWS`, and the second is a symbol declaration with the same name. The only sentence in the definition which has an `smtlib` attribute with this name is the following:
```
  symbol Lbl'Stop'WordStack'Unds'EVM-TYPES'Unds'WordStack{}() : SortWordStack{} [functional{}(), constructor{}(), org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/ttuegel/evm-semantics/.build/defn/haskell/evm-types.k)"), smtlib{}("_dotWS"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(362,26,362,59)"), injective{}()]
```
Do you have any idea how to reproduce this in a unit test?

virgil-serbanuta: Answered face to face.

kore/src/Kore/Step/SMT/Representation/Symbols.hs
```diff
@@ -71,8 +71,8 @@ buildRepresentations
     -> AST.UnresolvedDeclarations
 buildRepresentations indexedModule =
     listToDeclarations builtinDeclarations
-    `AST.mergePreferFirst` listToDeclarations constructorDeclarations
     `AST.mergePreferFirst` listToDeclarations smtlibDeclarations
+    `AST.mergePreferFirst` listToDeclarations constructorDeclarations
```
ttuegel: What is the effect of moving the `constructorDeclarations` to the end? I would have expected that we _prefer_ sending constructor declarations to the solver.

kore/src/Kore/Step/SMT/Representation/Symbols.hs
```diff
@@ -193,7 +193,11 @@ constructorDeclaration
         Attribute.Symbol.constructor attributes
     Attribute.Functional { isDeclaredFunctional } =
         Attribute.Symbol.functional attributes
-    encodedName = AST.encodable symbolConstructor
+    Attribute.Smtlib { getSmtlib } = Attribute.Symbol.smtlib attributes
+    encodedName =
```
ttuegel: Just to make sure I understand correctly: This encodes the contructor's name for the solver, _unless_ the constructor symbol was given an `smtlib` attribute by the user, in which case we use the provided name. Is that right? If so, I think that's a pretty important piece of logic, and even if the function is small, I would like us to lift it to the top-level and give documentation.

ana-pantilie: I didn't end up keeping these changes. We were debugging one of the tests and thought that this was causing an issue, but it wasn't the case.

kore/src/Kore/Step/SMT/Representation/Symbols.hs
ana-pantilie: If I can remember correctly, @virgil-serbanuta explained that swapping these two lines would result in the symbol declaration being represented as https://github.com/kframework/kore/blob/d13e1a0c3257f1e2e52764468b5e3a021832b8f1/kore/src/Kore/Step/SMT/AST.hs#L203 instead of as https://github.com/kframework/kore/blob/d13e1a0c3257f1e2e52764468b5e3a021832b8f1/kore/src/SMT/AST.hs#L179. I hope this is correct, if not, I'm sure @virgil-serbanuta has a much better idea of what is going on here.

virgil-serbanuta: Well, the current change moves the constructor declarations in the middle, precisely because that makes sure that we don't send smtlib declarations for constructors.

Constructors are declared when we declare the sort, the entry in `constructorDeclarations` is used only for bookkeeping, since we need to know what symbols does the SMT understand. The mechanism through which that happens is, indeed the `IndirectSymbolDeclaration` vs `FunctionDeclaration` thing that Ana mentioned.

PR #1515, opened 2020-02-05
Separate SubstitutionOrd from Ord

ttuegel:
This pull request uses a new class `SubstitutionOrd` to sort substitutions instead of relying on `Ord`, decoupling "generating fresh variables" from "choosing variables to instantiate".

---

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`

PR #1534, opened 2020-02-10
toRepresentationCondition: Use layoutOneLine

ttuegel:
layoutOneLine is faster than unparseToText, which does "smart" rendering
internally.

---

###### Reviewer checklist


ttuegel:
There is a much better way to do this once #1515 is merged.

Issue #1529, opened 2020-02-10
Kore.Internal.Substitution as a bounded context


ttuegel:
`Kore.Internal.Substitution.orderRenaming` and `Kore.Step.Simplification.AndTerms.variableFunctionAndEquals` separately maintain the same invariant of the `Substitution` type. Instead, we should treat `Substitution` as a bounded context and  maintain this invariant in a single place only.

---

- [x] Introduce an `Assignment` type such as `data Assignment variable = Assignment_ !(UnifiedVariable variable) !(TermLike variable)` with a hidden constructor and a smart constructor like `assign :: UnifiedVariable variable -> TermLike variable -> Assignment variable` that applies `orderRenaming`.
- [x] Add a pattern synonym for deconstructing (but not constructing!) `Assignment` and document the invariant that it always returns a properly-ordered assignment.
- [x] Optionally, rename `orderRenaming` to something like `normalOrder`, in case we ever need to deal with additional ordering problems.
- [x] Replace `(UnifiedVariable variable, TermLike variable)` with `Assignment variable` everywhere in `Kore.Internal.Substitution`.
- [x] Add a function `Kore.Internal.Pattern.assign :: UnifiedVariable variable -> TermLike variable -> Pattern variable` which
  - creates an `Assignment`,
  - unpacks the `Assignment` to get the `term` of the result, and
  - creates a `Substitution` from the `Assignment` which is the `substitution` of the result.

PR #1537, opened 2020-02-11
Separate substitution priority order from variable names

ttuegel:
Instead of using `Ord` to order variables for substitution, this pull request uses `SubstitutionOrd`. The class is essentially the same, but having a separate class allows distinguishing the concepts.

See also: #1529 

---

###### Reviewer checklist


kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -384,14 +381,13 @@ applyRulesSequence
...
-    initial
-    -- Wrap the rules so that unification prefers to substitute
-    -- axiom variables.
-    (map EqualityPattern.toAxiomVariables -> rules)
-  = matchRules sideCondition initial rules
+applyRulesSequence sideCondition initial rules =
+    matchRules sideCondition initial rules'
     >>= finalizeRulesSequence sideCondition initial
+  where
+    -- TODO (thomas.tuegel): Include freeVariables sideCondition
```
virgil-serbanuta: This TODO raises an interesting question:

This is called for a term somewhere inside the configuration, or inside the main condition, or something like that. If we want to avoid the variables in the side condition, shouldn't we also avoid the variables in the main configuration, but outside the term we're simplifying?

(there's probably a simple explanation for that, but it's easier to just ask)

ttuegel: I thought about this for a while. We do not need to avoid variables elsewhere in the configuration because:

1. There is no risk of confusing during rule application (matching, simplification, etc.) because the variable won't appear in this sub-term.
1. Any *new* variable introduced by the rule will be existentially quantified, so there is also no risk of confusion after the result is returned.

Issue #1491, opened 2020-01-31
Allow log entries to have an extent


ttuegel:
It would be useful to give log entries an (optional) extent, so that we can say things of child entries like "this happened while applying the rule at file.k:N:M"

See also: #1490

PR #1543, opened 2020-02-13
Issue 1491, part 2 (Refactored SMT to be based on LoggerT)

traiansf:
Part of #1491 
---

###### Reviewer checklist

### DIFF
--- a/kore/src/SMT.hs
+++ b/kore/src/SMT.hs
@@ -8,7 +8,7 @@ Maintainer  : thomas.tuegel@runtimeverification.com
 
 
 module SMT
-    ( SMT, SmtT, runSmtT
+    ( SMT
     , Solver
     , stopSolver
     , runSMT
@@ -30,7 +30,6 @@ module SMT
     , NoSMT (..), runNoSMT
     -- * Expressions
     , SExpr (..)
-    , SimpleSMT.Logger
     , SimpleSMT.nameFromSExpr
     , SimpleSMT.showSExpr
     , SimpleSMT.tBool
@@ -49,7 +48,6 @@ module SMT
 
 import Prelude
 
-import qualified Colog
 import Control.Concurrent.MVar
 import qualified Control.Exception as Exception
 import qualified Control.Monad as Monad
@@ -65,7 +63,6 @@ import Control.Monad.IO.Class
 import Control.Monad.IO.Unlift
     ( MonadUnliftIO (..)
     , UnliftIO (..)
-    , withRunInIO
     , withUnliftIO
     )
 import qualified Control.Monad.Morph as Morph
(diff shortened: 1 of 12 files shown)


kore/src/SMT.hs
```diff
@@ -219,16 +213,12 @@ class Monad m => MonadSMT m where
...
+newtype NoSMT a = NoSMT { getNoSMT :: Log.LoggerT IO a }
+    deriving (Functor, Applicative, Monad)
+    deriving (MonadCatch, MonadIO, MonadThrow, MonadLog)
 
-instance MonadLog NoSMT where
-    logEntry entry =
-        NoSMT $ ReaderT $ \logger ->
-            Colog.unLogAction logger (Log.toEntry entry)
+runNoSMT :: Log.LogAction IO Log.SomeEntry -> NoSMT a -> IO a
+runNoSMT logger noSMT = runReaderT (Log.getLoggerT (getNoSMT noSMT)) logger
```
virgil-serbanuta: some kind of runLoggerT instead of `runReaderT (Log.getLoggerT ???) logger`?

kore/src/SMT/SimpleSMT.hs
```diff
@@ -249,33 +253,32 @@ data Value =  Bool  !Bool           -- ^ Boolean value
...
 
-type Logger = Log.LogAction IO Log.SomeEntry
-
 -- | An interactive solver process.
 data Solver = Solver
     { hIn    :: !Handle
     , hOut   :: !Handle
     , hErr   :: !Handle
     , hProc  :: !ProcessHandle
-    , logger :: !Logger
```
ttuegel: We can keep this module in `IO` by introducing a new data type `SolverHandle` with just the four handle fields. Every function that returns a `Solver` now will return a `SolverHandle` instead, and nothing else in this module needs to change. In `SMT`, we can keep a `SolverHandle` instead of a `Solver`, and assemble the `Solver` from the `LogAction` and the `SolverHandle` when we call a function that needs it.

kore/src/SMT.hs
```diff
@@ -259,89 +249,70 @@ different threads may be interleaved; use 'inNewScope' to acquire exclusive
...
+        Smt $ ReaderT $ \r ->
             withUnliftIO $ \u ->
-                return (UnliftIO (unliftIO u . flip runReaderT r . runSmtT))
-
-type SMT = SmtT IO
+                return (UnliftIO (unliftIO u . flip runReaderT r . runSmt))
 
-withSolver' :: (Solver -> IO a) -> SMT a
-withSolver' action = SmtT $ do
+withSolver'
```
ttuegel: If we follow my suggestions for the `SimpleSMT` module, then this can be specialized to:

```.hs
withSolver' :: (Solver -> IO a) -> SMT a
```

and we can remove the dependency on `UnliftIO`.

PR #1549, opened 2020-02-15
Add makefile rule for running the REPL for spec tests

vladciobanu:
I noticed this seems to be missing. Our tests are obviously not using it, but I think it's nice to have for debugging and people wanting to try out our REPL.

---

###### Reviewer checklist


vladciobanu:
I think the following rule only works for files that are literally `*repl-spec.k`, whereas this works for any `-spec.k`. It should also not be picked up as a test. The point isn't to run tests but to allow us or other people to easily run any spec in our test suite using the `kore-repl`.

Issue #1551, opened 2020-02-16
Exists simplifier should respect Substitution invariants


ttuegel:
See also: https://runtimeverification.slack.com/archives/CC360GUTG/p1581688174300900

PR #1552, opened 2020-02-17
Use Target variable in exists simplifier

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1551

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/TermLike.hs
+++ b/kore/src/Kore/Step/Simplification/TermLike.hs
@@ -404,10 +404,10 @@ simplifyInternal term sideCondition = do
                     Exists.simplify
                         (targetSideCondition sideCondition)
                         (targetSimplifiedChildren simplifiedChildren)
-                return
-                    $ fmap
-                        (Pattern.mapVariables unTargetElement unTargetSet)
-                        targetedResults
+                let unTargetedResults =
+                        Pattern.mapVariables unTargetElement unTargetSet
+                        <$> targetedResults
+                return unTargetedResults
               where
                 refresh = Lens.over Binding.existsBinder refreshElementBinder
                 targetSideCondition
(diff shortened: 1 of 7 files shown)


kore/src/Kore/Variables/Target.hs
```diff
@@ -158,3 +159,13 @@ instance
     unparse (NonTarget var) = unparse var
     unparse2 (Target var) = unparse2 var
     unparse2 (NonTarget var) = unparse2 var
+
+targetIfEqual
+    :: Eq variable
+    => variable
```
ttuegel: I think this is only used on `ElementVariable`s anyway, and we should always try to keep the `ElementVariable` or `SetVariable` wrapper:

```suggestion
    => ElementVariable variable
```

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -389,11 +397,39 @@ simplifyInternal term sideCondition = do
...
                 Equals.simplify sideCondition =<< simplifyChildren equalsF
-            ExistsF exists ->
-                Exists.simplify sideCondition
-                    =<< simplifyChildren (refresh exists)
+            ExistsF exists -> do
+                simplifiedChildren <-
+                    simplifyChildren (refresh exists)
+                targetedResults <-
+                    Exists.simplify
+                        (targetSideCondition sideCondition)
```
ttuegel: Why don't we target the children before simplification? That is, add the `Target` and `NonTarget` wrappers in `refresh. That should eliminate the need to do any re-ordering at all in the `\exists` simplifier.

ana-pantilie: I tried doing that at first but I remember it resulted in some type issues, but maybe I did something wrong along the way and that's why it didn't work. I tried a few solutions and the current one seemed the cleanest, but I can revisit what you suggest if you think it's better.

ttuegel: > I remember it resulted in some type issues

That was expected, but let's merge what you have, and come back to it.

PR #1564, opened 2020-02-19
Remove UnliftIO from SMT

### LINKED ISSUE #1554: Remove UnliftIO from SMT
Everything in `SMT` that uses `MonadUnliftIO` should use `MonadMask` instead.

#### Rationale

`MonadMask` (or `MonadCatch` and `MonadThrow`) ...

1. ... are more widely used in the rest of the code.
1. ... are straightforward generalizations of functions in the standard library, so they should be more familiar to new contributors.
1. ... capture exactly the required functionality without any extra behaviors that might invite unintended use.

#### Plan

- [ ] Derive instances of `MonadMask` and its superclasses in `SMT`, as required.
- [ ] Inline `SMT.SimpleSMT.inNewScope` into [SMT.withSolver](https://github.com/kframework/kore/blob/b0746f4881ad8aa06f81a7c4b79dcb0ac902acc4/kore/src/SMT.hs#L297-L304), using [`Control.Monad.Catch.finally`](https://hackage.haskell.org/package/exceptions-0.10.4/docs/Control-Monad-Catch.html#v:finally), while removing `withRunInIO`.
- [ ] [`SMT.withSolverT'`](https://github.com/kframework/kore/blob/b0746f4881ad8aa06f81a7c4b79dcb0ac902acc4/kore/src/SMT.hs#L286-L289) shall use [`Control.Monad.Catch.bracket`](https://hackage.haskell.org/package/exceptions-0.10.4/docs/Control-Monad-Catch.html#v:bracket) instead of [`Control.Concurrent.MVar.withMVar`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Concurrent-MVar.html#v:withMVar). Refer to the definition of `withMVar` for its implementation in terms of `takeMVar` and `putMVar`.

andreiburdusa:
---
Fixes #1554 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -32,8 +32,8 @@ import qualified Control.Monad as Monad
 import Control.Monad.Catch
     ( MonadCatch
     )
-import Control.Monad.IO.Unlift
-    ( MonadUnliftIO
+import Control.Monad.IO.Class
+    ( MonadIO
     )
 import Control.Monad.Trans.Except
     ( runExceptT
@@ -198,7 +198,7 @@ exec
     ::  ( Log.WithLog Log.LogMessage smt
         , MonadProfiler smt
         , MonadSMT smt
-        , MonadUnliftIO smt
+        , MonadIO smt
         )
     => Limit Natural
     -> VerifiedModule StepperAttributes
@@ -235,7 +235,7 @@ execGetExitCode
     ::  ( Log.WithLog Log.LogMessage smt
         , MonadProfiler smt
         , MonadSMT smt
-        , MonadUnliftIO smt
+        , MonadIO smt
         )
     => VerifiedModule StepperAttributes
     -- ^ The main module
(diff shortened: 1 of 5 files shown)


kore/src/Kore/Exec.hs
```diff
@@ -32,8 +32,8 @@ import qualified Control.Monad as Monad
 import Control.Monad.Catch
     ( MonadCatch
     )
-import Control.Monad.IO.Unlift
-    ( MonadUnliftIO
+import Control.Monad.IO.Class
+    ( MonadIO
```
ttuegel: Let's add (in a separate pull request) `MonadIO` to `Prelude.Kore`.

andreiburdusa: I'll do that. But what do you mean by removing the dependency on `unliftio`?

ttuegel: I mean [removing the dependency](https://github.com/kframework/kore/blob/master/kore/package.yaml#L83) from the package. Is it still used anywhere else? It's OK if it is, I just wondered if you know.

PR #1565, opened 2020-02-19
Simplify claims with side conditions

### LINKED ISSUE #1547: Term not simplified before first step
Steps to reproduce:

1. Use this branch: https://github.com/virgil-serbanuta/evm-semantics/tree/hkg-haskell-issue-1
2. Run this: 
```
rm -rf .build/defn
make defn
kompile --debug --main-module ETHEREUM-SIMULATION --backend haskell --hook-namespaces KRYPTO --syntax-module ETHEREUM-SIMULATION .build/defn/haskell/driver.k --directory .build/defn/haskell -I .build/defn/haskell
```
3. Run this: 
```
kprove --haskell-backend-command "/home/???/kore/.build/kore/bin/kore-repl " tests/specs/erc20/hkg/allowance-spec.k -d ./.build/defn/haskell --def-module VERIFICATION --debug -v
```
4. Note that the `<callData>` cell contains (indirectly) the following term, which is marked as being simplified conditionally (`Sfc`):
```
/* Fn Sfc */
Lbl'Hash'getValue'LParUndsRParUnds'EDSL'Unds'Int'Unds'TypedArg{}(
    /* Fl Fn D Sfa */
    Lbl'Hash'address'LParUndsRParUnds'EDSL'Unds'TypedArg'Unds'Int{}(
        /* Fl Fn D Sfa */
        VarOWNER:SortInt{}
    )
)
```
However, we should be able to simplify that term with this rule from https://github.com/virgil-serbanuta/evm-semantics/blob/hkg-haskell-issue-1/edsl.md :
```
rule #getValue(#address( DATA )) => DATA
      requires #rangeAddress(DATA)
```
@yzhang90 says that we should be able to apply this using the condition from the initial term. Also this seems to be simplified after the first step.

ttuegel:
Include the `requires` clause of a claim as its side condition during initial simplification.

Fixes: #1547

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Rule/Simplify.hs
+++ b/kore/src/Kore/Step/Rule/Simplify.hs
@@ -31,7 +31,6 @@ import qualified Kore.Internal.MultiAnd as MultiAnd
     ( make
     )
 import qualified Kore.Internal.MultiOr as MultiOr
-import qualified Kore.Internal.OrPattern as OrPattern
 import Kore.Internal.Pattern
     ( Pattern
     )
@@ -68,6 +67,7 @@ import Kore.Step.Simplification.Simplify
     ( InternalVariable
     , MonadSimplify
     )
+import qualified Kore.Step.SMT.Evaluator as SMT.Evaluator
 import Kore.Syntax.Variable
     ( Variable
     )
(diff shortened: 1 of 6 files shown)


kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -131,18 +132,27 @@ instance SimplifyRuleLHS (ReachabilityRule Variable) where
...
+    . Lens.traverseOf RulePattern.leftPattern worker
   where
-    RulePattern {left, requires} = rule
+    worker, simplify, simplifyWithSolver
+        :: Pattern variable
+        -> BranchT simplifier (Pattern variable)
+    worker =
+        (return . Pattern.requireDefined)
+        >=> simplify
+        >=> simplifyWithSolver
```
virgil-serbanuta: We usually call `SMT.Evaluator.filterMultiOr` on the result of `simplifyTopConfiguration` instead of, say, `simplifyConditionsWithSmt` because we assume that the condition is already fully simplified (implied by the function name).

kore/test/Test/Kore/Step/Rule/Simplify.hs
```diff
@@ -223,3 +243,76 @@ runSimplifyRule rule =
...
+        left = Lens.view (field @"left") rule
+        sort = termLikeSort left
+
+    test
+        :: HasCallStack
+        => TestName
+        -> RulePattern Variable
+        -> [RulePattern Variable]
+        -> TestTree
+    test name (OnePathRule -> input) (map OnePathRule -> expect) =
```
virgil-serbanuta: You can leave it as-is if you want (this is a rather small test file), but I would be tempted to make it more obvious that this test uses the `g(a) => f(a)` mapping below, possibly by sending it as an argument.

PR #1571, opened 2020-02-20
Red for stuck nodes in the repl graph.

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Repl/Data.hs
+++ b/kore/src/Kore/Repl/Data.hs
@@ -374,7 +374,7 @@ helpText =
     \ was reached using the SMT solver or it was reached through the Remove \n\
     \ Destination step.\n\
     \(**) A green node represents the proof has completed on\
-    \ that respective branch.\
+    \ that respective branch. A red node represents a stuck configuration.\
     \\n\n\
     \Rule names can be added in two ways:\n\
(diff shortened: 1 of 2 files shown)


ana-pantilie:
Just to be consistent, we have an entry in the help text which documents green nodes https://github.com/kframework/kore/blob/fecc3bb99c65ced787e0cf6c4917845d4dcd881f/kore/src/Kore/Repl/Data.hs#L376, I think we should do the same for red nodes.

virgil-serbanuta:
> Just to be consistent, we have an entry in the help text which documents green nodes
> 
> https://github.com/kframework/kore/blob/fecc3bb99c65ced787e0cf6c4917845d4dcd881f/kore/src/Kore/Repl/Data.hs#L376
> 
> , I think we should do the same for red nodes.

Added to the previous footnote. Should I make a separate one instead?

ana-pantilie:
> > Just to be consistent, we have an entry in the help text which documents green nodes
> > https://github.com/kframework/kore/blob/fecc3bb99c65ced787e0cf6c4917845d4dcd881f/kore/src/Kore/Repl/Data.hs#L376
> > 
> > , I think we should do the same for red nodes.
> 
> Added to the previous footnote. Should I make a separate one instead?

It's good this way, thank you.

Issue #900, opened 2019-08-28
Group and apply rules according to their priority


ttuegel:
Group semantic rules by their `priority` attribute. Apply the rules in each group using `deriveSeq` or `derivePar` according to the claim type. Apply the groups in sequence.

See also: #859

ttuegel:
This task explicitly does not include #898.

ttuegel:
I reduced the appetite on this issue now because I just don't think it will take that long.

PR #1580, opened 2020-02-21
Group and apply rewrite rules according to their priority

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/900

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Strategies/Goal.hs
+++ b/kore/src/Kore/Strategies/Goal.hs
@@ -374,18 +374,20 @@ instance Goal (AllPathRule Variable) where
                 }
 
     strategy _ goals rules =
-        allPathFirstStep rewrites
+        allPathFirstStep priorityGroups
         :> Stream.iterate
             id
             ( allPathFollowupStep
                 coinductiveRewrites
-                rewrites
+                priorityGroups
             )
       where
-        rewrites =
+        priorityGroups' =
             groupSortOn
                 (RulePattern.getPriority . toRulePattern)
                 rules
+        priorityGroups = if null priorityGroups' then [[]] else priorityGroups'
+
         coinductiveRewrites =
             AllPathRewriteRule
             . RewriteRule
--- a/kore/src/Kore/Strategies/Goal.hs
+++ b/kore/src/Kore/Strategies/Goal.hs
@@ -29,6 +29,9 @@ import Prelude.Kore
 import Control.Exception
     ( throw
     )
+import Control.Monad
+    ( join
+    )
 import Control.Monad.Catch
     ( Exception (..)
     , MonadCatch
@@ -42,6 +45,10 @@ import Data.Coerce
     )
 import qualified Data.Default as Default
 import qualified Data.Foldable as Foldable
+import Data.List.Extra
+    ( groupSortOn
+    , sortOn
+    )
 import qualified Data.Set as Set
 import Data.Stream.Infinite
     ( Stream (..)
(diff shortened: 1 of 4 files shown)


traiansf:
Credit goes to @virgil-serbanuta for the idea of doing it in the strategy building phase.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -732,19 +742,27 @@ onePathFollowupStep claims axioms =
...
         , CheckGoalStuck
         , CheckGoalRemainder
         , Simplify
         , TriviallyValid
         , RemoveDestination
-        , DerivePar axioms
-        , Simplify
+        ]
+        <> groupStrategy axiomGroups <>
+        [ Simplify
```
virgil-serbanuta: The groupStrategy function will add `Simplify, TriviallyValid` at the end of each simplification, so I think that most of the time these will be redundant. To cover everything, maybe it would be better to add a special case to `groupStrategy` for an empty input list.

Also, `allPathFollowupStep` seems to go directly to `ResetGoal` after `groupStrategy`, they should do the same thing unless there is a reason to do otherwise.

kore/test/Test/Kore/Strategies/AllPath/Verification.hs
```diff
@@ -322,6 +323,46 @@ test_allPathVerification =
         assertEqual ""
             (Left $ Pattern.fromTermLike Mock.e)
             actual
+    , testCase "Priority: should apply first axiom" $ do
```
virgil-serbanuta: Since this is all-path, would it be more accurate to say `should also apply the first axiom`?

kore/test/Test/Kore/Strategies/OnePath/Verification.hs
```diff
@@ -402,6 +403,66 @@ test_onePathVerification =
...
+            [ simpleClaim Mock.a Mock.d
+            ]
+            []
+        assertEqual ""
+            (Right ())
+            actual
+    , testCase "Priority: should apply axiom with higher priority" $ do
+        -- Axioms:
+        --     a => b
+        --     b => c [priority(2)]
```
virgil-serbanuta: The comment does not describe the test.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -374,18 +374,20 @@ instance Goal (AllPathRule Variable) where
...
-                rewrites
+                priorityGroups
             )
       where
-        rewrites =
+        priorityGroups' =
             groupSortOn
                 (RulePattern.getPriority . toRulePattern)
                 rules
+        priorityGroups = if null priorityGroups' then [[]] else priorityGroups'
```
ttuegel: What's going on here? Why would `priorityGroups'` be empty?

kore/src/Kore/Strategies/Goal.hs
ttuegel: > To cover everything, maybe it would be better to add a special case to `groupStrategy` for an empty input list.

I agree. It looks to me like `groupStrategy` silently does something wrong if it's given an empty list. (If that's not true, then you can ignore everything that follows.) I think we should at least do one of the following:

1. Do the right thing with an empty list. (Best)
1. Take a `NonEmpty` list as its argument.
1. `assert` that its argument is non-empty. (Worst, but still better than silently doing the wrong thing.)

Issue #1574, opened 2020-02-20
Simplify map and set ceils


virgil-serbanuta:
We can simplify map ceils in the following way: if `a`, `b`, `c`, `d` are maps (either elements or map variables or opaque terms) then `#ceil(concat(a, b, c, d))` (sorry for the notation abuse, but `concat` is associative and commutative) is equal to `#ceil(concat(a, b)) and #ceil(concat(a, c)) and #ceil(concat(a, d)) and #ceil(concat(b, c)) and #ceil(concat(b, d)) and #ceil(concat(c, d))`.

Also, if `a` and `b` are elements, i.e. `a` is `e->f` and `b` is `g->h` then `#ceil(concat(a, b))` is `#ceil(a) and #ceil(b) and not (e == g)`. `#ceil(a)` is equal to `#ceil(e) and #ceil(f)`.

Also, if `a` and `b` are concrete elements, then, in the internal representation, we already checked that `not (e == g)`. Also, `#ceil(a)` is either `#ceil(f)` or `#top`, I'm not sure.

PR #1583, opened 2020-02-21
Simplify Ceil of Map and Set

ttuegel:
Fixes: #1574 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/Ceil.hs
+++ b/kore/src/Kore/Step/Simplification/Ceil.hs
@@ -17,9 +17,12 @@ module Kore.Step.Simplification.Ceil
 
 import Prelude.Kore
 
-import qualified Data.Bifunctor as Bifunctor
+import Control.Monad
+    ( zipWithM
+    )
 import qualified Data.Foldable as Foldable
 import qualified Data.Functor.Foldable as Recursive
+import qualified Data.List as List
 import qualified Data.Map.Strict as Map
 
 import qualified Kore.Attribute.Symbol as Attribute.Symbol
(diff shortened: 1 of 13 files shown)


kore/src/Kore/Step/Simplification/Ceil.hs
```diff
@@ -256,116 +263,187 @@ makeEvaluateBuiltin sideCondition (Domain.BuiltinList l) = Just $ do
...
+    ->  SideCondition variable
     ->  Domain.NormalizedAc
             normalized
             (TermLike Concrete)
             (TermLike variable)
-    -> Maybe (simplifier (OrCondition variable))
-makeEvaluateNormalizedAc
-    sideCondition
+    ->  Maybe (simplifier (OrCondition variable))
+makeEvaluateNormalizedAc mkOpaqueAc sideCondition normalizedAc =
```
virgil-serbanuta: This seems reasonable, but I was wondering if it would be worth moving the map/set code to a new file?

kore/test/Test/Kore/Step/Simplification/Ceil.hs
```diff
@@ -419,40 +423,6 @@ test_ceilSimplification =
                 , domainValueChild = mkStringLiteral "a"
                 }
         assertEqual "ceil(1)" expected actual
-    , testGroup "Builtin.Map"
```
virgil-serbanuta: I would also test this with regular testing. The property test below is nice, but it seems to me to be somewhat complex, and having some tests where it's obvious what happens may be valuable.

ttuegel:
Ah, the test failures seem to be due to insignificant changes in the golden files.

kore/test/Test/Kore/Step/Simplification/Ceil.hs
ttuegel: Done! And, it found a bug in the mock symbols I was using.

PR #1589, opened 2020-02-25
Exec and search with priority rules

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1587

Also modifies `priorityAttribute` to receive an Integer instead of a String.

Fixes default priorities bug in https://github.com/kframework/kore/pull/1580 as well.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step.hs
+++ b/kore/src/Kore/Step.hs
@@ -12,9 +12,9 @@ module Kore.Step
     , rewrite
     , simplify
     , rewriteStep
+    , priorityAllStrategy
+    , priorityAnyStrategy
     , transitionRule
-    , allRewrites
-    , anyRewrite
     , heatingCooling
       -- * Re-exports
     , RulePattern
@@ -29,6 +29,10 @@ import Prelude.Kore
 
 import qualified Control.Monad.Trans as Monad.Trans
 import qualified Data.Foldable as Foldable
+import Data.List.Extra
+    ( groupSortOn
+    , sortOn
+    )
 import qualified Data.Text.Prettyprint.Doc as Pretty
 import Numeric.Natural
     ( Natural
@@ -42,8 +46,10 @@ import qualified Kore.Step.Result as Result
     )
 import qualified Kore.Step.RewriteStep as Step
 import Kore.Step.RulePattern
-    ( RewriteRule (RewriteRule)
+    ( RewriteRule (..)
     , RulePattern
+    , ToRulePattern (..)
+    , getPriority
     , isCoolingRule
     , isHeatingRule
     , isNormalRule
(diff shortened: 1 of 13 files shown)


kore/src/Kore/Step.hs
```diff
@@ -163,13 +169,40 @@ anyRewrite
 anyRewrite rewrites =
     Strategy.any (rewriteStep <$> rewrites)
 
+priorityAllStrategy
+    :: ToRulePattern rewrite
+    => [rewrite]
+    -> Strategy (Prim rewrite)
+priorityAllStrategy rewrites =
+    Strategy.any (fmap allRewrites priorityGroups)
```
ttuegel: Is `Strategy.any` the right choice here? @traiansf

ttuegel: Also, if we replaced `allRewrites` with `anyRewrite`, then wouldn't this be a valid implementation of `priorityAnyStrategy`? If that is the case, could we parameterize this function by the "inner" strategy, so that we have a single implementation? Or in other words, make this a combinator?

kore/src/Kore/Step.hs
```diff
@@ -163,13 +169,40 @@ anyRewrite
...
+
 {- | Heat the configuration, apply a normal rewrite, and cool the result.
  -}
 -- TODO (thomas.tuegel): This strategy is not right because heating/cooling
 -- rules must have side conditions if encoded as \rewrites, or they must be
 -- \equals rules, which are not handled by this strategy.
 heatingCooling
-    :: (forall rewrite. [rewrite] -> Strategy (Prim rewrite))
+    ::  ( forall rewrite
+         .  ToRulePattern rewrite
```
ttuegel: Why this new constraint?

kore/src/Kore/Step.hs
traiansf: > Is Strategy.any the right choice here?

I think that it serves the purpose, at least if we assume the initial configuration to be concrete (i.e., no possibility of remainders).  My argument for this choice is that, at least in its current implementation, any would try each group in sequence, until it finds one matching; if there are multiple matches in that group, it will explore all of them, but would not try rules in subsequent groups.

If we would like to think about symbolic search instead, then we would probably end up with something similar to the all-path strategy.

> Or in other words, make this a combinator?

We used to have it written as a combinator, but we weren't sure there would be other uses for it; note also that sorting seems more natural than grouping for `priorityAnyStrategy`.  

But, of course, if you think it might be useful to have it as a combinator, we'll make it so :-)

kore/src/Kore/Step.hs
traiansf: If I remember correctly, I think this was needed to be able to combine heatingCooling with `priorityAllStrategy`, which has this additional constraint.

kore/src/Kore/Step/RulePattern.hs
```diff
@@ -303,8 +304,16 @@ isNormalRule RulePattern { attributes } =
         Attribute.Normal -> True
         _ -> False
 
-getPriority :: RulePattern variable -> Attribute.Priority
-getPriority = Attribute.priority . attributes
+getPriorityOfRule :: RulePattern variable -> Integer
+getPriorityOfRule RulePattern { attributes } =
+    if isOwise
```
ttuegel: Actually, I do have two requests:

1. Instead of hard-coding the values here, add top-level definitions `owisePriority, defaultPriority :: Integer`, and
1. Make `defaultPriority = 50`, see here: https://runtimeverification.slack.com/archives/C7E701MFG/p1582663748043400

kore/src/Kore/Step.hs
virgil-serbanuta: If we are going to keep Strategy.any here, should we rename it to Strategy.first, or something like that? We could also add an any function that just calls first.

kore/src/Kore/Step.hs
```diff
@@ -163,13 +169,40 @@ anyRewrite
...
+         => [rewrite] -> Strategy (Prim rewrite)
```
virgil-serbanuta: Here and above: Indentation should probably be a multiple of 4.

PR #1590, opened 2020-02-25
Bug: repl should not generate only one-path claims at exit

ana-pantilie:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Repl/State.hs
+++ b/kore/src/Kore/Repl/State.hs
@@ -24,7 +24,7 @@ module Kore.Repl.State
     , sortLeafsByType
     , generateInProgressClaims
     , currentClaimSort
-    , conjOfOnePathClaims
+    , conjOfClaims
     , appReplOut
     , replOut, replOutputToString
     ) where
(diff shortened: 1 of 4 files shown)


kore/src/Kore/Repl/State.hs
```diff
@@ -654,13 +654,13 @@ substituteAlias
...
     :: Claim claim
     => (claim, TermLike Variable)
-    -> Rule.OnePathRule Variable
-createOnePathClaim (claim, cpattern) =
-    Rule.OnePathRule
-    $ Rule.RulePattern
+    -> claim
+createClaim (claim, cpattern) =
+    fromRulePattern claim
+    Rule.RulePattern
```
vladciobanu: The formatting here looks a bit off since it's not indented compared to the previous line. It feels like a do block.

kore/src/Kore/Repl/State.hs
```diff
@@ -669,44 +669,43 @@ createOnePathClaim (claim, cpattern) =
         }
 
 conjOfOnePathClaims
```
vladciobanu: Was this not also meant to be polymorphic over the claim type, like the other changes?

PR #1591, opened 2020-02-25
Use applied rule in warnSimplificationWithRemainder

### LINKED ISSUE #1570: Warning for simplification with remainders should report which rule was applied

andreiburdusa:
---
Fixes #1570 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Log/WarnSimplificationWithRemainder.hs
+++ b/kore/src/Kore/Log/WarnSimplificationWithRemainder.hs
@@ -19,6 +19,7 @@ import Data.Typeable
 import qualified Generics.SOP as SOP
 import qualified GHC.Generics as GHC
 
+import qualified Kore.Attribute.Axiom as Attribute.Axiom
 import Kore.Internal.OrPattern
     ( OrPattern
     )
@@ -35,6 +36,10 @@ import Kore.Internal.TermLike
     )
 import qualified Kore.Internal.TermLike as TermLike
 import Kore.Internal.Variable
+import Kore.Step.EqualityPattern
+    ( EqualityRule (..)
+    )
+import qualified Kore.Step.EqualityPattern as EqualityPattern
 import Kore.Unparser
 import Log
 import Pretty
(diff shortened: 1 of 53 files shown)


kore/src/Kore/Step/Axiom/Registry.hs
```diff
@@ -162,10 +162,10 @@ equalitiesToEvaluators
   =
```
andreiburdusa: This function takes multiple rules, and `scatter`s them, so I try to figure out how to keep track of each rule

kore/src/Kore/Log/WarnSimplificationWithRemainder.hs
```diff
@@ -102,26 +107,24 @@ warnSimplificationWithRemainder
     => InternalVariable variable
     => TermLike variable  -- ^ input pattern
     -> SideCondition variable  -- ^ input condition
-    -> OrPattern variable  -- ^ results
     -> OrPattern variable  -- ^ remainders
+    -> EqualityRule Variable
```
ana-pantilie: I think you can remove the TODO above now.

kore/src/Kore/Step/Axiom/EvaluationStrategy.hs
```diff
@@ -101,8 +101,20 @@ simplificationEvaluation rule =
     BuiltinAndAxiomSimplifier $ \term condition -> do
         results' <- evaluateAxioms [rule] condition term
         let initial = Step.toConfigurationVariables (Pattern.fromTermLike term)
+            remainders' = Results.remainders results'
         Step.recoveryFunctionLikeResults initial results'
-        return $ Results.toAttemptedAxiom results'
+        let attemptedAxiom = Results.toAttemptedAxiom results'
+        case attemptedAxiom of
```
ana-pantilie: I think you can factor this `case` out to make the code more readable.

ttuegel: I agree, this should be something like `when (hasRemainder attemptedAxiom) $ warnSimplificationWithRemainder ...`

Issue #1556, opened 2020-02-18
kore-repl: save proof state


ttuegel:
Add a command to save the current proof state to a file so that `kore-repl` can quickly resume.

See also: https://runtimeverification.slack.com/archives/CC360GUTG/p1581946734321200

PR #1593, opened 2020-02-26
Repl: save partial proof command

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1556

---

###### Reviewer checklist


kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -992,6 +1004,75 @@ saveSession path =
...
+    currentIndex <- Lens.use (field @"claimIndex")
+    claims <- Lens.use (field @"claims")
+    maybeConfig <- getConfigAt maybeNode
+    case maybeConfig of
+        Nothing -> putStrLn' "Invalid node!"
+        Just (_, currentProofState) -> do
+            let config = unwrapConfig currentProofState
+                newClaim = createClaim currentClaim config
+                newTrustedClaims =
+                    makeTrusted <$> removeAt currentIndex claims
```
virgil-serbanuta: If this removes the current claim: except before the first step, the current claim should also be marked as trusted (in coinductive proofs, the claim being proven can be use after the first step).

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -992,6 +1004,75 @@ saveSession path =
...
+                newDefinition =
+                    createNewDefinition
```
virgil-serbanuta: I think the original spec module may contain things like symbol definitions, it would be nice to preserve them.

kore/src/Kore/Repl/Interpreter.hs
ana-pantilie: That's right, thanks for catching this!

kore/src/Kore/Repl/Interpreter.hs
ana-pantilie: I thought only claim definitions could appear in the spec module, I asked @traiansf about this, but maybe I misunderstood.

virgil-serbanuta: Ok, he should know more about this than me.

test/tiny/test-save-partial-proof.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+IGNORE=$($KORE_REPL test-a-to-c-vdefinition.kore -r --repl-script run-one-save-state-repl-script-spec.k.repl --module VERIFICATION --prove test-a-to-c-spec.kore --spec-module A-TO-C-SPEC)
```
virgil-serbanuta: Why the `IGNORE=$(...)` part?

ana-pantilie: The Makefile will dump all the output from the script into the `.golden` and `.out` files, and I wanted to ignore the output from the two repl processes since I'm only interested in `diff`-ing the config files. I tried to do that by redirecting the outputs to `/dev/null` (which seemed nicer) but for some reason unknown to me the script would complain it couldn't find the two config files. This seemed like a reasonable workaround which works.

Issue #1557, opened 2020-02-18
Attributes to replace #isConcrete


ttuegel:
The `#isConcrete` symbol is unsound in Kore. We will replace it everywhere with axiom attributes because it is sound for the backend to use attributes as hints when a rule should or should not be applied.

- `concrete(X, Y, Z, ...)`: the rule will apply only if variables `X`, `Y`, `Z`, etc. are instantiated at constructor-like patterns. The named variables must appear free in the rule; this will be checked in the frontend.
- `concrete()` or `concrete`: the existing behavior becomes a special case of the above; the rule will apply only if all variables are instantiated at constructor-like patterns.
- `concrete(X), concrete(Y)` will be equivalent to `concrete(X, Y)`.
- The opposite attribute `symbolic(X, Y, Z, ...)` will also be defined.
- It is an error if the same variable is mentioned in a `concrete` and `symbolic` attribute.

#### Plan

The backend will receive variable names as strings, so some refactoring will be required to properly verify axiom sentence attributes.

Each of the listed steps should be workable as a pull request in its own right.

- [x] Remove the [`ParseAttributes`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Attribute/Axiom.hs#L165) instance of `Kore.Attribute.Axiom.Axiom`. Lift the definition of `parseAttributes` to the top-level of that module as `parseAxiomAttributes`. Fix compilation errors by replacing `parseAttributes @Axiom` with `parseAxiomAttributes`.
- [x] Remove the [`ParseAttributes`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Attribute/Axiom/Concrete.hs#L56) instance of `Kore.Attribute.Axiom.Concrete.Concrete`. Lift the definition of `parseAttributes` to the top-level of that module as `parseConcreteAttribute`. Fix compilation errors by replacing `parseAttributes @Concrete` with `parseConcreteAttribute`.
- [x] Give `Concrete` a phantom `variable` type parameter, and do the same for `Axiom`. Fix compilation errors.
[...]

PR #1594, opened 2020-02-26
Begin replacing #isConcrete with axiom attributes

andreiburdusa:
---
This solves the first 3 list items from #1557 
###### Reviewer checklist


ttuegel:
About having to add a type parameter to `Axiom` everywhere, I suggest:

1. Add the phantom type parameter `variable` to `Concrete`. Fix the errors in `Kore.Attribute.Axiom.Concrete` by changing `Concrete` to `Concrete variable`, but fix the errors in all other modules by changing `Concrete` to `Concrete Variable`.
1. Change the `Concrete Variable` in `Axiom` to `Concrete variable` and add a type parameter to `Axiom`. Fix the errors in `Kore.Attribute.Axiom` by adding a type parameter, but fix the errors in all other modules by changing `Axiom symbol ` to `Axiom symbol Variable`.
1. Add `mapAxiomVariables` and `mapConcreteVariables` functions with types similar to [`mapFreeVariables`](https://github.com/kframework/kore/blob/bdfb0efa6538a72fee230ce34dc7db25dff03f34/kore/src/Kore/Attribute/Pattern/FreeVariables.hs#L87). These won't actually do anything yet.
1. Keep going "up" by changing `Axiom symbol Variable` to `Axiom symbol variable` in types until you get to anything in `Kore.IndexedModule` or `Kore.ASTVerifier`. In those hierarchies, we can keep `variable ~ Variable`. At some point, you will need the `mapAxiomVariables` function defined above.

I think that will get you through to the next step.

kore/src/Kore/Exec.hs
```diff
@@ -567,7 +567,7 @@ assertSomeClaims claims =
 makeClaim
     :: Goal.FromRulePattern claim
     => Goal.ToRulePattern claim
-    => (Attribute.Axiom Symbol, claim) -> claim
+    => (Attribute.Axiom Symbol Variable, claim) -> claim
```
andreiburdusa: I left this a `Variable`:
`fromRulePattern` has to return a `RulePattern Variable` because it's first parameter (`ruleType`) has that type, returned by `toRulePattern`

kore/src/Kore/Exec.hs
```diff
@@ -671,7 +671,10 @@ initializeProver definitionModule specModule maybeAlreadyProvenModule within =
         tools <- Simplifier.askMetadataTools
         let Initialized { rewriteRules } = initialized
             changedSpecClaims
-                :: [(Attribute.Axiom Symbol, MaybeChanged (ReachabilityRule Variable))]
+                ::  [   ( Attribute.Axiom Symbol Variable
```
andreiburdusa: `initializeProver` works with `VerifiedModule`s, whose `axiomAtts` is set to be `Axiom Symbol Variable`

kore/src/Kore/Step/Rule.hs
```diff
@@ -145,7 +145,7 @@ extractRewriteAxioms idxMod =
     mapMaybe extractRewriteAxiomFrom (indexedModuleAxioms idxMod)
 
 extractRewriteAxiomFrom
-    :: (Attribute.Axiom Internal.Symbol.Symbol, Verified.SentenceAxiom)
+    :: (Attribute.Axiom Internal.Symbol.Symbol Variable, Verified.SentenceAxiom)
```
andreiburdusa: `fromSentenceAxiom` returns an `Either ... (QualifiedAxiomPattern Variable)` so I thought the `sentence` should also depend on a `Variable`

PR #1608, opened 2020-02-28
Continue replacing #isConcrete with axiom attributes

andreiburdusa:
---
See #1557 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/ASTVerifier/SentenceVerifier.hs
+++ b/kore/src/Kore/ASTVerifier/SentenceVerifier.hs
@@ -502,4 +502,5 @@ parseAndVerifyAxiomAttributes indexModule attrs =
     parseAxiomAttributes' attrs >>= verifyAxiomAttributes indexModule
   where
     parseAxiomAttributes' =
-        Attribute.Parser.liftParser . Attribute.parseAxiomAttributes
+        Attribute.Parser.liftParser . Attribute.parseAxiomAttributes freeVariables'
+    freeVariables' = fmap (freeVariables . sentenceAxiomPattern . snd) . indexedModuleAxioms $ indexModule
--- a/kore/src/Kore/ASTVerifier/SentenceVerifier.hs
+++ b/kore/src/Kore/ASTVerifier/SentenceVerifier.hs
@@ -354,18 +354,24 @@ verifyAliasSentence sentence = do
     sortParams   = (aliasParams . sentenceAliasAlias) sentence
     expectedSort = sentenceAliasResultSort
 
-verifyAxioms :: [ParsedSentence] -> SentenceVerifier ()
-verifyAxioms =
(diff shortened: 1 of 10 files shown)


kore/src/Kore/Attribute/Axiom.hs
```diff
@@ -226,7 +231,7 @@ parseAxiomAttribute attr =
         Monad.>=> typed @Unit (parseAttribute attr)
         Monad.>=> typed @Idem (parseAttribute attr)
         Monad.>=> typed @Trusted (parseAttribute attr)
-        Monad.>=> typed @(Concrete Variable) (parseConcreteAttribute attr)
+        Monad.>=> typed @(Concrete Variable) (parseConcreteAttribute attr def)
```
ttuegel: This is the problem: You will need to pass in the `FreeVariables` to `parseAxiomAttribute` also. Sorry, I forgot that part.

kore/src/Kore/ASTVerifier/SentenceVerifier.hs
```diff
@@ -502,4 +502,5 @@ parseAndVerifyAxiomAttributes indexModule attrs =
     parseAxiomAttributes' attrs >>= verifyAxiomAttributes indexModule
   where
     parseAxiomAttributes' =
-        Attribute.Parser.liftParser . Attribute.parseAxiomAttributes
+        Attribute.Parser.liftParser . Attribute.parseAxiomAttributes freeVariables'
```
ttuegel: Here you need to use only the `FreeVariables` of the axiom that these attributes belong to. You will probably need to pass them in as another argument from the place where the sentences are verified.

kore/src/Kore/Attribute/Pattern/FreeVariables.hs
```diff
@@ -55,6 +56,9 @@ instance Hashable variable => Hashable (FreeVariables variable) where
     hashWithSalt salt (FreeVariables freeVars) =
         hashWithSalt salt (Set.toList freeVars)
 
+instance Default (FreeVariables variable) where
```
ttuegel: I think we should remove this instance to prevent confusion.

kore/src/Kore/Step/Axiom/Evaluate.hs
```diff
@@ -139,12 +141,20 @@ evaluateAxioms equalityRules sideCondition termLike
...
+        Attribute.Axiom.Concrete.freeVariables
         . Attribute.Axiom.concrete
         . EqualityPattern.attributes
         . getEqualityRule
 
+    freeVariablesOfRule =
+        freeVariables
+        . getEqualityRule
+
+    ruleIsConcrete rule =
```
ttuegel: For now, this should just be `not $ Set.null $ freeVariablesOfRule rule`.

PR #1609, opened 2020-02-28
Add a "release" flag

ttuegel:
The "release" flag is enabled by default. When disabled, the build is about 25%
faster, but the build products are 5-10% slower. For developers, this is a good
trade-off.

After this is merged, I will update the kframework/k release job to force the `release` flag. Then I will make another pull request to disable it by default in the backend, so that builds are faster for our developers.

---

###### Reviewer checklist


dwightguth:
You might consider making it configurable via a variable in the k frontend, with release being the default.

Issue #1605, opened 2020-02-27
 KItem "in" Set does not consider symbolic patterns


yzhang90:
The haskell backend failed to evaluate
```
?ACCT in SetItem(?ACCT)
```
to true.

PR #1616, opened 2020-03-02
Evaluate in hook for symbolic sets

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1605

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Builtin/Set.hs
+++ b/kore/src/Kore/Builtin/Set.hs
@@ -31,10 +31,6 @@ module Kore.Builtin.Set
     , unifyEquals
     ) where
 
--- import Kore.Unparser
---     ( unparseToString
---     )
-
 import Prelude.Kore
 
 import Control.Error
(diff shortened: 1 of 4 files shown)


kore/src/Kore/Builtin/Set.hs
```diff
@@ -279,12 +283,35 @@ evalIn =
...
+                    ifTrueAddDefinednessCondition result _set
+                bothConcrete = do
+                    _elem <- hoistMaybe $ Builtin.toKey _elem
+                    _set <- expectConcreteBuiltinSet Set.inKey _set
+                    (Builtin.appliedFunction . asExpandedBoolPattern)
+                        (Map.member _elem _set)
+            setSymbolic <|> bothSymbolic <|> bothConcrete
       where
         asExpandedBoolPattern = Bool.asPattern resultSort
+        ifTrueAddDefinednessCondition result setTerm =
```
ttuegel: Could we name this function something else, perhaps `returnIfTrueAndDefined`? To me, the name sounds like this function always returns a result, but only adds the definedness condition if the result is `True`, which would be wrong for several reasons. I had to read it several times before I understood. :laughing:

kore/src/Kore/Builtin/Set.hs
```diff
@@ -279,12 +283,35 @@ evalIn =
...
+            let condition =
+                    Conditional.fromPredicate
+                    $ makeCeilPredicate resultSort setTerm
+                trueWithCondition =
+                    Builtin.appliedFunction
+                    $ Pattern.andCondition
+                        (asExpandedBoolPattern result)
+                        condition
+             in if result
```
ttuegel: Could we lift this `if` to a guard on the function definition?

kore/test/Test/Kore/Builtin/Set.hs
```diff
@@ -232,17 +236,56 @@ test_inElement =
...
-                predicate = mkEquals_ patIn patTrue
-            (===) (Test.Bool.asPattern True) =<< evaluateT patIn
-            (===) Pattern.top                =<< evaluateT predicate
+                conditionTerm = mkAnd patTrue (mkCeil_ patElement)
+            actual <- evaluateT patIn
+            expected <- evaluateT conditionTerm
+            actual === expected
         )
+  where
+    -- genKeys :: Gen [TermLike Variable]
```
virgil-serbanuta: Do you still need this and the comment a few lines below?

ana-pantilie: These changes are outdated

Issue #1575, opened 2020-02-20
Better equality simplification for `or`


virgil-serbanuta:
Right now we simplify `a == b or c` to something like
```
a is bottom and b is bottom and c is bottom
or (a is not bottom
   and (b is not bottom or c is not bottom)
   and (b is not bottom implies a == b)
   and (c is not bottom implies a == c)
   )
```

That's usually fine since representing that `implies` as an `or` would lead to a combinatorial explosion. However, when we have only two terms, i.e. `a == b`, we use a simpler formula:
```
(a is bottom and b is bottom)
or (a is not bottom and b is not bottom and a == b)
```
So it's preferable to use the latter formula when possible.

In many cases, one of `a == b` and `a == c` will be bottom in a trivial way. But, if, say, `a == b` is bottom, then `a == b or c` is equivalent to `a == c`.

When checking `a == b or c or ...` we should first evaluate all the equalities, i.e. `a == b`, `a == c`, ... and we should remove the `or` terms where the equality is bottom. Only after this we should try to decide which simplification formula to use, and we should use the filtered equality results in that formula.

ttuegel:
I re-set the `appetite` label again because I think I was being too aggressive the other day.

virgil-serbanuta:
Summary: this was probably be fixed as well as possible by https://github.com/kframework/kore/pull/1604

Longer story:
First, above I made a mistake. When `a == b` is `⊥` then `a == (b ∨ c)` is the same as `¬⌈b⌉ ∧ (a == c)`.
Second, after the PR above, normal simplification should be enough, with the following simplification chain:

```
a == (b ∨ c)
-- Equals simplification
(¬⌈a⌉ ∧ ¬⌈b⌉ ∧ ¬⌈c⌉)
    ∨ (⌈a⌉
        ∧ (⌈b⌉ ∨ ⌈c⌉)
        ∧ (⌈b⌉ → a == b)
        ∧ (⌈c⌉ → a == c)
        )
-- a == b is ⊥
(¬⌈a⌉ ∧ ¬⌈b⌉ ∧ ¬⌈c⌉)
    ∨ (⌈a⌉
        ∧ (⌈b⌉ ∨ ⌈c⌉)
        ∧ (⌈b⌉ → ⊥)
        ∧ (⌈c⌉ → a == c)
        )
-- Implies simplification
(¬⌈a⌉ ∧ ¬⌈b⌉ ∧ ¬⌈c⌉)
    ∨ (⌈a⌉
[...]

PR #1618, opened 2020-03-02
Test equals simplification

virgil-serbanuta:
Fixes https://github.com/kframework/kore/issues/1575

---

###### Reviewer checklist


kore/test/Test/Kore/Step/Simplification/Equals.hs
```diff
@@ -225,7 +225,72 @@ test_equalsSimplification_Or_Pattern =
...
+                        { term = Mock.functionalConstr11 Mock.cg
+                        , predicate = makeTruePredicate_
+                        , substitution = mempty
+                        }
+                    , Conditional
+                        { term = Mock.functionalConstr10 Mock.ch
+                        , predicate = makeTruePredicate_
+                        , substitution = mempty
+                        }
+                    ]
```
ttuegel: It would appease hlint if we wrote,
```.hs
let original =
         Equals
            { equalsOperandSort = testSort
            , equalsResultSort = testSort
            , equalsFirst = first
            , equalsSecond = second
            }
    swapEquals equals@Equals { equalsFirst = equalsSecond', equalsSecond = equalsFirst' } =
        equals { equalsFirst = equalsFirst', equalsSecond = equalsSecond' }
actual1 <- evaluateOr original
...
actual2 <- evaluateOr (swapEquals original)
...
```

I think this would actually be clearer because it makes it explicit in the test that we are testing that we get the same outcome with the arguments swapped.

virgil-serbanuta: I did more than that, hopefully it's reasonable.

PR #1634, opened 2020-03-05
Use ActualEntry instead of SomeEntry in the logger

ana-pantilie:
Part of https://github.com/kframework/kore/issues/1491

- Add a data type data ActualEntry = ActualEntry { entryContext :: Seq SomeEntry, actualEntry :: SomeEntry }.
- Define instances From ActualEntry SomeEntry and From SomeEntry ActualEntry which forget the entryContext and add an empty entryContext respectively.
- LoggerT should carry a LogAction IO ActualEntry instead of LogAction IO SomeEntry. Use Colog.contramap (from @SomeEntry) and Colog.contramap (from @ActualEntry) as necessary to fix type errors.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Log.hs
+++ b/kore/src/Log.hs
@@ -33,6 +33,7 @@ import Prelude.Kore
 import Colog
     ( LogAction (..)
     , Severity (..)
+    , cmap
     , (<&)
     )
 import Control.Monad.Catch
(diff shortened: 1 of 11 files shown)


kore/src/Log.hs
```diff
@@ -228,8 +233,11 @@ instance MonadTrans LoggerT where
...
     :: Entry entry
-    => LogAction m SomeEntry
+    => LogAction m ActualEntry
     -> entry
     -> m ()
 logWith logger entry =
-    logger Colog.<& toEntry entry
+    fromLogAction logger Colog.<& toEntry entry
+
+fromLogAction :: From b a => LogAction m a -> LogAction m b
```
ttuegel: Please give an explicit quantifier,

```suggestion
fromLogAction :: forall a b. From b a => LogAction m a -> LogAction m b
```

and use type application on at least one of the quantified type variables everywhere that `fromLogAction` is used.

kore/src/Log/Entry.hs
```diff
@@ -72,5 +77,18 @@ entryTypeText :: SomeEntry -> Text
 entryTypeText (SomeEntry entry) =
     Text.pack . show . Reflection.typeOf $ entry
 
+data ActualEntry =
+    ActualEntry
+        { actualEntry :: SomeEntry
```
ttuegel: Please make these fields strict.

PR #1635, opened 2020-03-05
Define inequality of Bytes

### LINKED ISSUE #1623: Define inequality of Bytes
Equality of Bytes is defined, but not inequality, so that this simplifies to `\top`:

```
\equals(\dv{Bytes}("2387e975"), \dv{Bytes}("238e975"))
```

but this does _not_ simplify to `\bottom`:

```
\equals(\dv{Bytes}("2387e975"), \dv{Bytes}("f02ce809"))
```

---

- [ ] Add a unification case (for `\and` and `\equals`) to `Kore.Step.Simplification.AndTerms` that is `\bottom` when we have two different concrete Bytes values.

andreiburdusa:
---
Fixes #1623 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -30,6 +30,7 @@ import Control.Error
 import qualified Control.Error as Error
 import qualified Control.Monad.Trans as Monad.Trans
 import qualified Data.Foldable as Foldable
+import qualified Data.Functor.Foldable as Recursive
 import qualified Data.Text as Text
 import qualified Data.Text.Prettyprint.Doc as Pretty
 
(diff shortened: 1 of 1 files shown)


kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -682,3 +687,16 @@ functionAnd first second
...
+bytesDifferent
+    :: InternalVariable variable
+    => MonadUnify unifier
+    => TermLike variable
+    -> TermLike variable
+    -> MaybeT unifier (Pattern variable)
+bytesDifferent
+    first@(Recursive.project -> _ :< InternalBytesF (Const bytesFirst))
+    second@(Recursive.project -> _ :< InternalBytesF (Const bytesSecond))
+  | isConcrete first && isConcrete second && bytesFirst /= bytesSecond
```
ttuegel: An `InternalBytes` is always concrete:

```suggestion
  | bytesFirst /= bytesSecond
```

PR #1640, opened 2020-03-06
Use ActualEntry in the logger

ana-pantilie:
Part of https://github.com/kframework/kore/issues/1491

---

###### Reviewer checklist


kore/src/Kore/Log.hs
```diff
@@ -292,10 +297,16 @@ swappableLogger mvar =
...
-defaultLogPretty (SomeEntry entry) =
+defaultLogPretty :: ActualEntry -> Pretty.Doc ann
+defaultLogPretty ActualEntry { actualEntry = SomeEntry entry, entryContext } =
     header Pretty.<+> longDoc entry
   where
     severity = prettySeverity (entrySeverity entry)
     type' = Pretty.pretty (lookupTextFromTypeWithError $ toSomeEntryType entry)
-    header = severity Pretty.<+> Pretty.parens type' <> Pretty.colon
+    header =
+        Pretty.brackets (Pretty.hsep prettyContext)
```
ttuegel: We should use `Pretty.vsep` here and put the context before the brackets. My intention was to have something that looks like a "traditional" compiler error message:

```
while applying rule ...:
in side condition ...:
while doing something else ...:
Error: ...
    ...
```

kore/src/Kore/Log/DebugAxiomEvaluation.hs
```diff
@@ -123,6 +123,19 @@ data AxiomEvaluationState
 
 instance Entry DebugAxiomEvaluation where
     entrySeverity _ = Debug
+    shortDoc DebugAxiomEvaluation { state } =
+        Just $ Pretty.hcat ["DebugAxiomEvaluation_", showState]
+      where
+        showState =
+            case state of
```
ttuegel: This is user-facing text, so it should give more information than this. The `shortDoc` should be about one line long.

kore/src/Kore/Step/Function/Evaluator.hs
```diff
@@ -249,41 +249,42 @@ maybeEvaluatePattern
                 Profile.axiomEvaluation identifier
                 $ evaluator termLike sideConditions
             flattened <- case result of
-                AttemptedAxiom.NotApplicable -> do
+                AttemptedAxiom.NotApplicable ->
                     DebugAxiomEvaluation.notEvaluated
```
ttuegel: Why does this need to be wrapped around return? Nothing will be logged during this time, so this can just be a log entry (instead of a context).

kore/src/SMT.hs
```diff
@@ -229,7 +231,15 @@ instance MonadLog NoSMT where
         logAction <- ask
         let entryLogger = Log.fromLogAction @Log.ActualEntry logAction
         Trans.lift $ entryLogger Colog.<& Log.toEntry entry
-    logWhile _ = id
+    logWhile entry2 action =
```
ttuegel: This should call `logEntry` before wrapping `action`, right?

kore/src/Log.hs
```diff
@@ -225,7 +226,16 @@ instance Monad m => MonadLog (LoggerT m) where
...
+
+    logWhile entry2 action = do
+        logEntry entry2
+        LoggerT . local modifyContext $ getLoggerT action
+      where
+        modifyContext = Colog.cmap $ \entry1 ->
+            entry1
+                { entryContext =
+                    entryContext entry1 <> Seq.singleton (toEntry entry2)
+                }
```
ttuegel: You don't have to make this change, but it's the ideal use-case for lenses:

```suggestion
        modifyContext = Colog.cmap $ Lens.(<>~) (field @"entryContext") (Seq.singleton $ toEntry entry2)
```

ana-pantilie: Thanks for the suggestion! For some reason, I wasn't able to get this to compile, but I'll keep it in mind next time I need to implement this functionality.

kore/src/Kore/Log/DebugAxiomEvaluation.hs
ana-pantilie: I modified the instance for a few `State` cases for which I thought it would make sense to use `logWhile`, but please tell me if you had anything else in mind.

kore/src/Kore/Log.hs
ana-pantilie: I changed this to look something like the following for entries with `shortDoc`: 
```
kore-exec: [2020-03-12 12:29:53.188880665]
  While starting axiom evaluation of Lbl'Unds-GT-Eqls'Int'Unds' :
  While starting axiom evaluation of Lbl'Unds'orBool'UndsUnds'BOOL'Unds'Bool'Unds'Bool'Unds'Bool :
    Debug (DebugAxiomEvaluation): Ending with N/A: Lbl'Unds-GT-Eqls'Int'Unds'
```
Ordinary entries still look the same:
```
kore-exec: [2020-03-12 12:29:53.19308644] Debug (DebugAxiomEvaluation): Starting:
Lbl'Unds'orBool'UndsUnds'BOOL'Unds'Bool'Unds'Bool'Unds'Bool
```
Is this what you had in mind?

kore/src/Log.hs
ttuegel: I probably made an embarrassing typo somewhere, I'm sorry! :sweat_smile:

PR #1652, opened 2020-03-10
Add options to kore-repl as in kore-exec

### LINKED ISSUE #1626: kore-repl must take the same logging options as kore-exec
Actually, all the options should probably be the same, except those not related to the `--prove` mode.

See also: https://github.com/kframework/kore/issues/1625#issuecomment-594737556

andreiburdusa:
---
Fixes #1626 
###### Reviewer checklist


kore/app/repl/Main.hs
```diff
@@ -192,6 +205,7 @@ mainWithOptions
         , replScript
         , replMode
         , outputFile
+        , koreLogOptions
         }
   = do
     mLogger <- newMVar emptyLogger
```
ttuegel: We need the `LogAction` created from `koreLogOptions` to be put in the `MVar` here, instead of `emptyLogger`. Then we should probably rename `emptySwappableLogger` to `initialLogger`.

kore/app/repl/Main.hs
```diff
@@ -215,21 +226,32 @@ mainWithOptions
...
-                            specDefIndexedModule
-                            Nothing
-                            mLogger
-                            replScript
-                            replMode
-                            outputFile
+                $ withLogger koreLogOptions
+                    $ logActionToIO
+                        smtConfig
+                        (swappableLogger mLogger)
```
ttuegel:
```suggestion
                        initialLogger
```

kore/app/repl/Main.hs
```diff
@@ -215,21 +226,32 @@ mainWithOptions
...
-                            (declareSMTLemmas indexedModule)
-                        proveWithRepl
-                            indexedModule
-                            specDefIndexedModule
-                            Nothing
-                            mLogger
-                            replScript
-                            replMode
-                            outputFile
+                $ withLogger koreLogOptions
```
ttuegel: Where did `SMT.runSMT` go?

andreiburdusa: I used it inside `logActionToIO` at line 253

ttuegel: Could we rename that function to `runSMT`, or something like that? It's not clear from the name that it has anything to do with SMT.

Issue #1650, opened 2020-03-10
Keep Bottom nodes in execution graph


ttuegel:
The execution graph must keep Bottom nodes so that their parents are not considered unproven leaf nodes.

Fixes #1637

PR #1668, opened 2020-03-12
Refactor Kore.Unification.SubstitutionSimplifier

ttuegel:
The `SubstitutionSimplifier` will no longer throw errors at unresolved cycles. In fact, the type `SubstitutionError` is entirely removed. Dead code was also removed from the `Kore.Unification.SubstitutionNormalization` module, and the `Kore.Unification.UnifierImpl` module was found to be entirely dead.

---

###### Reviewer checklist


ttuegel:
> Why do we want to not throw errors at unresolved cycles anymore, and how does this impact the way we do simplification/unification and what is the relation to #1650?

This is not directly related to #1650, but #1670 is, and that pull request makes it practically impossible to throw the cyclic substitution error anymore. I thought it would be better to remove the effectively-dead code in a separate pull request to keep the Git history cleaner and so we could directly link any test failures to one set of changes or the other.

The reason it is safe to remove the error is basically: it has outlived its purpose. Originally, the error was to prevent the backend from looping forever while trying to apply a cyclic substitution. That's no longer a danger, because the `SubstitutionSimplifier` knows how to handle cyclic substitutions. See also: https://runtimeverification.slack.com/archives/CC360GUTG/p1583954104118100

PR #1670, opened 2020-03-12
Refactor UnificationProcedure

ttuegel:
~Please review #1668 first.~

This pull request removes unused `MonadUnify` constraints from code that applies `\rewrites` rules. The unifier introduces a redundant means of branching during rule application. Applying a rule should always return a single `Results` value, not a disjunction of `Results`es. This will eliminate some `error` cases in #1659.

I also added some proper log entries to reduce duplication between modules that apply `\rewrites` rules.

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/RewriteStep.hs
+++ b/kore/src/Kore/Step/RewriteStep.hs
@@ -243,7 +243,7 @@ applyRulesWithFinalizer
     .  InternalVariable variable
     => MonadUnify unifier
     => Finalizer unifier variable
-    -> UnificationProcedure
+    -> UnificationProcedure unifier
     -> [RulePattern variable]
     -- ^ Rewrite rules
     -> Pattern (Target variable)
@@ -266,7 +266,7 @@ applyRulesParallel
     :: forall unifier variable
     .  InternalVariable variable
     => MonadUnify unifier
-    => UnificationProcedure
+    => UnificationProcedure unifier
     -> [RulePattern variable]
     -- ^ Rewrite rules
     -> Pattern (Target variable)
@@ -283,7 +283,7 @@ applyRewriteRulesParallel
     :: forall unifier variable
     .  InternalVariable variable
     => MonadUnify unifier
-    => UnificationProcedure
+    => UnificationProcedure unifier
     -> [RewriteRule variable]
     -- ^ Rewrite rules
     -> Pattern variable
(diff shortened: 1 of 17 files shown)


kore/src/Kore/Step/RewriteStep.hs
```diff
@@ -156,8 +158,8 @@ finalizeAppliedRule renamedRule appliedConditions =
 
 finalizeRule
     ::  ( InternalVariable variable
-        , Log.WithLog Log.LogMessage unifier
-        , MonadUnify unifier
+        , MonadLog unifier
```
ana-pantilie: Not necessary, but would it be clearer if the variable was renamed to `simplifier` since the constraint is now `MonadSimplify`? Here and in the rest of the changes in this file.

kore/src/Kore/Step/Step.hs
```diff
@@ -172,18 +159,18 @@ class UnifyingRule rule where
 -- |Unifies/matches a list a rules against a configuration. See 'unifyRule'.
 unifyRules
     :: InternalVariable variable
-    => MonadUnify unifier
+    => MonadSimplify unifier
     => UnifyingRule rule
-    => UnificationProcedure
+    => UnificationProcedure unifier
```
ana-pantilie: Same as above.

kore/src/Kore/Step/Step.hs
```diff
@@ -394,23 +379,22 @@ The rule is considered to apply if the result is not @\\bottom@.
...
-    => MonadUnify unifier
+    => MonadSimplify unifier
     => SideCondition variable
     -- ^ Top-level conditions
     -> Maybe (Condition variable)
     -- ^ Initial conditions
     -> Condition variable
     -- ^ Unification conditions
-    -> unifier (OrCondition variable)
+    -> BranchT unifier (OrCondition variable)
```
ana-pantilie: Same as above.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -951,7 +950,7 @@ deriveResults goal results = do
             Result.mapConfigs
                 (flip (configurationDestinationToRule goal) destination)
                 (flip (configurationDestinationToRule goal) destination)
-                (Result.mergeResults results)
+                results
     results' <-
         traverseConfigs (mapRules onePathResults)
```
ana-pantilie: I know this isn't your change, but I just noticed this is still called `onePathResults`.

ana-pantilie:
Actually I just remembered, shouldn't new log entry types be added to `Kore.Log.Registry`?

ttuegel:
> Actually I just remembered, shouldn't new log entry types be added to `Kore.Log.Registry`?

Thanks, I missed that! Maybe we should think about making `SomeEntry` a sum type, so that it is harder to make this mistake.

PR #1674, opened 2020-03-16
Enabled concrete to take a list of arguments

traiansf:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Attributes.hs
+++ b/kore/src/Kore/Attribute/Attributes.hs
@@ -12,6 +12,7 @@ module Kore.Attribute.Attributes
     , attributePattern_
     , attributeString
     , attributeInteger
+    , attributeVariable
     ) where
 
 import Prelude.Kore
@@ -35,6 +36,9 @@ import qualified Kore.Attribute.Null as Attribute
 import Kore.Debug
 import Kore.Syntax
 import Kore.Unparser
+import Kore.Variables.UnifiedVariable
+    ( UnifiedVariable
+    )
 import qualified Pretty
 import qualified SQL
 
@@ -68,6 +72,9 @@ attributeString literal =
 
 attributeInteger :: Integer -> AttributePattern
 attributeInteger = attributeString . Text.pack . show
+
+attributeVariable :: UnifiedVariable Variable -> AttributePattern
+attributeVariable v = (asAttributePattern . VariableF. Const) v
 
 {-|'Attributes' corresponds to the @attributes@ Kore syntactic declaration.
 It is parameterized by the types of Patterns, @pat@.
--- a/kore/src/Kore/Attribute/Attributes.hs
+++ b/kore/src/Kore/Attribute/Attributes.hs
@@ -74,7 +74,7 @@ attributeInteger :: Integer -> AttributePattern
 attributeInteger = attributeString . Text.pack . show
 
 attributeVariable :: UnifiedVariable Variable -> AttributePattern
-attributeVariable v = (asAttributePattern . VariableF. Const) v
+attributeVariable v = (asAttributePattern . VariableF . Const) v
 
 {-|'Attributes' corresponds to the @attributes@ Kore syntactic declaration.
 It is parameterized by the types of Patterns, @pat@.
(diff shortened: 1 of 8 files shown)


traiansf:
This only handles the syntactic part, i.e., parsing and combining multiple concrete and symbolic attributes. 
Still to do (in another PR, I think): use the concrete and symbolic information in accepting (or not)  results of simplification axioms.

kore/src/Kore/Attribute/Axiom.hs
```diff
@@ -200,12 +205,12 @@ instance From symbol SymbolOrAlias => From (Axiom symbol variable) Attributes wh
             , from . owise
             ]
 
-instance SQL.Column (Axiom SymbolOrAlias variable) where
+instance SQL.Column (Axiom SymbolOrAlias Variable) where
```
ttuegel: Why does this change? (I think it's fine, I'm just not sure why.)

kore/src/Kore/Attribute/Attributes.hs
```diff
@@ -69,6 +73,9 @@ attributeString literal =
 attributeInteger :: Integer -> AttributePattern
 attributeInteger = attributeString . Text.pack . show
 
+attributeVariable :: UnifiedVariable Variable -> AttributePattern
+attributeVariable v = (asAttributePattern . VariableF. Const) v
```
ttuegel:
```suggestion
attributeVariable v = (asAttributePattern . VariableF . Const) v
```

kore/src/Kore/Attribute/Axiom/Concrete.hs
```diff
@@ -63,27 +74,42 @@ concreteSymbol =
...
+        -> [AttributePattern]
+        -> Concrete Variable
+        -> Parser (Concrete Variable)
+    parseApplication params args (Concrete concreteVars) = do
+        Parser.getZeroParams params
+        vars <- mapM getVariable args
+        mapM_ checkFree vars
+        let newVars =
+                if null vars then freeVariables else foldMap freeVariable vars
+        mapM_ (checkDups concreteVars) (getFreeVariables newVars)
```
ttuegel: Could you add a comment explaining why it is necessary to check for duplicates even if `vars` is empty? I had to read this a few times before it was clear to me.

kore/src/Kore/Attribute/Axiom/Concrete.hs
```diff
@@ -63,27 +74,42 @@ concreteSymbol =
...
+        return (Concrete $ concreteVars <> newVars)
+    checkFree :: UnifiedVariable Variable -> Parser ()
+    checkFree variable =
+        Monad.unless (isFreeVariable variable freeVariables)
+        $ Kore.Error.koreFail
+            ("expected free variable, found " ++ show variable)
+    checkDups :: FreeVariables Variable -> UnifiedVariable Variable -> Parser ()
+    checkDups freeVars var = Monad.when (isFreeVariable var freeVars)
+        $ Kore.Error.koreFail
+            ("duplicate concrete variable declaration for " ++ show var)
```
ttuegel: I don't think "declaration" is the right word, but I don't have a better suggestion right now.

kore/src/Kore/Attribute/Axiom/Concrete.hs
```diff
@@ -63,27 +74,42 @@ concreteSymbol =
...
+                if null vars then freeVariables else foldMap freeVariable vars
+        mapM_ (checkDups concreteVars) (getFreeVariables newVars)
+        return (Concrete $ concreteVars <> newVars)
+    checkFree :: UnifiedVariable Variable -> Parser ()
+    checkFree variable =
+        Monad.unless (isFreeVariable variable freeVariables)
+        $ Kore.Error.koreFail
+            ("expected free variable, found " ++ show variable)
+    checkDups :: FreeVariables Variable -> UnifiedVariable Variable -> Parser ()
+    checkDups freeVars var = Monad.when (isFreeVariable var freeVars)
```
ttuegel: There are two things about `checkDups` that I think we could do better:

1. It checks for `concrete(X), concrete(X)`, but it doesn't check for `concrete(X, X)`.
1. It doesn't report all the duplicated variables, for example in `concrete(X, Y), concrete(X, Y)` it would report only `X`.

kore/src/Kore/Attribute/Axiom.hs
traiansf: it's because the From instance for Concrete needs a `Variable`, because it needs `concreteAttribute`, which needs `UnifiedVariable Variable`

Issue #1183, opened 2019-10-30
Matching to iterate over Map and Set


ttuegel:
We need to support matching for functions to iterate over the keys of a Map or Set, as seen [here](https://github.com/kframework/evm-semantics/pull/532/files#diff-494b4c66d4128fe395a8c9607fab3629R1180-R1181).

~A sufficient and simple implementation is given [here](https://github.com/kframework/kore/issues/1045#issuecomment-539253453).~

The only thing we need to support right now is matching a single symbolic key with an opaque Map or Set term, something of the form `K |-> V M:Map`, against an arbitrary Map or Set. If `M` is also an opaque term in the Map or Set being matched, we should remove it from both and match the remainder. If `K` is a key in both, we should match the corresponding values and then match `M` against the remainder. Otherwise, we can pick an arbitrary key to match `K`, match the corresponding value to `V`, and match the remainder of the Map or Set to `M`.

See also: #1045

denis-bogdanas:
@tomtomjhj `Set2List` and iterating over set still doesn't work in functions if Set elements are symbolic. Is this intended?

ttuegel:
@denis-bogdanas No, can you provide a spec?

denis-bogdanas:
turns out only `Set2List` doesn't work in functions, the rest works:
```
module ISSUE
  imports COLLECTIONS
  imports STRING

  configuration <k> $PGM:KItem </k>
                <theSet> .Set </theSet>

    syntax KItem ::= #checkArgWithList ( String )
                   | "#included"

    rule <k> #checkArgWithList(ARG) => #included ...</k>
         <theSet> S </theSet>
      requires ARG #in Set2List(S)

    syntax Bool ::= KItem "#in" List                                                       [function, functional]
    rule X #in ListItem(A) L:List => X ==K A orBool X #in L                                [simplification]
    rule X #in .List => false                                                              [simplification]
endmodule
```

```
module INCLUDEDLIST-SPEC
  imports ISSUE

  rule <k> #checkArgWithList("foo") => #included </k>
       <theSet> SetItem("foo") SetItem(A:KItem) </theSet>
    requires A =/=K "foo"
endmodule
```
Result:
```
  #Not ( {
    "foo" #in Set2List ( SetItem ( "foo" )
    SetItem ( A ) )
  #Equals
    true
  } )
#And
  #Not ( {
    A
  #Equals
    "foo"
  } )
#And
  <generatedTop>
    <k>
[...]

PR #1675, opened 2020-03-16
Matching to iterate over Map and Set

ana-pantilie:
Needs https://github.com/kframework/kore/pull/1694

Fixes https://github.com/kframework/kore/issues/1183

---

###### Reviewer checklist


kore/test/Test/Kore/Step/Axiom/Matcher.hs
```diff
@@ -971,6 +963,23 @@ test_matching_Map =
         [ (ElemVar yInt, mkInt 2)
         , (ElemVar mMap, mkMap [(mkElemVar xInt, mkInt 0)] [mkElemVar nMap])
         ]
+    -- Map iteration matching
```
ttuegel: These are the correct tests.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -687,6 +696,39 @@ matchNormalizedAc pushValue wrapTermLike normalized1 normalized2
         _ -> empty
     lift $ Foldable.traverse_ pushValue concrete12
     lift $ Foldable.traverse_ pushValue abstractMerge
+  | otherwise = do
```
ttuegel: I'm not sure this should be an `otherwise` guard. The case we want to cover here is: The first Map is a single element (the key and value are both variables) and one opaque map (a variable). The second Map may be anything.

If the single key in the first Map is equal to any key in the second Map, then the corresponding values must match, and `frame1` must match the remainder of the second Map.

If the single key is not in the second Map, then match it with any (concrete or symbolic) key in the second map, and match the corresponding values. `frame1` also matches the remainder of the second Map.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -687,6 +696,39 @@ matchNormalizedAc pushValue wrapTermLike normalized1 normalized2
...
+      Monad.guard (null concrete1)
+      case opaque1 of
+          [frame1]
+            | [frame2] <- opaque2 -> do
```
ttuegel: I don't know what this case is for, but I don't think it should exist at all.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -687,6 +696,39 @@ matchNormalizedAc pushValue wrapTermLike normalized1 normalized2
...
+            | [element1] <- abstract1
+            , (key1, value1) <- Builtin.unwrapElement element1
+            , Just value2 <- Builtin.lookupSymbolicKeyOfAc key1 normalized2 -> do
+                lift $ pushValue (Pair value1 value2)
+                let ac2 =
+                        wrapTermLike
+                        $ Builtin.removeSymbolicKeyOfAc key1 normalized2
+                push (Pair frame1 ac2)
+            | [element1] <- abstract1
+            , (element2 : _) <- abstract2 -> do
```
ttuegel: I think this is only drawing `element2` from the symbolic elements of the second Map, right? It should also cover the concrete elements.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -659,7 +666,10 @@ matchNormalizedAc
         , MatchingVariable variable
         , Monad unifier
         )
-    =>  ( Pair (Builtin.Value normalized (TermLike variable))
+    =>  ( Pair (Builtin.Element normalized (TermLike variable))
+        -> MatcherT variable unifier ()
+        )
+    ->  ( Pair (Builtin.Value normalized (TermLike variable))
```
ttuegel: Let's add a type synonym for these,

```.hs
type Push variable unifier a = Pair a -> MatcherT variable unifier ()
```

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -668,31 +678,63 @@ matchNormalizedAc
...
-    lift $ Foldable.traverse_ pushValue concrete12
-    lift $ Foldable.traverse_ pushValue abstractMerge
+matchNormalizedAc pushElement pushValue wrapTermLike normalized1 normalized2
+    | [element1] <- abstract1
+    , [frame1] <- opaque1
+    , null concrete1 = do
+        let (key1, value1) = Builtin.unwrapElement element1
+        case Builtin.lookupSymbolicKeyOfAc key1 normalized2 of
+            Just value2 -> do
+                lift $ pushValue (Pair value1 value2)
```
ttuegel: It would be better to move `lift` to the start of the block, that is `lift $ do ...`

test/matching-inj/run.k
```diff
@@ -11,7 +12,10 @@ module RUN
   configuration <k> $PGM:Pgm </k>
 
   syntax AB ::= #lookup(AB, Map) [function]
-  rule #lookup(KEY:AB, KEY |-> VAL _) => VAL
+
+  rule #lookup(KEY1, KEY2 |-> VAL M) => #lookup(KEY1, M)
```
ttuegel: This function should not require iteration over the "tail" of the Map. The second rule below should be equivalent.

ana-pantilie: Hmm, so I tried to make sure it first tries to match the "normal" way, but if the "map iteration" case existed it would always try that case instead! I was seriously confused until I saw that during this test it tries to enter the map matching code twice, and it only succeeds the second time. This means I'd somehow need to make sure it tries to match "normally" multiple times before trying the "map iteration" code. Is this doable with the current matching algorithm? And I'm not even sure it makes sense for other cases.

ttuegel: I thought about this a bit, and I think what we should do is sort the constraints when matching. I will make another issue for that, and it can be handled in a separate pull request that we merge before this one.

ana-pantilie: Thanks!

PR #1682, opened 2020-03-18
Print Info when a strategy step starts

### LINKED ISSUE #1581: LOG: print Info when a strategy step starts
In the log, I can only see `DebugProofState` when a strategy step finishes. However, there is no log entry gets printed when a strategy step starts.

andreiburdusa:
---
Fixes #1581 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Log/InfoReachability.hs
+++ b/kore/src/Kore/Log/InfoReachability.hs
@@ -7,6 +7,8 @@ module Kore.Log.InfoReachability
     ( InfoReachability (..)
     , whileSimplify
     , whileRemoveDestination
+    , whileDeriveSeq
+    , whileDerivePar
     , prettyInfoReachabilityGoal
     , prettyInfoReachabilityGoalAndRules
     ) where
@@ -135,3 +137,21 @@ whileRemoveDestination
     -> log a
     -> log a
 whileRemoveDestination goal = logWhile (InfoRemoveDestination goal)
+
+whileDeriveSeq
+    :: MonadLog log
+    => Entry (InfoReachability goal)
+    => [Rule goal]
+    -> goal
+    -> log a
+    -> log a
+whileDeriveSeq rules goal = logWhile (InfoDeriveSeq rules goal)
+
+whileDerivePar
+    :: MonadLog log
+    => Entry (InfoReachability goal)
+    => [Rule goal]
+    -> goal
+    -> log a
+    -> log a
+whileDerivePar rules goal = logWhile (InfoDerivePar rules goal)
--- a/kore/src/Kore/Log/InfoReachability.hs
+++ b/kore/src/Kore/Log/InfoReachability.hs
@@ -1,0 +1,137 @@
+{- |
+Copyright   : (c) Runtime Verification, 2020
+License     : NCSA
+
+-}
+module Kore.Log.InfoReachability
+    ( InfoReachability (..)
+    , whileSimplify
+    , whileRemoveDestination
+    , prettyInfoReachabilityGoal
+    , prettyInfoReachabilityGoalAndRules
+    ) where
+
+-- TODO: Do not export prettyInfoReachabilityGoal.
+-- TODO: Do not export prettyInfoReachabilityGoalAndRules.
+
+import Prelude.Kore
+
+import Kore.Internal.Variable
+    ( Variable
+    )
+import Kore.Strategies.Rule
+import Log
+import qualified Pretty
+
+data InfoReachability goal
+    = InfoSimplify !goal
+    | InfoRemoveDestination !goal
+    | InfoDeriveSeq ![Rule goal] !goal
+    | InfoDerivePar ![Rule goal] !goal
+
+prettyInfoReachabilityGoal
+    :: Pretty.Pretty goal
+    => Pretty.Doc ann
+    -> goal
+    -> Pretty.Doc ann
+prettyInfoReachabilityGoal transition goal =
+    Pretty.vsep
+        [ "transition:"
+        , transition
+        , "goal:"
+        , Pretty.pretty goal
+        ]
+
+prettyInfoReachabilityGoalAndRules
+    :: Pretty.Pretty goal
+    => Pretty.Pretty rule
+    => Pretty.Doc ann
+    -> goal
+    -> [Rule goal]
+    -> (Rule goal -> rule)
+    -> Pretty.Doc ann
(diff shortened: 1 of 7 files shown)


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -622,6 +623,47 @@ data TransitionRuleTemplate monad goal =
         -> Strategy.TransitionT (Rule goal) monad (ProofState goal goal)
     }
 
+data InfoReachability goal 
+    = InfoSimplify !goal
```
andreiburdusa: I called this `InfoSimplify` instead of `Simplify` because otherwise the uses of `ProofState.Simplify` were ambiguous.
(same for the other 3 constructors)

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -622,11 +626,117 @@ data TransitionRuleTemplate monad goal =
         -> Strategy.TransitionT (Rule goal) monad (ProofState goal goal)
     }
 
+data InfoReachability goal
```
ttuegel: This is a log entry, please move it to `Kore.Log.InfoReachability`.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -622,11 +626,117 @@ data TransitionRuleTemplate monad goal =
...
+            rules
+            (getRewriteRule . unRuleAllPath)
+    pretty (InfoDerivePar rules goal) =
+        prettyInfoReachabilityGoalAndRules
+            "DerivePar"
+            goal
+            rules
+            (getRewriteRule . unRuleAllPath)
+
+instance Entry (InfoReachability (ReachabilityRule Variable)) where
```
ttuegel: Please implement `shortDoc` for each constructor. It should be a short, one-line description for each. It doesn't need to include all the details. For example,

```.hs
shortDoc (InfoSimplify _) = "While simplifying the configuration"
```

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -622,11 +626,117 @@ data TransitionRuleTemplate monad goal =
...
+    entrySeverity _ = Info
+
+instance Entry (InfoReachability (OnePathRule Variable)) where
```
ttuegel: Are the instances for `OnePathRule` and `AllPathRule` actually used? I thought we would only need `ReachabilityRule`.

andreiburdusa: The `Goal` instance of `OnePathRule Variable` uses `transitionRuleTemplate`
which uses `whileSimplify`
which uses `logWhile :: Entry entry => entry -> m a -> m a`

andreiburdusa:
I tried to create two new modules `Kore.Strategies.TransitionRule` and `Kore.Log.InfoReachability` but then I wouldn't be able to solve circular dependencies:

```
data InfoReachability goal
    = InfoSimplify !goal
    | InfoRemoveDestination !goal
    | InfoDeriveSeq ![Rule goal] !goal
    | InfoDerivePar ![Rule goal] !goal
```
depends on the instances of
```
class Goal goal where
    data Rule goal
    ...
    transitionRule :: ...
    ...
```
that define `Rule` and `transitionRule` (at the same time)

But when defining `transitionRule` for `instance Goal (AllPathRule Variable)`, `transitionRuleTemplate` is used, which should use `whileSimplify`, defined as `whileSimplify goal = logWhile (**InfoSimplify** goal)`

The problem is that definitions for `data Rule` and `transitionRule` can't be given in different modules, so we get the following circular dependency:
`data ***InfoReachability*** goal` -> `data Rule goal` linked to `transitionRule` -> `transitionRuleTemplate` -> `whileSimplify` -> `InfoSimplify` which is a constructor of `data ***InfoReachability*** goal`

ttuegel:
I extracted `InfoReachability` from `Kore.Strategies.Goal`. There were different ways to do this, but I chose to extract `Rule` from the `Goal` class, making it a stand-alone datatype family. We also need to add `InfoReachability` to `Kore.Log.Registry`, but we have to remove the type parameter first. There are a few things left for you to do to make that happen:

- [ ] Remove the logging from `transitionRuleTemplate` and make a transition rule transformer `logTransitionRule :: (Prim goal -> ProofState goal goal -> Strategy.TransitionT (Rule goal) m (ProofState goal goal)) -> (Prim goal -> ProofState goal goal -> Strategy.TransitionT (Rule goal) m (ProofState goal goal))` which wraps its argument with the logging functions.
- [ ] Remove the parameter from `InfoReachability` and specialize its type to `goal ~ ReachabilityRule Variable`.
- [ ] Finally add `InfoReachability` to `Kore.Log.Registry`.

kore/src/Kore/Log/InfoReachability.hs
```diff
@@ -0,0 +1,115 @@
...
+    , whileSimplify
+    , whileRemoveDestination
+    , whileDeriveSeq
+    , whileDerivePar
+    , prettyInfoReachabilityGoal
+    , prettyInfoReachabilityGoalAndRules
+    ) where
+
+-- TODO: Do not export prettyInfoReachabilityGoal.
+-- TODO: Do not export prettyInfoReachabilityGoalAndRules.
```
ttuegel: Please fix these TODOs.

kore/test/Test/Kore/Strategies/AllPath/AllPath.hs
```diff
@@ -427,7 +432,7 @@ newtype AllPathIdentity a = AllPathIdentity { unAllPathIdentity :: Identity a }
 
 instance MonadLog AllPathIdentity where
```
ttuegel: This instance should no longer be necessary.

andreiburdusa:
```
transitionRule
        :: (MonadCatch m, MonadSimplify m)
        => Prim goal
        -> ProofState goal goal
        -> Strategy.TransitionT (Rule goal) m (ProofState goal goal)
```
is used [here](https://github.com/kframework/kore/blob/0ca0bbf5e5a041dc61da252e1eb97c6ae1299523/kore/test/Test/Kore/Strategies/AllPath/AllPath.hs#L423), and `MonadSimplify` requires `MonadLog`. Does it mean it's still necessary?

ttuegel: Yes, it is.

PR #1685, opened 2020-03-18
implemented substitution checks for concrete and symbolic attributes

traiansf:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Axiom/Matcher.hs
+++ b/kore/src/Kore/Step/Axiom/Matcher.hs
@@ -98,6 +98,11 @@ import Pair
 
 -- * Matching
 
+type MatchResult variable =
+    ( Predicate variable
+    , Map.Map (UnifiedVariable variable) (TermLike variable)
+    )
+
 {- | Dispatch a single matching constraint.
 
 @matchOne@ is the heart of the matching algorithm. Each matcher is applied to
@@ -135,21 +140,11 @@ matchIncremental
     .  (MatchingVariable variable, MonadSimplify unifier)
     => TermLike variable
     -> TermLike variable
-    -> unifier
-        (Maybe
-            ( Predicate variable
-            , Map (UnifiedVariable variable) (TermLike variable)
-            )
-        )
+    -> unifier (Maybe (MatchResult variable))
 matchIncremental termLike1 termLike2 =
     Monad.State.evalStateT matcher initial
   where
-    matcher :: MatcherT variable unifier
-        (Maybe
-            ( Predicate variable
-            , Map (UnifiedVariable variable) (TermLike variable)
-            )
-        )
+    matcher :: MatcherT variable unifier (Maybe (MatchResult variable))
     matcher = pop >>= maybe done (\pair -> matchOne pair >> matcher)
 
     initial =
@@ -166,13 +161,7 @@ matchIncremental termLike1 termLike2 =
     free2 = (getFreeVariables . freeVariables) termLike2
 
     -- | Check that matching is finished and construct the result.
-    done
-        :: MatcherT variable unifier
-            (Maybe
-                ( Predicate variable
-                , Map (UnifiedVariable variable) (TermLike variable)
-                )
-            )
+    done :: MatcherT variable unifier (Maybe (MatchResult variable))
     done = do
         MatcherState { queued, deferred } <- Monad.State.get
         let isDone = null queued && null deferred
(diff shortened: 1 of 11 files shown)


traiansf:
Part of #1557

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -432,28 +436,59 @@ matchRule
...
-    -- Unify the left-hand side of the rule with the term of the initial
-    -- configuration.
-    let ruleLeft = matchingPattern rule
-    unification <- unifyPatterns ruleLeft initialTerm >>= maybe empty return
+    unification <- unifyPatterns >>= maybe empty return
     -- Combine the unification solution with the rule's requirement clause,
     let requires = precondition rule
     unification' <-
         evaluateRequires sideCondition initialCondition unification requires
+    checkConcrete concreteVars symbolicVars unification'
```
ttuegel: We should not evaluate the `requires` if the required variables are not symbolic or concrete, so this check should move up.

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -432,28 +436,59 @@ matchRule
...
+                Just t -> p t
+            isNotConcrete = check (not . TermLike.isConstructorLike)
+            isNotSymbolic = check TermLike.isConstructorLike
+            notConcretes =
+                filter isNotConcrete (Set.toList concretes)
+            notSymbolics =
+                filter isNotSymbolic (Set.toList symbolics)
+        Monad.unless (null notConcretes && null notSymbolics)
+            $ Unifier.explainAndReturnBottom
+                "Substitution not satisfying concrete/symbolic constraints"
```
ttuegel: This explanation could be more useful. Which constraints, specifically?

kore/src/Kore/Step/Step.hs
```diff
@@ -155,6 +156,9 @@ class UnifyingRule rule where
         -> rule variable1
         -> rule variable2
 
+    -- | Axiom attributes for this rule
+    ruleAttributes :: rule variable -> Attribute.Axiom TermLike.Symbol variable
```
ttuegel: This method ties us to rules that have the same representation. Instead, let's add a method like:

```.hs
checkInstantiation :: rule variable -> Map (UnifiedVariable variable) (TermLike variable) -> [InstantiationFailure variable]
```

where `InstantiationFailure` is a sum type of the ways that instantiating the rule can fail. By not tying us to a particular representation of rules, this gives us more freedom in the future. We are also separating concerns by extracting this logic from `EquationalStep`.

kore/src/Kore/Step/EquationalStep.hs
traiansf: Well, it was up; however, we need to normalize the substitution before the check and this means doing a predicate simplification.  We could probably perform two simplifications, one before, with just substitution, then perform the checks, then add the requires and resimplify.

Would you rather have that?

ttuegel: The matcher _always_ returns a normalized substitution, it should not be necessary to do any simplification after.

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -140,11 +135,21 @@ matchIncremental
     .  (MatchingVariable variable, MonadSimplify unifier)
     => TermLike variable
     -> TermLike variable
-    -> unifier (Maybe (Condition variable))
+    -> unifier
+        (Maybe
+            ( Predicate variable
+            , Map (UnifiedVariable variable) (TermLike variable)
+            )
```
ttuegel: We should add a type synonym for this.

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -429,26 +429,24 @@ matchRule
...
     -> unifier (UnifiedRule variable (rule variable))
 matchRule sideCondition initial rule = do
-    unification <- unifyPatterns >>= maybe empty return
+    -- Unify the left-hand side of the rule with the term of the initial
+    -- configuration.
+    (substPredicate, substitutionMap) <- unifyPatterns >>= maybe empty return
+    -- check instantiation
+    let failures = checkInstantiation rule substitutionMap
+        unification =
+            from (Substitution.fromMap substitutionMap) <> from substPredicate
```
ttuegel: Because this uses `from`, please give a type signature for `unification`.

kore/src/Kore/Step/EquationalStep.hs
```diff
@@ -462,6 +460,14 @@ matchRule sideCondition initial rule = do
             ruleLeft
             initialTerm
 
+    ruleWithSubstitution
+        :: Condition variable -> unifier (UnifiedRule variable (rule variable))
+    ruleWithSubstitution unification
```
ttuegel: The name of this helper doesn't give any clue that it evaluates the `requires`.

Issue #1648, opened 2020-03-09
No branching in K-EQUAL hooks


ttuegel:
After kframework/k#1151

The implementations of the K-EQUAL hooks should not branch, but expect concrete arguments.

---

- [ ] Concrete implementation of `==K` in the backend.
  The built-in implementation of `==K` should apply only to concrete (constructor-like) arguments. Given concrete arguments, its value should be `true` or `false` by comparing patterns using Haskell equality.
- [ ] Symbolic implementation of `==K` in the frontend.
  The rules are `rule A ==K B => true requires true #And {A #Equals B}` and `rule A ==K B => false [owise]`. These should go in a `K-EQUAL-SYMBOLIC` module with the `symbolic` and `kore` attributes to avoid interfering with the LLVM or Java backends.

ttuegel:
Fixed in #1785.

PR #1693, opened 2020-03-23
Evaluate BOOL.and by unification

ttuegel:
This implements internally the rule

```
\equals(andBool(X, Y), true) => \and(\equals(X, true), \equals(Y, true))
```

See also: #1648

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Builtin/Bool.hs
+++ b/kore/src/Kore/Builtin/Bool.hs
@@ -46,6 +46,10 @@ import qualified Data.Text as Text
 import qualified Text.Megaparsec as Parsec
 import qualified Text.Megaparsec.Char as Parsec
 
+import Kore.Attribute.Hook
+    ( Hook (..)
+    )
+import qualified Kore.Attribute.Symbol as Attribute.Symbol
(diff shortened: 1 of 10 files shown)


kore/src/Kore/Builtin/Bool.hs
```diff
@@ -165,3 +174,62 @@ builtinFunctions =
...
+ -}
+data BoolAnd term =
+    BoolAnd
+        { symbol :: !Symbol
+        , operand1, operand2 :: !term
+        }
+
+{- | Match the @BOOL.and@ hooked symbol.
+ -}
+matchBoolAnd :: TermLike variable -> Maybe (BoolAnd (TermLike variable))
```
traiansf: So this kind of constructs would be used instead of pattern synonyms?
I could get used to that...

ttuegel: Yes. I wanted to use a pattern synonym, but then I can't use the guard to check that the symbol is hooked to `BOOL.and`. (Actually, there is a way to "fake" guards using view patterns and tuples, but that starts to get complicated fast.)

Issue #1687, opened 2020-03-20
Repl: cannot redirect/pipe `claim` command


ana-pantilie:
`claim > file`, `claim >> file` and `claim | exec` do not get parsed by the `repl` parser.

If fixed, https://github.com/kframework/kore/blob/6e1ab7cf8ac172346b2103c64a69f3b9200c03b0/dist/kast.kscript#L1 should also be changed to `alias kclaim = claim | kast -i kore -o pretty -d . /dev/stdin`.

PR #1697, opened 2020-03-24
Repl: fix claim command parser

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1687

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Repl/Parser.hs
+++ b/kore/src/Kore/Repl/Parser.hs
@@ -251,7 +251,7 @@ claimIndexParser :: Parser ClaimIndex
 claimIndexParser = ClaimIndex <$$> Char.string "c" *> decimal
 
 ruleNameParser :: Parser RuleName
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Repl/Parser.hs
```diff
@@ -251,7 +251,7 @@ claimIndexParser :: Parser ClaimIndex
 claimIndexParser = ClaimIndex <$$> Char.string "c" *> decimal
 
 ruleNameParser :: Parser RuleName
-ruleNameParser = RuleName <$$> quotedOrWordWithout ""
+ruleNameParser = RuleName <$$> quotedOrWordWithout "|>"
```
andreiburdusa: So the problem was `|` parsed as part of the command and not as piping operator?

ana-pantilie: That's right.

Issue #1639, opened 2020-03-05
Better map unification needed


daejunpark:
A better map unification is needed especially for the Beacon chain verification.

Consider the following definition and its specs:

h.k:
```
module H

imports BOOL
imports INT
imports MAP

configuration <T> <k> $PGM:Cmd </k> <m> .Map </m> </T>

syntax Cmd ::= foo(Int)
rule <k> foo(X) => . ... </k>
     <m>
       .Map => (X |-> true)
     </m>

endmodule
```

h1-spec.k:
```
module H1-SPEC
imports H

rule <k> foo(X) => . </k>
     <m>
       .Map => (?Y |-> true)
     </m>
     requires X ==Int ?Y

endmodule
```

h2-spec.k:
```
module H2-SPEC
imports H

rule <k> foo(X +Int 1) => . </k>
     <m>
       .Map => (?Y |-> true)
     </m>
     requires X +Int 1 ==Int ?Y

endmodule
```

h3-spec.k:
```
module H3-SPEC
imports H

rule <k> foo(X) => . </k>
     <m>
       .Map => (?Y -Int 1 |-> true)
     </m>
     requires X ==Int ?Y -Int 1

endmodule
```

h4-spec.k:
```
module H4-SPEC
imports H

rule <k> foo(X +Int 1) => . </k>
     <m>
       .Map => (?Y -Int 1 |-> true)
     </m>
     requires X +Int 1 ==Int ?Y -Int 1

endmodule
```

The first two specs (h1-spec and h2-spec) are passed, but the last two (h3-spec and h4-spec) are failed.  I guess that the failure of h4-spec is due to the lack of a proper support for map unification.  But, the failure of h3-spec is quite surprising, because h2-spec and h3-spec are just dual, but only h2-spec is passed.

The Beacon chain verification requires the backend to be able to prove all the four specs.  Note that the equivalent expressions (e.g., `X +Int 1`, and `?Y -Int 1` in h4-spec) for the key are more complicated in the Beacon chain verification. For example, they are `X +Int N` and `(X /Int 32) *Int 32 + 31`,  or  `((X + 1) /Int 32) *Int 32 + 31` and `(X /Int 32) *Int 32 + 31`.

```
$ for i in h[1-4]-spec.k; do echo $i; kprove $i; done
h1-spec.k
#True

h2-spec.k
#True

h3-spec.k
<T>
  <k>
    .
  </k>
  <m>
    ?Y -Int 1 |-> true
  </m>
</T>

[...]

PR #1701, opened 2020-03-24
Added SMT conversion for int and bool existential quantifiers

traiansf:
I was a little reluctant to make the translation work for any sort, but I'm willing to implement it for all sorts if you think that it should work.

Fixes Issue #1639

---

###### Reviewer checklist


kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -219,6 +222,29 @@ translatePredicate translateUninterpreted predicate =
...
+        varTerm = mkElemVar existsVariable
+        varSort = sortedVariableSort (getElementVariable existsVariable)
+        tools :: SmtMetadataTools Attribute.Symbol
+        tools = given
+        Attribute.Sort { hook = Hook { getHook } } =
+            sortAttributes tools varSort
+
+    translateExists varSort varTerm predTerm
+      = do
+        var <- translateUninterpreted varSort varTerm
```
ttuegel: I'm a little uncertain about this; doesn't `translateUninterpreted` declare a global constant in the SMT-LIB translation?

traiansf: Yes, it does. And in this case it's useless. However, it seems to play well with the fact that this is the regular way to translate variables. If we want to translate them in a different way, we would need to change the way we translate patterns completely, to keep track of free and bound variables. Do you want me to try and do that?

traiansf: @ttuegel I've addressed your concerns by adding an extra argument to `translateUninterpreted` to select whether the variable should be declared or not.  Please take another look.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -219,6 +223,28 @@ translatePredicate translateUninterpreted predicate =
       where
         applicationChildrenSorts = termLikeSort <$> applicationChildren
 
+    translatePredicateExists Exists { existsVariable, existsChild } =
+        case  getHook of
+            Just builtinSort
+              | builtinSort == Builtin.Bool.sort
+              -> translateExists SMT.tBool varTerm existsChild
+              | builtinSort == Builtin.Int.sort
+              -> translateExists SMT.tInt varTerm existsChild
```
virgil-serbanuta: I think the SMT should also be able to easily handle sorts that are constructor-based without recursive constructors, but it's probably not worth the trouble to add that now.

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+    :: InternalVariable variable
+    => SMT.MonadSMT m
+    => TermLike.ElementVariable variable
+    -> Translator m variable SExpr
+lookupVariable var = do
+    VarContext { freeVars, quantifiedVars } <- State.get
+    maybe empty return $
+        (SMT.Atom . smtName <$> Map.lookup var quantifiedVars)
+        <|>
+        Map.lookup var freeVars
```
ttuegel: Even though this function is short, I had to read it through several times. I found it difficult to follow because the handling of free and quantified variables is intertwined.

```suggestion
lookupVariable var =
    lookupQuantifiedVariable <|> lookupFreeVariable
  where
    lookupQuantifiedVariable = do
        VarContext { quantifiedVars } <- State.get
        hoistMaybe $ SMT.Atom . smtName <$> Map.lookup var quantifiedVars
    lookupFreeVaraible = do
        VarContext { freeVars } <- State.get
        hoistMaybe $ Map.lookup var freeVars
```

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+    => SMT.MonadSMT m
+    => MonadLog m
+    => SExpr  -- ^ type name
+    -> TermLike.ElementVariable variable  -- ^ variable to be declared
+    -> Translator m variable SExpr
+declareVariable t variable = do
+    n <- Counter.increment
+    let varName = "<" <> Text.pack (show n) <> ">"
+    var <- SMT.declare varName t
+    State.modify' (Lens.over (field @"freeVars") $ Map.insert variable var)
```
ttuegel:
```suggestion
   field @"freeVars" Lens.%= Map.insert variable var
```

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+translateSmtEncoding
+    quantifiedVars
+    SmtEncoding { smtName = funName, boundVars }
+  =
+    maybe empty (return . SimpleSMT.fun funName) boundEncodings
+  where
+    boundEncodings =
+        traverse
+            (fmap (SMT.Atom . smtName) . (`Map.lookup` quantifiedVars))
+            boundVars
```
ttuegel: I think point-free style is not making this clearer:

```suggestion
        for boundVars $ \name -> SMT.Atom . smtName <$> Map.lookup name quantifiedVars
```

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+            cache = SmtEncoding { smtName, smtType = t, boundVars }
+        _ <- SMT.declareFun
+            SMT.FunctionDeclaration
+            { name = smtName
+            , inputSorts =
+                smtType <$> mapMaybe (`Map.lookup` boundVarsMap) boundVars
+            , resultSort = t
+            }
+        State.modify'
+            ( Lens.over (field @"terms") $ Map.insert boundPat cache )
```
ttuegel:
```suggestion
        field @"terms" Lens.%= Map.insert boundPat cache
```

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+    let varName = "<" <> Text.pack (show n) <> ">"
+        smtVar = SimpleSMT.const varName
+    State.modify'
+        (Lens.over (field @"quantifiedVars") $ Map.insert var
+            SmtEncoding
+            { smtName = varName
+            , smtType
+            , boundVars = []
+            }
+        )
```
ttuegel:
```suggestion
    field @"quantifiedVars" Lens.%= Map.insert var
        SmtEncoding
        { smtName = varName
        , smtType
        , boundVars = []
        }
```

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+    VarContext { quantifiedVars, terms } <- State.get
+    let freeVars = getFreeVariables $ TermLike.freeVariables pat
+        boundVarsMap =
+            Map.filterWithKey
+                (\k _ -> ElemVar k `Set.member` freeVars)
+                quantifiedVars
+        boundVars = Map.keys boundVarsMap
+        boundPat = TermLike.mkExistsN boundVars pat
+    lookupUninterpreted boundPat quantifiedVars terms
+        <|> declareUninterpreted boundPat boundVars boundVarsMap
```
ttuegel: Why does `declareUninterpreted` need `boundVars` in addition to `boundVarsMap`, given that the former is just the keys of the latter?

kore/src/Kore/Step/SMT/Lemma.hs
```diff
@@ -72,41 +74,49 @@ declareSMTLemmas m = do
...
        , Monad m
        )
     => SExpr  -- ^ type name
-    -> p  -- ^ uninterpreted pattern
-    -> Translator m p SExpr
-translateUninterpreted t pat | isVariable pat =
+    -> TranslateItem variable -- ^ uninterpreted pattern
+    -> Translator m variable SExpr
+translateUninterpreted _ (QuantifiedVariable _) =
+    error "Lemma encodings don't support quantified variables (yet)"
```
ttuegel: Why not `empty` instead of `error`? That would match the existing behavior, which rejects quantifiers anyway.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -219,7 +235,37 @@ translatePredicate translateUninterpreted predicate =
...
+        :: SExpr -> ElementVariable variable -> p -> Translator m variable SExpr
+    translateExists varSort var predTerm
+      = do
+        oldVar <- State.gets (Map.lookup var . quantifiedVars)
+        smtVar <- translateTerm varSort (QuantifiedVariable var)
+        smtPred <- translatePredicatePattern predTerm
+        State.modify'
+            ( Lens.over (field @"quantifiedVars")
+            $ maybe (Map.delete var) (Map.insert var) oldVar
+            )
```
ttuegel:
```suggestion
        field @"quantifiedVars" Lens.%= maybe (Map.delete var) (Map.insert var) oldVar
```

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -237,14 +283,30 @@ translatePredicate translateUninterpreted predicate =
         Attribute.Sort { hook = Hook { getHook } } =
             sortAttributes tools sort
 
+data SmtEncoding variable = SmtEncoding
+    { smtName :: Text
+    , smtType :: SExpr
+    , boundVars :: [ElementVariable variable]
```
ttuegel: These fields should be strict.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -237,14 +283,30 @@ translatePredicate translateUninterpreted predicate =
         Attribute.Sort { hook = Hook { getHook } } =
             sortAttributes tools sort
 
+data SmtEncoding variable = SmtEncoding
```
ttuegel: This needs documentation. In particular: What is this the encoding _of_?

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -237,14 +283,30 @@ translatePredicate translateUninterpreted predicate =
...
 -- ----------------------------------------------------------------
 -- Translator
-type VarContext p = Map p (SExpr, SExpr)
+data VarContext variable
+  = VarContext
+    { terms :: !(Map (TermLike variable) (SmtEncoding variable))
+    , freeVars :: !(Map (ElementVariable variable) SExpr)
+    , quantifiedVars :: !(Map (ElementVariable variable) (SmtEncoding variable))
+    }
+    deriving (Eq, GHC.Generic, Show)
```
ttuegel:
```suggestion
    = VarContext
        { terms :: !(Map (TermLike variable) (SmtEncoding variable))
        , freeVars :: !(Map (ElementVariable variable) SExpr)
        , quantifiedVars :: !(Map (ElementVariable variable) (SmtEncoding variable))
        }
    deriving (Eq, GHC.Generic, Show)
```

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -237,14 +283,30 @@ translatePredicate translateUninterpreted predicate =
...
+    { smtName :: Text
+    , smtType :: SExpr
+    , boundVars :: [ElementVariable variable]
+    }
+    deriving (Eq, GHC.Generic, Show)
+
 -- ----------------------------------------------------------------
 -- Translator
-type VarContext p = Map p (SExpr, SExpr)
+data VarContext variable
```
ttuegel: I think we should call this `TranslatorState`, or something like that, because it emphasizes that it's related to the translator, and because it doesn't only concern variables.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -219,7 +235,37 @@ translatePredicate translateUninterpreted predicate =
       where
         applicationChildrenSorts = termLikeSort <$> applicationChildren
 
-    translatePattern :: Sort -> p -> Translator m p SExpr
+    translatePredicateExists
+        :: Exists Sort variable p -> Translator m variable SExpr
+    translatePredicateExists Exists { existsVariable, existsChild } =
+        case  getHook of
```
ttuegel:
```suggestion
        case getHook of
```

deps/k_release
```diff
@@ -1 +1 @@
-https://github.com/kframework/k/releases/download/v5.0.0-b3f3e28
+https://github.com/kframework/k/releases/download/v5.0.0-80b5cf9
```
ttuegel: Did you mean to change the K release?

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -59,6 +68,11 @@ import SMT
     )
 import qualified SMT
 
+
+data TranslateItem variable
+    = QuantifiedVariable (ElementVariable variable)
+    | UninterpretedTerm (TermLike variable)
```
ttuegel: These constructors should be strict.

kore/src/Kore/Step/SMT/Evaluator.hs
traiansf: true, I just had it around and thought to use it. I've changed it as you suggested.

kore/src/Kore/Step/SMT/Lemma.hs
traiansf: ok... I just thought we might want to find out if anyone tries to declare a lemma with quantifiers. If we use empty here, it would be silently ignored.

deps/k_release
traiansf: yes, I think I meant to.  But I don't remember exactly why. I think that the tests weren't working as expected without the latest changes in the frontend.

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
-        logVariableAssignment n
-        return var
-    logVariableAssignment n =
-        logDebug
-        . Text.pack . show
-        . Pretty.nest 4 . Pretty.sep
-        $ [Pretty.pretty n, "|->", unparse pat]
+        logVariableAssignment n boundPat
+        let smtName = "<" <> Text.pack (show n) <> ">"
+            cache = SmtEncoding { smtName, smtType = t, boundVars }
```
virgil-serbanuta: Would s/cache/cached/ be better?

kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -173,30 +180,114 @@ goTranslatePredicate
...
+        _ <- SMT.declareFun
+            SMT.FunctionDeclaration
+            { name = smtName
+            , inputSorts =
+                smtType <$> mapMaybe (`Map.lookup` boundVarsMap) boundVars
```
virgil-serbanuta: Why is mapMaybe reasonable here? Why not mapM?

traiansf:
@ttuegel @virgil-serbanuta I think this is much better now; thanks for your feedback! could you please take another look?

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -219,7 +240,39 @@ translatePredicate translateUninterpreted predicate =
...
+      where
+        existsBuiltinSort = case getHook of
+            Just builtinSort
+              | builtinSort == Builtin.Bool.sort
+              -> translateExists SMT.tBool existsVariable existsChild
+              | builtinSort == Builtin.Int.sort
+              -> translateExists SMT.tInt existsVariable existsChild
+            _ -> empty
+        existsConstructorSort = do
+            smtSort <- hoistMaybe $translateSort varSort
```
virgil-serbanuta, line 256: space after $

PR #1703, opened 2020-03-25
LoggerT: Keep an explicit context stack instead of a closure

ttuegel:
Keeping a closure which implicitly contains the context stack is less efficient
than explicitly carrying the context stack.

---

###### Reviewer checklist


kore/src/Log.hs
```diff
@@ -214,38 +225,60 @@ instance MonadLog log => MonadLog (ReaderT a log)
 
 instance MonadLog log => MonadLog (Strict.StateT state log)
 
+data LoggerEnv monad =
+    LoggerEnv
+        { logAction :: !(LogAction monad ActualEntry)
```
ana-pantilie: Maybe I'm not understanding this right, but since we're moving the context in the env and out of the `LogAction`, shouldn't the type for the `LogAction` be `SomeEntry` instead of `ActualEntry`?

ttuegel: The `LogAction` consumes an `ActualEntry`, so this requires us to give it the context whenever we log an entry. If it was a `SomeEntry` `LogAction`, we wouldn't have anywhere to put the context.

Issue #1560, opened 2020-02-18
Unwrap EqualityPattern


ttuegel:
[`EqualityRule`](https://github.com/kframework/kore/blob/dafda53fac500e02a598fafafec2ecdac01756c7/kore/src/Kore/Step/EqualityPattern.hs#L139-L141) exists only to wrap `EqualityPattern`. We should remove the wrapper and just rename `EqualityPattern` to `EqualityRule`. Not only does the wrapper seem pointless, but `Pattern` isn't the correct name; it's a `Rule`.

PR #1711, opened 2020-03-30
Begin to collect the scattered parts of equation application

ttuegel:
The parts of the backend related to applying equations are scattered through several parts of the code. Here, I begin to collect them under `Kore.Equation`.

See also: #1560 

---

###### Reviewer checklist


ttuegel:
Eventually, the old types will disappear, but temporarily you will see conversions in some places.

PR #1714, opened 2020-03-30
Concrete implementation of ==K hook; internal simplification rules

ana-pantilie:
~Needs https://github.com/kframework/kore/pull/1772~
Needs https://github.com/kframework/kore/pull/1774

Fixes https://github.com/kframework/kore/issues/1648

- `==K` hook is evaluated for constructor-like input only
- adds internal simplification rules for `bool` and `kequal`
- fixes unit tests
- adds new integration test `not-kequal` (reduced example from `evm-semantics`)

---

###### Reviewer checklist


ttuegel:
We should make the new integration tests independent of the frontend, so that we do not have to wait for the frontend release to pass before we can merge this pull request. The `evm-semantics` regression tests in this repository do the same thing. There is a [script](https://github.com/kframework/kore/blob/master/test/regression-evm-semantics-39dd1e5/generate.sh) which generates the Kore files from the K files, and we commit the Kore files into the repository. You should make a copy of that script and modify it for the new tests. For each test, the script should invoke `kollect test-name kprove ...` (fill in `...` with the arguments to `kprove`). Then, you would run the script with the correct frontend version on your `PATH` and with `KORE` pointing to your clone of the Kore repository. We should also commit the K source files to the repository (so that the tests can be re-generated) but these should go in a subdirectory so that the `Makefile` doesn't try to run them automatically. (It will run the generated tests instead.)

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -207,7 +213,34 @@ termAndEquals unifyChildren a b =
...
+        Condition.fromPredicate
+        $ makeEqualsPredicate_ (falseTerm boolTerm) term
+    falseTerm term = asInternal (termLikeSort term) False
+    worker termLike1 trueTerm
+      | Just BoolNot { operand } <- matchBoolNot termLike1
+      , isFunctionPattern termLike1
+      , Just value2 <- matchBool trueTerm
+      , value2
+      = lift $ do
+        condition <- unifyChildren' trueTerm operand
```
ttuegel, line 242: This is not right. The terms are `notBool operand` and `true`, and then this unifies `operand` with `true`. It should unify `operand` with `false`. Actually, it can be a little more general: given `notBool operand` and (`true` or `false`) it should unify `operand` with (`false` or `true`).

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -207,7 +213,34 @@ termAndEquals unifyChildren a b =
...
+    => TermSimplifier variable unifier
+    -> TermLike variable
+    -> TermLike variable
+    -> MaybeT unifier (Pattern variable)
+termNotBool unifyChildren a b =
+    worker a b <|> worker b a
+  where
+    unifyChildren' boolTerm term =
+        Pattern.withoutTerm <$> unifyChildren (falseTerm term) term
+        <|> return (equalsCondition boolTerm term)
```
ttuegel, line 231: There should not be the alternative (`<|>`) case here. If this is called as part of the `And` simplifier, it gives the wrong result, and if it is called in the `Equals` simplifier, the fall-back happens automatically.

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -206,3 +209,96 @@ neqKey = "KEQUAL.neq"
...
+    worker termLike1 termLike2
+      | Just KEqual { operand1, operand2 } <- matchKEqual termLike1
+      , isFunctionPattern termLike1
+      , Just value2 <- Bool.matchBool termLike2
+      , value2
+      , atLeastOneSymbolic operand1 operand2
+      = lift $ do
+        condition <- unifyChildren' operand1 operand2
+        pure Conditional
+            { term = mkTop (termLikeSort termLike1)
```
ttuegel, line 262: The term should be `termLike2`; what you have here is only correct when we have `(_ ==K _) = true`, but not when we have `(_ ==K _) ∧ true`. The second form is likely to appear in `owise` rules.

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -206,3 +209,96 @@ neqKey = "KEQUAL.neq"
...
+
+termKEqualsFalse
+    :: forall variable unifier
+    .  InternalVariable variable
+    => MonadUnify unifier
+    => TermSimplifier variable unifier
+    -> TermLike variable
+    -> TermLike variable
+    -> MaybeT unifier (Pattern variable)
+termKEqualsFalse unifyChildren a b =
```
ttuegel, line 276: We should combine this with `termKEqualsTrue`.

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -206,3 +209,96 @@ neqKey = "KEQUAL.neq"
...
+    worker termLike1 termLike2
+      | Just KEqual { operand1, operand2 } <- matchKEqual termLike1
+      , isFunctionPattern termLike1
+      , Just value2 <- Bool.matchBool termLike2
+      , not value2
+      , atLeastOneSymbolic operand1 operand2
+      = lift $ do
+        condition <- unifyChildren' operand1 operand2
+        pure Conditional
+            { term = mkTop (termLikeSort termLike1)
```
ttuegel, line 300: The same comment, about the term, from `termKEqualsTrue` applies here.

kore/test/Test/Kore/Builtin/KEqual.hs
```diff
@@ -55,30 +55,6 @@ test_KEqual =
         actual <- evaluate original
         assertEqual' "" expect actual
 
-    , testCaseWithSMT "kseq(x, dotk) equals kseq(x, dotk)" $ do
```
ttuegel, line 58: Why are these tests removed?

ana-pantilie: I initially removed them because they were dependent on `==K` simplification and that was handled by applying the frontend rules. It looks like they fail even with the internal implementation, I'm looking into what might be wrong.

ana-pantilie: Update: the term needed to be an `and true ...` or an `equals true ...` in order for the internal simplification to work; I'll update the tests.

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -206,3 +208,61 @@ neqKey = "KEQUAL.neq"
...
+    => TermSimplifier variable unifier
+    -> TermLike variable
+    -> TermLike variable
+    -> MaybeT unifier (Pattern variable)
+termKEquals unifyChildren a b =
+    worker a b <|> worker b a
+  where
+    unifyChildren' term1 term2 =
+        unificationPredicate term1 term2
+        <|> return (makeEqualsPredicate_ term1 term2)
```
ttuegel, line 242: We should not have this fallback case because it prevents the unification procedure from failing correctly.

kore/src/Kore/Step/Simplification/Simplify.hs
```diff
@@ -154,7 +155,8 @@ class (WithLog LogMessage m, MonadSMT m, MonadProfiler m)
 
     simplifyCondition
         :: InternalVariable variable
-        => SideCondition variable
+        => NotSimplifier m2
```
ttuegel, line 158: We don't want to add this argument here, because how the instance of `MonadSimplify` chooses to simplify `\not` is an implementation detail: it could do anything or nothing at all, and we don't want that to leak out into the interface.

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -107,14 +108,15 @@ import Kore.Variables.UnifiedVariable
...
     )
 
-newtype SubstitutionSimplifier simplifier =
+newtype SubstitutionSimplifier simplifier1 simplifier2 =
     SubstitutionSimplifier
         { simplifySubstitution
             :: forall variable
             .  InternalVariable variable
-            => SideCondition variable
+            => NotSimplifier simplifier1
```
ttuegel, line 116: This argument should not be here. If `SubstitutionSimplifier` needs a `NotSimplifier`, then it should capture it when it is created, i.e. the `NotSimplifier` should be an argument to `substitutionSimplifier`.

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -155,7 +167,8 @@ newtype MakeAnd monad =
         { makeAnd
             :: forall variable
             .  InternalVariable variable
-            => TermLike variable
+            => NotSimplifier monad
```
ttuegel, line 170: Same as for `SubstitutionSimplifier`: if `MakeAnd` needs a `NotSimplifier`, it should capture that dependency when it is created.

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -208,7 +210,27 @@ termAndEquals unifyChildren a b =
...
+    worker a b <|> worker b a
+  where
+    makeNot :: TermLike variable -> Bool -> TermLike variable
+    makeNot term value =
+        asInternal (termLikeSort term) (not value)
+    worker termLike1 boolTerm
+      | Just BoolNot { operand } <- matchBoolNot termLike1
+      , isFunctionPattern termLike1
+      , Just value <- matchBool boolTerm
+      = lift $ unifyChildren (makeNot boolTerm value) operand
```
ttuegel, line 233: I think we should inline `makeNot` here:

```suggestion
    worker termLike1 boolTerm
      | Just BoolNot { operand } <- matchBoolNot termLike1
      , isFunctionPattern termLike1
      , Just value <- matchBool boolTerm
      = lift $ do
        let notValue = asInternal (termLikeSort boolTerm) (not value)
        unifyChildren notValue operand
```

kore/src/Kore/Builtin/KEqual.hs
```diff
@@ -206,3 +203,48 @@ neqKey = "KEQUAL.neq"
...
+      , Just value2 <- Bool.matchBool termLike2
+      = lift $ do
+        solution <-
+            fmap OrPattern.fromPatterns
+            <$> Unify.gather
+            $ unifyChildren operand1 operand2
+        finalSolution <-
+            if value2
+                then return solution
+                else notSimplifier SideCondition.top solution
```
ttuegel, line 248: This is not right when we are doing `\and` simplification: the term should always be `value2` (`true` or `false`) because the unification problem is something like:

```
\and(true, X ==K Y)
```

and `solution` is the result of simplifying `\and(X, Y)`. This is fine for `\equals` simplification, because the unification problem is like:

```
\equals(true, X ==K Y)
```

and `solution` is the result of simplifying `\equals(X, Y)` so that the `term` is always `\top` or `\bottom`. I think it will solve the problem you are seeing with `with-config` tests if we restrict this unification case to `\equals` simplification.

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -218,6 +231,8 @@ andEqualsFunctions = fmap mapEqualsFunctions
     , (BothT,   \_ _ _ -> constructorAndEqualsAssumesDifferentHeads, "constructorAndEqualsAssumesDifferentHeads")
     , (BothT,   \_ _ s -> overloadedConstructorSortInjectionAndEquals s, "overloadedConstructorSortInjectionAndEquals")
     , (BothT,   \_ _ s -> Builtin.Bool.termAndEquals s, "Builtin.Bool.termAndEquals")
+    , (BothT,   \_ _ s -> Builtin.Bool.termNotBool s, "Builtin.Bool.termNotBool")
+    , (BothT,   \_ _ s -> Builtin.KEqual.termKEquals s notSimplifier, "Builtin.KEqual.termKEquals")
```
ttuegel, line 235: I think this change will solve the problem you are having:

```suggestion
    , (EqualsT,   \_ _ s -> Builtin.KEqual.termKEquals s notSimplifier, "Builtin.KEqual.termKEquals")
```

PR #1720, opened 2020-04-01
Enable new equation application

### LINKED ISSUE #1473: Hot-spot: Kore.Step.EquationalStep.applyRulesSequence
This function is a hot-spot because it is the top of the inner loop of function evaluation. The generated code is more complex than it needs to be because this is not yet specialized to function evaluation; it still refers to `Kore.Step.Step`.

### LINKED ISSUE #1560: Unwrap EqualityPattern

### LINKED ISSUE #1602: Apply defining equations in parallel
Because we reject definition rules which leave remainders, we need to apply equations to the configuration in parallel, rather than in sequence; that is, each rule applied to the term, rather than to the remainder of the previous rule. Each rule is still expected to apply without remainders. We should issue a warning about non-determinism if multiple defining equations (with the same priority) match the term.

---

- [ ] Apply equations in parallel.
- [ ] For simplification rules, `priority` is not semantic. These rules just get ordered.
- [ ] Verification error if a definition rule has `priority`.

### LINKED ISSUE #1689: Logging applied function rules
The log entry `DebugAppliedRule` is logged whenever the rule gets matched. 
We also want to log if the rule application succeeds, here: https://github.com/kframework/kore/blob/bc4cd5ded05e1a3dbeb744027818d326f69f5f68/kore/src/Kore/Step/Axiom/EvaluationStrategy.hs#L89
We should also log if the rule gets skipped, in the `| otherwise` case.

In order to log this information, `evaluateAxioms` should also return which rule got applied.

ttuegel:
This pull request brings the new equation application code in `Kore.Equation.Application` up to feature parity with the old equation application code, which is replaced everywhere. To make this pull request easier to review, I will remove the old code in a separate pull request which only deletes code.

Fixes #1689 
Fixes #1602 
Fixes #1560 
Fixes #1473 

---

###### Reviewer checklist


ttuegel:
@traiansf I asked you to review this because I want to be absolutely sure that I've got the side conditions handled correctly. In the old equation code, you had written some things to handle remainders correctly when the `requires` is not implied by the side condition. The new equation code demands that the `requires` clause is implied by the side condition, so all the code you wrote went away. I just want to double-check that is correct.

kore/src/Kore/Equation/Equation.hs
```diff
@@ -262,3 +273,32 @@ substitute assignments equation =
...
   where
     Equation { requires, left, right, ensures, attributes } = equation
+
+{- | The list of identifiers for an 'Equation'.
+
+The identifiers are:
+
+* the rule label
+* symbol names on the left-hand side
+* symbol 'Klabel's on the left-hand side
```
traiansf, line 283: why do we need symbol KLabels? Aren't symbol names enough?

traiansf: Reading below, they seem to be useful for identifying the rule when rewriting. Is that the case?

kore/src/Kore/Log/DebugAppliedRewriteRules.hs
```diff
@@ -54,10 +54,8 @@ import Log
 
 data DebugAppliedRewriteRules =
     DebugAppliedRewriteRules
-        { configuration
-            :: !(Pattern Variable)
-        , appliedRewriteRules
-            :: ![UnifiedRule Variable (RewriteRule Variable)]
+        { configuration :: Pattern Variable
```
ana-pantilie, line 57: Why did you remove the strictness annotations here and below?

ttuegel:
> However, I was not able to locate the remainder-related code you mentioned and your related changes.

@traiansf Yes, I deleted all of the remainder-related code because the method of applying equations can't produce remainders anymore. That's what I wanted to check with you about.

kore/src/Kore/Equation/Equation.hs
ttuegel: A typical user won't know the symbol names, just the K labels. Allowing both ways of specifying a rule is an affordance to users and developers both.

kore/src/Kore/Log/DebugAppliedRewriteRules.hs
ttuegel: The log entry data types shouldn't be strict because we don't want the fields to be evaluated if that entry type is disabled, and if an entry type is enabled, we want evaluation to happen in the logging thread, not in the main thread.

virgil-serbanuta: Since the standard thing to do is to have strict fields, and this explanation might not be easy to discover for someone who doesn't already know it, would it be worth to add a code comment?

Issue #1628, opened 2020-03-04
kore-repl: stepf should advance to the branching point


ttuegel:
See also: #1625

PR #1729, opened 2020-04-03
Advance current node when running stepf.

mcalancea:
Addresses https://github.com/kframework/kore/issues/1628. Comments document the expected behaviour. 

LE: An image illustrating the aim of the integration test: Running `stepf 11` advances the proof to node `10` which is branching at the moment. Afterwards, running `stepf 4` would advance it to `18` because `10` is no longer branching, its other children having finished their proofs during these `4` steps.

###### Reviewer checklist


traiansf:
I wonder
1. whether there is a place describing the REPL commands. Then some documentation (maybe your Haskell comments)  should be added there
2. whether it would be possible to test this somehow. I remember @ana-pantilie has written some tests for the REPL.  Besides guarding for regressions, they might complement the documentation.

mcalancea:
> > I wonder
> > ```
> > 1. whether there is a place describing the REPL commands. Then some documentation (maybe your Haskell comments)  should be added there
> > 
> > 2. whether it would be possible to test this somehow. I remember @ana-pantilie has written some tests for the REPL.  Besides guarding for regressions, they might complement the documentation.
> > ```
> 
> Yes, please modify the `helpText` in `Data.hs` accordingly. Also, please add (at least) one test in `Test.Kore.Repl.Interpreter.hs` which checks if the current node has been modified correctly. Also, you can unit test graph related functions in `Test.Kore.Repl.Graph.hs`.

Thanks for the comments! 

I was hesitant to modify the help text because it looks like it is generally meant to be very succinct. Presently, it doesn't even specify that `step` modifies the current node and `stepf` doesn't. The exact behaviour is rather verbose and I think it might become cumbersome for people to scroll around the message, given that we print all of it every time. (It might be a good idea for help to take a command as an argument? or have a separate `help --verbose command` which expands on the details). In any case, I would like suggestions regarding how much info should go into the message. 

The stuff in `Test.Kore.Repl.Graph.hs` seems to concern `Graphs` exclusively, without any business logic, while my code checks proof-states of nodes. Should it still go there?

I'll work out some tests for `Test.Kore.Repl.Interpreter.hs`.

ana-pantilie:
@mcalancea You can either describe the behavior very generally right next to `stepf`, but if you think that would be confusing you could just add a footnote at the bottom of the help text describing the behavior in more detail. Regarding `Graph.hs`, you can unit test `InnerGraph` functions there as well, even though we currently don't have tests which do that.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1249,6 +1256,35 @@ recursiveForcedStep n node
...
+    | otherwise =
+        case interestingResults of
+            []  -> Nothing
+            [a] -> Just a
+            _   -> Just node
+    where
+      gnode = unReplNode node
+      nodeIsBottom = (==) (getNodeState graph gnode) Nothing
+      sucResults = fmap (getInterestingBranchingNode (n - 1) graph . ReplNode)
+                        (Graph.suc graph gnode)
```
ana-pantilie, line 1283: Please change the indentation here. See https://github.com/kframework/kore/wiki/Coding-Standards#line-splitting

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1249,6 +1256,35 @@ recursiveForcedStep n node
         SingleResult sr -> (recursiveForcedStep $ n-1) sr
         BranchResult xs -> Foldable.traverse_ (recursiveForcedStep (n-1)) xs
 
+{- | Returns the first interesting branching node encountered by
+exploring the proof graph for 'n' steps over all branches, starting
+from 'node'. If no such node exists, it tries to return the only existing
+non-bottom leaf. If no such leaf exists, it returns Nothing.
+-}
+
```
ana-pantilie, line 1264: No newline here. Also, I'd move this function to `State.hs`.

test/imp/test-stepf.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+IGNORE=$($KORE_REPL sum-spec-vdefinition.kore -r --repl-script run-stepf-repl-script-spec.k.repl --module VERIFICATION --prove sum-spec.kore --spec-module SUM-SPEC)
+
+if grep -q "Config at node 10" shouldBeAt10.out; then
+    if grep -q "Config at node 18" shouldBeAt18.out; then
+        diff finished.out finished.out.golden
+    else
+        echo "Current node is not qual to expected value 10 after second call to stepf."
```
ana-pantilie, line 9: I think you have a typo here.

test/imp/test-stepf.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+IGNORE=$($KPROVE --haskell-backend-command "$KORE_REPL -r --repl-script run-stepf-repl-script-spec.k.repl" -d . -m VERIFICATION sum-spec.k)
```
ana-pantilie, line 3: Please change this to use the new make rule from https://github.com/kframework/kore/pull/1742.

test/imp/test-stepf.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if grep -q "Config at node 10" shouldBeAt10.out; then
+    if grep -q "Config at node 18" shouldBeAt18.out; then
+        diff finished.out finished.out.golden
```
ana-pantilie, line 7: I think not using the `test-stepf.sh.out.golden` file here is confusing, and you'd need to manually regenerate the `finished.out.golden` if you ever change the `Proof completed...` text. Would replacing this line with `cat finished.out` work?

include.mk
```diff
@@ -24,6 +24,7 @@ KOMPILE = $(K_RELEASE_BIN)/kompile
 KRUN = $(K_RELEASE_BIN)/krun
 export KRUN
 KPROVE = $(K_RELEASE_BIN)/kprove
+export KPROVE
```
ana-pantilie, line 27: I don't think you need this anymore.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1249,6 +1256,7 @@ recursiveForcedStep n node
         SingleResult sr -> (recursiveForcedStep $ n-1) sr
         BranchResult xs -> Foldable.traverse_ (recursiveForcedStep (n-1)) xs
 
+
```
ana-pantilie, line 1259: This newline should be removed.

test/imp/Makefile
```diff
@@ -2,3 +2,4 @@ DEF = imp
 KORE_MODULE = IMP
 GOLDEN += sum-save-proofs-spec.k.save-proofs.kore.golden
 include $(CURDIR)/../include.mk
+.PHONY: run-repl-script
```
ana-pantilie, line 5: I don't think this works :disappointed:. I'll try to figure out what's going on with the repl-script make rule in a follow-up PR.

test/imp/test-stepf.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+#IGNORE=$($KPROVE --haskell-backend-command "$KORE_REPL -r --repl-script run-stepf-repl-script-spec.k.repl" -d . -m VERIFICATION sum-spec.k)
```
ana-pantilie, line 3: You forgot to remove this.

test/imp/run-stepf-repl-script-spec.k.repl
```diff
@@ -0,0 +1,5 @@
+stepf 11
```
ana-pantilie, line 1: You don't use this file anymore, right?

ttuegel:
> Also, I see the integration test ran fine on Jenkins, except for this warning `make[3]: warning: jobserver unavailable: using -j1. Add '+' to parent make rule.`. @ttuegel do you think this could cause issues?

Yes, this means that the test script is calling `make`, which it must not do.

test/imp/Makefile
ttuegel: Please no `.PHONY` targets in the tests; these have to be re-run every time you `make test`. File targets are only rebuilt when the test fails or `kore-exec` is updated, which is necessary when iterating through many broken tests to fix them.

ana-pantilie:
> > Also, I see the integration test ran fine on Jenkins, except for this warning `make[3]: warning: jobserver unavailable: using -j1. Add '+' to parent make rule.`. @ttuegel do you think this could cause issues?
> 
> Yes, this means that the test script is calling `make`, which it must not do.

I see. I suggested using `make` here in order to automate running the `repl` with a script file, otherwise we would have needed to store the definition for `imp` even though the test doesn't actually test anything related to the `kore` output, so it should work the same regardless of the frontend version. Would it be fine to just store the (~3k line) definition and run `kore-repl` directly, or do you have any idea how to automate running the `repl` with `kprove` and making sure `kompile` was run before without using `make`?  @ttuegel

test/imp/Makefile
ana-pantilie: This was an attempt to solve a different issue, related to the `run-repl-script` Makefile rule I added yesterday. There's something not working as intended which I need to figure out.

ttuegel:
> do you have any idea how to automate running the `repl` with `kprove` and making sure `kompile` was run before without using `make`?

Yes, that is what [this target](https://github.com/kframework/kore/blob/master/test/include.mk#L117) does. There should be a spec named `something-repl-script-spec.k`; it can be a link to another spec, if they are the same. The `kore-repl` script should be named `something-repl-script-spec.k.repl`. The target to run the test is `something-repl-script-spec.k.out`, and `something-repl-script-spec.k.out.golden` will create the golden file for comparison.

ana-pantilie:
> > do you have any idea how to automate running the `repl` with `kprove` and making sure `kompile` was run before without using `make`?
> 
> Yes, that is what [this target](https://github.com/kframework/kore/blob/master/test/include.mk#L117) does. There should be a spec named `something-repl-script-spec.k`; it can be a link to another spec, if they are the same. The `kore-repl` script should be named `something-repl-script-spec.k.repl`. The target to run the test is `something-repl-script-spec.k.out`, and `something-repl-script-spec.k.out.golden` will create the golden file for comparison.

Wouldn't that just run the spec with the script? Sorry, I forgot to explain the intention of the test in my previous reply. The test should check that after `stepf n`, a specific node is reached. So some additional logic is required, as in the `test-stepf.sh` script added by @mcalancea .

ttuegel:
> The test should check that after `stepf n`, a specific node is reached.

The test in this pull request does that by checking the output from the REPL, right? That output will be in the golden file.

ana-pantilie:
> The test in this pull request does that by checking the output from the REPL, right? That output will be in the golden file.

Hmm, yes, that should work, and it shouldn't require printing the config anymore, because the repl prints the current node in the prompt. It looks like we were overthinking this test. 

I tried this out but I think there are some issues in the rules in the Makefile. [Here](https://github.com/kframework/kore/blob/c12bfca964733d163a86e4679be31e16b0057f1a/test/include.mk#L99) we use `KPROVE_OPTS`, and this always calls `kore-exec`, see [link](https://github.com/kframework/kore/blob/c12bfca964733d163a86e4679be31e16b0057f1a/include.mk#L32). This means that our current `repl-script` tests don't get run with the `repl` either.

PR #1738, opened 2020-04-07
SQL: Allow recursive datatypes

ttuegel:
This pull request refactors `SQL` to allow recursive datatypes.

1. The separate `SQL.Table` and `SQL.Column` modules are combined into `SQL` to prevent orphan instances when defining columns as foreign keys.
1. The above requires refactoring `SQL.SOP` so that it does not depend on `Column`, because `Table` depends on `SQL.SOP`.
1. `defineColumn` takes an additional argument: the name of the table where the column will reside. This is used to avoid re-defining the table for a recursive datatype when defining the self-referencing key column.
1. Instances of `Table` and `Column` are added for `[a]` and `NonEmpty a`.

---

###### Reviewer checklist


kore/src/SQL/SOP.hs
```diff
@@ -214,43 +234,45 @@ and tag columns are used as foreign keys into the constructor tables.
...
-createTableGeneric
-    :: forall proxy table
-    .  Typeable table
-    => (SOP.HasDatatypeInfo table, SOP.All2 Column (SOP.Code table))
-    => proxy table
-    -> SQL ()
-createTableGeneric proxy = createTableGenericAux (tableNameTypeable proxy) proxy
+    createTable'
+        :: forall fields
+        .  SOP.All SOP.Top fields
```
ana-pantilie, line 257: I noticed there are multiple constraints such as this in the code. Are they still necessary? Doesn't it mean we aren't imposing any real constraints on the `fields`?

ttuegel: Yes, they are necessary. This constraint means that `fields` is a type-level list of types `[Type]` with a known "shape", so that if we have a proxy type `Proxy fields`, we can construct any `fields`-based type, in particular `NP f fields` (the **N**-ary **P**roduct of `fields`).

ana-pantilie: Right, I see, thanks!

Issue #1731, opened 2020-04-03
Forbid injections into hooked sorts


ttuegel:
Sort injections into hooked sorts are invalid for the same reason that constructor declarations in hooked sorts are invalid. We should check this whenever we validate an axiom with the `subsort` attribute.

PR #1759, opened 2020-04-22
Added task for verifying definitions

traiansf:
This is a support task for Issue #1731 

---

###### Reviewer checklist


traiansf:
The idea of the new `verify` target is to trigger the (internal) verification of the definition.  This is done by calling the `kore-parser` with the `--verify` switch.

To allow this target to be run only for certain definitions, a file `<definition>.k.verify` must be created (can be empty) to trigger it.

PR #1762, opened 2020-04-23
Forbid injections into hooked sorts.

mcalancea:
Solves https://github.com/kframework/kore/issues/1731. 

###### Reviewer checklist


kore/src/Kore/ASTVerifier/AttributesVerifier.hs
```diff
@@ -132,6 +136,17 @@ verifyNoHookAttribute attributes = do
...
+    :: MonadError (Error VerifyError) error
+    => [Kore.Attribute.Sort.Sort]
+    -> error ()
+verifyNoHookedSupersort supersortsAtts = do
+    let isHooked = (getHasDomainValues . hasDomainValues) <$> supersortsAtts
+    case elem True isHooked of
+        False ->
+            return ()
+        True ->
+            koreFail "Hooked sorts cannot be supersorts."
```
ttuegel, line 149: We can write this more succinctly than the `case` statement using `when`:

```suggestion
    when (or isHooked) $ koreFail "Hooked sorts may not have subsorts."
```

I also suggest changing the error message slightly to emphasize that the problem is really with the subsorts. Declaring a subsort is like declaring a new constructor in the supersort. The hooked sort is something like `Int` or `Bool`; we don't allow the user to declare new constructors in these sorts.

PR #1778, opened 2020-04-29
Reject RewriteRule when lhs and rhs are the same

### LINKED ISSUE #1656: Reject semantic rules with the same left- and right-hand sides
The backend should reject rewrite rules with the same left- and right-hand sides. These rules always loop in the backend, and represent either a problem with the semantics or with the frontend.

andreiburdusa:
---
Fixes #1656 
###### Reviewer checklist


kore/src/Kore/Step/RulePattern.hs
```diff
@@ -873,3 +874,20 @@ instance UnifyingRule RewriteRule where
...
+  where
+    RulePattern { left, antiLeft, requires, rhs } =
+        getRewriteRule rewriteRule
+    lhs
+      | Just antiLeftTerm <- antiLeft
+      = TermLike.mkAnd
+            (TermLike.mkNot antiLeftTerm)
+            (TermLike.mkAnd (Predicate.unwrapPredicate requires) left)
+      | otherwise
+      = TermLike.mkAnd (Predicate.unwrapPredicate requires) left
```
ttuegel, line 893: As there is an `rhsToTerm` I think we should extract this into `lhsToTerm`.

kore/src/Kore/Step/RulePattern.hs
```diff
@@ -873,3 +874,20 @@ instance UnifyingRule RewriteRule where
     mapRuleVariables mapElemVar mapSetVar (RewriteRule rule) =
         RewriteRule (mapRuleVariables mapElemVar mapSetVar rule)
     {-# INLINE mapRuleVariables #-}
+
+lhsEqualsRhs
+    :: InternalVariable variable
+    => RewriteRule variable
```
ttuegel, line 880: This should only take a `RulePattern`.

kore/src/Kore/Exec.hs
```diff
@@ -641,6 +648,7 @@ initialize verifiedModule within = do
         simplifyToList rule = do
             simplified <- simplifyRuleLhs rule
             return (MultiAnd.extractPatterns simplified)
+    maybe (pure ()) errorRewriteRuleId (find lhsEqualsRhs rewriteRules)
```
ttuegel, line 651: `maybe (pure ())` is `traverse_`:

```suggestion
    find lhsEqualsRhs rewriteRules & traverse_ errorRewriteRuleId
```

kore/src/Kore/Log/ErrorRewriteRuleId.hs
```diff
@@ -0,0 +1,42 @@
...
+import Log
+
+newtype ErrorRewriteRuleId =
+    ErrorRewriteRuleId { getRule :: RewriteRule Variable }
+
+instance Pretty ErrorRewriteRuleId where
+    pretty errorRewriteRule =
+        Pretty.vsep
+            [ "Found semantic rule with the same left- and right-hand side:"
+            , Pretty.pretty . getRewriteRule . getRule $ errorRewriteRule
```
ttuegel, line 33: Let's print the original source location here, instead of the full rule. We should also add another line of explanation, such as:
```suggestion
            , Pretty.pretty . getRewriteRule . getRule $ errorRewriteRule
            , "Execution would not terminate when the rule applies."
```

kore/src/Kore/Log/ErrorRewriteRuleId.hs
```diff
@@ -0,0 +1,42 @@
...
+    ( RewriteRule
+    , getRewriteRule
+    )
+import Kore.Syntax.Variable
+    ( Variable
+    )
+
+import Log
+
+newtype ErrorRewriteRuleId =
```
ttuegel, line 26: I think the name of this error is slightly misleading because it suggests that the rewrite rule has an "identity" form, but actually the right-hand side is under `\next`. I would prefer if we called this `ErrorRewriteLoop`; the name suggests what the problem is.

PR #1789, opened 2020-05-02
Interpretation of function rules

ttuegel:
This documents the new interpretation of K function rules in Kore.

---

###### Reviewer checklist


traiansf:
Looks good to me.  Maybe an example for priority/owise rules would help.  I was initially surprised that the anti-left patterns only quantify over the original variables in the LHS, but it actually makes sense at a closer look.

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+
+This axiom does not imply `fun(A) = C` or `fun(B) = C`,
+and therefore cannot be used to evaluate `fun`.
+
+### Priority
+
+The `priority` attribute is not properly supported.
+The `owise` attribute is supported, but its implementation is inefficient:
+work is duplicated by re-checking that the other rules in the family do not apply.
+
```
grosu, line 96: Will this affect proofs later on, when we'll generate proof objects?  We may want to apply the priority/owise as kompile-time, to work with all backends.  We can add attributes if needed to tell backends "optimize".

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+## Background
+
+The family of K rules defining `fun`,
+
+```.k
+rule fun(φ₁ᵢ({Y}), φ₂ᵢ({Y}), ...) => ψᵢ({Y}) requires Preᵢ({Y}) ensures Postᵢ({Y})
+```
+
+is interpreted in Kore as
```
grosu, line 26: This way of writing it confused me when I first read this document.  I had to come back here to add this comment later, once I understood what was going on.  What I understood first was that this is how function rules are interpreted in Kore and that's it, nothing to do about it.  So I would say that that's how they are interpreted currently, as of May 2020, which is problematic, as explained below.

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+
+as shorthand for two rules,
+
+```.k
+rule fun(A) = C
+rule fun(B) = C
+```
+
+Under the current interpretation, this produces an axiom
+
```
grosu, line 83: Why?  The current interpretation does not talk about merging rules (btw, `=` should be `=>` in the rules above).

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+```.kore
+axiom fun(A ∨ B) = C.
+```
+
+This axiom does not imply `fun(A) = C` or `fun(B) = C`,
+and therefore cannot be used to evaluate `fun`.
+
```
grosu, line 90: Overall, it was not clear what the problem was above.  I know what's going on now from the slack discussion, but I would not figure it out from the text above only.

grosu: Maybe clarifying the previous comment will also fix this.

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+axiom \implies(Preᵢ({Y}) ∧ Argsᵢ({X}, {Y}) ∧ Prioᵢ({X}), (fun(X₁, X₂, ...) = ψᵢ({Y})) ∧ Postᵢ({Y}))
+```
+
+where
+
+```.kore
+Argsᵢ({X}, {Y}) = (X₁ ∈ φ₁ᵢ({Y})) ∧ (X₂ ∈ φ₂ᵢ({Y})) ∧ ...
+Prioᵢ({X}) = ∧ⱼ ¬ ∃ {Y}. Preⱼ({Y}) ∧ Argsⱼ({X}, {Y})
+    for all j that priority(rule j) < priority(rule i).
+```
```
grosu, line 117: The above is pretty tricky, I would add a sentence or two explaining what is going on.  Maybe also some generic/artificial example.  Also, I would emphasize that the key insight of the new interpretation is that it defines the function element-wise, which is likely what the user intended and solves all the problems due to partiality, including those we discussed on slack wrt `fun(A #or B)` (eg, when `A` or `B` is empty/undefined) --- I would also capture that discussion somewhere, to make it clear to first-time readers of this what the problem is, probably above when you discuss the `#or` problem?.

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+
+is interpreted in Kore as
+
+```.kore
+axiom \implies(X ∈ concatMap(elementMap(_1, _2), M:Map), sizeMap(X) = +Int(1, sizeMap(M))).
+```
+
+If `concatMap(_, _)` is undefined,
+then the pre-condition `X ∈ concatMap(_, _)` is not satisfied;
+therefore, the rule is sound.
```
grosu, line 137: ... (ie, the axiom is `true` and thus one cannot infer or use the equality as a consequence)

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,176 @@
...
+= \implies(X ∈ A, fun(X) = C) ∧ \implies(X ∈ B, fun(X) = C)
+```
+
+so that the original axiom is equivalent to two axioms, as intended:
+
+```.kore
+axiom \implies(X ∈ A, fun(X) = C)
+axiom \implies(X ∈ B, fun(X) = C).
+```
+
```
grosu, line 173: Nice!  My only comment here is that in order for the reader to appreciate this, he must first understand why the previous interpretation didn't work well.  I added a comment about that above.

design-decisions/2020-05-02-function-rules.md
ttuegel: >  We may want to apply the priority/owise as kompile-time, to work with all backends.

The new encoding of function rules would require us to apply priority at kompile-time, whereas the current encoding does not allow it.

ttuegel:
@traiansf I added an example of priority rules, as you requested. Could you take a look?

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,261 @@
...
+which is intended to be equivalent to the pair of rules above.
+Under the current interpretation, this rule produces an axiom,
+
+```.kore
+axiom \implies(\top, (fun(A ∨ B) = C) ∧ \top)
+```
+
+which is **not** equivalent to the first interpretation.
+Specificalyl, the first interpretation is satisfied if and only if
```
ana-pantilie, line 101:
```suggestion
Specifically, the first interpretation is satisfied if and only if
```

design-decisions/2020-05-02-function-rules.md
```diff
@@ -0,0 +1,261 @@
...
+        \and(
+            _<=_<_(X₁, X₂, X₃) = false,
+            \top
+        )
+    )
+axiom
+    \implies(
+        \and(
+            \top,
+            \top,
```
ana-pantilie, line 242: I'm reading this as `\and(\top, \top, ...)`. Should one of the `\top` be removed?

ttuegel: No, there will be three pre-conditions now: the actual `requires`, the `Args`, and the `Prio`. Only the `Prio` can be omitted if the rule has the lowest priority (that is, if there are no other rules to apply first).

PR #1792, opened 2020-05-05
Remove Kore.Internal.SideCondition.andCondition

### LINKED ISSUE #1733: Audit: Remove Kore.Internal.SideCondition.andCondition
The function `andCondition` exists to create de-normalized `SideConditions`. De-normalized `SideConditions` should be forbidden, and therefore this function should not exist.

andreiburdusa:
---
Fixes #1733 
###### Reviewer checklist


kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -161,12 +161,13 @@ andCondition
     => SideCondition variable
     -> Condition variable
     -> SideCondition variable
-andCondition SideCondition {assumedTrue} newCondition =
-    SideCondition
+andCondition initial@SideCondition {assumedTrue} newCondition =
+    if isNormalized result then result else initial
```
ttuegel, line 165:
```suggestion
    assert (isNormalized result) result
```

andreiburdusa:
What I learned from running [`functions`](https://github.com/kframework/kore/tree/master/test/functions)`/positive-spec.k` in the repl, both before and after removing `andCondition`:

before

after

Both go to node 1 the same way, using [Axiom 0](https://github.com/kframework/kore/blob/master/test/functions/functions.k#L8).

Then, from node 1:
```
  <T>
    <k> next true </k>
    <n> N:Int </n>
  </T>
#And
  { N >Int 0 #Equals true }
```
both use `Claim 0`:
```
  #Ceil ( positive ( N ) )
#And
  <T>
    <k> next positive ( N ) </k>
    <n> N:Int </n>
  </T>
```
`→ ◇w`
```
<T>
  <k> end </k>
  <n> N:Int </n>
</T>
```

But instead of getting node 2 with the configuration
```
  <T>
    <k> end </k>
    <n> N:Int </n>
 </T>
#And { N >Int 0 #Equals true }
```

After removing `andCondition` it gets node 2:
```
  #Ceil ( positive ( N ) )
#And
  <T>
    <k> end </k>
    <n> N:Int </n>
  </T>
#And { N >Int 0 #Equals true }
#And { true #Equals positive ( N ) }
```

I think the difference is in applying the [rule](https://github.com/kframework/kore/blob/master/test/functions/functions.k#L12)
```
rule positive(M) => true
  requires M >Int 0
```

ttuegel:
We should check the call sites of `applyEquation` to see what `SideCondition` it is called with.

ttuegel:
> We should check the call sites of `applyEquation` to see what `SideCondition` it is called with.

Sorry, I meant `attemptEquation` here.

andreiburdusa:
> > We should check the call sites of `applyEquation` to see what `SideCondition` it is called with.
> 
> Sorry, I meant `attemptEquation` here.

Is this the `SideCondition` we are looking for?
```
\and{_PREDICATE{}}(
    /* term: */
    /* D Sfa */ \top{_PREDICATE{}}(),
\and{_PREDICATE{}}(
    /* predicate: */
    /* Sfc */
    \ceil{SortBool{}, _PREDICATE{}}(
        /* Fn Sfc */
        Lblpositive'LParUndsRParUnds'FUNCTIONS'Unds'Bool'Unds'Int{}(
            /* Fl Fn D Sfa */ VarN:SortInt{}
        )
    ),
    /* substitution: */
    /* Spa */
    \equals{SortGeneratedCounterCell{}, _PREDICATE{}}(
        /* Fl Fn D Sfa */ VarDotVar0:SortGeneratedCounterCell{},
        /* Fl Fn D Sfa */
        Lbl'-LT-'generatedCounter'-GT-'{}(
            /* Fl Fn D Sfa */ VarDotVar00:SortInt{}
        )
    )
))
```

It's passed to `attemptEquation` inside `definitionEvaluation`

ttuegel:
The problem is because we do not pass in a `SideCondition` when applying a semantic rule. The constraint

```
true #Equals positive ( N )
```

originates when we apply the claim `next positive(N) => end` to the configuration `next true`. We need to use the top-level `SideCondition`

```
N >Int 0 #Equals true
```

when we simplify the `positive` constraint. The backend does not yet support re-simplifying the top-level `SideCondition` with itself, so we have to complete the simplification of `positive` _before_ we add that constraint to the top-level `SideCondition` (that is: before we finish applying the claim).

andreiburdusa:
Now there is only one unit test failing.

The reason is that the rule
```
claim {}
/* Spa */
    \implies{testSort{}}(
        /* Spa */
        \and{testSort{}}(
            /* Spa */
            \equals{testSort{}, testSort{}}(
                /* Fl Fn D Sfa */ x:testSort{},
                /* Fl Fn D Sfa Cl */ a{}()
            ),
            /* Fl Fn D Sfa */ x:testSort{}    <---- this is the only difference
        ),
        /* Spa */
        weakExistsFinally{testSort{}}(
            /* Spa */
            \and{testSort{}}(
                /* Spa */
                \not{testSort{}}(
                    /* Spa */
                    \equals{testSort{}, testSort{}}(
                        /* Fl Fn D Sfa */ x:testSort{},
                        /* Fl Fn D Sfa Cl */ a{}()
                    )
                ),
                /* Fl Fn D Sfa */ x:testSort{}
            )
        )
    )
[]
```
becomes 
```
    claim {}
    /* Spa */
        \implies{testSort{}}(
            /* Spa */
            \and{testSort{}}(
                /* Spa */
                \equals{testSort{}, testSort{}}(
                    /* Fl Fn D Sfa */ x:testSort{},
                    /* Fl Fn D Sfa Cl */ a{}()
                ),
                /* Fl Fn D Sfa Cl */ a{}()    <---- this is the only difference
            ),
            /* Spa */
            weakExistsFinally{testSort{}}(
                /* Spa */
                \and{testSort{}}(
                    /* Spa */
                    \not{testSort{}}(
                        /* Spa */
                        \equals{testSort{}, testSort{}}(
                            /* Fl Fn D Sfa */ x:testSort{},
                            /* Fl Fn D Sfa Cl */ a{}()
                        )
                    ),
                    /* Fl Fn D Sfa */ x:testSort{}
                )
            )
        )
    []
```
when given as an argument to [`runOnePathSteps`](https://github.com/kframework/kore/blob/master/kore/test/Test/Kore/Strategies/OnePath/Step.hs#L773)

I think `requires` is applied to `left`.

kore/src/Kore/Step/Simplification/SubstitutionSimplifier.hs
```diff
@@ -195,9 +194,7 @@ simplifyAnds MakeAnd { makeAnd } (NonEmpty.sort -> patterns) =
             AndF And { andFirst, andSecond } ->
                 foldM simplifyAnds' intermediate [andFirst, andSecond]
             _ -> do
-                let sideCondition =
-                        SideCondition.topTODO
-                        `SideCondition.andCondition` intermediateCondition
+                let sideCondition = SideCondition.topTODO
```
ttuegel, line 197: This is fine for now, but let's make a follow-up pull request to add a `SideCondition` argument to `simplifyAnds` and remove this `topTODO`.

Issue #1801, opened 2020-05-07
Errors appear twice in the log


ttuegel:
Errors appear twice in the log: first the actual error entry appears, then this:

```
kore-exec: [2020-05-06 14:42:39.976271596] Error (ErrorException):
    Error! Please report this.
    CallStack (from HasCallStack):
      error, called at src/Kore/Log/ErrorRewritesInstantiation.hs:169:5 in kore-0.20.0.0-7yFXMqqWkjaHW436PDhtRm:Kore.Log.ErrorRewritesInstantiation
```

We should combine the `CallStack` with the actual error entry, and omit the second entry.

PR #1822, opened 2020-05-15
Handle error entry types as exceptions

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1801

---

###### Reviewer checklist


kore/app/exec/Main.hs
```diff
@@ -416,6 +425,18 @@ mainWithOptions execOptions = do
     KoreExecOptions { koreSearchOptions } = execOptions
     KoreExecOptions { koreMergeOptions } = execOptions
 
+    handleErrorRewriteLoop
+        :: ErrorRewriteLoop -> Main ExitCode
+    handleErrorRewriteLoop entry = do
+        logEntry entry
+        return $ ExitFailure 1
```
ttuegel, line 432: We will only need one handler here, if we make sure that our error entries are always wrapped in `SomeEntry`:

```suggestion
    handleSomeEntry
        :: SomeEntry -> Main ExitCode
    handleSomeEntry (SomeEntry entry) = do
        logEntry entry
        return $ ExitFailure 1
```

kore/src/Kore/Log/ErrorRewriteLoop.hs
```diff
@@ -27,23 +38,32 @@ import Kore.Syntax.Variable
...
-newtype ErrorRewriteLoop =
-    ErrorRewriteLoop { getRule :: RewriteRule Variable }
+data ErrorRewriteLoop =
+    ErrorRewriteLoop
+        { rule :: RewriteRule Variable
+        , errorCallStack :: CallStack
+        }
+    deriving (Show)
+
+instance Exception ErrorRewriteLoop
```
ttuegel, line 48: We can use a custom `Exception` instance to ensure that the entry is always wrapped:

```suggestion
instance Exception ErrorRewriteLoop where
    toException = toException . SomeEntry
    fromException exn = fromException exn >>= \(SomeEntry entry) -> pure entry
```

We will also have to add `instance Exception SomeEntry` where `SomeEntry` is defined.

kore/src/Kore/Log/ErrorRewriteLoop.hs
```diff
@@ -27,23 +38,32 @@ import Kore.Syntax.Variable
...
-        Pretty.vsep
+    pretty ErrorRewriteLoop { rule, errorCallStack } =
+        Pretty.vsep $
             [ "Found semantic rule with the same left- and right-hand side at:"
             , Pretty.pretty
-                . sourceLocation . attributes . getRewriteRule . getRule
-                $ errorRewriteRule
+                . sourceLocation . attributes . getRewriteRule $ rule
             , "Execution would not terminate when the rule applies."
+            , "Aborting execution."
```
ttuegel, line 57: We can remove this. It was only there because I didn't know what else to pass to `error`. :smile: 

```suggestion
```

Issue #1807, opened 2020-05-08
Interpretation of function rules


ttuegel:
We must implement the changes to function rules described in #1789.

Although we will now recognize the priority form of equation rules, we will probably not implement full support yet.

---

- [ ] Add fields to `Equation` to hold the `argument` and `antiLeft` predicates.
- [ ] Extract the `argument` and `antiLeft` patterns from the definition when reading the equation.
- [ ] Add a step after matching the `Equation` that simplifies the `argument` predicate and adds it to the match result before we check the match result.
- [ ] Duplicate the function tests in `kore-test` so that we run all tests with the old and new form of the equations.

PR #1829, opened 2020-05-18
Equation: support new interpretation of function rules

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1807

---

###### Reviewer checklist


kore/src/Kore/Equation/Application.hs
```diff
@@ -174,6 +179,28 @@ attemptEquation sideCondition termLike equation =
...
+    applyAndSelectMatchResult results =
+        Foldable.foldr1
+            takeFirstSuccess
+            ( whileApplyMatchResult
+            . applyMatchResult equationRenamed
+            <$> results
+            )
+    takeFirstSuccess first second = do
+        lift (runExceptT first)
+        >>= either (const second) (const first)
```
ttuegel, line 202: This is not quite right: it will call `first` twice if it succeeds. What we really want to do is:
```suggestion
    takeFirstSuccess first second = catchError first (const second)
```

kore/src/Kore/Equation/Application.hs
```diff
@@ -516,6 +594,8 @@ data ApplyMatchResultError variable
     -- term was required.
     | NotMatched (SomeVariableName variable)
     -- ^ The variable was not matched.
+    | NonTargetSubstitution (SomeVariableName variable)
```
ttuegel, line 597: I think we should call this `NonMatchingSubstitution` instead because having this substitution means that the solution is not a matching solution (and the documentation should say that).

kore/src/Kore/Equation/Application.hs
```diff
@@ -148,11 +156,17 @@ attemptEquation
...
-            & whileApplyMatchResult
+            if isTop argument
+                then do
+                    matchResult <- match left termLike & whileMatch
+                    applyMatchResult equationRenamed matchResult
+                        & whileApplyMatchResult
+                else
+                    applyAndSelectMatchResult
+                    =<< simplifyArgumentWithResult argument antiLeft
+                    =<< whileMatch (match left termLike)
```
ttuegel, line 169: Can we refactor this so that it is clear that both branches call `match` first? And, let's wrap the entire block in `whileMatch`, because `simplifyArgumentWithResult` is a sort of generalized matching procedure.

kore/src/Kore/Equation/Application.hs
```diff
@@ -176,6 +210,40 @@ attemptEquation sideCondition termLike equation =
...
+        -> MatchResult (Target variable)
+        -> t simplifier [MatchResult (Target variable)]
+    simplifyArgumentWithResult
+        argument
+        antiLeft
+        (matchPredicate, matchSubstitution)
+      =
+        -- TODO: toMap is unsafe;
+        -- is it impossible to happen (if the frontend generates equations
+        -- correctly), or should we add a new AttemptEquationError case?
```
ttuegel, line 226: Yes, this is impossible, but let's add a `HasCallStack` constraint so that we get a reasonable message if it ever fails.

kore/src/Kore/Equation/Application.hs
```diff
@@ -176,6 +210,40 @@ attemptEquation sideCondition termLike equation =
...
+
+    mkSubstitution =
+        Substitution.fromMap
+        . Map.fromList
+        . fmap
+            (\(variableName, term)->
+                let variableSort = TermLike.termLikeSort term
+                 in (Variable { variableName, variableSort }, term)
+            )
+        . Map.toList
```
ttuegel, line 245: This is just a generic conversion from `Map` to `Substitution`, right? Could we move this to an instance of `From` in `Kore.Internal.Substitution`?

We can also use `Map.toAscList` and `Map.fromAscList` here, because this operation can't change the order of the variables.

kore/src/Kore/Equation/Application.hs
```diff
@@ -250,6 +319,15 @@ applyMatchResult equation matchResult@(predicate, substitution) = do
       | otherwise
       = empty
 
+    checkNonTargetVariables
+      | Just variable <-
+          Foldable.find
+            Target.isSomeNonTargetName
+            (Map.keys substitution)
+      = [NonTargetSubstitution variable]
```
ttuegel, line 327: Can we generate an error here for _all_ `NonTarget` substitutions?

kore/src/Kore/Equation/Equation.hs
```diff
@@ -82,6 +82,8 @@ import qualified SQL
 
 data Equation variable = Equation
     { requires :: !(Predicate variable)
+    , argument :: !(Predicate variable)
```
ttuegel, line 85: Should this be a `Maybe`, like `antiLeft`?

kore/src/Kore/Equation/Simplification.hs
```diff
@@ -68,20 +68,31 @@ simplifyEquation
...
+                , ensures
+                , attributes } = equation
         simplified <- simplifyTermLike left >>= expectSingleResult
         let Conditional { term, predicate, substitution } = simplified
         Monad.guard (isTop predicate)
         let subst = Substitution.toMap substitution
             left' = TermLike.substitute subst term
             requires' = TermLike.substitute subst <$> requires
+            argument' = TermLike.substitute subst <$> argument
+            antiLeft' = (fmap . fmap ) (TermLike.substitute subst) antiLeft
```
ttuegel, line 88:
```suggestion
            antiLeft' = (fmap . fmap) (TermLike.substitute subst) antiLeft
```

kore/test/Test/Kore/Equation/Application.hs
```diff
@@ -421,6 +588,45 @@ axiom_
     -> Equation'
 axiom_ left right = axiom left right (makeTruePredicate sortR)
 
+functionAxiomNEW
```
ttuegel, line 591: Let's call this `functionAxiomUnification`, to emphasize that it constructs a function axiom using unification for the arguments.

kore/test/Test/Kore/Equation/Sentence.hs
```diff
@@ -48,6 +48,24 @@ test_fromSentenceAxiom =
                     , ensures = wrapPredicate ensures
                     }
         assertions original equation
+    , testCase "New equation form: ⌈f(x))⌉ → f(x) = g(x) ∧ ⌈h(x)⌉" $ do
```
ttuegel, line 51:
```suggestion
    , testCase "New equation form: ⌈f(x)⌉∧⌈y ∈ x⌉ → f(y) = g(x) ∧ ⌈h(x)⌉" $ do
```

test/set/set-iteration/test-2-spec.k
```diff
@@ -14,5 +14,8 @@ module TEST-2-SPEC
 
   rule <k> runTest( f( SetItem(X:Element) SetItem(Y:Element) ) ) </k>
     => <k> doneTest( SetItem(g(X)) SetItem(g(Y)) ) </k>
+    requires
+      (X =/=K Y)
+      andBool (g(X) =/=K g(Y))
```
ttuegel, line 19: Why is this needed?

test/set/set-iteration/test.k
```diff
@@ -5,13 +5,13 @@ module TEST
 
   syntax Element ::= "a" | "b" | "c"
 
-  syntax Element ::= g(Element) [function]
+  syntax Element ::= g(Element) [function, functional]
```
ttuegel, line 8: Why is this needed?

test/set/set-iteration/test-2-spec.k
ana-pantilie: This is needed to ensure the set is defined. Before, when we used matching, this condition didn't need to be true.

kore/src/Kore/Equation/Equation.hs
ana-pantilie: Yes, you're right, thanks for noticing this!

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -963,6 +968,246 @@ test_simplificationIntegration =
         assertEqual "" expected actual
     ]
 
+test_simplificationIntegrationNEW :: [TestTree]
```
ana-pantilie, line 971: TODO: rename this and other new test suites

kore/src/Kore/Equation/Application.hs
```diff
@@ -168,6 +188,26 @@ attemptEquation sideCondition termLike equation =
...
+        :: [MatchResult (Target variable)]
+        -> ExceptT
+            (AttemptEquationError variable)
+            simplifier
+            (Equation variable, Predicate variable)
+    applyAndSelectMatchResult [] =
+        whileApplyMatchResult
+        $ applyMatchResult
+            equationRenamed
+            (Predicate.makeFalsePredicate_, mempty)
```
ttuegel, line 201: This case should be handled as if `matchIncremental` had returned `Nothing`, by throwing `MatchError`. `applyMatchResult` does not handle the `\bottom` predicate in any special way, so if unification of the arguments fails, then we would call `checkRequires` without instantiating the equation. This might account for the performance problem you found with `wasm-semantics`.

kore/src/Kore/Equation/Application.hs
```diff
@@ -168,6 +188,26 @@ attemptEquation sideCondition termLike equation =
...
+    applyAndSelectMatchResult results =
+        Foldable.foldr1
+            takeFirstSuccess
+            ( whileApplyMatchResult
+            . applyMatchResult equationRenamed
+            <$> results
+            )
```
ttuegel, line 208:
```suggestion
        whileApplyMatchResult $ Foldable.foldr1
            takeFirstSuccess
            (applyMatchResult equationRenamed <$> results)
```

kore/src/Kore/Equation/Application.hs
```diff
@@ -176,6 +216,28 @@ attemptEquation sideCondition termLike equation =
         debugAttemptEquationResult equation result
         return result
 
+    simplifyArgumentWithResult
+        :: HasCallStack
+        => MonadTrans t
+        => Predicate (Target variable)
+        -> Maybe (Predicate (Target variable))
+        -> MatchResult (Target variable)
+        -> t simplifier [MatchResult (Target variable)]
```
ttuegel, line 225: Isn't `simplifyArgumentWithResult` called only in a particular instance of `t`? In that case, we should not make `t` a variable here, because it may prevent specialization. It would be enough just to remove the signature.

kore/src/Kore/Equation/Application.hs
```diff
@@ -176,6 +216,28 @@ attemptEquation sideCondition termLike equation =
...
+      =
+        lift $
+            let toMatchResult Conditional { predicate, substitution } =
+                    (predicate, Substitution.toMap substitution)
+             in Substitution.mergePredicatesAndSubstitutions
+                    sideCondition
+                    ([argument, matchPredicate] <> maybeToList antiLeft)
+                    [from @_ @(Substitution _) matchSubstitution]
+                    & Logic.observeAllT
+                    & (fmap . fmap) toMatchResult
```
ttuegel, line 239:
```suggestion
        lift $ do
            let toMatchResult Conditional { predicate, substitution } =
                    (predicate, Substitution.toMap substitution)
            Substitution.mergePredicatesAndSubstitutions
                sideCondition
                ([argument, matchPredicate] <> maybeToList antiLeft)
                [from @_ @(Substitution _) matchSubstitution]
                & Logic.observeAllT
                & (fmap . fmap) toMatchResult
```

kore/src/Kore/Equation/Application.hs
ana-pantilie: Thanks for finding this! I hadn't thought this was wrong. I'm unable to revisit the wasm tests currently because of the various frontend issues, though.

PR #1830, opened 2020-05-18
Respect the identity of variables, Part 1

### LINKED ISSUE #1545: Respect the identity of variables
In `Kore.Syntax.Variable`, we have this definition of a variable:

```.hs
data Variable = Variable
    { variableName :: !Id
    , variableCounter :: !(Maybe (Sup Natural))
    , variableSort :: !Sort
    }
```

`Eq` and `Ord` are defined as if this is a _value_ type (in domain-driven design terminology). In fact, if the `variableName` and `variableCounter` of two variables are the same, they are regarded as the same variable; then if their `variableSort` differs, they are not _different_, but _inconsistent_. That is, `Variable` should be an _entity_ type (domain-driven design terminology again) whose unique identifier is the combination of `variableName` and `variableCounter`.

If we move `variableSort` out of `Variable`, then `FreshPartialOrd` (introduced in #1539) becomes a simple upper semilattice. Alternatively, we might combine `variableName` and `variableCounter` into a `VariableName` type, and define an upper semilattice on this type only (instead of a class) and provide optics for renaming other variable types.

ttuegel:
This pull request begins to fix #1545. Most of the changes in this pull request are simply consequences of the changes in [`Kore.Syntax.Variable`](https://github.com/kframework/kore/pull/1830/files#diff-f297c554f543b5360a2cdb2d2f960203), so I recommend starting there.

### Summary

- Added type `Variable1` which carries a variable name and sort, representing an occurrence of a variable in a Kore pattern. Unlike `Variable`, this type is parameterized in the variable name type.
- Added a `VariableName` type, representing the basic syntax of variable names.
- Added `ElementVariableName` and `SetVariableName` wrappers.
- Added `SomeVariableName`, which may be an `ElementVariableName` or a `SetVariableName`.
- Added `AdjSomeVariableName`, which represents mappings preserving the element- or set-ness of a variable. The name of the type owes to the _adjunction_ between `SomeVariableName` and `AdjSomeVariableName`. This is not as complicated as it sounds, but it bears close attention. Most of the changes in the pull request are related to introducing this type.
- Added class `NamedVariable` for variables, representing the isomorphism between `_ Variable` and `Variable1 _`.

In the follow-up to this pull request, `Variable1 _` will replace `_ Variable` and the `NamedVariable` class will be obsolete. I wanted to get this in as a separate pull request because I think adding `AdjSomeVariableName` is significant in its own right.

---

###### Reviewer checklist


kore/src/Kore/Equation/Equation.hs
```diff
@@ -226,22 +228,44 @@ refreshVariables
...
-        mapElemVars elemVar =
-            case Map.lookup (ElemVar elemVar) rename of
-                Just (ElemVar elemVar') -> elemVar'
-                _ -> elemVar
-        mapSetVars setVar =
-            case Map.lookup (SetVar setVar) rename of
-                Just (SetVar setVar') -> setVar'
-                _ -> setVar
-        subst = TermLike.mkVar <$> rename
+    let rename =
```
traiansf, line 231: I can't help notice that the complexity of the code has grown by a non-trivial margin. From 10 lines with easy-to-read-and-write code, we got to more than 30 of code with lenses

ttuegel: Yes, it's unfortunate, but temporary: the lenses are a consequence of switching between representations of variables, and after I finish the next part of the changes, then all the lenses go away.

PR #1831, opened 2020-05-19
\in simplifier should evaluate \ceil of \and

ana-pantilie:
---

###### Reviewer checklist


kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -521,14 +520,11 @@ test_simplificationIntegration =
...
-                            (makeInPredicate Mock.boolSort
-                                (mkCeil Mock.setSort Mock.cf)
-                                (mkCeil Mock.setSort Mock.cg)
-                            )
-                            (makeInPredicate_
-                                (mkCeil Mock.setSort Mock.cf)
-                                (asInternal (Set.fromList []))
+                            (makeAndPredicate
+                                (makeCeilPredicate Mock.boolSort Mock.cf)
+                                (makeCeilPredicate Mock.boolSort Mock.cg)
```
ttuegel, line 525: I don't think this is right. It should be `makeCeilPredicate _ (mkAnd Mock.cf Mock.cg)`, otherwise there is nothing that constrains `cf` and `cg` with respect to each other.

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -557,12 +553,26 @@ test_simplificationIntegration =
...
+                                    )
+                                )
+                                (makeAndPredicate
+                                    (makeCeilPredicate Mock.otherSort
+                                        (Mock.tdivInt mkTop_ mkTop_)
+                                    )
+                                    (makeCeilPredicate Mock.otherSort
+                                        (mkNu iz (Mock.builtinInt 595))
+                                    )
+                                )
```
ttuegel, line 572: This is redundant. The first condition ensures that each of these subterms is defined. Do you know where this is coming from?

kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -586,17 +596,30 @@ test_simplificationIntegration =
...
```
ttuegel, line 619: This is redundant, as above.

kore/test/Test/Kore/Step/Simplification/Integration.hs
ana-pantilie: I was under the impression that `ceil(\and(a, b))` will also generate `\and(\ceil(a), \ceil(b))` if it can't infer that `a` and `b` are defined: https://github.com/kframework/kore/blob/dceca511119b17383c6f07e98914a30e4815de32/kore/src/Kore/Step/Simplification/Ceil.hs#L350 Am I misunderstanding this? I checked for this example and the terms appear in the `childCeils`.

kore/test/Test/Kore/Step/Simplification/Integration.hs
ana-pantilie: Isn't `\ceil(\and(\ceil(a), \ceil(b)))` equivalent to `\and(\ceil(a), \ceil(b))`?

ttuegel: Yes, you're right: I mis-read the original as `\in(cf, cg)` not `\in(\ceil(cf), \ceil(cg))`.

kore/test/Test/Kore/Step/Simplification/Integration.hs
ttuegel: OK, it looks like this is doing what is intended.

PR #1859, opened 2020-06-02
TermLike: Remove externalizeFreshVariables

ttuegel:
Any numeral suffix of a variable name is used to initialize the internal counter. That makes it impossible to have a variable name collision during unparsing, so `externalizeFreshVariables` is no longer needed.

---

###### Reviewer checklist


ttuegel:
@ana-pantilie This should fix the bug you were seeing, where variable names were unparsed inconsistently in an error message.

PR #1864, opened 2020-06-03
Remove hie.yaml

ttuegel:
hie-bios is now smart enough to infer the correct settings with Stack using
haskell-ide-engine or ghcide. A multi-component configuration for Cabal is
provided at hie-cabal.yaml; the user can link it to hie.yaml if desired.

---

###### Reviewer checklist


ttuegel:
@ana-pantilie If you are using HIE or ghcide, could you check that the updated instructions in `README.md` work?

ana-pantilie:
> @ana-pantilie If you are using HIE or ghcide, could you check that the updated instructions in `README.md` work?

I think it's breaking on the test files. The error is `cannot satisfy package -hedgehog-1.0.2`.

ttuegel:
I think that Stack doesn't install dependencies when it gets run by the language server, so you may need to run `stack build --test --bench --only-dependencies` before it will work. If that installs the missing dependencies, I will add it to the instructions.

PR #1867, opened 2020-06-04
Add case to instance Synthetic ConstructorLike (Application Symbol)

### LINKED ISSUE #1817: Overloaded constructors are not always constructor-like
Overloaded constructors are not constructor-like if they appear under a sort injection, but they are currently treated as other constructors.

andreiburdusa:
---
Fixes #1817 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
+++ b/kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
@@ -95,6 +95,11 @@ instance Synthetic ConstructorLike (Application Symbol) where
         , childrenAreConstructorLike =
             ConstructorLike . Just $ ConstructorLikeHead
 
+        | Symbol.isConstructor symbol
+        , childrenAreConstructorLike
+        , childrenAreNotSortInjections =
+            ConstructorLike . Just $ SortInjectionHead
+
         -- Checks that the children of a sort injection are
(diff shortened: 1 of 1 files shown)


kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
```diff
@@ -95,6 +95,11 @@ instance Synthetic ConstructorLike (Application Symbol) where
         , childrenAreConstructorLike =
             ConstructorLike . Just $ ConstructorLikeHead
 
+        | Symbol.isConstructor symbol
```
ttuegel, line 98: This should only apply when the symbol is an _overloaded_ constructor.

kore/src/Kore/Attribute/Pattern/ConstructorLike.hs
```diff
@@ -95,6 +95,11 @@ instance Synthetic ConstructorLike (Application Symbol) where
...
+        , childrenAreConstructorLike
+        , childrenAreNotSortInjections =
```
ttuegel, line 100: When the constructor is overloaded, the children are _expected_ to be sort injections.

ttuegel:
I'm going to go ahead and close this because it looks like we cannot do it right now.

PR #1870, opened 2020-06-05
Simplification rule for in .Set

### LINKED ISSUE #1862: Simplify SET.in
We should add a builtin simplification rule equivalent to

```.k
rule X in .Set => false
```

This rule will help keep term sizes small. This simplification rule must be built-in because the frontend generates more `Set` sorts than just those found in `domains.k`.

traiansf:
Fixes #1862 

---

###### Reviewer checklist


kore/test/Test/Kore/Builtin/Set.hs
```diff
@@ -149,10 +150,17 @@ import Test.SMT hiding
...
 
 genKeys :: Gen [TermLike VariableName]
-genKeys = Gen.subsequence (concreteKeys <> symbolicKeys)
+genKeys = Gen.subsequence (concreteKeys <> symbolicKeys <> functionalKeys)
 
 genKey :: Gen (TermLike VariableName)
-genKey = Gen.element (concreteKeys <> symbolicKeys)
+genKey = Gen.element (concreteKeys <> symbolicKeys <> functionalKeys)
+
+
```
ttuegel, line 158:
```suggestion
```

PR #1884, opened 2020-06-12
Kore.Exec.exec: Do not explicitly construct the execution graph

### LINKED ISSUE #1875: Do not keep execution graph in memory during execution
During ordinary concrete and symbolic execution (not search) we should not keep the execution graph in memory. This accounts for some of the slow tests in #1349.

See also: #1795

ttuegel:
Improves concrete execution performance by up to 15%.

Fixes #1875 

---

###### Reviewer checklist


kore/src/Kore/Exec.hs
```diff
@@ -220,15 +212,36 @@ exec
...
-        let
-            Execution { executionGraph } = execution
-            finalConfig = pickLongest executionGraph
-            finalTerm =
-                forceSort patternSort
-                $ Pattern.toTermLike finalConfig
+        initialized <- initialize verifiedModule
+        let Initialized { rewriteRules } = initialized
+        finalConfig <-
+            getFinalConfigOf $ do
```
andreiburdusa, line 218:
```suggestion
        (_, finalConfig) <-
            Logic.observeT $ do
```

ttuegel: [`observeT`](https://hackage.haskell.org/package/logict-0.7.0.2/docs/Control-Monad-Logic.html#v:observeT) invokes `fail` if there is not at least one result, but I can change this to be somewhat more idiomatic.

PR #1885, opened 2020-06-12
Add Kore.Log.DebugUnification

ttuegel:
---

###### Reviewer checklist


ttuegel:
I cannot update the pull request description, so here it is:

This pull request adds a `DebugUnification` log entry type and restricts unification logging to the cases which are actually interesting, namely:

1. The solution of the entire unification problem, or
1. The exact sub-problem which could not be solved.

Unification logging is only enabled for _unification_, not for equality or conjunction simplification, because there are too many expected unsolvable cases in those for logging to be useful.

This is important for performance: In actual programs and proofs, I have seen speed-up of 15-35%, and the unit tests are 70% faster.

---

###### Reviewer checklist

kore/src/Log.hs
```diff
@@ -233,18 +227,21 @@ instance MonadLog log => MonadLog (ReaderT a log)
...
 
 newtype LoggerT m a =
     LoggerT { getLoggerT :: ReaderT (LoggerEnv m) m a }
     deriving (Functor, Applicative, Monad)
     deriving (MonadIO, MonadThrow, MonadCatch, MonadMask)
 
+data LoggerEnv monad =
+    LoggerEnv
+        { logAction :: !(LogAction monad ActualEntry)
+        , context :: ![SomeEntry]
```
ana-pantilie, line 241: Why is the context a list now?

kore/src/Log.hs
```diff
@@ -266,12 +263,11 @@ instance Monad m => MonadLog (LoggerT m) where
         lift $ someLogAction <& toEntry entry
     {-# INLINE logEntry #-}
 
-    logWhile !entry2 action = do
+    logWhile entry2 action = do
```
ana-pantilie, line 266: Why is `entry2` no longer strict?

kore/src/Log/Entry.hs
```diff
@@ -92,15 +88,14 @@ entryTypeText (SomeEntry entry) =
 data ActualEntry =
     ActualEntry
         { actualEntry :: !SomeEntry
-        , entryContext :: !(Seq SomeEntry)
+        , entryContext :: ![SomeEntry]
```
ana-pantilie, line 91: Same question as above.

kore/src/Kore/Log/DebugUnification.hs
```diff
@@ -0,0 +1,118 @@
...
+import Log
+import Pretty
+    ( Pretty
+    )
+import qualified Pretty
+
+data DebugUnification
+    = DebugUnificationWhile !WhileDebugUnification
+    | DebugUnificationSolved UnificationSolved
+    | DebugUnificationUnsolved !UnificationUnsolved
```
andreiburdusa, line 31: How did you decide which fields are strict and which are not?

ttuegel: The fields that are `data` are strict, but the `newtype` field is not. The `data` types themselves have lazy fields, so if they were bound lazily here, there would be two levels of laziness; binding them strictly here makes sure there is only one level of indirection. `newtype` doesn't have lazy or strict fields, so the lazy binding at `DebugUnificationSolved` is the only level of laziness.

This reminds me that I need to finish my laziness-by-example blog post. :sweat_smile:

kore/src/Log.hs
ttuegel: The list has better sharing. When we add an entry to the context, the `tail` of the list is always the old context. `Seq` uses a tree internally, so every 2-3 new entries, it allocates a new branch of the tree which isn't shared.

kore/src/Log/Entry.hs
ttuegel: The `ActualEntry` changes because the `LoggerEnv` needs to change, to avoid converting between types when constructing the `ActualEntry`.

kore/src/Log.hs
ttuegel: We want the entries to be evaluated lazily. This probably wasn't a problem in practice, because all the entries seem to have lazy fields, but it is just a precaution.

PR #1881, opened 2020-06-11
Remove UnificationError

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1767

---

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`

PR #1891, opened 2020-06-17
Use unification in removeDestination

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1412

~Needs https://github.com/kframework/kore/pull/1881~

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -950,53 +952,31 @@ withConfiguration configuration =
...
-            error . show . Pretty.vsep $
-            [ "Unifying the terms of the configuration and the\
-            \ destination has unexpectedly produced more than one\
-            \ unification case."
-            , Pretty.indent 2 "Unification cases:"
-            ]
-            <> fmap
-                (Pretty.indent 4 . unparse)
-                (Foldable.toList unifiedConfigs)
+        simplifyConditionalTermToOr sideCondition (mkIn_ configTerm destTerm)
```
ttuegel, line 972: We should use the correct result sort here:

```suggestion
        simplifyConditionalTermToOr sideCondition (mkIn configSort configTerm destTerm)
```

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1008,6 +988,15 @@ removalPredicate
           , Pretty.indent 4 (unparse destTerm)
           ]
   where
+    evaluateMultipleExists patt = do
+        let extraElemVariables =
```
ttuegel, line 992: I don't think this is quantifying over the correct variables. I think we have an explicit list of variables to quantify in the claim.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1008,6 +988,15 @@ removalPredicate
...
+                getExtraElemVariables configuration patt
+            remainderPattern =
+                Pattern.fromCondition
+                . Conditional.withoutTerm
+                $ (const <$> destination <*> patt)
+        Exists.makeEvaluate
```
ttuegel, line 998: I think we should use `Exists.simplifyEvaluated` here, so that we don't need the `traverse` + `OrPattern.flatten` combination above. (`traverse` on an `OrPattern` is not a good idea.) Using `simplifyEvaluated`, we can simply fold over the list of variables to be quantified.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -1011,17 +1001,23 @@ removalPredicate
     Conditional { term = destTerm } = destination
     (configTerm, configPredicate) = Pattern.splitTerm configuration
     sideCondition = SideCondition.assumeTrueCondition configPredicate
+    configSort = termLikeSort configTerm
+    remainderPattern patt =
+        Pattern.fromCondition
+        . Conditional.withoutTerm
+        $ (const <$> destination <*> patt)
```
ttuegel, line 1008: I know you didn't originally write this part, but you might consider: 

```suggestion
        Pattern.fromCondition
        $ on (<>) Conditional.withoutTerm destination patt
```

I think it's not obvious that `(const <$> _ <*> _)` is the `Semigroup` action on the conditions.

PR #1893, opened 2020-06-17
Efficient and complete implication check design

ttuegel:
This documents a more efficient and complete design for implication checking.

---

###### Reviewer checklist


docs/2020-06-17-Checking-Implication.md
```diff
@@ -0,0 +1,161 @@
...
+
+```
+φ(X) → ∃ Y. ψ(X, Y)
+```
+
+then the claim is proven.
+Otherwise, the algorithm continues with
+
+```
+φ(X) → • ∘ ∃ Y. ψ(X, Y)
```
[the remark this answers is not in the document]
ttuegel: That's not true anymore. At one point in the past, the backend continued with 

```
φ(X) ∧ ¬ ∃ Y. ⌈φ(X) ∧ ψ(X, Y)⌉ → • ∘ ∃ Y. ψ(X, Y).
```

I think that's formally equivalent to what you wrote. The backend now reports any configuration as "stuck" unless the implication condition `¬ ∃ Y. ⌈φ(X) ∧ ψ(X, Y)⌉ ` is exactly `\top` (in which case we keep rewriting) or `\bottom` (in which case the proof is done). I agree that this is technically incomplete, but so far it has served fine. Do you think it is important that we handle this now? If so, I should add it to the document, because there are again important ways to optimize the implementation.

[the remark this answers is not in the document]
traiansf: Probably not important at the moment; however, we should have a note about it somewhere in case it ever becomes relevant.

docs/2020-06-17-Checking-Implication.md
```diff
@@ -0,0 +1,161 @@
...
+```
+⌈t₁(X) ∧ t₂(X, Y)⌉ → ⌈t₁(X)⌉
+```
+
+so that if unification succeeds, there is no need to explicitly consider `⌈t₁(X)⌉`.
+If unification fails, then the negative conjunct is `\top` so that the implication is satisified when `⌈t₁(X)⌉ ∧ P₁(X)` is unsatisfiable.
+Now, the implication is checked in two cases:
+after a configuration is rewritten (in which case the configuration may be rewritten again)
+or after a configuration can no longer be rewritten (a remainder pattern).
+In the case of a rewritten configuration, when unification fails, there is no need to check the rest of the implication pattern because the condition `⌈t₁(X)⌉ ∧ P₁(X)` is ensured by subsequent rewrites.
```
virgil-serbanuta, line 129: This is true because the backend stops the proof and returns an error whenever the z3 solver can't figure out whether a predicate is satisfiable or unsatisfiable, right?

docs/2020-06-17-Checking-Implication.md
```diff
@@ -0,0 +1,161 @@
...
+and `P₁(X)` and `P₂(X, Y)` are predicates:
+
+```
+\not(∃ Y. ⌈φ(X) ∧ ψ(X, Y)⌉) ∧ ⌈φ(X)⌉
+===
+\not(∃ Y. ⌈t₁(X) ∧ t₂(X, Y)⌉ ∧ P₁(X) ∧ P₂(X, Y)) ∧ ⌈t₁(X)⌉ ∧ P₁(X)
+===
+\not((∃ Y. ⌈t₁(X) ∧ t₂(X, Y)⌉ ∧ P₂(X, Y)) ∧ P₁(X)) ∧ ⌈t₁(X)⌉ ∧ P₁(X)
+===
+\or(\not(∃ Y. ⌈t₁(X) ∧ t₂(X, Y)⌉ ∧ P₂(X, Y)), ¬ P₁(X)) ∧ ⌈t₁(X)⌉ ∧ P₁(X)
```
virgil-serbanuta, line 110: s/not/¬/? Or, perhaps, s/¬/not/ in ¬ P₁(X)?

docs/2020-06-17-Checking-Implication.md
```diff
@@ -0,0 +1,161 @@
...
+===
+\or(\not(∃ Y. ⌈t₁(X) ∧ t₂(X, Y)⌉ ∧ P₂(X, Y)) ∧ ⌈t₁(X)⌉ ∧ P₁(X), ¬ P₁(X) ∧ ⌈t₁(X)⌉ ∧ P₁(X))
+===
```
ana-pantilie, line 113: I found it pretty hard to read these two lines. Maybe it would be clearer if the `\or` and `\not` would be indented in some way?

docs/2020-06-17-Checking-Implication.md
```diff
@@ -0,0 +1,161 @@
...
+
+```
+(¬ ⌈φ(X) ∧ ∃ Y. ψ(X, Y)⌉) ∧ ⌈φ(X)⌉
+===
+(¬ ∃ Y. ⌈φ(X) ∧ ψ(X, Y)⌉) ∧ ⌈φ(X)⌉.
+```
+
+## Optimization
+
+We can optimize this solution the typical case where
```
ana-pantilie, line 93: Is there a word missing here?

docs/2020-06-17-Checking-Implication.md
ttuegel: It's true because the pattern we are rewriting is `t₁ ∧ P₁`, and if we add the new condition then we would be rewriting `t₁ ∧ P₁ ∧ ⌈t₁⌉ ∧ P₁`, which represents the same pattern. I should rewrite this section to be clearer about that.

PR #1894, opened 2020-06-18
Add built-in rule for symbolic ==Int

### LINKED ISSUE #1886: No built-in rule for symbolic ==Int
After adding a custom `==Int` rule to ERC0 verifier, we now see stuck terms like this:
```
#buf ( 32 , #if ?ACCT1 ==Int ?ACCT1 andBool ?ACCT ==Int ?ACCT #then ?SYMB_INT #else #lookup ( ?STORAGE0 , keccak ( #buf ( 32 , ?ACCT1 ) ++ #buf ( 32 , keccak ( #buf ( 32 , ?ACCT ) ++ b"0000000000000000000000000000000000000000000000000000000000000001" ) ) ) ) #fi )
```
https://github.com/runtimeverification/erc20-verification/pull/65
https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/erc20-verification/detail/PR-65/4/pipeline/

All custom equality rules:
https://github.com/runtimeverification/erc20-verification/blob/9b50cc43e6daaf925ecaf4384bc0d85f7fd9d9e9/specs/symbolic/resources/symb-test-verification.k#L46-L53

andreiburdusa:
---
Fixes #1886 
###### Reviewer checklist


kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,31 @@ builtinFunctions =
...
+        Condition.fromPredicate (makeCeilPredicate resultSort termLike)
+    returnPattern = return . flip Pattern.andCondition conditions
+    conditions = foldMap mkCeilUnlessDefined arguments
+
+    concrete = do
+        _intLeft <- expectBuiltinInt eqKey _intLeft
+        _intRight <- expectBuiltinInt eqKey _intRight
+        _intLeft == _intRight
+            & Bool.asPattern resultSort
+            & returnPattern
```
ttuegel, line 351: There is no need to even consider the extra side-conditions in the concrete case.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,31 @@ builtinFunctions =
...
+
+    symbolicReflexivity =
+        case (Recursive.project _intLeft, Recursive.project _intRight) of
+            (_ :< VariableF varLeft, _ :< VariableF varRight) ->
+                varLeft == varRight
+                    & Bool.asPattern resultSort
+                    & returnPattern
```
ttuegel, line 358: This is not right: `varLeft == varRight` _implies_ that `varLeft ==Int varRight`, but not the other way around! We cannot say anything when `varLeft /= varRight`.

Because variables are always defined, there is also no need to include the side conditions here. On the other hand, if we do include the stronger side conditions above, then we do not need to restrict to variables here; we can compare any two symbolic patterns.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,31 @@ builtinFunctions =
     log2 n
         | n > 0 = Just (smallInteger (integerLog2# n))
         | otherwise = Nothing
+
+evalEq :: Builtin.Function
+evalEq resultSort arguments@[_intLeft, _intRight] =
+    concrete <|> symbolicReflexivity
+  where
+    mkCeilUnlessDefined termLike
+      | TermLike.isDefinedPattern termLike = Condition.topOf resultSort
```
ttuegel, line 340: We actually need a stronger condition here: the arguments must be `functional`. If the arguments could take multiple values, then we would have to return multiple values. Consider:

```
\or(1, 2) ==Int 1 = \or(true, false)
```

kore/test/Test/Kore/Builtin/Int.hs
```diff
@@ -136,6 +138,18 @@ testComparison symb impl =
   where
     name = expectHook symb
 
+test_reflexivity_symbolic :: TestTree
+test_reflexivity_symbolic =
+    testPropertyWithSolver (Text.unpack name) $ do
```
ttuegel, line 143: This is an ordinary integration test, not a integration property test. We should use `HUnit` instead, to avoid running the test 100 times.

Please also add a _unit_ (not integration) test to show that `x ==Int y` is inconclusive.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,32 @@ builtinFunctions =
...
+        _intRight <- expectBuiltinInt eqKey _intRight
+        _intLeft == _intRight
+            & Bool.asPattern resultSort
+            & return
+
+    symbolicReflexivity =
+        let (_ :< patternLeft) = Recursive.project _intLeft
+            (_ :< patternRight) = Recursive.project _intRight
+        in
+        if patternLeft == patternRight then
```
ttuegel, line 357: I don't think we need to `project` the patterns here:

```suggestion
        if _intLeft == _intRight then
```

kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,32 @@ builtinFunctions =
...
+    conditions = foldMap mkCeilUnlessDefined arguments
+
+    concrete = do
+        _intLeft <- expectBuiltinInt eqKey _intLeft
+        _intRight <- expectBuiltinInt eqKey _intRight
+        _intLeft == _intRight
+            & Bool.asPattern resultSort
+            & return
+
+    symbolicReflexivity =
```
ttuegel, line 353: Even though we add side conditions to make sure the inputs are defined, we still need to make sure that the inputs are function-like patterns:

```suggestion
    symbolicReflexivity =
        Control.Monad.guard (TermLike.isFunctionPattern _intLeft)
        -- Do not need to check _intRight because we only return a result
        -- when _intLeft and _intRight are equal.
```

kore/src/Kore/Builtin/Int.hs
```diff
@@ -325,3 +331,32 @@ builtinFunctions =
...
+evalEq :: Builtin.Function
+evalEq resultSort arguments@[_intLeft, _intRight] =
+    concrete <|> symbolicReflexivity
+  where
+    mkCeilUnlessDefined termLike
+      | TermLike.isFunctionalPattern termLike = Condition.topOf resultSort
+      | otherwise =
+        Condition.fromPredicate (makeCeilPredicate resultSort termLike)
+    returnPattern = return . flip Pattern.andCondition conditions
+    conditions = foldMap mkCeilUnlessDefined arguments
```
ttuegel, line 344: Let's move these closer to the definition of `symbolicReflexivity` because that is the only place they are used.

kore/src/Kore/Builtin/Int.hs
andreiburdusa: Actually, `mkCeilUnlessDefined` uses `isFunctionalPattern`. I forgot to change its name :sweat_smile: . Should I make it check only if the inputs are defined and use the guard you suggested for function-like check?

Issue #1815, opened 2020-05-13
Do not rename SideCondition variables when applying equations


ttuegel:
`checkRequires` renames the variables of the `SideCondition` before sending its query to the solver. Instead, we should rename the variables in the result _before_ calling `checkRequires`: if the check passes, we will need the renamed equation anyway, and even if the check does not pass, the equation's result is almost always smaller than the `SideCondition` and faster to rename.

ttuegel:
I am electing to postpone this after #1545.

ttuegel:
This is already half-way done: The `requires` is already renamed. Next, we need to change every function in `Kore.Equation.Application` which takes a `SideCondition (Target variable)` argument to take a `SideCondition variable` argument instead. These functions are:

- [ ] `attemptEquation`
- [ ] `checkRequires`
- [ ] `targetEquationVariables`

PR #1895, opened 2020-06-19
Do not rename SideCondition variables when applying equations

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1815

---

###### Reviewer checklist


kore/src/Kore/Equation/Application.hs
```diff
@@ -228,7 +228,7 @@ attemptEquation sideCondition termLike equation =
             let toMatchResult Conditional { predicate, substitution } =
                     (predicate, Substitution.toMap substitution)
             Substitution.mergePredicatesAndSubstitutions
-                sideCondition
+                SideCondition.top
```
ana-pantilie, line 231: I assumed it would be fine to not include the side condition here, since we use it in `checkRequires`. Is this correct or am I overlooking something?

kore/src/Kore/Equation/Application.hs
```diff
@@ -222,14 +222,14 @@ attemptEquation sideCondition termLike equation =
     simplifyArgumentWithResult
         argument
         antiLeft
-        (matchPredicate, matchSubstitution)
```
ttuegel, line 225: Will this drop the `matchPredicate`? I don't see where it's included, otherwise.

ana-pantilie: Yes, I can just make the function take only the substitution if it's clearer.

ttuegel: We can't drop the `matchPredicate`, though. We just want to postpone checking it until the `checkRequires` stage.

ana-pantilie: Huh, for some reason I was convinced we already use it in `checkRequires`. You're right, thank you for noticing this!

PR #1898, opened 2020-06-19
Merge priority rules

virgil-serbanuta:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Internal/Alias.hs
+++ b/kore/src/Kore/Internal/Alias.hs
@@ -78,7 +78,6 @@ instance
     Synthetic (FreeVariables variable) (Application (Alias patternType))
   where
     -- TODO (thomas.tuegel): Consider that there could be bound variables here.
-    -- Also, the alias could introduce its own free variables when expanding.
     synthetic = Foldable.fold
     {-# INLINE synthetic #-}
 
--- a/kore/src/Kore/Internal/Alias.hs
+++ b/kore/src/Kore/Internal/Alias.hs
@@ -78,6 +78,7 @@ instance
     Synthetic (FreeVariables variable) (Application (Alias patternType))
   where
     -- TODO (thomas.tuegel): Consider that there could be bound variables here.
+    -- Also, the alias could introduce its own free variables when expanding.
     synthetic = Foldable.fold
     {-# INLINE synthetic #-}
 
--- a/kore/src/Kore/Step/AntiLeft.hs
+++ b/kore/src/Kore/Step/AntiLeft.hs
@@ -1,0 +1,418 @@
+{-|
+Copyright   : (c) Runtime Verification, 2020
+License     : NCSA
+-}
+
+module Kore.Step.AntiLeft
+    ( AntiLeft (..)
+    , antiLeftPredicate
+    , forgetSimplified
+    , mapVariables
+    , parse
+    , substitute
+    , toTermLike
+    ) where
+
+import Prelude.Kore
+
+import Control.DeepSeq
+    ( NFData
+    )
+import Data.Foldable
+    ( fold
+    , toList
+    )
+import Data.List
+    ( foldl'
+    )
+import Data.Map
+    ( Map
+    )
+import qualified Data.Map as Map
+import Data.Set
+    ( Set
+    )
+import qualified Data.Set as Set
+import qualified Generics.SOP as SOP
+import qualified GHC.Generics as GHC
+
+import Debug
+    ( Debug
+    , Diff
+    )
+import Kore.Attribute.Pattern.FreeVariables
+    ( HasFreeVariables (freeVariables)
+    , bindVariables
+    )
+import qualified Kore.Attribute.Pattern.FreeVariables as FreeVariables
+    ( toNames
+    )
+import Kore.Internal.Predicate
+    ( Predicate
+    , makeAndPredicate
+    , makeCeilPredicate_
+    , makeMultipleExists
+    , makeMultipleOrPredicate
+    , makeOrPredicate
+    )
+import qualified Kore.Internal.Predicate as Predicate
+    ( forgetSimplified
+    , mapVariables
+    , substitute
+    , wrapPredicate
+    )
+import Kore.Internal.TermLike
+    ( pattern And_
+    , pattern ApplyAlias_
+    , pattern Bottom_
+    , pattern Exists_
+    , pattern Or_
+    , TermLike
+    , mkAnd
+    , mkElemVar
+    )
+import qualified Kore.Internal.TermLike as TermLike
+    ( forgetSimplified
+    , mapVariables
+    , substitute
+    )
+import Kore.Internal.Variable
+    ( InternalVariable
+    )
+import Kore.Step.Simplification.ExpandAlias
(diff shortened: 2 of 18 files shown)


test/priority/priority.k
```diff
@@ -2,10 +2,19 @@ module PRIORITY
...
   rule one => three [priority(2)]
 
+  rule start(X) => one       requires X <Int 0  [priority(1)]
+  rule start(X) => one       requires X >=Int 0 andBool X <Int 5  [priority(1)]
+  rule start(X) => middle(X) requires X <Int 10 [priority(2), label(first)]
+  rule start(X) => one                          [priority(3)]
+
+  rule middle(X) => one requires X <Int 6 [priority(1)]
+  rule middle(X) => two requires X >Int 2 [priority(2), label(second)]
+  rule middle(X) => one                   [priority(3)]
```
ehildenb, line 16:
```suggestion
  rule middle(X) => one                   [priority(3), label(third)]
```

Add this label, and make a test `first` followed by `third`.

kore/src/Kore/Step/Rule.hs
```diff
@@ -266,18 +269,37 @@ complexRewriteTermToRule attributes pat =
...
-                RewriteRule RulePattern
-                    { left = lhs
-                    , antiLeft = Just antiLeft
-                    , requires = Predicate.wrapPredicate requires
-                    , rhs = termToRHS rhs
-                    , attributes
-                    }
+            rhs -> case AntiLeft.parse antiLeft of
+                Nothing -> (error . show. Pretty.vsep)
+                    [ "Cannot parse antileft term"
```
hjorthjort, line 274:
```suggestion
                    [ "Cpuld not parse antileft term"
```

"Cannot" sounds to me like it's unsupported, I would have been confused by this message, especially since I don't know what antileft really means.

kore/src/Kore/Internal/Alias.hs
```diff
@@ -78,6 +78,7 @@ instance
     Synthetic (FreeVariables variable) (Application (Alias patternType))
   where
     -- TODO (thomas.tuegel): Consider that there could be bound variables here.
+    -- Also, the alias could introduce its own free variables when expanding.
```
ttuegel, line 81: It is not allowed for an alias to introduce free variables.

kore/src/Kore/Step/AntiLeft.hs
```diff
@@ -0,0 +1,417 @@
...
+    , SomeVariableName (SomeVariableNameElement)
+    , Variable (variableName)
+    , mapElementVariable
+    )
+import Kore.Variables.Fresh
+    ( FreshPartialOrd
+    , refreshElementVariable
+    )
+
+data AntiLeftLhs variable = AntiLeftLhs
```
ttuegel, line 97: Please add documentation explaining what this type represents. I think it's the left-hand side of one rule used in the anti-left pattern? But, I don't know why the alias and its expansion are both present.

kore/src/Kore/Step/AntiLeft.hs
```diff
@@ -0,0 +1,417 @@
...
+
+instance SOP.HasDatatypeInfo (AntiLeftLhs variable)
+
+instance Debug variable => Debug (AntiLeftLhs variable)
+
+instance (Debug variable, Diff variable) => Diff (AntiLeftLhs variable)
+
+data AntiLeft variable = AntiLeft
+    { aliasTerm :: !(TermLike variable)
+    , maybeInner :: !(Maybe (AntiLeft variable))
```
ttuegel, line 121: Could you add documentation to the fields of this type? Particularly `maybeInner`; it's not clear to me what this is.

kore/src/Kore/Step/AntiLeft.hs
```diff
@@ -0,0 +1,417 @@
...
+    -> AntiLeftLhs variable
+substituteLeft subst antiLeft@(AntiLeftLhs _ _ _ _) =
+    case antiLeft of
+        AntiLeftLhs {aliasTerm, existentials, predicate, term} ->
+            AntiLeftLhs
+                { aliasTerm = TermLike.substitute subst' aliasTerm
+                , existentials
+                , predicate = Predicate.substitute subst' predicate
+                , term = TermLike.substitute subst' term
+                }
```
ttuegel, line 221: Because there can only be one case alternative, we might use `where` instead, but this is not very important and I leave it up to you.

kore/src/Kore/Step/AntiLeft.hs
```diff
@@ -0,0 +1,417 @@
...
+    stripExistentials
+        :: TermLike variable -> ([ElementVariable variable], TermLike variable)
+    stripExistentials (Exists_ _ var term) = (var : vars, remaining)
+      where
+        (vars, remaining) = stripExistentials term
+    stripExistentials term = ([], term)
+
+{-| Creates the AntiLeft predicate as described in
+docs/2020-06-30-Combining-Priority-Axioms.md
+-}
```
ttuegel, line 325: In the document, I think what you have below appears negated, is that right?

test/priority/priority.k
```diff
@@ -2,10 +2,19 @@ module PRIORITY
...
   imports BOOL
 
-  syntax S ::= "one" | "two" | "three" | "four" | "five" | "six" | "seven"
+  syntax S ::= "one" | "two" | "three" | "four" | "five" | "six" | "seven" | start(Int) | middle(Int)
   rule one => two [priority(1)]
   rule one => three [priority(2)]
 
+  rule start(X) => one       requires X <Int 0  [priority(1)]
+  rule start(X) => one       requires X >=Int 0 andBool X <Int 5  [priority(1)]
+  rule start(X) => middle(X) requires X <Int 10 [priority(2), label(first)]
```
ttuegel, line 11: What happens to the anti-left of this rule when it is combined with the other rules?

kore/src/Kore/Step/AntiLeft.hs
virgil-serbanuta: Better now?

kore/src/Kore/Step/AntiLeft.hs
virgil-serbanuta: I removed the alias, it made sense in an older version of this PR, but doesn't anymore.

I added some explanations, is it more clear now?

kore/src/Kore/Step/AntiLeft.hs
virgil-serbanuta: That seems better, fixed.

kore/src/Kore/Step/AntiLeft.hs
virgil-serbanuta: I added some explanations, hopefully it's more clear now.

test/priority/priority.k
virgil-serbanuta: I'm not fully sure I understood what you meant. Please tell me if the paragraph below does not answer your question.

When `first` is combined with `second` or `third`, its antileft is unchanged. Both of the test output files above contain `/*  */ \not{SortGeneratedTopCell{}}(/*  */ priorityLE1{}())`, which should be exactly the antileft of `first`.

PR #1904, opened 2020-06-22
Priority rewrite encoding

virgil-serbanuta:
I didn't find a design document for this issue, so I created one. Does everyone agree with the current solution?

---

###### Reviewer checklist


docs/2020-06-22-priority-rewrites.md
```diff
@@ -0,0 +1,229 @@
...
+    ∧ ¬ ∃ X₁ . ⌈ C(Y) ∧ φ₁(X₁) ⌉
+    ∧ ψ₂(X)
+===
+// ⌈⌉ commutes with ∃
+∃ X .
+    ⌈ C(Y) ∧ φ₂(X) ⌉
+    ∧ ¬ ⌈ ∃ X₁ . C(Y) ∧ φ₁(X₁) ⌉
+    ∧ ψ₂(X)
+===
+// ∃ X₁ . A ∧ B(X)  ===  A ∧ ∃ X₁ . B(X)
```
traiansf, line 120: ` X₁` => `X` ?

docs/2020-06-22-priority-rewrites.md
```diff
@@ -0,0 +1,229 @@
...
+∃ X .
+    ⌈ C(Y) ∧ φ₂(X) ⌉
+    ∧ ¬ ⌈ C(Y) ∧ ∃ X₁ . φ₁(X₁) ⌉
+    ∧ ψ₂(X)
+===
+// ⌈ A ⌉ ∧ P === ⌈ A ∧ P ⌉
```
traiansf, line 126: if P is a predicate

docs/2020-06-22-priority-rewrites.md
```diff
@@ -0,0 +1,229 @@
...
+rules at previous priorities:
+```
+φ₁(X) ⇒ ψ₁(X) requires P₁(X)
+...
+φn(X) ⇒ ψn(X) requires Pn(X)
+```
+and we encode the rule as:
+```
+    φ(X) ∧ P(X)
+    ∧ (¬ ∃ X₁ . φ₁(X₁) ∧ P₁(X₁))
```
traiansf, line 168: maybe move the not outside of the parens?  similarly below

docs/2020-06-22-priority-rewrites.md
```diff
@@ -0,0 +1,229 @@
...
+```
+φ₁(X) ⇒ ψ₁(X) requires P₁ [priority(p₁)]
+φ₂(X) ⇒ ψ₂(X) requires P₂ [priority(p₂)]
+```
+and, if we also know that `p₁ < p₂`, we could replace them with
+```
+φ₁(X) ⇒ ψ₁(X) requires P₁
+φ₂(X) ⇒ ψ₂(X)
+    requires P₂(X)
+        ∧ ¬ ∃ X₁ . ⌈φ₂(X) ∧ φ₁(X₁)⌉
```
traiansf, line 189: shouldn't `P1` be inside the negation, as well?

ttuegel:
Does this describe priority rules as they are currently implemented, or as we think they should be implemented?

virgil-serbanuta:
> Does this describe priority rules as they are currently implemented, or as we think they should be implemented?

The "Priority expression (non-predicate/term)" section describes the rules as implemented. The "Priority predicate" section describes an alternative. I am not sure how they should be implemented, but I'm leaning towards keeping the current implementation, since it seems cleaner (hence the "I recommend using the priority expression." sentence at the end of the document).

I'm hoping this will be a discussion about how they should be implemented (unless everyone agrees with my recommendation). Regardless of how it goes, I should replace the recommendation with something else before merging.

traiansf:
I guess the main question here is: "which of the alternatives benefits us more?"
The current one is nice because it allows us to easily identify and discard the priority conditions.  
What benefits does the other alternative bring?

virgil-serbanuta:
I think that the current one is also somewhat cleaner from a theoretical point of view, which usually means it's more resilient to changes in assumptions (e.g. the LHS being function-like).

I think that the other one (the predicate-based one) could also be implemented in a way that makes it easier to discard the priority conditions. It would also make it somewhat easier (but not that much easier) to implement rule merging and it would mean the LHS has a simpler overall structure ("term and predicate", as opposed to "term and predicate and stuff"). This simpler structure would only be relevant to tools which don't want to handle priority rules in a special way (not sure which, perhaps new backends?). This option also has an explicit cost (we need to implement it).

The differences might not be that big, so it might not matter much which one we choose, and, as I mention, I somewhat prefer the current option.

virgil-serbanuta:
BTW, intuitively the second option shouldn't need the extra assumption on the LHS, but I didn't figure out yet how to do that.

ttuegel:
> I think that the current one is also somewhat cleaner from a theoretical point of view, which usually means it's more resilient to changes in assumptions (e.g. the LHS being function-like).

I guess, besides that, it also matches what the backend produces for remainder patterns when rewriting. That means: we don't actually have to consider the priority pattern at runtime because we know that the remainder pattern covers it.

docs/2020-06-22-Rewrite-Rule-Priorities.md
```diff
@@ -0,0 +1,236 @@
...
+where the highest priority is the one with the lowest number.
+
+As an example, if `A(X)` rewrites to
+* `B(X)` if `X<0`
+* `C(X)` if `X≥0 ∧ X<1`
+* `D(X)` if `X≥1`,
+we could write the following rules:
+
+```
+rule A(X) ⇒ B(X) requires X <Int 0 [priority(1)]
```
yzhang90, line 29: why all the three rules rewrite to B(X)?

virgil-serbanuta: Copy-paste error. Fixed.

docs/2020-06-22-Rewrite-Rule-Priorities.md
```diff
@@ -0,0 +1,236 @@
...
+===
+// A ∧ ¬ (A ∧ B) = A ∧ ¬ B
+∃ X . ⌈ C(Y) ∧ φ₂(X) ∧ P₂(X) ∧ ¬ ∃ X₁ . C(Y) ∧ φ₁(X₁) ∧ P₁(X₁) ⌉ ∧ ψ₂(X)
+```
+which is the same as above (above `P₁` and `P₂` are included in `φ₁` and `φ₂`).
+
+### Discussion
+
+Using the priority expression (the option that's currently implemented)
+is cleaner form a theoretical point of view since it requires that the
```
ttuegel, line 217: I'm not sure, but that the significance here is:

```suggestion
is cleaner form a theoretical point of view since it requires only that the
```

docs/2020-06-22-Rewrite-Rule-Priorities.md
```diff
@@ -0,0 +1,236 @@
...
+configuration is function-like, and, arguably, not even that is needed,
+i.e. the priority expression could work as the definition of applying priority
+rules. It also matches what the Haskell backend produces for remainder pattern
+when rewriting.
+
+The priority predicate requires, additionally, that the left hand side of any
+priority rule is function-like. In practice, they should
+be functional, so this should not matter much.
```
ttuegel, line 225: Actually, this is a significant limitation: It would prevent using `#Or` patterns in the frontend with `priority` rules.

docs/2020-06-22-priority-rewrites.md
virgil-serbanuta: Fixed

docs/2020-06-22-priority-rewrites.md
virgil-serbanuta: Fixed.

docs/2020-06-22-priority-rewrites.md
virgil-serbanuta: Fixed.

docs/2020-06-22-priority-rewrites.md
virgil-serbanuta: Fixed.

docs/2020-06-22-Rewrite-Rule-Priorities.md
virgil-serbanuta: Thanks, fixed.

docs/2020-06-22-Rewrite-Rule-Priorities.md
virgil-serbanuta: I added that to the discussion.

Issue #1916, opened 2020-06-24
`#Ceil` rule which depends on top constraint not applied


denis-bogdanas:
To reproduce: https://github.com/runtimeverification/erc20-verification/pull/75
Stuck part, in top constraint:
```
  #Not ( #Ceil ( #lookup ( ?STORAGE0 , keccak ( #buf ( 32 , ?ACCT ) ++ b"0000000000000000000000000000000000000000000000000000000000000000" ) ) ) )
#And
...
```

Rule that should apply, in `evm-symb-testing.md`
```
    rule #Ceil(#lookup(STORAGE, _))                  => #True
      requires #isValidStorage(STORAGE)                                         [anywhere, simplification]
```
Top constraint that should match rule side condition:
```
#And
  {
    true
  #Equals
    #isValidStorage ( ?STORAGE0 )
  }
```

ttuegel:
The problem is that we use `simplifyConditionalTermToOr` [here](https://github.com/kframework/kore/blob/68c549b01e21356fc82f93fb8b7b62b4470208e2/kore/src/Kore/Strategies/Goal.hs#L982), which simplifies the term but does not fully simplify the resulting conditions. We should use `Kore.Step.Simplification.Pattern.simplify` instead.

PR #1927, opened 2020-06-25
Fully simplify conditions from implication checking

ttuegel:
Fixes: #1916 

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -812,11 +817,13 @@ removeDestination lensRulePattern mkState goal =
...
-            let configAndRemoval =
-                    fmap (configuration <*) removal
+            let configAndRemoval = fmap (configuration <*) removal
+                sideCondition =
+                    Pattern.withoutTerm configuration
+                    & SideCondition.fromCondition
             simplifiedRemoval <-
                 simplifyConditionsWithSmt
-                    SideCondition.top
+                    sideCondition
```
ana-pantilie, line 826: Why is the side condition necessary here?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -979,14 +986,29 @@ removalPatterns
...
+                simplifyEvaluatedMultiPredicate
+                    sideCondition
+                    (MultiAnd.make
+                        [ unifiedConditions
+                        , destinationConditions
+                        ]
+                    )
             let remainderPatterns =
-                    fmap remainderPattern unifiedConfigs
+                    fmap Pattern.fromCondition_ remainderConditions
```
ana-pantilie, line 1011: Ah, so we also needed to first fully simplify the destination and all the resulting conditions together before calling `simplifyEvaluated` for `Exists` and `Not`, right?

kore/test/Test/Kore/Strategies/Goal.hs
```diff
@@ -0,0 +1,41 @@
...
+                Pattern.withCondition
+                    (Mock.f (mkElemVar Mock.x))
+                    (Condition.fromPredicate
+                        $ makeCeilPredicate Mock.testSort
+                        $ Mock.f (mkElemVar Mock.x)
+                    )
+            destination = Pattern.fromTermLike (mkElemVar Mock.y)
+            existentials = [Mock.y]
+
+        _output <- removalPatterns destination configuration existentials
```
ana-pantilie, line 35: Shouldn't this be compared to the expected result?

kore/src/Kore/Strategies/Goal.hs
ttuegel: This side condition will be used for function evaluation while simplifying the argument, before `simplifyConditionsWithSmt` sends the result to the solver.

kore/src/Kore/Strategies/Goal.hs
ttuegel: Yes, that's right: the simplifiers for `Exists` and `Not` assume that their arguments are fully simplified, but they don't check. What makes it even harder is that _usually_ the patterns are fully simplified at this point, but in a few cases they need extra function or simplification rules to be applied.

PR #1932, opened 2020-06-29
Refactor reachability proof strategies

ttuegel:
---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -345,56 +322,60 @@ instance ClaimExtractor OnePathRule where
...
-                coinductiveRewrites
-                priorityGroups
-            )
+    simplify = simplify' _Unwrapped
+    checkImplication = checkImplication' _Unwrapped
+    isTriviallyValid = isTriviallyValid' _Unwrapped
+    applyClaims claims = deriveSeqAllPath (map goalToRule claims)
+
+    applyAxioms axiomss = \goal ->
+        foldM applyAxioms1 (GoalRemainder goal) axiomss
```
ana-pantilie, line 337: This implementation for `applyAxioms` is different from the `OnePathRule` instance because of the priority groups?

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -202,32 +184,53 @@ unprovenNodes executionGraph =
...
         => goal -> Rule goal
     goalToRule = coerce
 
-    transitionRule
-        :: (MonadCatch m, MonadSimplify m)
-        => Prim goal
-        -> ProofState goal goal
-        -> Strategy.TransitionT (Rule goal) m (ProofState goal goal)
+    -- TODO (thomas.tuegel): isTriviallyValid should be part of
+    -- checkImplication. why is this so slow?
```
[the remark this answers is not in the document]
ttuegel: Oh, I didn't mean to commit that comment. The code isn't slow; that's my lament that Emacs is impossibly slow on my monster 16-core desktop, while it's still fast on my 10 year old laptop. I was running Emacs' built-in profiler while I typed that out (and waited about 10 seconds for the text to appear). I guess I forgot to delete it after. :sweat_smile:

kore/src/Kore/Strategies/Goal.hs
ttuegel: That's right. I guess I should probably make it the same for symmetry, but when applying the rules in sequence, it doesn't make a difference anyway.

kore/src/Kore/Strategies/Goal.hs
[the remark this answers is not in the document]
ana-pantilie: Oh, that's weird. :confused: Also, I had no idea Emacs has a builtin profiler. Apparently Vim has one too, that's good to know. :smile:

kore/src/Kore/Strategies/Goal.hs
ana-pantilie: I think it's fine the way it is, I was just making sure I understand the code properly.

Issue #1911, opened 2020-06-24
Infer definedness of sub-terms


ttuegel:
In the [implication checking](https://github.com/kframework/kore/blob/master/docs/2020-06-17-Checking-Implication.md#summary) document, I mentioned that:

> If we checked `⌈t(X)⌉ ∧ P(X)` [during the implication check], then we could infer that the configuration is defined, allowing us to do subsequent rewriting without generating extra definedness conditions. The efficiency of this method merits investigation.

We will add an internal node to `TermLike` called `Defined`. `Defined` will have a single child. As a pattern, it will take all its synthetic attributes from the child, except that the `Defined` node is always a defined pattern. During the implication check, after we check that the configuration's term is defined, we will push a `Defined` node down onto the term. We will push the `Defined` node down onto subterms as appropriate; for example,

```.kore
#defined(\and(φ₁, φ₂)) = #defined(\and(#defined(φ₁), #defined(φ₂)))
#defined(σ(φ₁, ...)) = #defined(σ(#defined(φ₁), ...))  // remove the outer-most #defined if σ is a functional symbol
```

We will need to add unification and matching cases which strip the virtual `#defined` node from the pattern.

The `\ceil` simplifier already takes advantage of the definedness pattern attribute. The backend will now check definedness of the term in advance and avoid re-checking terms that are already known to be defined.

PR #1947, opened 2020-07-02
TermLike.TermLike: add Defined node to TermLike

ana-pantilie:
First part of https://github.com/kframework/kore/issues/1911

---

###### Reviewer checklist


kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1435,6 +1440,68 @@ mkEvaluated
     -> TermLike variable
 mkEvaluated = updateCallStack . synthesize . EvaluatedF . Evaluated
 
+mkDefined
+    :: forall variable
+    .  HasCallStack
+    => InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+mkDefined = updateCallStack . Recursive.cata worker
```
ttuegel, line 1449: We should use `Recursive.unfold` here. When the input and output of an operation are both co-/recursive, we have the a choice to `fold` or `unfold`; the second one produces the result lazily (incrementally) so it composes better with other transformations.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1435,6 +1440,68 @@ mkEvaluated
...
+            VariableF (Const variable) ->
+                if isElementVariable variable
+                    then synthesize patt
+                    else mkDefinedAtTop (synthesize patt)
+            EndiannessF _ -> mkDefinedAtTop (synthesize patt)
+            SignednessF _ -> mkDefinedAtTop (synthesize patt)
+            InjF _ -> mkDefinedAtTop (synthesize patt)
+            DefinedF _ -> synthesize patt
+
+mkDefinedAtTop :: Ord variable => TermLike variable -> TermLike variable
```
ttuegel, line 1502: Please add a comment distinguishing this from `mkDefined`.

kore/src/Kore/Step/Simplification/Defined.hs
```diff
@@ -0,0 +1,28 @@
...
+-- as 'Defined'. Does not do any actual simplification.
+simplify
+    :: InternalVariable variable
+    => Defined (OrPattern variable)
+    -> OrPattern variable
+simplify Defined { getDefined = defined } =
+    OrPattern.fromTermLike
+    $ markSimplified
+    $ mkDefined
+    $ OrPattern.toTermLike defined
```
ttuegel, line 28: This will incorrectly mark the `predicate` and `substitution` as defined, which effectively deletes them. We want to map `mkDefined` over only the `TermLike` terms.

kore/test/Test/Kore/Internal/TermLike.hs
```diff
@@ -422,3 +423,50 @@ test_renaming =
...
+                                (Mock.functional11
+                                    (defined (Mock.f mkTop_))
+                                )
+                            )
+                        )
+                        mkTop_
+                    )
+            actual = mkDefined term
+        assertEqual "" expected actual
+    , testCase "Nested or, ceils, functional symbol" $ do
```
ttuegel, line 452: This test case is not correct: `mkDefined` should not distribute over `\or`.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1435,6 +1441,97 @@ mkEvaluated
...
+                        then
+                            mkApplySymbol
+                                applicationSymbolOrAlias
+                                (fmap worker applicationChildren)
+                        else
+                            mkDefinedAtTop
+                                ( mkApplySymbol
+                                    applicationSymbolOrAlias
+                                    (fmap worker applicationChildren)
+                                )
```
ttuegel, line 1484:
```suggestion
                    (if isFunctional applicationSymbolOrAlias then id else mkDefinedAtTop)
                    (mkApplySymbol
                                applicationSymbolOrAlias
                                (fmap worker applicationChildren)
                    )
```

kore/src/Kore/Step/Simplification/Defined.hs
```diff
@@ -0,0 +1,28 @@
...
+-- from simplification as 'Defined'. Does not
+-- do any actual simplification.
+simplify
+    :: InternalVariable variable
+    => Defined (OrPattern variable)
+    -> OrPattern variable
+simplify Defined { getDefined = defined } =
+    Pattern.markSimplified
+    . fmap mkDefined
+    <$> defined
```
ttuegel, line 28: This should check if the child is already known to be defined, and remove the outer marker if it is. This could happen if we mark something defined, and later simplify the child pattern in a way that we can otherwise infer it is defined.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -114,6 +114,7 @@ translatePredicateWith translateTerm predicate =
     translatePredicatePattern pat =
         case Cofree.tailF (Recursive.project pat) of
             EvaluatedF child -> translatePredicatePattern (getEvaluated child)
+            DefinedF child -> translatePredicatePattern (getDefined child)
```
ttuegel, line 117: We should also add a case to `translatePattern`, below.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -292,6 +293,9 @@ translatePredicateWith translateTerm predicate =
                         smtSort <- hoistMaybe $ translateSort sort
                         translateUninterpreted smtSort pat
                     ApplySymbolF app -> translateApplication app
+                    DefinedF (Defined child) -> do
+                        smtSort <- hoistMaybe $ translateSort sort
+                        translateUninterpreted smtSort child
```
ttuegel, line 298: We want `translatePattern` to behave as if the `Defined` isn't there:

```suggestion
                    DefinedF (Defined child) -> translatePattern sort child
```

Issue #1665, opened 2020-03-11
Proof stuck on infeasible state


daejunpark:
kprove failed to prove [process-slots-base1-spec](https://github.com/runtimeverification/beacon-chain-verification/blob/master/dynamic/process-slots-base1-spec.k), outputing the following stuck term, which is clearly bottom:
```
  ( #Forall x . #Ceil ( DotVar2 x ) )
#And
  #Not ( ( #Forall x . #Ceil ( DotVar2 x ) )
  #And
    #Not ( {
      Slot
    #Equals
      Slot +Int 1
    } ) )
...
```

To reproduce:
```
$ git clone git@github.com:runtimeverification/beacon-chain-verification.git
$ git checkout 9ec687a7ab0a7fc8748e4861a6d429aeb58761ea
$ cd dynamic
$ make deps
$ make build-static
$ kprove -d .build/defn/static -m VERIFICATION process-slots-base1-spec.k
```

daejunpark:
related: https://github.com/kframework/k/issues/1157

ttuegel:
An `in_keys` symbol for Map cells was added in kframework/k#1161. We need to:

- [ ] Add axioms for `in_keys` to `domains.k`.
- [ ] Add `\ceil` axioms for Map to `domains.k` in terms of `in_keys`.
- [ ] Emit axioms for this symbol in `ModuleToKORE`.
- [ ] Emit `\ceil` axioms for Map cells in `ModuleToKORE`.
- [ ] Disable and remove the backend's internal `\ceil(Map)` simplification.

ttuegel:
@ana-pantilie We can write the rules for `inKeys` in this `owise` style:

```
rule inKeys(K, K |-> _ M:Map) => true
rule inKeys(K, M:Map) => false [owise]
```

ttuegel:
I think we should cut the scope of the solution to this issue because we have exceeded our appetite for it. Instead of relying on simplification rules defined in the frontend to replace the backend's internal simplifier for Maps, I suggest:

- [ ] Merge the open pull request(s?) that define axioms for `in_keys`.
- [ ] Instead of [creating a `\forall` pattern](https://github.com/kframework/kore/blob/master/kore/src/Kore/Builtin/AssocComm/CeilSimplifier.hs#L140), apply the [`in_keys` symbol](https://github.com/kframework/kore/blob/master/kore/src/Kore/Builtin/Map/Map.hs#L128) for the relevant sort.

[...]

PR #1950, opened 2020-07-03
Prove infeasible states

ttuegel:
Fixes: #1665 

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -559,6 +587,7 @@ reachabilityFirstStep =
         , CheckGoalStuck
         , CheckGoalRemainder
         , Simplify
+        , InferDefined
         , TriviallyValid
```
ana-pantilie, line 591: Is the `TriviallyValid` transition necessary anymore?

PR #1951, opened 2020-07-03
Show elapsed time in timestamps log

### LINKED ISSUE #1942: Log timestamps should show time elapsed since execution began
Displaying the timestamps as "absolute" date and time is not really useful. It would be more useful to display "microseconds since startup", for example.

andreiburdusa:
---
Fixes #1942 
###### Reviewer checklist


kore/src/Kore/Log.hs
```diff
@@ -284,30 +276,19 @@ makeKoreLogger exeName timestampSwitch logToText =
             typeName = type' someEntry'
     indent = Pretty.indent 4
 
+getElapsedTime :: MonadIO io => io RtsTime
+getElapsedTime =
+    liftIO $ fmap (toMicroseconds . Stats.elapsed_ns) Stats.getStats
```
ana-pantilie, line 281: Are you sure this gets the correct time? It seems like [elapsed_ns](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Stats.html#v:elapsed_ns) is specific to garbage collection. Also, are we sure we want to use this package? This is in the description of the module:
```
This module provides access to internal garbage collection and memory usage statistics. These statistics are not available unless a program is run with the -T RTS flag.

This module is GHC-only and should not be considered portable.
```

ana-pantilie: I took another look after you told me you used the functions from our `Stats` module, but I'm still unsure of how safe this is. It looks like `getStats` explicitly triggers a GC, and this side effect might impact execution in an unintended way. @ttuegel , when you have time, could you take a quick look as well?

ttuegel: That's right, you can't use `Stats` here. Look at how [`clockSomethingIO`](https://github.com/kframework/kore/blob/ce2b5beeab5cd7f7373a10ef9b2b62fd28ea6123/kore/app/share/GlobalMain.hs#L422) is implemented. The start time needs to be acquired before the logger is created.

kore/app/repl/Main.hs
```diff
@@ -89,8 +94,8 @@ data KoreReplOptions = KoreReplOptions
     , koreLogOptions   :: !KoreLogOptions
     }
 
-parseKoreReplOptions :: Parser KoreReplOptions
-parseKoreReplOptions =
+parseKoreReplOptions :: TimeSpec -> Parser KoreReplOptions
```
ttuegel, line 97: Please document what this function argument represents.

kore/src/Kore/Log/KoreLogOptions.hs
```diff
@@ -106,6 +111,7 @@ instance Default KoreLogOptions where
             , logEntries = mempty
             , debugSolverOptions = def @DebugSolverOptions
             , exeName = ExeName mempty
+            , startTime = fromNanoSecs 0
```
ttuegel, line 114: I don't really like this as a default because it doesn't make any sense to have time start at the beginning of the Unix epoch. Can we remove the `Default` instance?

andreiburdusa: Do you mean removing the `Default` instance for `KoreLogOptions`?

andreiburdusa: I spent some time trying to remove the `Default` instance and it doesn't seem straightforward. I'll try again tomorrow

andreiburdusa:
@ttuegel 
I attempted to remove the `Default` instance for `KoreLogOptions`. For the moment I'm still using `fromNanoSecs 0` as default time in `Test.Kore.Repl.Parser` but I can try to use the actual time if you say it's worthed.

kore/src/Kore/Log.hs
```diff
@@ -230,26 +221,32 @@ makeKoreLogger
...
-                TimestampsEnable ->
-                    Just $ Pretty.brackets (formattedTime localTime)
+                TimestampsEnable -> Just $
+                    Pretty.brackets
+                        $ Pretty.hsep
+                            [ "Microseconds since startup:"
+                            , Pretty.pretty
+                                $ (toMicroseconds . toNanoSecs) elapsedTime
+                            ]
+        toMicroseconds = (`div` 1000)
```
ttuegel, line 248:
```suggestion
                TimestampsEnable -> Just $
                    Pretty.brackets (Pretty.pretty $ toMicroSecs elapsedTime)
        toMicroSecs = (`div` 1000) . toNanoSecs
```

kore/src/Kore/Log.hs
```diff
@@ -285,28 +282,22 @@ makeKoreLogger exeName timestampSwitch logToText =
...
-getLocalTime =
-    liftIO $ utcToLocalTime <$> getCurrentTimeZone <*> getCurrentTime
-
--- Formats the local time using the provided format string.
-formatLocalTime :: IsString s => String -> LocalTime -> s
-formatLocalTime format = fromString . formatTime defaultTimeLocale format
+withTimestamp :: MonadIO io => TimeSpec -> ActualEntry -> io WithTimestamp
+withTimestamp startTime msg = liftIO $ do
+    currentTime <- getTime Monotonic
+    pure $ WithTimestamp msg (diffTimeSpec currentTime startTime)
```
ttuegel, line 288: Let's use the `currentTime` here, and calculate the difference when the log message is emitted. Then, we don't need to inject the `startTime` so deeply.

```suggestion
withTimestamp :: MonadIO io => ActualEntry -> io WithTimestamp
withTimestamp msg = liftIO $ do
    currentTime <- getTime Monotonic
    pure $ WithTimestamp msg currentTime
```

Issue #1897, opened 2020-06-19
Integrate Kore profiling with speedscope


ttuegel:
We should generate [collapsed-stack profiles](https://github.com/jlfwong/speedscope/wiki/Importing-from-custom-sources#brendan-greggs-collapsed-stack-format) for Kore to be used with tools like [speedscope](https://www.speedscope.app/).

ttuegel:
It may actually be simpler to generate Speedscope's own JSON format, we shouldn't be picky.

PR #1966, opened 2020-07-09
Kore profiling

ttuegel:
Fixes: #1897 

- Removed `kore-profiler`.
- Added `kore-prof` to turn eventlog files into speedscope files. Use `--matching` or `--not-matching` to filter entries in the profile.
- Replaced `MonadProfiler` with `MonadProf`, which is a simpler interface.
- When eventlog is enabled (with `+RTS -l` at runtime) all profiling events are enabled. The profile is filtered by `kore-prof` after `kore-exec` is finished, so it is not required to re-run `kore-exec` to filter events differently.
- Profiling code is moved out of the domain model and into the integration code. Not only does this separate responsibilities, but it makes the domain code open to extension.

---

###### Reviewer checklist


kore/test/Test/Kore/Step/Rule/Simplify.hs
```diff
@@ -293,34 +304,60 @@ test_simplifyClaimRule =
...
+            $ termLike
+      where
         applyReplacements
             :: InternalVariable variable
-            => TermLike variable
+            => [(TermLike VariableName, TermLike VariableName)]
+            -> TermLike variable
             -> TermLike variable
-        applyReplacements zero =
+        applyReplacements replacements zero =
```
ana-pantilie, line 360: I know this isn't your change, but what are these replacements?

ttuegel: This test uses a mock simplifier: instead of actually running the simplifier, it just replaces certain patterns in the result. This is the list of patterns to replace, and their replacements.

kore/test/Test/Kore/Step/Function/Integration.hs
```diff
@@ -600,9 +599,10 @@ test_functionIntegration =
         :: BuiltinAndAxiomSimplifierMap
         -> TermLike VariableName
         -> IO (Pattern VariableName)
-    evaluate functionIdToEvaluator patt =
-        runSimplifier Mock.env { simplifierAxioms = functionIdToEvaluator }
-        $ TermLike.simplify patt SideCondition.top
+    evaluate functionIdToEvaluator termLike =
+        runSimplifier Mock.env { simplifierAxioms = functionIdToEvaluator } $ do
+            patterns <- TermLike.simplify SideCondition.top termLike
+            pure (OrPattern.toPattern patterns)
```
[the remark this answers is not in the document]
ttuegel: No, it's exactly the same. Thanks for pointing that out!

PR #1968, opened 2020-07-10
Refactoring: remove localSimplifierAxioms method from MonadSimplify

ana-pantilie:
---

###### Reviewer checklist


kore/test/Test/Kore/Step/Simplification/TermLike.hs
```diff
@@ -41,22 +39,3 @@ simplifyEvaluated original =
...
-        TestSimplifier . localSimplifierAxioms f . getTestSimplifier
-    askMemo = TestSimplifier (Memo.liftSelf TestSimplifier <$> askMemo)
-    askInjSimplifier = TestSimplifier askInjSimplifier
-    askOverloadSimplifier = TestSimplifier askOverloadSimplifier
-    simplifyCondition sideCondition condition =
-        Logic.mapLogicT TestSimplifier
-        (simplifyCondition sideCondition condition)
-
-    -- Throw an error if any term would be simplified.
-    simplifyTermLike = undefined
```
ana-pantilie, line 62: The compiler was complaining that `TestSimplifier` isn't used anywhere. @ttuegel you added this in the profiling PR, right?

ttuegel:
I sat down with this after #1972 merged, and I realized that this won't work the way we want because `simplifyTermLike` calls `askSimplifierAxioms` and _both_ will be under a `lift`. So, I'm going to close this until we think of something better to do instead.

PR #1974, opened 2020-07-13
Begin to plug the profiling gaps in the \ceil simplifier

### LINKED ISSUE #1896: Profile \ceil simplifier in eventlog
We should add eventlog profiling events to the `\ceil` simplifier to make this information available to users. We would also like this information to investigate alternative strategies for managing `\ceil` conditions.

ttuegel:
Fixes #1896 

---

###### Reviewer checklist


kore/src/Kore/Step/Simplification/Equals.hs
```diff
@@ -58,8 +58,7 @@ import Kore.Step.Simplification.AndTerms
     ( maybeTermEquals
     )
 import {-# SOURCE #-} qualified Kore.Step.Simplification.Ceil as Ceil
-    ( makeEvaluate
-    , makeEvaluateTerm
+    ( makeEvaluateTerm
```
ana-pantilie, line 61: Why is this still needed?

kore/src/Kore/Step/Simplification/CeilSimplifier.hs
```diff
@@ -105,3 +121,15 @@ runCeilSimplifierWith
...
 {-# INLINE runCeilSimplifierWith #-}
+
+ceilSimplifierTermLike
+    ::  InternalVariable variable
+    =>  MonadSimplify simplifier
+    =>  CeilSimplifier
+            (ReaderT (SideCondition variable) simplifier)
+            (TermLike variable)
+            (OrCondition variable)
+ceilSimplifierTermLike =
```
ana-pantilie, line 132: Where is this used, or where/why will this be needed?

ttuegel: It's not, thanks for spotting that!

kore/src/Kore/Step/Simplification/Equals.hs
ttuegel: It's not either, and if I also remove the import from `Kore.Builtin.Builtin`, then we can remove the `.hs-boot` file for `Ceil`! :tada:

ttuegel:
> I couldn't find anything else to refactor, except removing some outdated comments.

That sounds great! "Cleanup" might be better than "refactor" anyway. I think cleanup can include refactoring.

PR #1980, opened 2020-07-15
Missing instance MonadProf for StateT

ttuegel:
---

###### Reviewer checklist

### DIFF
--- a/kore/src/Prof.hs
+++ b/kore/src/Prof.hs
@@ -23,6 +23,12 @@ import Control.Monad.Trans.Except
 import Control.Monad.Trans.Reader
     ( ReaderT
     )
+import qualified Control.Monad.Trans.State.Lazy as Lazy
+    ( StateT
+    )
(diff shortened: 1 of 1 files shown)


kore/src/Prof.hs
```diff
@@ -62,6 +68,20 @@ instance (MonadMask prof, MonadProf prof) => MonadProf (ReaderT r prof) where
     traceEvent name = lift (traceEvent name)
     {-# INLINE traceEvent #-}
 
+instance
+    (MonadMask prof, MonadProf prof)
+    => MonadProf (Lazy.StateT s prof)
+  where
+    traceEvent name = lift (traceEvent name)
```
andreiburdusa, line 75: Is `traceEvent` similar to `traceM`?

ttuegel: `traceEvent` sends a message to the eventlog instead of to `stderr` like `traceM`.

PR #1982, opened 2020-07-15
Log the depth of execution and proof graphs

### LINKED ISSUE #1883: Log number of steps in execution
We should log the number of steps after execution or proving. The log entry should have `info` severity.

ttuegel:
Fixes #1883 

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -483,18 +483,16 @@ ruleOnePathToRuleReachability
     -> Rule ReachabilityRule
 ruleOnePathToRuleReachability = coerce
 
-type TransitionRule m goal =
-    Prim
-    -> ProofState goal
-    -> Strategy.TransitionT (Rule goal) m (ProofState goal)
+type TransitionRule m rule state =
+    Prim -> state -> Strategy.TransitionT rule m state
```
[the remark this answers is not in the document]
ttuegel: Before, `rule` was fixed to `Rule state`. `trackProofDepth` augments `state` to be `(ProofDepth, state)`, so I would have to write an type instance `Rule (ProofDepth, state)` and define a `Goal` instance for it.

kore/src/Kore/Strategies/Verification.hs
```diff
@@ -264,54 +268,49 @@ verifyClaim
...
-        acts
+        do
+            (proofDepth, proofState) <- acts
+            let maybeUnproven = ProofState.extractUnproven proofState
+            Foldable.for_ maybeUnproven $ \unproven -> do
+                infoUnprovenDepth proofDepth
+                Monad.Except.throwError . OrPattern.fromPattern
+                    $ getConfiguration unproven
+            pure proofDepth
+        & Logic.observeAllT
```
andreiburdusa, line 307: Would it be a good idea for me to use `runLogicT` here and get only the maximum value?
Something like
`Logic.runLogicT logic (fmap . (<>)) (pure $ ProofDepth 0)`
where `logic` is the `do` block

ttuegel: It might be a good idea to do something else, but that will build a big thunk on the heap before it starts computing the maximum depth.

ttuegel: On second thought, this doesn't seem to cause any change in performance, so I'd like to keep it simple until we see any problems.

PR #1983, opened 2020-07-15
Allow redirection and piping with aliases

### LINKED ISSUE #1960: kore-repl: redirection doesn't work with alias
Suppose we have the following alias in a preloaded repl script:
```
alias konfig = config | kast -i kore -o pretty -d .build/defn/haskell /dev/stdin
```
Then the following command does NOT redirect the output but just prints to the stdout.
```
Kore (N)> konfig > filename
Config at node N is:
...
```

However, if the aliased command is inlined, then the redirection works:
```
Kore (N)> config | kast -i kore -o pretty -d .build/defn/haskell /dev/stdin > filename
Redirected output to "filename".
```

andreiburdusa:
---
Fixes #1960 
###### Reviewer checklist


kore/test/Test/Kore/Repl/Parser.hs
```diff
@@ -449,6 +451,30 @@ aliasesWithArgs =
     alias name arguments command =
         Alias $ AliasDefinition { name, arguments, command }
 
+pipeRedirectAliasTests :: [ParserTest ReplCommand]
+pipeRedirectAliasTests =
```
ana-pantilie, line 455: Could you add a similar test for append `>>`?

Issue #1934, opened 2020-06-29
Explain error when Z3 crashes


ttuegel:
See also: #1929

PR #1998, opened 2020-07-20
Handle Z3 exceptions

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1934

---

###### Reviewer checklist

### DIFF
--- a/kore/src/SMT/SimpleSMT.hs
+++ b/kore/src/SMT/SimpleSMT.hs
@@ -287,7 +287,7 @@ data SolverException =
     deriving (Show, Typeable)
 
 instance Exception.Exception SolverException where
-    displayException (SolverException { exitCode, someException }) =
+    displayException SolverException { exitCode, someException } =
         (show . Pretty.vsep . catMaybes)
         [ Just "Error while communicating with the solver:"
         , Just $ Pretty.indent 4 $ prettyException someException
--- a/kore/src/SMT/SimpleSMT.hs
+++ b/kore/src/SMT/SimpleSMT.hs
@@ -26,6 +26,7 @@ module SMT.SimpleSMT
     , simpleCommand
     , simpleCommandMaybe
     , loadFile
+    , SolverException (..)
 
     -- ** S-Expressions
     , SExpr(..)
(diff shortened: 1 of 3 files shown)


kore/src/SMT/SimpleSMT.hs
```diff
@@ -344,18 +352,38 @@ command solver c =
...
-    send solver (List [Atom "exit"])
-    ec <- waitForProcess hProc
-    let handler :: X.IOException -> IO ()
-        handler ex = (debug solver . Text.pack) (show ex)
-    X.handle handler $ do
-        hClose hIn
-        hClose hOut
-        hClose hErr
-    return ec
+newtype SolverException = SolverException String
```
ttuegel, line 355: Why not,

```suggestion
newtype SolverException = SolverException (Maybe ExitCode)
```

kore/src/SMT/SimpleSMT.hs
```diff
@@ -344,18 +352,38 @@ command solver c =
...
+    Exception.throwM
+        ( Exception.SomeException . SolverException
+        $ "The Z3 process has been unexpectedly\
+        \ terminated with exit code: "
+        <> maybe "none" show solverExitCode
+        )
+
+stop :: Solver -> IO ExitCode
+stop solver@Solver { solverHandle = SolverHandle { hIn, hOut, hErr, hProc } } =
+    Exception.handle
```
ttuegel, line 374: We should also add this handler to `send` and `recv`.

kore/src/SMT/SimpleSMT.hs
```diff
@@ -266,6 +274,26 @@ data SolverHandle = SolverHandle
...
+newtype SolverException = SolverException (Maybe ExitCode)
+    deriving (Show, Typeable)
+
+instance Exception.Exception SolverException where
+    displayException (SolverException solverExitCode) =
+        "Solver exception: The Z3 process has been unexpectedly\
+        \ terminated with exit code: "
+        <> maybe "none" show solverExitCode
+
+throwSolverException :: ProcessHandle -> IOException -> IO a
```
ttuegel, line 286: Is there a reason not to catch all exceptions (`SomeException` instead of `IOException`)?

ana-pantilie: I was thinking that this way we're fairly sure that something's wrong with accessing the solver, and it's not some unrelated exception.

ttuegel: That's true; we should ignore `AsyncException`, but I think it's OK because we are re-throwing the exception anyway. Let me take care of it as part of "clean-up".

PR #2000, opened 2020-07-21
No bug report on UserInterrupt

### LINKED ISSUE #1996: Do not create a bug report when user interrupts backend
The backend perceives the interrupt signal as an exception and generates a bug report. This is not the desired behavior.

andreiburdusa:
---
Fixes #1996 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/BugReport.hs
+++ b/kore/src/Kore/BugReport.hs
@@ -16,6 +16,10 @@ import Prelude.Kore
 
 import qualified Codec.Archive.Tar as Tar
 import qualified Codec.Compression.GZip as GZip
+import Control.Exception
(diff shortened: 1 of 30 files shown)


kore/src/Kore/BugReport.hs
```diff
@@ -79,7 +83,7 @@ writeBugReportArchive base tar = do
 {- | Run the inner action with a temporary directory holding the bug report.
 
 The bug report will be saved as an archive if that was requested by the user, or
-if there is an error in the inner action.
+if there is an error in the inner action other than `UserInterrupt`.
```
ttuegel, line 86:
```suggestion
if there is an error in the inner action other than 'UserInterrupt'.
```

Issue #1913, opened 2020-06-24
Update LTS Haskell


ttuegel:
We are still using LTS Haskell 14, while LTS Haskell 16 has been released.

PR #2001, opened 2020-07-21
Update LTS Haskell

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1913

---

###### Reviewer checklist


kore/src/Debug.hs
```diff
@@ -316,8 +314,18 @@ datatypeInfoCofreeF =
...
+            ((strictnessInfoForX :* (strictnessInfoForY :* Nil)) :* Nil)
+        )
   where
     constrInfo = SOP.Infix ":<" SOP.RightAssociative 5
+    -- TODO: is the strictness info correct?
+    strictnessInfoForX =
+        SOPMeta.StrictnessInfo
+            SOPMeta.NoSourceUnpackedness
+            SOPMeta.NoSourceStrictness
+            SOPMeta.DecidedLazy
```
ttuegel, line 327: The strictness info is arbitrary because it doesn't affect how we print anything. It would be better to change `debugPrecAux` to take a custom type with only the required information.

kore/src/Debug.hs
```diff
@@ -194,8 +192,8 @@ debugPrecAux datatypeInfo (SOP sop) =
     constrs :: NP ConstructorInfo xss
     constrs =
         case datatypeInfo of
-            SOP.ADT     _ _ cs -> cs
-            SOP.Newtype _ _ c  -> c :* Nil
+            SOP.ADT     _ _ cs _ -> cs
+            SOP.Newtype _ _ c    -> c :* Nil
```
ttuegel, line 196: It would probably be better to change `debugPrecAux` to take only an `NP ConstructorInfo xss` instead of the full `DatatypeInfo`, because we only use one field.

ttuegel:
@ana-pantilie I got `kore` to build with GHC 8.10.1, but not `ghcide`, `hlint`, or `stylish-haskell` due to some dependency failure. I think I may be the only one affected by that.

PR #2005, opened 2020-07-22
Warn when productivity is low

### LINKED ISSUE #1988: Warn about poor performance when productivity is low
It is very unusual that productivity (MUT time / Total time) drop below 90% (i.e. that GC time rise above 10%). Using the `Stats` interface, the backend should issue a warning if this threshold is crossed at the end of execution. This will help detect performance problems early.

andreiburdusa:
---
Fixes #1988 
###### Reviewer checklist

### DIFF
--- a/kore/src/Stats.hs
+++ b/kore/src/Stats.hs
@@ -9,7 +9,6 @@ module Stats
     , getStats
     , writeStats
     , readStats
-    , warnIfLowProductivity
     ) where
 
 import Prelude.Kore
@@ -19,14 +18,10 @@ import Data.Aeson
     , ToJSON
     )
 import qualified Data.Aeson as Aeson
-import Data.Text
-    ( pack
-    )
 import Data.Word
 import qualified Generics.SOP as SOP
 import qualified GHC.Generics as GHC
 import qualified GHC.Stats as GHC
-import Log
 import qualified System.Mem as System
 
 import Debug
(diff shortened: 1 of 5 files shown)


andreiburdusa:
> Is it possible to use the logger to issue a warning for this?

That would require `IO` to be an instance of `MonadLog` and by looking at the class definition I think it was intended to instantiate only transformers :thinking:, but I don't know

ana-pantilie:
`LoggerT` is `MonadIO` so you could get the stats and log them inside the logger, but it might be that you would want to access this information only at the very end of execution. Would that be the case here?

andreiburdusa:
> `LoggerT` is `MonadIO` so you could get the stats and log them inside the logger, but it might be that you would want to access this information only at the very end of execution. Would that be the case here?

Oh, I thought by "logger" you mean the entire logging mechanism.

@ttuegel Can the warning be moved as Ana suggested?

kore/src/Stats.hs
```diff
@@ -82,3 +87,11 @@ readStats filePath =
     either errorWith return =<< Aeson.eitherDecodeFileStrict filePath
   where
     errorWith message = error ("readStats: " ++ message)
+
+warnIfLowProductivity :: MonadLog log => MonadIO log => log ()
+warnIfLowProductivity = do
+    Stats { gc_cpu_ns, cpu_ns } <- liftIO getStats
+    when (gc_cpu_ns * 10 > cpu_ns) $
+        logWarning
```
ana-pantilie, line 95: Is there any reason we don't want to add a separate entry type, `WarnIfLowProductivity`, instead of logging this as a `LogMessage`?

andreiburdusa: parsimony :))
I'll add a new entry type :+1:

PR #2006, opened 2020-07-23
No bug report on ExitSuccess

### LINKED ISSUE #1899: kore-repl: exiting successful proof generates bug report archive

andreiburdusa:
---
Fixes #1899 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/BugReport.hs
+++ b/kore/src/Kore/BugReport.hs
@@ -110,8 +110,8 @@ withBugReport exeName bugReport act =
         case exitCase of
             ExitCaseSuccess _ -> optionalWriteBugReport tmpDir
             ExitCaseException someException
-              | Just ExitSuccess == fromException someException ->
-                    -- ^^^ user exits the repl after the proof was finished
+              | Just ExitSuccess == fromException someException
+                    {- User exits the repl after the proof was finished -} ->
                     optionalWriteBugReport tmpDir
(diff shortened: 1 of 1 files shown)


kore/src/Kore/BugReport.hs
```diff
@@ -109,6 +110,8 @@ withBugReport exeName bugReport act =
         case exitCase of
             ExitCaseSuccess _ -> optionalWriteBugReport tmpDir
             ExitCaseException someException
+              | Just ExitSuccess == fromException someException ->
```
ana-pantilie, line 113: It's possible the constructor names are confusing here, or I'm missing some information, but how can a crash triggered by an exception be a successful run?

andreiburdusa: When a command returns the status `SuccessStop` we call `exitSuccess` which is equivalent to `exitWith ExitSuccess`. My understanding of [`exitWith`](http://localhost:8000/file/home/andrei/.stack/programs/x86_64-linux/ghc-tinfo6-8.6.5/share/doc/ghc-8.6.5/html/libraries/base-4.12.0.0/System-Exit.html#v:exitWith) is that an exception is thrown.

ana-pantilie: Ah, I see. Since this is `repl` specific, could you add a comment here describing that this corresponds to the user exiting the `repl` after the proof was finished?

PR #2011, opened 2020-07-27
SET.difference for symbolic values

### LINKED ISSUE #1993: Set difference `-Set` doesn't work for symbolic values
I get this sort of stuck terms in top condition:
```
  #Not ( #Ceil (
    SetItem ( ?ACCT:Int )
    ( SetItem ( ?ACCT:Int ) -Set SetItem ( ?ACCT:Int ) )
  ) )
```

Please also check `|Set`

Not blocking but required to eliminate a workaround in ERC20.

andreiburdusa:
---
Fixes #1993 
###### Reviewer checklist


andreiburdusa:
I used a small `.k` file to test this and the [productivity](https://github.com/kframework/kore/issues/1988) was down to 61%

kore/test/Test/Kore/Builtin/Set.hs
```diff
@@ -425,6 +428,35 @@ test_difference =
...
+            let patDifference =
+                    mkApplySymbol
+                        differenceSetSymbol
+                        [ xSingleton `concatSet` zeroSingleton
+                            `concatSet` oneSingleton
+                        , xSingleton `concatSet` zeroSingleton
+                        ]
+            actual <- evaluate patDifference
+            assertEqual "" expect actual
+        )
```
ttuegel, line 445: This needs additional test cases.

If a symbolic key is found in the Set, we can remove it, but if it is not found, then we cannot conclude anything:

```
[X 1] -Set [X, Y] === [1] -Set [Y]
```

If a key might not be defined, then we need a `\ceil` condition:

```
[f(X) 1] -Set [f(X)] === [1] \and \ceil f(X)
```

kore/src/Kore/Builtin/Set.hs
```diff
@@ -335,9 +339,83 @@ evalDifference resultSort [_set1, _set2] = do
...
+    differenceSetSymbol =
+        binarySymbol "differenceSet" resultSort
+        & Internal.Symbol.hook "SET.difference"
+    differenceSet
+        :: InternalVariable variable
+        => TermLike variable
+        -> TermLike variable
+        -> TermLike variable
+    differenceSet set1 set2 =
+        TermLike.mkApplySymbol differenceSetSymbol [set1, set2]
```
ttuegel, line 418: You cannot copy these from the test suite. These are stand-ins for data supplied by the user, it cannot be hardcoded in the backend. You need the original symbol. `evalDifference` is eventually called by [`Kore.Builtin.Builtin.functionEvaluator`](https://github.com/kframework/kore/blob/292626439babae4b65bc4f9b3f6606fd3235c9d1/kore/src/Kore/Builtin/Builtin.hs#L271), which only passes the result sort. (That function call happens further down in this file.)

Instead, we should add a new function to `Kore.Builtin.Builtin`, something like this:

```.hs
applicationEvaluator :: _ -> BuiltinAndAxiomSimplifier
applicationEvaluator impl =
    applicationAxiomSimplifier evaluator
  where
    evaluator
        :: InternalVariable variable
        => MonadSimplify simplifier
        => Application Symbol (TermLike variable)
        -> simplifier (AttemptedAxiom variable)
    evaluator (_ :< app) = do
        let app' = fmap TermLike.removeEvaluated app
        getAttemptedAxiom (impl app' >>= appliedFunction)
```

PR #2015, opened 2020-07-27
checkSubstitutionCoverage: check for constructor-like configuration

### LINKED ISSUE #2010: Unable to narrow on results of uninterpretted functions
The following definition crashes with (ErrorRewritesInstantiation)

```
module TEST
    imports BOOL
    imports INT

    syntax MaybeInt ::= "Some" Int
                      | "None"
    syntax MaybeInt ::= f() [function, functional, no-evaluators]

    configuration <k> f() </k>

    rule <k> None => true ... </k>
    rule <k> Some I => I ... </k>
endmodule
```

We expect it to narrow to two cases:

* one with `f() ==K None` and the K cell contains `true`
* one where `f() :=K Some I:Int` and the K cell contains `I`

andreiburdusa:
---
Fixes #2010 
###### Reviewer checklist


andreiburdusa:
I used a small `.k` file to test this and the [productivity](https://github.com/kframework/kore/issues/1988) was down to 87%

kore/src/Kore/Log/ErrorRewritesInstantiation.hs
```diff
@@ -162,4 +170,11 @@ checkSubstitutionCoverage configuration solution
...
     isCoveringSubstitution = Set.null missingVariables
-    isSymbolic = Foldable.any isSomeConfigVariableName substitutionVariables
+    isSymbolic =
+        Foldable.any isSomeConfigVariableName substitutionVariables
+        || isNothing
+            ( getConstructorLike
+            . Attribute.constructorLikeAttribute
+            . extractAttributes
+            $ term configuration
+            )
```
ttuegel, line 180: I think we only need to check that the configuration is not constructor-like:

```suggestion
    isSymbolic = (not . isConstructorLike) (term configuration)
```

test/imp/disjunction-in-simplification-spec.k.out.golden
```diff
@@ -1,10 +1,22 @@
...
-  <k>
-    x = choice ( a , b ); ~> .
-  </k>
-  <state>
-    a |-> A:Int
-    b |-> B:Int
-    x |-> _0
-  </state>
-</T>
+  #Not ( #Exists I . {
```
ttuegel, line 1: This test is not right. I think we should remove it.

ttuegel:
One more thing: can you please add the example from the issue as an integration test?

PR #2016, opened 2020-07-27
Infer defined configurations

### LINKED ISSUE #1911: Infer definedness of sub-terms

ttuegel:
After #2013 
After #2017 
Fixes #1911 

---

###### Reviewer checklist


kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -1242,6 +1210,153 @@ test_functionAnd =
...
+            [ testCase "\\and(partial, defined)" $ do
+                let expect = [Pattern.fromTermLike partial]
+                (actualAnd, actualUnify) <- simplifyUnify partial defined
+                assertEqual "" expect actualAnd
+                assertEqual "" expect actualUnify
+            , testCase "\\and(defined, partial)" $ do
+                let expect = [Pattern.fromTermLike defined]
+                (actualAnd, actualUnify) <- simplifyUnify defined partial
+                assertEqual "" expect actualAnd
+                assertEqual "" expect actualUnify
```
ana-pantilie, line 1228: Could you add a comment quickly explaining why we return the pattern on the left here? I assume this has to do with the decision in `equalAndEquals`, right? Or have I gotten this completely wrong?

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1557,6 +1558,14 @@ mkDefined = updateCallStack . worker
 mkDefinedAtTop :: Ord variable => TermLike variable -> TermLike variable
 mkDefinedAtTop = synthesize . DefinedF . Defined
 
+unDefined :: TermLike variable -> TermLike variable
```
ana-pantilie, line 1561: Maybe it would be useful to document this as well, as we have `mkDefinedAtTop` and `mkDefined`?

PR #2017, opened 2020-07-27
Add context to some error messages

ttuegel:
---

###### Reviewer checklist


kore/src/Kore/Internal/Substitution.hs
```diff
@@ -419,7 +433,7 @@ wrap xs = Substitution xs
 -- this unless you are sure you need it.
 unsafeWrap
     :: HasCallStack
-    => Ord variable
+    => InternalVariable variable
     => [(SomeVariable variable, TermLike variable)]
```
andreiburdusa, line 437: I see Ana left a "TODO" comment about a posible refactoring on line 431. I could take care of it. What do you think @ana-pantilie @ttuegel ?

ana-pantilie: We already have `unsafeWrapFromAssignments`, and I think a major difference between these two is that `unsafeWrap` doesn't enforce the variable renaming invariant. I believe the question here is if we want to be able to create a `Substitution` like this, even for testing (where `unsafeWrap` is most widely used). @ttuegel what do you think? Is there any reason to keep `unsafeWrap` the way it is?

kore/src/Kore/Internal/Substitution.hs
```diff
@@ -437,14 +451,15 @@ unsafeWrap =
         -- and if this is an element variable substitution, the substitution
         -- must be defined.
         & assert (not $ isElementVariable var && isBottom termLike)
```
ttuegel, line 453:
```suggestion
        -- TODO (thomas.tuegel): isBottom -> isDefinedPattern
        & assert (not $ isElementVariable var && isBottom termLike)
```

Issue #2028, opened 2020-07-29
Rising memory use while proving


ttuegel:
Proving with `kore-exec` should have approximately constant memory use because the execution graph is not retained, but memory use rises with time. Memory used is similar to `kore-repl`, which retains the execution graph. Possible causes:

- The execution graph is accidentally retained.
- The cause of rising memory use isn't the execution graph.

We should run a heap profile to determine how memory use increases with time. After this is fixed, we should run integration tests with very tight heap limits to prevent the problem from recurring.

PR #2040, opened 2020-08-05
Kore.Step.Transition: Lazy orElse

ttuegel:
Implementing `ifte` for `TransitionT` allows us to write a simple, lazy implementation of `orElse`, which improves memory use (#2028). Normally, we think of laziness as increasing memory use, but the problem here is partial laziness: the former implementation of `orElse` would only force the _spine_ of the list and not the values (that is, `_ : _ : _ : []`) so we would end up with a large list of unevaluated thunks. This fully-lazy implementation of `orElse` only forces conses `(:)` on-demand.

I wrote this in a TDD style as a demonstration, but I also caught a bug doing it that way.

---

###### Reviewer checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -172,6 +172,11 @@
 
 - ignore: {name: "Redundant compare", within: [Kore.Syntax.Id]}
 
+- ignore:
+    name: "Use void"
+    within:
+      - Test.Kore.Step.Transition
+
 
 # Define some custom infix operators
 # - fixity: infixr 3 ~^#^~
--- a/kore/src/Kore/Step/Transition.hs
+++ b/kore/src/Kore/Step/Transition.hs
@@ -10,7 +10,9 @@ License     : NCSA
 
 module Kore.Step.Transition
     ( TransitionT (..)
+    , Transition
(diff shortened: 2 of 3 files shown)


ttuegel:
> Thanks for taking the time to explain your method and to splitting this up nicely into several commits. I see that red-green is standard terminology in TDD, but out of curiosity, what does AUR stand for?

Long story short, `AUR` stands for "refactor". The standard terminology is "red-green-refactor". To make the commit messages line up in the terminal, I use `GRN` for "green", but `REF` looks too much like `RED` when I'm scanning quickly. Some people say "red-green-gold" instead of "red-green-refactor", to keep the "color" theme. So, I might have done `RED-GRN-GLD`, but I still think that `GLD` looks too much like `GRN`. So, I went back to Latin ("aurum"): `RED-GRN-AUR`. I can glance at the last commit message and remember where I am in the cycle. This is my own personal quirk; there's no need for anyone else to adopt (or even understand) it! :smile:

ana-pantilie:
Oh, I see! That's a good idea. Also, "aur" is the Romanian word for gold. :smile:

Issue #2056, opened 2020-08-11
Unwrap Defined Map when evaluating MAP.concat


ttuegel:
When evaluating `MAP.concat` (during simplification and unification), we should unwrap a `Defined` term if its child is a builtin Map. If the child is not a builtin Map, then we should leave the `Defined` wrapper in place.

This can interfere with Map unification, causing bugs [like this](https://github.com/kframework/evm-semantics/pull/891#issuecomment-672020373).

This also applies to `SET.concat`.

PR #2063, opened 2020-08-12
Remove Defined wrapper around builtin collections when normalizing

ttuegel:
Fixes: #2056 
Fixes: #2057 

Ideally, there would only be one place this needs to change (one place where the internal representations of collections are constructed) but I'm not sure how to accomplish that right now.

---

###### Reviewer checklist


ttuegel:
> Are there any cases where we would like to parse a collection from a term and not normalize it?

Yes. When we are normalizing a builtin, some of the `opaque` terms may have been evaluated so that they have become builtins themselves. But, we don't want to normalize all the `opaque` terms, because any term which was not evaluated will just be wrapped again.

Issue #1278, opened 2019-11-20
Simplify right-hand side of claims


ttuegel:
Should we simplify the right-hand side of claims at startup? Does it matter, if we implement #1270?

yzhang90:
Suppose we have a claim
```
 rule <k> #start => #end </k> <a> X  => #fun2(X)</a>
```
and semantics rule
```
rule <k> #start => #step1 </k> <a> X </a> requires X > 0
```
In the `removeDestination` step after applying the semantics rule, even if we do unification first, and the unification in the <k> cell fails, we still need to evaluate  `#fun2(X)`, is it correct?

But in most cases, #fun2 would be either unsimplifiable or its argument is concrete. If it is the latter case, function memoization should be helpful. 

I would postpone the implementation of this until we met some concrete examples where simplifying  RHS will greatly improve the performance.

ttuegel:
Related: #1278, #1304

ttuegel:
We discussed this today. We might consider evaluating functions on the right-hand side, but avoiding other simplification which could cause branching.

ttuegel:
> Does it matter, if we implement #1270?

The answer to this question, now that #1270 is finished, is "yes, it matters".

ttuegel:
We presently cache all function evaluations with concrete (constructor-like) arguments. In the actual cases I have seen, this suffices to speed up proving, but I would like to investigate a little before closing this issue.

ttuegel:
#1945 documents a case where a valid claim cannot be proven because the right-hand side is not evaluated. Specifically, an `ensures` condition must be used to apply a user-defined equation.

ttuegel:
The solution here is to modify the internal representation of claims as follows:

```.hs
data ProtoReachabilityClaim =
    ProtoReachabilityClaim
    { left :: !(Pattern RewritingVariableName)
    , existentials :: ![RewritingVariable]
    , right :: !(OrPattern RewritingVariableName)
    }
```

When simplifying the left-hand side of the claim, the proof will branch when it encounters a disjunction. When simplifying the right-hand side of the claim, the backend will store any disjunction in the `OrPattern`. The `Simplify` step of the proof should simplify both sides, using the `requires` as a side-condition for simplifying the right-hand side.

ehildenb:
[...]

PR #2080, opened 2020-08-18
Simplify right-hand side of claims

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1278

After https://github.com/kframework/kore/pull/2025

---

###### Reviewer checklist


kore/src/Kore/Strategies/Goal.hs
```diff
@@ -846,16 +846,35 @@ simplify'
...
+    simplifiedDestination <-
+        OrPattern.observeAllT
+        $ Logic.scatter destination
+        >>= Pattern.simplify requiresSideCondition
+    let goal' =
+            Lens.set
+            (lensClaimPattern . field @"right")
+            (OrPattern.flatten simplifiedDestination)
+            goal
+    simplifyLeftHandSide goal'
```
ttuegel, line 866: Shouldn't we simplify the left-hand side first? We want the condition on the left-hand side to be fully simplified before we use it as a `SideCondition`. Also, we might be able to remove some branches, in which case it isn't necessary to simplify the right-hand side.

kore/src/Kore/Strategies/Goal.hs
```diff
@@ -846,16 +846,35 @@ simplify'
...
+            SideCondition.assumeTrueCondition leftCondition
+    simplifiedDestination <-
+        OrPattern.observeAllT
+        $ Logic.scatter destination
+        >>= Pattern.simplify requiresSideCondition
+    let goal' =
+            Lens.set
+            (lensClaimPattern . field @"right")
+            (OrPattern.flatten simplifiedDestination)
+            goal
```
ttuegel, line 865:
```suggestion
    goal' <-
        Lens.forOf (lensClaimPattern . field @"right") goal
        $ OrPattern.observeAllT
        $ Logic.scatter destination
        >>= Pattern.simplify requiresSideCondition
        >>= Logic.scatter
```

Also, I'm not sure `OrPattern.flatten` should exist. :thinking:

PR #2086, opened 2020-08-21
Make ErrorDecidePredicateUnknown an actual Error

andreiburdusa:
---

###### Reviewer checklist


kore/src/Kore/Log/ErrorDecidePredicateUnknown.hs
```diff
@@ -46,11 +56,10 @@ instance Entry ErrorDecidePredicateUnknown where
...
 
 errorDecidePredicateUnknown
-    :: MonadLog log
-    => InternalVariable variable
+    :: InternalVariable variable
     => NonEmpty (Predicate variable)
     -> log ()
 errorDecidePredicateUnknown predicates' =
-    logEntry ErrorDecidePredicateUnknown { predicates }
+    throw ErrorDecidePredicateUnknown { predicates }
```
ttuegel, line 63: This should be `throwM` from `Control.Monad.Catch`.

ttuegel (changes requested):
Except for the one comment, everything looks good. I'm not surprised by the tests that need the solver, those tests are more like integration tests than unit tests anyway.

ttuegel:
@andreiburdusa If this worked using `throw`, then let's forget about using `throwM`. Using `throw` will make it hard to add context to the error message, but we have exceeded the appetite for such a small issue as this.

kore/src/Kore/Log/ErrorDecidePredicateUnknown.hs
```diff
@@ -29,6 +34,11 @@ newtype ErrorDecidePredicateUnknown =
         }
     deriving (Show)
 
+instance Exception ErrorDecidePredicateUnknown where
+    toException = toException . SomeEntry
+    fromException exn =
+        fromException exn >>= \entry -> fromEntry entry
```
ttuegel, line 40:
```suggestion
    fromException exn = fromException exn >>= fromEntry
```

kore/test/Test/Kore/Step/Rule/Combine.hs
```diff
@@ -379,11 +380,20 @@ applyAlias name aliasRight =
...
+    $ runSimplifier Mock.env
+    $ mergeRules (rule :| rules)
+runMergeRulesNoSMT [] = error "Unexpected empty list of rules."
+
 runMergeRules
     :: [RewriteRule VariableName]
     -> IO [RewriteRule VariableName]
 runMergeRules (rule : rules) =
-    runNoSMT
+    runSMT (pure ())
```
ttuegel, line 396: Why doesn't this need the declarations from `Mock.env`?

ttuegel (approved):
I approved, but I would like to know why the tests don't need the SMT declarations.

kore/test/Test/Kore/Step/Rule/Combine.hs
andreiburdusa: I didn't investigate in depth, but it might be that it uses `runSimplifier Mock.env` on the next line

andreiburdusa: I didn't investigate in depth, but i think it might be that it uses `runSimplifier Mock.env` below

PR #2136, opened 2020-09-11
Infer that the right-hand side of claims is defined

### LINKED ISSUE #2110: Infer that the right-hand side of claims is defined
We already did this for the left-hand side, but at the time, we did not keep the right-hand side simplified.

See also: #1103

andreiburdusa:
---
Fixes #2110 
###### Reviewer checklist


kore/test/Test/Kore/Step/Rule/Simplify.hs
```diff
@@ -301,6 +314,38 @@ test_simplifyRule_OnePathRule =
             )
 
         assertEqual "" expected actual
+
+    , testCase "Case where f(x) in rhs is defined;\
```
andreiburdusa, line 318: I added this tests before I started working on the feature, as similar to the lhs case, but now I don't think it's needed anymore because of the updates to https://github.com/kframework/kore/pull/2136/files#diff-b59258b15feafb833a251d4b7ac31319R247

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -179,11 +179,20 @@ simplifyClaimRule =
     worker :: ClaimPattern -> LogicT simplifier ClaimPattern
     worker claimPattern = do
         let lhs = ClaimPattern.left claimPattern
-        simplified <- simplify lhs
-        let substitution = Pattern.substitution simplified
-            lhs' = simplified { Pattern.substitution = mempty }
+            rhs = ClaimPattern.right claimPattern
+        simplifiedLhs <- simplify lhs
+        simplifiedRhs <- OrPattern.filterOr <$> traverse simplify rhs
```
ttuegel, line 184: @ana-pantilie Aren't we already simplify the right-hand side of the claim somewhere else?

kore/src/Kore/Step/Rule/Simplify.hs
```diff
@@ -179,11 +179,20 @@ simplifyClaimRule =
...
+        let substitution =
+                Pattern.substitution simplifiedLhs
+                <> foldMap Pattern.substitution (MultiOr.getMultiOr simplifiedRhs)
```
ttuegel, line 187: We cannot add the substitution from the right-hand side to the global substitution. We can't apply it to the left-hand side, and it isn't right to drop it from the right-hand side (below) either.

andreiburdusa: Initially I didn't do that, but I tried it out of other ideas

andreiburdusa: Initially I didn't do that, but I tried it out of other ideas

kore/src/Kore/Step/Rule/Simplify.hs
ana-pantilie: Yes, in the `simplify` transition of the strategy: https://github.com/kframework/kore/blob/c28200738e0a5e8c2b9aaf6a6046b996f68082e3/kore/src/Kore/Strategies/Goal.hs#L838

test/imp/sum-save-proofs-spec.k.save-proofs.kore.golden
```diff
@@ -145,64 +145,135 @@ module haskell-backend-saved-claims-43943e50-f723-47cd-99fd-07104d664c6d
...
+                    ),
+                    /* Fl Fn D Spa */
+                    Lbl'-LT-'generatedCounter'-GT-'{}(
+                        /* Fl Fn D Sfa */ Var'Unds'DotVar3:SortInt{}
+                    )
+                ),
+                /* Spa */
+                \ceil{SortGeneratedTopCell{}, SortGeneratedTopCell{}}(
+                    /* Fn Spa */
+                    Lbl'-LT-'generatedTop'-GT-'{}(
```
ttuegel, line 214: I know this hasn't been simplified because `Lbl'-LT-'generatedTop'-GT-'{}` is a functional constructor.

ttuegel (changes requested):

ttuegel (approved):

Issue #1037, opened 2019-09-30
Surprising Functor, Traversable instances for OrPattern, OrPredicate


ttuegel:
The `Functor` and `Traversable` instances of `OrPattern` and `OrPredicate` do not preserve non-redundancy (under idempotence) of `\or` given a non-injective map, e.g.

```.hs
fmap (const z) [x, y] = [z, z]
```

This is not a _bug_ exactly, just an annoyance, but it is a thoroughly-solvable annoyance which has bitten us multiple times.

---

- [ ] Make `OrPattern` and `OrPredicate` `newtype`s over `MultiOr`.
- [ ] Define [`MonoFunctor`](https://www.stackage.org/haddock/lts-14.7/mono-traversable-1.0.12.0/Data-MonoTraversable.html#t:MonoFunctor) and `MonoTraversable` instances which use `MultiOr.filterOr` to remove redundancies.

PR #2139, opened 2020-09-14
Remove MultiAnd's Functor instance

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/1037

After https://github.com/kframework/kore/pull/2134

---

###### Reviewer checklist


kore/src/Kore/Internal/MultiAnd.hs
```diff
@@ -73,8 +76,7 @@ A non-empty 'MultiAnd' would also have a nice symmetry between 'Top' and
 newtype MultiAnd child = MultiAnd { getMultiAnd :: [child] }
     deriving (Eq, Ord, Show)
     deriving (Semigroup, Monoid)
-    deriving (Functor, Applicative, Monad, Alternative)
```
ttuegel, line 76: Oh, I had no idea this type defined an `Alternative` instance. That's really bad!

kore/src/Kore/Internal/MultiOr.hs
```diff
@@ -194,7 +194,7 @@ distributeAnd =
     foldr (crossProductGeneric and') (singleton MultiAnd.top)
   where
     and' term ma =
-        term : MultiAnd.extractPatterns ma & MultiAnd.make
+        term : Foldable.toList ma & MultiAnd.make
```
ttuegel, line 197: This would leave our options open in the future to pick a more efficient implementation:

```suggestion
        MultiAnd.singleton term <> ma
```

kore/src/Kore/Step/Simplification/And.hs
```diff
@@ -179,7 +179,11 @@ makeEvaluateNonBool notSimplifier sideCondition patterns = do
             let (term1, condition1) = Pattern.splitTerm pattern1
             unified <- termAnd notSimplifier term1 term2
             pure (Pattern.andCondition unified condition1)
-    unified <- Foldable.foldlM unify Pattern.top (term <$> patterns)
+    unified <-
+        Foldable.foldlM
+            unify
+            Pattern.top
+            (term <$> Foldable.toList patterns)
```
ttuegel, line 186:
```suggestion
            (MultiAnd.map term patterns)
```

ttuegel (approved):

kore/src/Kore/Step/Simplification/And.hs
ana-pantilie: If I remember correctly this will make some tests fail. I believe it's because there are patterns with the same term and MultiAnd will remove the duplicates. For some reason this seemed reasonable to me, but now that I think about it more maybe I should investigate and see if there's something else going wrong.

kore/src/Kore/Internal/MultiOr.hs
ana-pantilie: Wow, so this change made one of the `distributeAnd` unit tests fail, and it's because we're deriving the `Semigroup` (and `Monoid`) instance for `MultiAnd`. It's good that we caught this now.

kore/src/Kore/Step/Simplification/And.hs
ana-pantilie: I took another look at this and the unit test would fail because the resulting pattern would have sort `PREDICATE_` instead of the right sort. This is because when `MultiAnd` is `\top`, we lose all sort information.

ttuegel (approved):

PR #2154, opened 2020-09-16
mkDefined doesn't preserve "simplified" attribute

andreiburdusa:
---

###### Reviewer checklist


ttuegel:
I realized what's going on here: we shouldn't use `mkApplySymbol` (and the other `mk_`) functions in `mkDefined`, because these calculate new attributes and reset the `simplified` attribute. Instead, we should take the original attributes and update only the `defined` field.

kore/src/Kore/Internal/TermLike.hs
```diff
@@ -1470,22 +1470,38 @@ mkDefined = worker
...
+        let (attrs :< termF) = Recursive.project term
          in case termF of
-                AndF And { andFirst, andSecond } ->
+                AndF And { andSort, andFirst, andSecond } ->
                     mkDefined1
-                        ( mkAnd
-                            (worker andFirst)
-                            (worker andSecond)
+                        ( Recursive.embed
+                            ( attrs
```
ttuegel, line 1478: You will need to update the `defined` field of `attrs`.

kore/src/Kore/Internal/TermLike/TermLike.hs
```diff
@@ -798,6 +800,13 @@ traverseVariablesF adj =
 extractAttributes :: TermLike variable -> Attribute.Pattern variable
 extractAttributes (TermLike (attrs :< _)) = attrs
 
+markDefined :: TermLike variable -> TermLike variable
```
ttuegel, line 803: This function should not exist. The `defined` attribute should only be constructed through its `Synthetic` instance. Otherwise, it isn't safe to modify the `TermLike` after marking it defined.

ttuegel (changes requested):

kore/src/Kore/Internal/TermLike.hs
ttuegel:
```.hs
attrs' = attrs { defined = synthetic (Cofree.headF <$> And andSort
                                        (worker andFirst)
                                        (worker andSecond)) }
```

ttuegel (approved):

Issue #2146, opened 2020-09-15
Simplifying without a SideCondition should not reset simplification markers


ttuegel:
Patterns should not be re-simplified if the current `SideCondition` is the same as, or is _implied by_ the `SideCondition` that the pattern was already simplified with. We don't need to do anything complicated to detect the implication; in practice, this only happens when the current `SideCondition` is `\top`.

---

- [ ] Add a function `implies :: SideCondition variable -> SideCondition variable -> Bool`. The only case we need to handle is `implies _ Top = True`.
- [ ] When we check that the side condition has changed, we actually should check if `implies old new` and not re-simplify it that's true.

ttuegel:
This issue can't be solved because it would give inconsistent `simplified` attributes. Instead, we will move forward with implementing caching in the simplifier, which makes this issue irrelevant.

PR #2156, opened 2020-09-18
Do not resimplify if new side condition is top

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/2146

~Needs https://github.com/kframework/kore/pull/2169~
Needs https://github.com/kframework/kore/pull/2170

---

###### Reviewer checklist


ttuegel:
Your changes seem reasonable to me, so I guess we need to investigate why the given pattern isn't simplified. Maybe start by adding some unit tests? The output Kore pattern should have `/* comments */` indicating each simplified sub-term.

kore/src/Kore/Internal/Pattern.hs
```diff
@@ -135,6 +136,24 @@ isSimplified sideCondition (splitTerm -> (t, p)) =
     TermLike.isSimplified sideCondition t
     && Condition.isSimplified sideCondition p
 
+-- | Checks whether a pattern has simplified children. A term with
+-- a conjunction at the top is simplified if its subterms are simplified.
+hasSimplifiedChildren
+    :: SideCondition.Representation -> Pattern variable -> Bool
+hasSimplifiedChildren sideCondition patt =
+    let term1 = term patt
```
ttuegel, line 144: Is it necessary to check the children of the `term` here? I thought it would only be necessary to check the children of the `predicate`.

kore/src/Kore/Attribute/Pattern/Simplified.hs
```diff
@@ -13,6 +13,11 @@ module Kore.Attribute.Pattern.Simplified
     , simplifiedConditionally
     , simplifiableConditionally
     , unparseTag
+    -- * For testing
```
ttuegel, line 16: I don't think any of these exports are things that need to be hidden, so I don't think we need to express reservations about exporting them:
```suggestion
```

ttuegel (approved):
This looks good to me, but I had one question; something I was unsure about.

Could you also add a comment to `Kore.Step.Simplification.TermLike.simplify`, explaining that all the `term`s are simplified, but the `predicate`s may only be conjunctions of simplified clauses?

kore/test/Test/Kore/Step/Simplification/IntegrationProperty.hs
```diff
@@ -72,7 +72,7 @@ test_simplifiesToSimplified =
         simplified <- catch
             (evaluateT (Pattern.fromTermLike term))
             (exceptionHandler term)
-        (===) True (OrPattern.isSimplified sideRepresentation simplified)
+        (===) True (OrPattern.hasSimplifiedChildren sideRepresentation simplified)
```
ana-pantilie, line 75: @ttuegel Based on the discussion during the meeting, this should actually be `isSimplified`, right? Further down in this file, we're using the `Pattern` simplifier to simplify this, so we should make sure that the whole `Pattern` is simplified at the end.

ttuegel:
Alright, I think I finally understand what's happening from your unit test. Is this how you understand the problem, too? We start with a term like:

```
\equals(
    f(z), // simplified with condition (1)
    ctor() // simplified with condition (2)
) // simplified with unknown condition
```

Before this pull request, we would simplify the children first with `\top`:

```
// after simplifying children, before simplifying \equals:
\equals(
    f(z), // simplified with \top
    ctor(z) // simplified with \top
)
```

so that the final result is `simplified with \top`.

After the changes in this pull request, we get this intermediate result:

```
// after simplifying children, before simplifying \equals:
\equals(
    f(z), // simplified with (1)
    ctor(z) // simplified with (2)
)
```

so that the final result is `simplified with unknown condition`.

Here are some things we can do:

1. If `TermLike.simplify` would return `simplified with unknown condition`, then rewrite the attribute to have the current condition, even if that is `\top`.
2. Find out where we are simplifying with `\top` as a side condition (besides the first step, obviously) and stop doing that.
3. Don't worry about re-simplifying in this situation. The problem will go away on its own when we implement caching in the simplifier.

What do you think?

PR #2162, opened 2020-09-22
Use DeriveAnyClass

ttuegel:
The `DeriveAnyClass` extension cuts down on boilerplate by letting us derive empty instances.

---

###### Reviewer checklist


kore/src/Kore/Attribute/Functional.hs
```diff
@@ -25,27 +26,16 @@ Note: This attribute is also used to annotate axioms stating functionality
...
-    (<>) (Functional a) (Functional b) = Functional (a || b)
-
-instance Monoid Functional where
-    mempty = Functional False
+    deriving (Eq, Ord, Show)
+    deriving (GHC.Generic)
+    deriving anyclass (Hashable, NFData)
+    deriving anyclass (SOP.Generic, SOP.HasDatatypeInfo)
+    deriving anyclass (Debug, Diff)
+    deriving (Semigroup, Monoid) via Monoid.Any
```
ana-pantilie, line 34: So both `Functional` and `Any` are newtypes over `Bool`, and you can derive `Functional`'s `Monoid` instance from the one defined on `Any`. This is nice!

kore/src/Kore/Reachability/Claim.hs
```diff
@@ -208,25 +208,13 @@ data AppliedRule claim
...
-
-instance SOP.HasDatatypeInfo claim => SOP.HasDatatypeInfo (AppliedRule claim)
-
-instance
-    ( Debug claim
-    , SOP.HasDatatypeInfo claim
-    , Debug (Rule claim)
-    , SOP.HasDatatypeInfo (Rule claim)
-    ) => Debug (AppliedRule claim)
+instance (Debug claim, Debug (Rule claim)) => Debug (AppliedRule claim)
```
ana-pantilie, line 213: We can't use deriving here because of the constraints? Does this have anything to do with `Rule` being a type family?

kore/src/Kore/Reachability/Claim.hs
```diff
@@ -208,25 +208,13 @@ data AppliedRule claim
...
 
 instance
-    ( Diff claim
-    , Debug claim
-    , SOP.HasDatatypeInfo claim
-    , Diff (Rule claim)
-    , Debug (Rule claim)
-    , SOP.HasDatatypeInfo (Rule claim)
+    ( Diff claim, Debug claim
+    , Diff (Rule claim), Debug (Rule claim)
```
ana-pantilie, line 217: Same question as above.

kore/src/Kore/Syntax/Application.hs
```diff
@@ -79,20 +71,18 @@ data Application head child = Application
...
+    deriving anyclass (Hashable, NFData)
+    deriving anyclass (SOP.Generic, SOP.HasDatatypeInfo)
 
 instance (Debug head, Debug child) => Debug (Application head child)
 
 instance
-    ( Debug head, Debug child, Diff head, Diff child )
+    ( Debug head, Diff head
+    , Debug child, Diff child
+    )
```
ana-pantilie, line 85: Same question as above.

ana-pantilie (approved):
This looks great to me! I only had a few questions.

andreiburdusa (approved):

kore/src/Kore/Reachability/Claim.hs
ttuegel: Yes, the compiler can't infer the constraint because it involves a type family.

kore/src/Kore/Syntax/Application.hs
ttuegel: The compiler can't infer this constraint, but it's for a different reason: There is an overloaded instance of `Debug` for `String` (so that the instance for `String` is different from the instance for `[a]`). There is a `[child]` in `Application`, so instance resolution is ambiguous. If `child ~ Char` then the compiler could pick the `String` instance, but not in the general case. By adding the `Debug child` constraint, we force the compiler to pick the general instance.

PR #2165, opened 2020-09-23
Normalize order of terms in \equals

### LINKED ISSUE #2137: Normalize order of terms in \equals
During simplification, the backend should normalized the order of terms in `\equals(_, _)` to remove duplicate clauses.

andreiburdusa:
---
Fixes #2137 
###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -717,11 +717,12 @@ functionAnd first second
     & pure
   | otherwise = empty
   where
-    (first', second') =
-        case compareForEquals first second of
-            LT -> (first, second)
-            _  -> (second, first)
+    (first', second') = minMaxBy compareForEquals first second
 
+
+{- | Gives an order for terms in \equals(_, ) to avoid duplicate clauses
+
+-}
 compareForEquals
     :: InternalVariable variable
     => TermLike variable
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -720,7 +720,9 @@ functionAnd first second
     (first', second') = minMaxBy compareForEquals first second
 
 
-{- | Gives an order for terms in \equals(_, ) to avoid duplicate clauses
+{- | Normal ordering for terms in @\equals(_, _)@.
+
+The normal ordering is arbitrary, but important to avoid duplication.
 
 -}
 compareForEquals
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -728,9 +728,9 @@ compareForEquals
     -> TermLike variable
     -> Ordering
 compareForEquals first second
-      | isConstructorLike first = LT
-      | isConstructorLike second = GT
-      | otherwise = compare first second
+  | isConstructorLike first = LT
+  | isConstructorLike second = GT
+  | otherwise = compare first second
 
 bytesDifferent
     :: InternalVariable variable
--- a/kore/src/Kore/Step/Simplification/AndTerms.hs
+++ b/kore/src/Kore/Step/Simplification/AndTerms.hs
@@ -17,6 +17,7 @@ module Kore.Step.Simplification.AndTerms
     , functionAnd
     , equalsFunctions
     , andFunctions
+    , compareForEquals
     ) where
 
 import Prelude.Kore
@@ -716,10 +717,20 @@ functionAnd first second
     & pure
   | otherwise = empty
   where
-    (first', second')
-      | isConstructorLike first = (first, second)
-      | isConstructorLike second = (second, first)
-      | otherwise = minMax first second
+    (first', second') =
+        case compareForEquals first second of
+            LT -> (first, second)
+            _  -> (second, first)
+
+compareForEquals
+    :: InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+    -> Ordering
+compareForEquals first second
+      | isConstructorLike first = LT
+      | isConstructorLike second = GT
+      | otherwise = compare first second
 
 bytesDifferent
     :: InternalVariable variable
--- a/kore/src/Kore/Step/Simplification/Exists.hs
+++ b/kore/src/Kore/Step/Simplification/Exists.hs
@@ -245,8 +245,9 @@ matchesToVariableSubstitution
   , Substitution.null boundSubstitution
   , not (TermLike.hasFreeVariable (inject $ variableName variable) term)
   = do
-    matchResult <- matchIncremental first second
-    case matchResult of
+    matchResultFS <- matchIncremental first second
+    matchResultSF <- matchIncremental second first
+    case matchResultFS <|> matchResultSF of
         Just (Predicate.PredicateTrue, results) ->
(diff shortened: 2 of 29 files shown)


andreiburdusa:
The order in [Kore.Step.Simplification.Equals.simplify](https://github.com/kframework/kore/pull/2165/files#diff-ad8c8821fcc726f3ca662ce3f975422fR169) might not be consistent with the order in [Kore.Step.Simplification.AndTerms.functionAnd](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/AndTerms.hs#L722).
I think the problem is that the arguments of `makeEvaluate` [here](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/Equals.hs#L206) are converted from `OrPattern` to `Pattern`, leading to different order: see [this failing test](https://github.com/kframework/kore/pull/2165/files#diff-f67488b99b5783d15108b7fca05d8340R58)

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -716,10 +717,20 @@ functionAnd first second
...
+
+compareForEquals
+    :: InternalVariable variable
+    => TermLike variable
+    -> TermLike variable
+    -> Ordering
+compareForEquals first second
+      | isConstructorLike first = LT
+      | isConstructorLike second = GT
+      | otherwise = compare first second
```
ttuegel, line 733:
```suggestion
  | isConstructorLike first = LT
  | isConstructorLike second = GT
  | otherwise = compare first second
```

kore/src/Kore/Step/Simplification/Exists.hs
```diff
@@ -245,8 +245,9 @@ matchesToVariableSubstitution
   , Substitution.null boundSubstitution
   , not (TermLike.hasFreeVariable (inject $ variableName variable) term)
   = do
-    matchResult <- matchIncremental first second
-    case matchResult of
+    matchResultFS <- matchIncremental first second
+    matchResultSF <- matchIncremental second first
+    case matchResultFS <|> matchResultSF of
```
ttuegel, line 250: Why is this necessary?

ttuegel (changes requested):

kore/test/Test/Kore/Step/Simplification/Exists.hs
```diff
@@ -280,7 +280,7 @@ test_makeEvaluate =
                         [(inject Mock.x, gOfA)]
                     }
         assertEqual "exists reevaluates" expect actual
-    , testCase "exists matches equality if result is top" $ do
+    , testCase "qqexists matches equality if result is top" $ do
```
ana-pantilie, line 283:
```suggestion
    , testCase "exists matches equality if result is top" $ do
```

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -716,10 +717,20 @@ functionAnd first second
...
   where
-    (first', second')
-      | isConstructorLike first = (first, second)
-      | isConstructorLike second = (second, first)
-      | otherwise = minMax first second
+    (first', second') =
+        case compareForEquals first second of
+            LT -> (first, second)
+            _  -> (second, first)
+
```
ana-pantilie, line 724: Can this be factored out? It seems to be the same logic applied both here and below.

kore/src/Kore/Step/Simplification/Exists.hs
andreiburdusa: Without it, the `actual` result from [this test](https://github.com/kframework/kore/blob/master/kore/test/Test/Kore/Step/Simplification/Exists.hs#L283) would be
```
\and{testSort{}}(
    /* term: */
    /* Created: Test.Kore.Step.Simplification.Exists.f */
    /* Fn Spa */
    f{}(/* Fl Fn D Sfa Cl */ a{}()),
\and{testSort{}}(
    /* predicate: */
    /* Sfa */
    \exists{testSort{}}(
        x:testSort{},
        /* Sfa */
        \equals{testSort{}, testSort{}}(
            /* Fn Sfa */ f{}(/* Fl Fn D Sfa Cl */ a{}()),
            /* Fn Sfa */ f{}(/* Fl Fn D Sfa */ x:testSort{})
        )
    ),
    /* substitution: */
    /* Spa */
    \equals{testSort{}, testSort{}}(
        /* Fl Fn D Sfa */ y:testSort{},
        /* Fn Sfa */ f{}(/* Fl Fn D Sfa Cl */ a{}())
    )
))
```
instead of having the expected `\top` predicate.

kore/src/Kore/Step/Simplification/AndTerms.hs
andreiburdusa: The result of the comparison is also needed [here](https://github.com/kframework/kore/pull/2165/files#diff-ad8c8821fcc726f3ca662ce3f975422fR171), but not its arguments, so I thought it's better to have `compareForEquals` compute the `Ordering` and let each function use it the way it needs.

kore/src/Kore/Step/Simplification/Exists.hs
ttuegel: This predicate can't exist at this point, because it's not fully simplified:

```
        \equals{testSort{}, testSort{}}(
            /* Fn Sfa */ f{}(/* Fl Fn D Sfa Cl */ a{}()),
            /* Fn Sfa */ f{}(/* Fl Fn D Sfa */ x:testSort{})
        )
```

So, we will leave this as-is for the purposes of this pull request, but please add a `TODO` to this function (`matchesToVariableSubstitution`) saying that it must go away.

kore/src/Kore/Step/Simplification/AndTerms.hs
ttuegel: Please add a function `minMaxBy :: (a -> a -> Ordering) -> (a, a) -> (a, a)` to `Prelude.Kore` like [`sortBy`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:sortBy). Here we can have `minMaxBy compareForEquals ...` and below we will have `minMaxBy (on compareForEquals OrPattern.toTermLike) ...`.

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -716,10 +717,20 @@ functionAnd first second
...
+compareForEquals
```
ttuegel, line 725: Please add documentation for this function, in particular, what does "for equals" mean in this context?

ttuegel (changes requested):

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -716,10 +717,21 @@ functionAnd first second
...
   | otherwise = empty
   where
-    (first', second')
-      | isConstructorLike first = (first, second)
-      | isConstructorLike second = (second, first)
-      | otherwise = minMax first second
+    (first', second') = minMaxBy compareForEquals first second
+
+
+{- | Gives an order for terms in \equals(_, ) to avoid duplicate clauses
```
ttuegel, line 723:
```suggestion
{- | Normal ordering for terms in @\equals(_, _)@.

The normal ordering is arbitrary, but important to avoid duplication.
```

ttuegel (approved):

PR #2169, opened 2020-09-25
Use hasSimplifiedChildren in term simplifier check

ana-pantilie:
---

###### Reviewer checklist


ttuegel (approved):

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -443,7 +445,7 @@ ensureSimplifiedResult
     -> OrPattern variable
     -> simplifier (OrPattern variable)
 ensureSimplifiedResult repr termLike results
-  | isSimplified results = pure results
+  | hasSimplifiedChildren results = pure results
```
ttuegel, line 448: We should revert this change, because the `TermLike` simplifier should actually return a fully-simplified condition.

PR #2170, opened 2020-09-25
Move simplifyConjunctionByAssumption to Condition simplifier

ana-pantilie:
---

###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`

PR #2172, opened 2020-09-29
Reset the SMT solver at fixed intervals

### LINKED ISSUE #2125: Reset the SMT solver at fixed intervals
Follow-up: #2087 

Restarting the SMT solver at the start of each proof step is not sufficient if there are deeply-recursive symbolic function calls. The encapsulation of the SMT interface is also slightly broken by the need to call `reset` from the reachability proof module.

Instead, the SMT solver interface should keep an internal counter of the number of calls to `withSolver`. After that number exceeds some limit, the solver should reset itself (at the end of a call to `withSolver`, when we are guaranteed that the solver is not in use).

MirceaS:
---
Fixes #2125 
###### Reviewer checklist

### DIFF
--- a/kore/src/SMT.hs
+++ b/kore/src/SMT.hs
@@ -315,7 +315,7 @@ instance MonadSMT SMT where
             SMT $ Reader.local
                 (\handle -> handle {mSolverHandle = mvar})
                 (Exception.finally (getSMT smt) $ do
-                    Trans.liftIO $ tryPutMVar mvar solverHandle
+                    _ <- Trans.liftIO $ tryPutMVar mvar solverHandle
                     Trans.liftIO $ pop solver
                     State.modify (+ 1)
                     counter <- State.get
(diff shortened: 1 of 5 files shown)


kore/src/SMT.hs
```diff
@@ -321,14 +324,22 @@ instance MonadSMT SMT where
...
+                        counter
                     )
                     (Trans.liftIO $ pop solver)
-                )
+            SMT $ State.modify (+ 1)
+            counterNew <- SMT State.get
+            -- Due to an issue with the SMT solver, we need to
+            -- reinitialise it after a number of runs, specified here.
+            -- This number can be adjusted based on experimentation
+            when (counterNew >= 100) reinit
```
ttuegel, line 341:
```suggestion
            when (counterNew >= 100) $ (SMT . RWST) $ \ _ _ ->
                    runRWST
                        (getSMT reinit)
                        (SolverInitAndHandle userInit mvar config)
                        counter
```

MirceaS:
Note: The code currently assumes that `smt` is well behaved in the sense that it puts back the shared variable so that the potential `reinit` doesn't block

ttuegel (changes requested):
I requested some changes, but I'm going to take care of them right now as part of my clean-up because I think we need it for #2170.

ttuegel (approved):

PR #2197, opened 2020-10-16
Save proofs even when successful

### LINKED ISSUE #2183: Save proofs even when successful
`--save-proofs` should save the trusted claims even when the specification is successfully proved.

ttuegel:
Fixes #2183 

---

###### Reviewer checklist

### DIFF
--- a/kore/src/Kore/Reachability/Prove.hs
+++ b/kore/src/Kore/Reachability/Prove.hs
@@ -189,8 +189,9 @@ type ProvenClaims = MultiAnd SomeClaim
 -- | The result of proving some claims.
 data ProveClaimsResult =
     ProveClaimsResult
-    { stuckClaim :: !StuckClaims
-    -- ^ The claims which were stuck, if any.
+    { stuckClaims :: !StuckClaims
+    -- ^ The conjuction of stuck claims, that is: of claims which must still be
+    -- proven. If all claims were proved, then the remaining claims are @\\top@.
     , provenClaims :: !ProvenClaims
     -- ^ The conjunction of all claims which were proven.
     }
@@ -222,7 +223,7 @@ proveClaims
         & runExceptT
         & flip runStateT (MultiAnd.make stillProven)
     pure ProveClaimsResult
-        { stuckClaim = either id (const MultiAnd.top) result
+        { stuckClaims = either id (const MultiAnd.top) result
         , provenClaims
         }
   where
--- a/kore/src/Kore/Reachability/Prove.hs
+++ b/kore/src/Kore/Reachability/Prove.hs
@@ -176,13 +176,16 @@ newtype AlreadyProven = AlreadyProven {getAlreadyProven :: [Text]}
 newtype StuckClaim =
     StuckClaim { getStuckClaim :: OrPattern VariableName }
 
+type ProvenClaims = MultiAnd SomeClaim
+
+-- | The result of proving some claims.
 data ProveClaimsResult =
     ProveClaimsResult
     { stuckClaim :: !(Maybe StuckClaim)
-    , provenClaims :: !(MultiAnd SomeClaim)
+    -- ^ The first claim which was stuck, if any.
+    , provenClaims :: !ProvenClaims
+    -- ^ The conjunction of all claims which were proven.
     }
-
-type ProvenClaims = MultiAnd SomeClaim
 
 proveClaims
     :: forall simplifier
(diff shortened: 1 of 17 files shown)


ttuegel:
This should just about be ready. I only need to add some documentation for types I added.

test/save-proofs/test-2-spec.k
```diff
@@ -10,6 +10,8 @@ endmodule
 module TEST-2-SPEC
   imports VERIFICATION
 
+  // This claim would fail, but it is loaded from the saved claims.
   rule BB(X:Int) => CC(X:Int)
+  // This claim requires the first claim.
```
ana-pantilie, line 15: Is this test missing an expected output?

test/save-proofs/test-1-spec.k
```diff
@@ -1,13 +1,12 @@
-// Two claims, one passes and is saved to a file, the other one fails.
 module VERIFICATION
   imports SAVE-PROOFS
 
   rule BB(X:Int) => CC(X:Int)
 endmodule
 
+// This claim is proven and saved.
```
ana-pantilie, line 7: Where is it saved?

kore/test/Test/Kore/Reachability/Prove.hs
```diff
@@ -69,308 +61,127 @@ import Test.Tasty.HUnit.Ext
...
+                        Unlimited
+                        (Limit 3)
+                        [ simpleAxiom Mock.a Mock.a ]
+                        [ mkSimpleClaim Mock.a Mock.b ]
+                        []
+                    assertEqual "" MultiAnd.top actual
+        in
+        [ mkTest "OnePath" simpleOnePathClaim
+        , mkTest "AllPath" simpleAllPathClaim
+        ]
```
ana-pantilie, line 184: Any reason why we don't want the mixed cases anymore?

kore/test/Test/Kore/Reachability/Prove.hs
```diff
@@ -402,28 +211,7 @@ test_reachabilityVerification =
...
-            , simpleAxiom Mock.a Mock.c
-            ]
-            [ simpleOnePathClaim Mock.a (mkOr Mock.b Mock.c)
-            , simpleOnePathClaim Mock.a (mkOr Mock.b Mock.c)
-            ]
-            []
-        assertEqual ""
-            (Right ())
-            actual
+        assertEqual "" MultiAnd.top actual
```
ana-pantilie, line 214: Could these tests be combined like the tests above were?

ana-pantilie (commented):
I added some questions.

In addition, I saw that you cleaned up the tests in `kore/test/Test/Kore/Reachability/Prove.hs`, but some of them kept the previous structure (I added a comment on one instance of this). Is there any reason for this, or could we (the reviewers) try to clean up those tests as well, as part of the reviewer refactoring?

kore/src/Kore/Reachability/Prove.hs
```diff
@@ -160,31 +165,35 @@ The action may throw an exception if the proof fails; the exception is a single
...
+    isBottom = const False
+
+type StuckClaims = MultiAnd StuckClaim
+
+type ProvenClaims = MultiAnd SomeClaim
+
+-- | The result of proving some claims.
+data ProveClaimsResult =
+    ProveClaimsResult
+    { stuckClaim :: !StuckClaims
```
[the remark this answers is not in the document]
ttuegel: > Maybe a middle ground solution would be `stuckClaims`, with the check in Main (line 703) renamed to `noStuckClaims` (or something similar) in a where clause.

That sounds good to me!

> Another question, in what situation can we have more than one stuck claim? I thought we stop execution when we find the first one.

When we exceed the breadth limit, we will present all the branches for the user to help with debugging.

kore/test/Test/Kore/Reachability/Prove.hs
ttuegel: The only thing that these mixed cases actually tested is that the backend doesn't try to use one-path claims to prove all-path claims, or vice versa. There are already tests specifically for that, and I didn't think it made sense to duplicate that check too many times.

kore/test/Test/Kore/Reachability/Prove.hs
ttuegel: Yes, I cleaned up the test suite some more.

ana-pantilie (approved):

andreiburdusa (approved):

Issue #2204, opened 2020-10-16
kore-exec: --strategy all


ttuegel:
- [ ] The default `--strategy` option should be `all`.
- [ ] The `heat-cool` strategy options should be removed.
- [ ] The code associated with those strategies should also be removed.

PR #2209, opened 2020-10-20
kore-exec: make --strategy all default; clean-up 

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/2204

---

###### Reviewer checklist


kore/app/exec/Main.hs
```diff
@@ -369,15 +369,13 @@ parseKoreExecOptions startTime =
             <> long "strategy"
             -- TODO (thomas.tuegel): Make defaultStrategy the default when it
```
ttuegel, line 370: You can remove this `TODO` also. :smiley:

ttuegel (approved):
You can also remove the `TODO` I indicated. This looks good! I realized that the integration test I was thinking of is actually for the _other_ `--strategy` issue, the one which @andreiburdusa is working on. So we don't need a test for this.

Issue #2207, opened 2020-10-19
Solver reset interval should be configurable


ttuegel:
- [ ] Replace [this constant](https://github.com/kframework/kore/blob/289f9d520e6a506e9ea6f1558589e576af3a5c1b/kore/src/SMT.hs#L353) with a field in the [SMT configuration](https://github.com/kframework/kore/blob/289f9d520e6a506e9ea6f1558589e576af3a5c1b/kore/src/SMT.hs#L426).
- [ ] Add a command-line option `--solver-reset-interval` which takes a positive integer and sets the appropriate value in the configuration.

ttuegel:
@ehildenb This would allow you to tweak the reset interval for the SMT solver and work around the issues you're having with KEVM. In the worst case, you could set `--solver-reset-interval 1` and every query would run in a fresh solver (at some performance penalty).

ehildenb:
@ttuegel maybe instead if the backend detects a solver anomaly, it should reset the solver immediately and re-try the query?

You could still set a manual reset interval, but I think the backend should try to recover from solver anomalies automatically if possible.

PR #2220, opened 2020-10-22
kore-exec: Add option --solver-reset-interval

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/2207

---

###### Reviewer checklist

### DIFF
--- a/kore/app/exec/Main.hs
+++ b/kore/app/exec/Main.hs
@@ -167,6 +167,7 @@ import Kore.Unparser
 import Options.SMT
     ( KoreSolverOptions (..)
     , Solver (..)
+    , ensureSmtPreludeExists
     , parseKoreSolverOptions
     , unparseKoreSolverOptions
     , writeKoreSolverFiles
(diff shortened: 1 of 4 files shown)


kore/src/SMT.hs
```diff
@@ -435,6 +445,8 @@ data Config =
         -- ^ optional log file name
         , timeOut :: TimeOut
         -- ^ query time limit
+        , resetInterval :: ResetInterval
```
ana-pantilie, line 448: Why aren't the fields in `Config` strict?

kore/app/exec/Main.hs
```diff
@@ -355,12 +362,24 @@ parseKoreExecOptions startTime =
...
+
     SMT.Config { timeOut = defaultTimeOut } = SMT.defaultConfig
-    readSMTTimeOut = do
-        i <- auto
-        if i <= 0
-            then readerError "smt-timeout must be a positive integer."
-            else return $ SMT.TimeOut $ Limit i
+    SMT.Config { resetInterval = defaultResetInterval } = SMT.defaultConfig
+
+    readPositiveInteger ctor optionName = do
```
ana-pantilie, line 369: Some of the functions here (including this one) are needed in `kore-repl` as well, so they appear duplicated there. Could we have a clean-up issue to move them all to a separate, common module (maybe `GlobalMain.hs`)? Or would that not work for some reason?

kore/src/SMT.hs
ttuegel: This was probably a mistake.

kore/app/exec/Main.hs
ttuegel: We should just write a separate command-line option parser for `SMT.Config`. It can go in `Options.SMT` in the library. Please do this as part of this pull request.

I don't want to put new things in `GlobalMain`; actually, I want to move everything out of there.

ttuegel (approved):

PR #2224, opened 2020-10-23
Retry SMT query once if it fails

ana-pantilie:
---

###### Reviewer checklist


kore/src/Kore/Log/ErrorDecidePredicateUnknown.hs
```diff
@@ -58,7 +58,7 @@ instance Entry ErrorDecidePredicateUnknown where
 errorDecidePredicateUnknown
     :: InternalVariable variable
     => NonEmpty (Predicate variable)
-    -> log ()
+    -> log Bool
```
ttuegel, line 61: This function never returns, so the return type can be anything:

```suggestion
    -> log a
```

kore/src/Kore/Log/WarnRetrySolverQuery.hs
```diff
@@ -0,0 +1,68 @@
...
+        Pretty.vsep $
+        [ "The SMT solver failed to solve the following query:"
+        , Pretty.indent 2 "Decide predicate:"
+        , Pretty.indent 4 (unparse predicate)
+        , Pretty.indent 2 "with side conditions:"
+        ]
+        <> fmap (Pretty.indent 4 . unparse) sideConditions
+        <> ["The SMT solver will be reset and the query\
+            \ will be tried one more time."
+           ]
```
ttuegel, line 48: Let's change the wording here, now that the warning is only shown after the second query succeeds (otherwise an error is thrown).

ttuegel (approved):
I approved, with two small suggestions.

PR #2231, opened 2020-10-29
Do not ignore `--strategy all`

andreiburdusa:
---
Fixes #2205 
###### Reviewer checklist

- [ ] Test coverage: `stack test --coverage`
- [ ] Public API documentation: `stack haddock`

Issue #2205, opened 2020-10-18
`--strategy all` is ignored


nishantjr:
This happens for both non-deterministic semantics, and when multiple branches are available for narrowing.
We expect both the following tests to result in a disjunction of configurations (two for the first, three for the second).

```
$ cat test.k
module TEST
    imports INT
    configuration <k> init </k>
    syntax KItem ::= "init"
    rule <k> init => 1 </k>
    rule <k> init => 2 </k>
endmodule

$ kompile test.k --backend haskell
$ krun --haskell-backend-command 'kore-exec --strategy all'
<k>
  2 ~> .
</k>
```

```
$ cat test.k 
module TEST
    imports INT
    configuration <k> init </k>
    syntax KItem ::= "init" | "one" | "two"
    rule <k> init => ?_:Int </k>
    rule <k> 1 => one </k>
    rule <k> 2 => two </k>
endmodule
$ KORE_EXEC_OPTS="--strategy all" krun
  <k>
    two ~> .
  </k>
#And
  {
    'QuesUnds'0
  #Equals
    2
  }
```

ttuegel:
- [This](https://github.com/kframework/kore/blob/11e10fbb61a31ef555f15d44721dc6253ab787e7/kore/src/Kore/Exec.hs#L284) should change to `OrPattern` (the configurations)
- [Apply](https://github.com/kframework/kore/blob/11e10fbb61a31ef555f15d44721dc6253ab787e7/kore/src/Kore/Exec.hs#L291) the `getExitCode` symbol to each configuration.
- Call `MultiOr.gather` [here](https://github.com/kframework/kore/blob/11e10fbb61a31ef555f15d44721dc6253ab787e7/kore/src/Kore/Exec.hs#L290).
- In the [case](https://github.com/kframework/kore/blob/11e10fbb61a31ef555f15d44721dc6253ab787e7/kore/src/Kore/Exec.hs#L293), we should return the exit code if all the branches agree about the exit code, otherwise we [return](https://github.com/kframework/kore/blob/11e10fbb61a31ef555f15d44721dc6253ab787e7/kore/src/Kore/Exec.hs#L297) `111`

PR #2234, opened 2020-11-01
Symbolic execution and search should apply rules correctly

ana-pantilie:
Part of https://github.com/kframework/kore/issues/2205

~Needs https://github.com/kframework/kore/pull/2231~
~Needs https://github.com/kframework/kore/pull/2261~

---

###### Reviewer checklist


kore/src/Kore/Step.hs
```diff
@@ -118,6 +136,59 @@ transitionRule =
...
+    transitionRuleWorker Begin (Remaining _) = empty
+    transitionRuleWorker Begin state = pure state
+
+    transitionRuleWorker Simplify (Rewritten patt) =
+        Rewritten <$> transitionSimplify patt
+    transitionRuleWorker Simplify (Remaining patt) =
+        Remaining <$> transitionSimplify patt
+
+    transitionRuleWorker (Rewrite rule) (Rewritten patt) =
+        transitionRewrite rule patt
```
ttuegel, line 161: We should not rewrite states that have already been rewritten. Only one rewrite rule should apply in each `transitionRule` step.

kore/src/Kore/Step.hs
```diff
@@ -153,6 +224,7 @@ priorityAllStrategy
     => [rewrite]
     -> Strategy (Prim rewrite)
 priorityAllStrategy rewrites =
+    -- TODO: should this be Strategy.all?
```
ttuegel, line 227: Neither `all` nor `first` is really correct. The rules should be grouped by priority and applied in parallel or sequence in the transition rule, just the same as is done in `Kore.Reachability.Claim`.

kore/src/Kore/Step.hs
ana-pantilie: This would mean that we need another state, something like `Claimed` from the reachability `ClaimState`, right?

ttuegel: I think `Remainder` serves that purpose here. We just should not apply rules to the `Rewritten` part.

test/issue-2010/1.test.out.golden
```diff
@@ -1,3 +1,19 @@
...
+  #And
+    #Not ( {
+      None
+    #Equals
+      f ( )
+    } )
+  #And
+    <k>
+      f ( ) ~> .
+    </k>
```
ana-pantilie, line 15: I'm not 100% certain of the behavior `K` is expected to have here. These are two behaviors I could think of, and I don't know which is expected:

1. Since `Some` and `None` are the only (visible) constructors for `MaybeInt`, this branch should be `\\bottom`.
2. Since `K` sorts are open, we do not assume that the only way to construct sort `MaybeInt` is with `Some` or with `None`, so this branch should exist.

ttuegel: (1) is the correct behavior, and we rely on the SMT solver for this, so something is not being sent correctly to the solver.

kore/app/exec/Main.hs
```diff
@@ -349,13 +342,13 @@ parseKoreExecOptions startTime =
...
-            <> value ("all", priorityAllStrategy)
+            <> value ("all", All)
             <> help "Select rewrites using STRATEGY."
             )
       where
         strategies =
-            [ ("any", priorityAnyStrategy)
-            , ("all", priorityAllStrategy)
+            [ ("any", Any)
+            , ("all", All)
```
ttuegel, line 351: Because this uses an enumeration type now, there is no reason to keep the `String` part of the tuple.

kore/src/Kore/Exec.hs
```diff
@@ -213,14 +209,15 @@ exec
         , MonadProf smt
         )
     => Limit Natural
+    -> Limit Natural
     -> VerifiedModule StepperAttributes
     -- ^ The main module
-    -> ([Rewrite] -> [Strategy (Prim Rewrite)])
-    -- ^ The strategy to use for execution; see examples in "Kore.Step.Step"
+    -> ExecutionMode
+    -- ^ The execution mode
```
ttuegel, line 216: I don't think this documentation is helpful. The reader can always jump to the documentation for `ExecutionMode`!

```suggestion
```

kore/src/Kore/Step.hs
```diff
@@ -49,119 +57,152 @@ import qualified Kore.Step.Simplification.Pattern as Pattern
...
 import qualified Kore.Step.Strategy as Strategy
-import qualified Kore.Step.Transition as Transition
 import qualified Kore.Unification.Procedure as Unification
 
-{- | A strategy primitive: a rewrite rule or builtin simplification step.
- -}
-data Prim rewrite = Simplify | Rewrite !rewrite
+{- | The program's state during symbolic execution.
+-}
+data ProgramState a = Start !a | Rewritten !a | Remaining !a
```
ttuegel, line 77: Could you please add documentation for each constructor?

kore/src/Kore/Step.hs
```diff
@@ -49,119 +57,152 @@ import qualified Kore.Step.Simplification.Pattern as Pattern
...
+limitedExecutionStrategy :: Limit Natural -> [Strategy Prim]
+limitedExecutionStrategy depthLimit =
+    Limit.takeWithin depthLimit (toList executionStrategy)
+
+{- The primitive transitions for the symbolic execution strategy.
+-}
+data Prim
+    = Begin
+    | Simplify
+    | ApplyRewrites
```
ttuegel, line 113: Unless this name is already taken, I suggest changing to a simple imperative verb to match the other constructors:

```suggestion
    | Rewrite
```

kore/src/Kore/Step.hs
```diff
@@ -49,119 +57,152 @@ import qualified Kore.Step.Simplification.Pattern as Pattern
...
+
+    transitionRuleWorker All ApplyRewrites (Remaining patt) =
+        transitionAllRewrite patt
+    transitionRuleWorker All ApplyRewrites (Start patt) =
+        transitionAllRewrite patt
+    transitionRuleWorker Any ApplyRewrites (Remaining patt) =
+        transitionAnyRewrite patt
+    transitionRuleWorker Any ApplyRewrites (Start patt) =
+        transitionAnyRewrite patt
+    transitionRuleWorker _ ApplyRewrites state = pure state
```
ttuegel, line 158: We can reduce duplication with a simple helper function. Explicitly matching on all the `ProgramState` constructors makes the logic clearer and protects us if we ever add constructors to `ProgramState`.

```suggestion
    transitionRuleWorker mode ApplyRewrites (Remaining patt) =
        transitionRewrite mode patt
    transitionRuleWorker mode ApplyRewrites (Start patt) =
        transitionRewrite mode patt
    transitionRuleWorker _ ApplyRewrites state@(Rewritten patt) = pure state
        
    transitionRewrite All = transitionAllRewrite
    transitionRewrite Any = transitionAnyRewrite
```

ttuegel (changes requested):

ttuegel (approved):

PR #2252, opened 2020-11-06
Parse left- and right-assoc

### LINKED ISSUE #2124: Update syntax and parser for left- and right-assoc notations
See kframework/k#1538

ttuegel:
Fixes #2124 

---

###### Reviewer checklist


kore/src/Kore/Repl/Parser.hs
```diff
@@ -52,7 +57,16 @@ import qualified Kore.Log as Log
 import qualified Kore.Log.Registry as Log
 import Kore.Repl.Data
 
-type Parser = Parsec String String
+type Parser = Parsec ReplParseError String
+
+newtype ReplParseError = ReplParseError { unReplParseError :: String }
```
ana-pantilie, line 62: What are the benefits of using a newtype with an `IsString` instance here rather than `String` directly?

kore/src/Kore/Syntax/Pattern.hs
```diff
@@ -238,13 +260,178 @@ instance TopBottom (Pattern variable annotation) where
...
     defineColumn tableName _ = SQL.defineColumn tableName (SQL.Proxy @Text)
     toColumn = SQL.toColumn . Pretty.renderText . Pretty.layoutOneLine . unparse
 
+instance
+    From
+        (And Sort (Pattern variable Attribute.Null))
+        (Pattern variable Attribute.Null)
+  where
+    from = Recursive.embed . (:<) Attribute.Null . from
+    {-# INLINE CONLIKE from #-}
```
ana-pantilie, line 273: I don't think I fully understand what `CONLIKE` does from the GHC documentation, I'll keep in mind to try to learn more about it, but could you give me some hints related to why it was necessary here? Did something not work properly or is there something specific about this code which makes a pragma like this necessary?

ana-pantilie (approved):
The parser looks much cleaner to me now. I'm wondering why the syntax pattern type used to take a type variable in the first place. I have some questions but they're mostly for my own understanding, and I also have a couple of minor suggestions.

kore/src/Kore/Parser/Lexer.hs
```diff
@@ -370,51 +292,82 @@ isIdChar c = isIdFirstChar c || isIdOtherChar c
...
 {- | Parses a 'Sort' 'Id'
 
 @
 <sort-id> ::= <id>
 @
 -}
-sortIdParser :: Parser Id
-sortIdParser = idParser
+parseSortId :: Parser Id
+parseSortId = parseId <?> "sort identifier"
```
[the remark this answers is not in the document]
ttuegel: Yes, the error message will be:

```
unexpected ...
expected sort identifier
```

kore/src/Kore/Repl/Parser.hs
ttuegel: When we use `String` directly, we have to declare an orphan instance.

kore/src/Kore/Syntax/Pattern.hs
ttuegel: `CONLIKE` is short for "constructor-like". This tells GHC that this function is inexpensive to compute, like a constructor. This will make the compiler very aggressive about inlining it. I'm using it here because I know that all these functions put together are just applying several constructors on top of the argument.

ttuegel:
> I'm also wondering why the sentence type used to take a type variable in the first place.

It was due to the class `AsSentence`, which I removed. We needed all the branches of the `Sentence` sum type to have the same parameters because there was once a functional dependency on this class. Without the functional dependency, we would need a type annotation in every place that `asSentence` was used. I was able to replace it with `From`, and I only had to add two or three type applications to make everything typecheck.

MirceaS (approved):

kore/src/Kore/Parser/ParserUtils.hs
```diff
@@ -61,22 +51,6 @@ does not have any available characters.
...
-takeWhile = takeWhileP Nothing
-
-{-|'endOfInput' is similar to Attoparsec's 'endOfInput'. It matches only the
-end-of-input position.
--}
-endOfInput :: Parser ()
-endOfInput = eof
-
-instance ShowErrorComponent String where
-    showErrorComponent str = str
```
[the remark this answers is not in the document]
ttuegel: They are all either re-exporting a library function under a non-standard name, or they are redefining a library function.

kore/src/Kore/Repl/Parser.hs
ana-pantilie: Oh, that's right, we needed the `ShowErrorComponent` instance.

kore/src/Kore/Syntax/Pattern.hs
ana-pantilie: I understand now, thank you!

PR #2257, opened 2020-11-10
Use 8-bit encoding of Bytes

### LINKED ISSUE #2256: Change concrete syntax of Bytes
We will change the concrete syntax of Bytes domain values to align with llvm-backend. At present, we use a big-endian base-16 encoding. Instead, we will change to a base-256 representation. The text representation of Kore domain values will continue to accept printable ASCII characters and escape sequences.

ttuegel:
Fixes #2256 

---

###### Reviewer checklist


test/regression-c-semantics/.gitignore
```diff
@@ -0,0 +1 @@
+definition.kore
```
MirceaS, line 1:
```suggestion
definition.kore

```
Don't we want a new line at the end here?

test/regression-c-semantics/test-regression-c-semantics.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+$KORE_PARSER --no-print-definition --no-print-pattern definition.kore --pattern executable.kore --module C
```
MirceaS, line 3: I thought we need to specify the module with = (`--module=C`)

MirceaS (approved):

test/regression-c-semantics/test-regression-c-semantics.sh
ttuegel: Both are accepted.

PR #2258, opened 2020-11-11
Parse signed hex values

### LINKED ISSUE #2251: STRING.string2base doesn't handle negative hex values
An expression like `String2Base("-10", 16)` wont evaluate and gives me a `#Bottom` when I try to krun it.

MirceaS:
---
Fixes #2251 

As can be seen from the code, the decimal case just above my change already contains `Text.signed` which leads me to believe that for the hex case it was left out on purpose so may need further investigation.

###### Reviewer checklist


ttuegel (changes requested):
Please write a unit test.

ttuegel (approved):

PR #2261, opened 2020-11-11
SMT.Translate.translatePattern: add fallback for translateApplication

ana-pantilie:
---

###### Reviewer checklist


kore/test/Test/Kore/Step/SMT/Translate.hs
```diff
@@ -158,6 +158,18 @@ test_translatePredicateWith =
...
             & Test.SMT.runNoSMT
         assertEqual "" actual1 actual2
+    , testCase "b = a, both constructors" $
+            translating (peq Mock.b Mock.a)
+        `yields`
+            (var 0 `eq` var 1)
+    , testCase "f() = a, f functional, a constructor" $
+            translating (peq Mock.functional00 Mock.a)
+        `yields`
+            (var 0 `eq` var 1)
```
ana-pantilie, line 168: I wasn't able to get the constructors to not get translated as `var n`. I tried adding a `translatingSMT` which would use the solver with the `MockSymbols` environment, to which I added a constructor declaration for `a`. Anyway, I think that it's not that important, as this successfully tests that `\equals a b` gets translated correctly even if `translateApplication` fails.

ttuegel: I wouldn't expect `translatingSMT` to help because we don't need access to the actual solver for this. I investigated further, and it turns out that the SMT declarations builder requires constructors to be declared twice: once as part of their sort and once as symbols. Could you please add a comment to this test, saying that it doesn't return the correct result for this reason? Also, please remove the `smtConstructor` parts from `MockSymbols`, because that function doesn't do anything.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -297,17 +297,23 @@ translatePredicateWith translateTerm predicate =
...
             _ -> case Cofree.tailF $ Recursive.project pat of
-                    VariableF _ -> do
-                        smtSort <- hoistMaybe $ translateSort sort
-                        translateUninterpreted smtSort pat
-                    ApplySymbolF app -> translateApplication app
+                    VariableF _ -> translateUninterpreted'
+                    ApplySymbolF app
+                      | isFunctionalPattern pat ->
+                          translateApplication app
+                          <|> translateUninterpreted'
```
ttuegel, line 304: Why can't we add this to `translateApplication` itself?

kore/test/Test/Kore/Step/SMT/Translate.hs
ana-pantilie: Thank you for looking into this!

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -230,29 +228,39 @@ translatePredicateWith translateTerm predicate =
...
+        maybeSort
+        original
         Application
             { applicationSymbolOrAlias
             , applicationChildren
             }
-      = do
-        let translated = translateSymbol applicationSymbolOrAlias
-        sexpr <- maybe empty return translated
-        when (isNothing translated)
```
ana-pantilie, line 248: From my understanding, if `translated` is `Nothing` this gets discarded before logging the warning. My changes should fix this.

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -230,29 +228,39 @@ translatePredicateWith translateTerm predicate =
...
-            $ warnSymbolSMTRepresentation applicationSymbolOrAlias
-        children <- zipWithM translatePattern
-            applicationChildrenSorts
-            applicationChildren
-        return $ shortenSExpr (applySExpr sexpr children)
+      | isFunctionalPattern original =
+        translateInterpretedApplication
+        <|> translateUninterpreted'
+      | otherwise =
+        translateInterpretedApplication
```
ana-pantilie, line 247: Before, we would fall back to `translateUninterpreted` for `translateInt` and `translateBool` without first checking if the pattern is functional. I assumed this wasn't correct, and fixed it here. This is reflected in the tests by the need to add some test data with the `functional` attribute.

kore/src/Kore/Step/SMT/Translate.hs
ana-pantilie: Thanks for suggesting this, I made the change and added some comments on the PR about the additional things I had to modify resulting from this.

ttuegel (approved):

PR #2292, opened 2020-11-26
Infer that terms are defined to send to SMT solver

### LINKED ISSUE #2287: Infer that terms are defined to send to SMT solver
The backend should use the `Defined` marker when it sends terms to the SMT solver, to infer that terms are defined. Specifically, if the backend passes [here](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/SMT/Translate.hs#L313), then [this check](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/SMT/Translate.hs#L243) should be skipped.

---

Reported by @denis-bogdanas

ana-pantilie:
Fixes #2287

1. Can now declare symbols to be used when translating test data for the solver.
2. Added unit tests as described in the issue. The first two tests had to be commented out, because of [this](https://github.com/kframework/kore/pull/2292/files#diff-2b0fe8f29c8bde92c2551fcd883166718a5af5dd720ebf8cb577cba91b00d510R345) issue. The other two should have only passed after these changes, but they used to pass before as well. Changing [this](https://github.com/kframework/kore/pull/2292/files#diff-2b0fe8f29c8bde92c2551fcd883166718a5af5dd720ebf8cb577cba91b00d510R349) to `empty` makes all new unit tests pass, though.
3. Further changes should be described in the issue.

---

###### Review checklist


kore/src/Kore/Step/SMT/Evaluator.hs
```diff
@@ -195,7 +194,7 @@ decidePredicate predicates =
     query :: MaybeT simplifier Result
     query =
         SMT.withSolver $ evalTranslator $ do
-            tools <- Simplifier.askMetadataTools
+            tools <- lift Simplifier.askMetadataTools
```
ttuegel, line 197: Why not move this outside `SMT.withSolver` instead?

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -82,6 +92,9 @@ data TranslateItem variable
     = QuantifiedVariable !(ElementVariable variable)
     | UninterpretedTerm !(TermLike variable)
 
+type TranslateTerm variable m =
+    SExpr -> TranslateItem variable -> Translator variable m SExpr
```
ttuegel, line 96: Is a `TranslateTerm` function actually allowed to depend on the type of `m`?

kore/src/Kore/Step/SMT/Translate.hs
```diff
@@ -228,28 +314,43 @@ translatePredicateWith translateTerm predicate =
...
+          | (isDefined && isFunctionPattern original)
+          || isFunctionalPattern original =
+              translateInterpretedApplication
+              <|> translateUninterpreted'
+          | otherwise =
+              -- TODO: this is not right, function-like
+              -- symbols should be sent to the solver only if
+              -- we know they are defined. Arbitrary symbols
+              -- should never be sent to the solver.
+              translateInterpretedApplication
```
ttuegel, line 349:
```suggestion
      =
        -- TODO: This would send interpreted symbols to the solver
        -- even if they may not be defined. We should only send symbols
        -- we know to be defined.
        translateInterpretedApplication
        -- TODO: Move call to guardLocalFunctionalPattern inside translateUnintepreted'.
        <|> (guardLocalFunctionalPattern >> translateUninterpreted')
        
        translateApplicationWorker assumeDefined
      where
        guardLocalFunctionalPattern
          | isFunctionalPattern original = return ()
          | otherwise = do
            TranslatorEnv { assumeDefined } <- ask
            Monad.guard (assumeDefined && isFunctionPattern original)
```

kore/src/Kore/Step/SMT/Translate.hs
ana-pantilie: I'm not entirely sure I understand the question right, but [translateTerm](https://github.com/kframework/kore/blob/deba447b057f856f5141e8ba18d42561d0c456eb/kore/src/Kore/Step/SMT/Evaluator.hs#L222) is the function we eventually pass to `translatePredicateWith`, and it's polymorphic in `m`.

ttuegel (approved):

PR #2296, opened 2020-12-02
Improve the performance of the Kore parser

### LINKED ISSUE #2189: Improve the performance of the Kore parser
The Kore parser runs out of memory with the C semantics.

See also: runtimeverification/c-semantics#172

MirceaS:
Fixes #2189 

---

###### Review checklist

### DIFF
--- a/kore/app/share/GlobalMain.hs
+++ b/kore/app/share/GlobalMain.hs
@@ -49,6 +49,7 @@ import Data.Text
     ( Text
     , pack
     )
+import qualified Data.Text.IO as Text
 import Data.Time.Format
     ( defaultTimeLocale
     , formatTime
@@ -511,12 +512,14 @@ parseDefinition :: FilePath -> Main ParsedDefinition
 parseDefinition = mainParse parseKoreDefinition
 
 mainParse
-    :: (FilePath -> String -> Either String a)
+    :: (FilePath -> Text -> Either String a)
     -> String
     -> Main a
 mainParse parser fileName = do
     contents <-
-        clockSomethingIO "Reading the input file" $ liftIO $ readFile fileName
+        Text.readFile fileName
+        & liftIO
+        & clockSomethingIO "Reading the input file"
     parseResult <-
         clockSomething "Parsing the file" (parser fileName contents)
     case parseResult of
--- a/kore/src/Kore/Parser/Lexer.hs
+++ b/kore/src/Kore/Parser/Lexer.hs
@@ -46,6 +46,9 @@ import Data.Map.Strict
     ( Map
     )
 import qualified Data.Map.Strict as Map
+import Data.Text
+    ( Text
+    )
 import qualified Data.Text as Text
 import Text.Megaparsec
     ( SourcePos (..)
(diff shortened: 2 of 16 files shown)


kore/app/share/GlobalMain.hs
```diff
@@ -511,12 +512,14 @@ parseDefinition :: FilePath -> Main ParsedDefinition
...
-    :: (FilePath -> String -> Either String a)
+    :: (FilePath -> Text -> Either String a)
     -> String
     -> Main a
 mainParse parser fileName = do
     contents <-
-        clockSomethingIO "Reading the input file" $ liftIO $ readFile fileName
+        Text.readFile fileName
+        & liftIO
+        & clockSomethingIO "Reading the input file"
```
MirceaS, line 522: @ttuegel Why are these the other way round now?

ttuegel:
I made some changes to your parser performance pull request which reduce the memory use significantly. (For some reason we were still using String. Never use String.) I would like you to try a few more things:
1. Avoid using Text.cons in the parser. Use lookAhead or match instead. Text.cons causes copying, but the tokens we are parsing are already in contiguous blocks of memory.
2. Refactor parseAnyId so that it only calls parseIntoId once. Right now, each branch calls parseIntoId separately. That incurs up to three calls to getSourcePos per token parsed.
3. Investigate the performance of validation.

kore/app/share/GlobalMain.hs
ttuegel: I changed it from `($)` to `(&)` so that the action (`Text.readFile`) would appear before the wrappers.

kore/src/Kore/Parser/Lexer.hs
```diff
@@ -224,17 +223,17 @@ genericIdRawParser
...
-    -> Parser String
+    -> Parser Text
 genericIdRawParser isFirstChar isBodyChar idKeywordParsing = do
-    c <- Parser.satisfy isFirstChar <?> "first identifier character"
-    cs <- Parser.takeWhileP (Just "identifier character") isBodyChar
-    let genericId = c : cs
-        keywordsForbidden = idKeywordParsing == KeywordsForbidden
+    (genericId, _) <- Parser.match
+        $ (Parser.satisfy isFirstChar <?> "first identifier character")
+        >> Parser.takeWhileP (Just "identifier character") isBodyChar
```
ttuegel, line 230: I think this is easier to read:

```suggestion
        (genericId, _) <- Parser.match $ do
        _ <- Parser.satisfy isFirstChar <?> "first identifier character"
        _ <- Parser.takeWhileP (Just "identifier character") isBodyChar
        pure ()
```

kore/src/Kore/Parser/Lexer.hs
```diff
@@ -357,42 +358,16 @@ parseSetId =
...
 
-symbolIdRawParser :: Parser String
+symbolIdRawParser :: Parser Text
 symbolIdRawParser = do
     c <- peekChar'
     if c == '\\'
-    then do
-        skipChar '\\'
-        (c :) <$> parseIdRaw KeywordsPermitted
+    then fst <$> Parser.match
```
ttuegel, line 367: Using `match` here makes the look-ahead (above) redundant.

ttuegel (changes requested):
Please remove the redundant look-ahead, and then we can merge this.

ttuegel (approved):

PR #2303, opened 2020-12-08
Always output fully-simplified states

### LINKED ISSUE #2299: Always output fully-simplified states
When running under `krun`, the backend uses an execution strategy where the program configuration is simplified immediately before rewriting. The result of rewriting is not simplified because the backend "knows" that it will just be simplified at the start of the next step. As a consequence, "remaining" configurations are not simplified, nor are states that have hit the `--depth` limit.

Reported by @ehildenb.

---

### Plan

- [x] Write unit tests to show these properties. (It is enough to inspect the `Stream` produced by `executionStrategy` and `limitedExecutionStrategy`.)
    - All final configurations are simplified under the `--depth` option.
    - Any stuck configuration is simplified.
    - Initial input states are allowed to be un-simplified, i.e. `--depth 0` _does_ perform simplification.
    - Every strategy step should `Rewrite` exactly once, because `search` depends on it.
- [x] Modify [`executionStrategy`](https://github.com/kframework/kore/blob/3fd34f99a05745bbc792538042c0513724c3cf37/kore/src/Kore/Step.hs#L104):
    - Move `Simplify` after `Rewrite`.
    -  Add a single step to the front of the stream which runs `Begin` and `Simplify` only.
- [x] Modify [`limitedExecutionStrategy`](https://github.com/kframework/kore/blob/3fd34f99a05745bbc792538042c0513724c3cf37/kore/src/Kore/Step.hs#L113) so that the special first step is not included in the depth limit.

emarzion:
Fixes #2299 

Update `executionStrategy` so that `Simplify` always appears after `Rewrite`. If execution is stopped by the depth limit, fully-simplified states will be printed for the user.

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Step.hs
+++ b/kore/src/Kore/Step.hs
@@ -102,19 +102,21 @@ retractRemaining _ = Nothing
 -}
 executionStrategy :: Stream (Strategy Prim)
 executionStrategy =
-    (Strategy.sequence . fmap Strategy.apply)
-        [ Begin
-        , Simplify
-        , Rewrite
-        , Simplify
-        ]
-     Stream.:> (
-    (Strategy.sequence . fmap Strategy.apply)
-        [ Begin
-        , Rewrite
-        , Simplify
-        ]
-    & Stream.iterate id)
+    step1 Stream.:> Stream.iterate id stepN
+  where
+    step1 =
+        (Strategy.sequence . fmap Strategy.apply)
+            [ Begin
+            , Simplify
+            , Rewrite
+            , Simplify
+            ]
+    stepN =
+        (Strategy.sequence . fmap Strategy.apply)
+            [ Begin
+            , Rewrite
+            , Simplify
+            ]
 
 limitedExecutionStrategy :: Limit Natural -> [Strategy Prim]
 limitedExecutionStrategy depthLimit =
@@ -126,7 +128,7 @@ data Prim
     = Begin
     | Simplify
     | Rewrite
-    deriving (Eq,Show)
+    deriving (Eq, Show)
 
 {- The two modes of symbolic execution. Each mode determines the way
     rewrite rules are applied during a rewrite step.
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Step.hs
```diff
@@ -106,8 +106,15 @@ executionStrategy =
...
+        , Simplify
         ]
-    & Stream.iterate id
+     Stream.:> (
+    (Strategy.sequence . fmap Strategy.apply)
+        [ Begin
+        , Rewrite
+        , Simplify
+        ]
+    & Stream.iterate id)
```
ttuegel, line 117: Using `where` would follow the style guide a bit better:

```.hs
    step1 Stream.:> Stream.iterate id stepN
  where
    step1 =
        (Strategy.sequence . fmap Strategy.apply)
            [ Begin
            , Simplify
            , Rewrite
            , Simplify
            ]
    stepN =
        (Strategy.sequence . fmap Strategy.apply)
            [ Begin
            , Rewrite
            , Simplify
            ]
```

kore/test/Test/Kore/Step.hs
```diff
@@ -298,6 +301,33 @@ test_stepStrategy =
...
+prop_alwaysRewrite :: Natural -> Bool
+prop_alwaysRewrite depthLimit =
+    all hasRewrite $ takeWithin (Limit depthLimit) (toList executionStrategy)
+
+getPrims :: Strategy Prim -> [Prim]
+getPrims = \case
+    Strategy.Seq s1 s2 -> getPrims s1 ++ getPrims s2
+    Strategy.Apply p -> [p]
+    Strategy.Continue -> []
+    _ -> [] -- ?
```
ttuegel, line 323: Let's write a function `isLastSimplify :: Strategy Prim -> Bool` in the same spirit as `hasRewrite`, so that we can cover all the cases better.

ttuegel (changes requested):

ttuegel (changes requested):
Please also have a look at our [style guide](https://github.com/kframework/kore/wiki/Coding-Standards).

ttuegel (approved):
I wrote a summary in the pull request this time, but please do that on your next pull request.

PR #2315, opened 2020-12-15
Keep table of term replacements in SideCondition, use in term simplifier

### LINKED ISSUE #2249: Lift simplifyConjunctions into the SideCondition
The table of term replacements used in `simplifyConjunctions` should be kept in the `SideCondition`. Instead of running `simplifyConjunctions` in the `Condition` simplifier, the `TermLike` and `Predicate` simplifiers will look up their argument in the `SideCondition`.

This will enable us to apply assumptions from the `SideCondition` while applying equations, which is not currently possible.

---

1.  [x] Make `SideCondition` a `MultiAnd` of `Predicate`.
    Appetite: S. This is not the final form it will take, but this change will expose any places in the code where we make assumptions about the form of `SideCondition`. At this point, we are still only using the `SideCondition` to send to the solver. There should be instances `From (Predicate variable) (SideCondition variable)` and `From (SideCondition variable) (Predicate variable)`. These instances must continue to work through the following steps.
1.  [ ] Specialize for looking up `Predicate`.
    Appetite: S. ~The `SideCondition` should contain a `HashSet` for looking up if a predicate is `\top` or `\bottom`.~ The `SideCondition` should contain a `HashMap` from `TermLike variable` to the same, storing the map used [here](https://github.com/kframework/kore/blob/deba447b057f856f5141e8ba18d42561d0c456eb/kore/src/Kore/Step/Simplification/Condition.hs#L209) so it's not computed every time. The `Predicate` simplifier should look up its argument in the `SideCondition`. Before #1442: We need to set the sort to `predicateSort` when storing or looking up a `Predicate`. After #1442: Predicate sorts will be erased automatically.
1.  [ ] Specialize the representation of `\equals`.
    Appetite: S. The `SideCondition` should contain a `HashMap` for looking up `TermLike`s for substitution.
1.  [ ] Combine `SideCondition`s with normalization.
    Appetite: S. Combining `SideCondition`s (by `sappend` or `mappend`) should apply the substitutions from each to the other so that the result is normalized. ~Note: This makes #1416 trivial.~

ana-pantilie:
Fixes #2249

~Needs https://github.com/kframework/kore/pull/2346~

- Changes representation of `SideCondition`: it now contains a `HashMap (TermLike variable) (TermLike variable)`
- Moves `simplifyConjunctionWithAssumption` to `SideCondition`, uses it in both the `Condition` simplifier and when constructing `SideCondition`s
- Replaces terms in the `TermLike` simplifier, keeps simplifying until the current term cannot be replaced anymore
- `Pattern` simplifier uses a `\\top` `SideCondition` by default, uses a different `SideCondition` in special cases only
- Fixes simplification in `checkImplicationWorker`
- Modifies `one-path` unit test which now detects a `\\bottom` RHS
- Adds unit tests for `TermLike` replacement during simplification
- Updates one integration test `.golden` file (comments changed)

---

###### Review checklist


kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -303,11 +309,14 @@ simplify sideCondition = \termLike ->
           | Pattern.isSimplified sideConditionRepresentation result
             && isTop resultTerm
             && resultSubstitutionIsEmpty
+            && SideCondition.cannotReplace sideCondition (Pattern.term result)
```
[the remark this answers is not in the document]
ana-pantilie: I think we would want to continue simplification until no more replacements from the `SideCondition` can apply. I added such a check in every place where we exit the term simplifier.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -4,6 +4,7 @@ License     : NCSA
 -}
 
 -- For instance Applicative:
+{-# LANGUAGE Strict               #-}
 {-# LANGUAGE UndecidableInstances #-}
```
ttuegel, line 8: Is this actually necessary?

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -208,3 +351,225 @@ toRepresentation
...
+ -}
+cannotReplaceTerm
+    :: InternalVariable variable
+    => SideCondition variable
+    -> TermLike variable
+    -> Bool
+cannotReplaceTerm SideCondition { replacements } term =
+    HashMap.lookup term replacements & isNothing
+
+data DoubleMap variable = DoubleMap
```
ttuegel, line 376: This isn't a generic type, so it should not have a generic name like `DoubleMap`. Let's call it `Assumptions` instead.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -208,3 +351,225 @@ toRepresentation
...
+    { termLikeMap :: HashMap (TermLike variable) (TermLike variable)
+    , predMap :: HashMap (Predicate variable) (Predicate variable)
```
ttuegel, line 378: Let's use the full name of the type:

```suggestion
    , predicateMap :: HashMap (Predicate variable) (Predicate variable)
```

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -208,3 +351,225 @@ toRepresentation
...
+            case predF of
+                Predicate.CeilF ceil_ -> 1 + sum (size <$> ceil_)
+                Predicate.EqualsF equals_ -> 1 + sum (size <$> equals_)
+                Predicate.FloorF floor_ -> 1 + sum (size <$> floor_)
+                Predicate.InF in_ -> 1 + sum (size <$> in_)
+                _ -> 1 + sum predF
+
+    assume
+        :: Predicate variable ->
+        StateT (DoubleMap variable) Changed ()
```
ttuegel, line 431:
```suggestion
        :: Predicate variable
        -> StateT (DoubleMap variable) Changed ()
```

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -208,3 +351,225 @@ toRepresentation
...
+
+    applyAssumptionsWorker
+        :: DoubleMap variable
+        -> Predicate variable
+        -> Changed (Predicate variable)
+    applyAssumptionsWorker assumptions original
+      | Just result <- HashMap.lookup original (predMap assumptions) = Changed result
+
+      | HashMap.null (termLikeMap assumptions') &&
+        HashMap.null (predMap assumptions') = Unchanged original
```
ttuegel, line 470:
```suggestion
      | HashMap.null (termLikeMap assumptions')
      , HashMap.null (predMap assumptions') = Unchanged original
```

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -208,3 +351,225 @@ toRepresentation
...
+          Predicate.CeilF ceil_ -> Predicate.CeilF <$> traverse
+            (applyAssumptionsWorkerTerm (termLikeMap assumptions')) ceil_
+          Predicate.FloorF floor_ -> Predicate.FloorF <$> traverse
+            (applyAssumptionsWorkerTerm (termLikeMap assumptions')) floor_
+          Predicate.EqualsF equals_ -> Predicate.EqualsF <$> traverse
+            (applyAssumptionsWorkerTerm (termLikeMap assumptions')) equals_
+          Predicate.InF in_ -> Predicate.InF <$> traverse
+            (applyAssumptionsWorkerTerm (termLikeMap assumptions')) in_
+          _ -> traverse (applyAssumptionsWorker assumptions') replaceIn
+        )
```
ttuegel, line 482: The [layout rule](https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-210002.7) makes the parentheses unnecessary here:

```suggestion
      | otherwise =
        case replaceIn of
            Predicate.CeilF ceil_ -> Predicate.CeilF <$> traverse
              (applyAssumptionsWorkerTerm (termLikeMap assumptions')) ceil_
            Predicate.FloorF floor_ -> Predicate.FloorF <$> traverse
              (applyAssumptionsWorkerTerm (termLikeMap assumptions')) floor_
            Predicate.EqualsF equals_ -> Predicate.EqualsF <$> traverse
              (applyAssumptionsWorkerTerm (termLikeMap assumptions')) equals_
            Predicate.InF in_ -> Predicate.InF <$> traverse
              (applyAssumptionsWorkerTerm (termLikeMap assumptions')) in_
            _ -> traverse (applyAssumptionsWorker assumptions') replaceIn
```

We should also extract `applyAssumptionsWorkerTerm (termLikeMap assumptions')` and `applyAssumptionsWorker assumptions'` to make this block easier to read.

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -311,11 +321,14 @@ simplify sideCondition = \termLike ->
...
             && isTop resultTerm
             && resultSubstitutionIsEmpty
+            && SideCondition.cannotReplaceTerm sideCondition (Pattern.term result)
           = return (OrPattern.fromPattern result)
           | Pattern.isSimplified sideConditionRepresentation result
             && isTop resultPredicate
+            && SideCondition.cannotReplaceTerm sideCondition (Pattern.term result)
           = return (OrPattern.fromPattern result)
           | isTop resultPredicate && resultTerm == originalTerm
+            && SideCondition.cannotReplaceTerm sideCondition (Pattern.term result)
```
ttuegel, line 331: This whole block should use commas instead of `&&`.

ttuegel (changes requested):

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -454,36 +451,36 @@ simplifyConjunctionByAssumption (toList -> andPredicates) =
...
-        & maybe (Unchanged original) (Changed . synthesize)
+      | Just result <- HashMap.lookup original (predicateMap assumptions) = Changed result
+
+      | HashMap.null (termLikeMap assumptions')
+      , HashMap.null (predicateMap assumptions') = Unchanged original
+
+      | otherwise =
+          case replaceIn of
+              Predicate.CeilF ceil_ -> Predicate.CeilF <$> traverse
+                (applyAssumptionsWorkerTerm (termLikeMap assumptions')) ceil_
```
ttuegel: Let's add a local definition `applyAssumptionsWorkerTerm' = applyAssumptionsWorkerTerm (termLikeMap assumptions')` to avoid some of this duplication:

```suggestion
              Predicate.CeilF ceil_ ->
                  Predicate.CeilF <$> traverse applyAssumptionsWorkerTerm' ceil_                
```

Please check the indentation around the guards and case block here, too.

ttuegel (changes requested):

ttuegel (approved):

andreiburdusa (approved):

PR #2333, opened 2020-12-22
Unify Equation arguments at initialization

### LINKED ISSUE #2326: Unify Equation arguments at initialization
Function definitions are instantiated by unifying their arguments, to allow `\or` patterns on the left-hand side. We can solve the unification problems at initialization to produce a conjunction of equations that can each be applied by simple matching. This would improve performance because our matching algorithm re-orders constraints while our unification algorithm does not.

---

- [ ] Make [`Kore.Equation.Simplification.simplifyEquation`](https://github.com/kframework/kore/blob/770f15d72a8e8f3f364ed666e8c193a198852c88/kore/src/Kore/Equation/Simplification.hs#L70) return a `MultiAnd` of equations.
- [ ] Use `Logic.scatter` instead of [`expectSingleResult`](https://github.com/kframework/kore/blob/770f15d72a8e8f3f364ed666e8c193a198852c88/kore/src/Kore/Equation/Simplification.hs#L104).
- [ ] Also simplify the `argument` (if any) and apply the substitution to the whole `Equation`.

ana-pantilie:
Fixes #2326

- simplify the `left` and the `argument` together in `simplifyEquation`; if the `argument` is `Nothing` the `left` is simplified by itself
- apply the subsitutions to create new equations; the new equations will have their `argument`s erased
- collect all the equations in the map

---

###### Review checklist


ttuegel (changes requested):
Please add some tests, to show that `simplifyEquation` does the right thing. We should at least check that: it moves the argument patterns to the left-hand side of the equation, and it handles `\or` in the `argument` patterns.

kore/src/Kore/Equation/Simplification.hs
```diff
@@ -54,63 +60,78 @@ simplifyExtractedEquations
...
 {- | Simplify an 'Equation' using only Matching Logic rules.
 
-The original rule is returned unless the simplification result matches certain
-narrowly-defined criteria.
+It attempts to unify the argument of the equation, creating new
+equations where the argument is substituted in the rest of the
+resulting equations, and the argument is removed.
+
+If any of the patterns resulting from simplifying the term and the
+argument contain a predicate which is not null, 'simplifyEquation'
```
ttuegel, line 76:
```suggestion
argument contain a predicate which is not 'Top', 'simplifyEquation'
```

kore/src/Kore/Equation/Simplification.hs
```diff
@@ -54,63 +60,78 @@ simplifyExtractedEquations
...
+is exited early and the original equation is returned.
```
ttuegel, line 77:
```suggestion
returns the original equation.
```

kore/test/Test/Kore/Equation/Simplification.hs
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+import Test.Tasty
+import Test.Tasty.HUnit.Ext
+
+test_simplifyEquation :: [TestTree]
+test_simplifyEquation =
+    [ testGroup "Unification equations: in all cases the argument gets\
+                \ simplified and substituted into the left term; the argument\
+                \ is then removed."
```
ttuegel, line 31: I think this would be better as a comment, so that it does not take up so much space in the test output.

```suggestion
    [ testGroup "Unify arguments"
        -- In all cases the argument gets simplified and substituted into the left term;
        -- the argument is then removed.
```

ttuegel (approved):
Approved, pending my suggestions.

PR #2335, opened 2020-12-23
kore-repl: represent Remaining states correctly

### LINKED ISSUE #2328: kore-repl: intermediate state represented as stuck
When running `test/itp-nth-ancestor/nth1-spec.k` with `kore-repl`, the final proof graph looks like:
stuck-state-before-proven

This means that one of the intermediate states (node 7) is considered `stuck` by  `kore-repl`, before being proven after the following step. Also, we should investigate why there are two `Simpl/RD` transitions, with node 7 at the middle, instead of a single `Simpl/RD` transition. We should probably rename `Simpl/RD` as well.
These issues have probably appeared because we changed the proof transition system without updating how we represent this information in `kore-repl.`

- [ ] [ClaimState.Remaining](https://github.com/kframework/kore/blob/34acb288b5c0ec734be2b133b6d9f23b36652ef2/kore/src/Kore/Reachability/ClaimState.hs#L37) is not a final stuck proof state anymore, so we should not represent it as such in `kore-repl`
- [ ] Investigate the `Simpl/RD` issue

ana-pantilie:
Fixes #2328

- `Remaining` nodes are no longer represented as stuck to the user
- edges can now be labled `Remaining`, `CheckImplication` or if a rule applied, with the rule id/name
- fixes `tryf` command:
    - adds case for when claim gets proven; adds test for this case
    - if the resulting claim state is `Stuck` or `Remaining` it means the rule cannot apply
- cleans up `runStepper`

---

###### Review checklist


ana-pantilie:
I think the first `Simpl/RD` edge from the issue appears because node 7 is a `Remaining` state, meaning that no rule was applied, and we only detect that the implication holds during the following step. I'll investigate further.

andreiburdusa (approved):

emarzion (approved):

PR #2346, opened 2021-01-12
unifyKequalsEq should apply for both \and and \equals

ana-pantilie:
`unifyKequalsEq` should apply for both `\\and` and `\\equals`. This is needed for https://github.com/kframework/kore/issues/2249, in which we remove all cases where we use a `Condition` as a `SideCondition` to simplify itself. Before, a `==K` term in `test/itp-nth-ancestor/nth1-spec.k` would not be simplified at the right moment.

---

###### Review checklist


ttuegel (changes requested):
Could you add tests for this? We should test that this unification case is applied for `\equals` and `\and`, and we should test that it is _not_ applied when the `_==K_` term is not functional.

ttuegel (approved):

Issue #2351, opened 2021-01-14
Remove Defined marker


ttuegel:
After #2249, we can modify the `SideCondition` to keep a set of all defined terms. Instead of looking for the `Defined` marker, we can look in the `SideCondition`.

---

- [x] Add a `HashSet` of defined `TermLike`s to `SideCondition`. (appetite: S)
- [x] Implement a function `assumeDefined :: TermLike variable -> SideCondition variable` which assumes the definedness of all sub-terms that are implied by the original term. In other words, if `t` is the original term, we assume a sub-term `s` is defined if `\ceil(t) \implies \ceil(s)`. Take care with Map and Set to minimize storage! If the size of a Map or set is `n`, we only want `n^2` entries and not `n!`. (appetite: M)
- [x] Implement a function `isDefined :: SideCondition variable -> TermLike variable -> Bool`. Again, take care with Map and Set. (appetite: M)
- [ ] Remove `Defined`. Replace `makeDefined` with `assumeDefined`. Use `isDefined` instead of checking pattern attributes. (appetite: M)

ttuegel:
The `\ceil` of a map or set reduces like this:

``` k
\defined: \ceil(X |-> U  Y |-> V  Z |-> W  M)
/* pair-wise \ceil of elements */ ∧ \ceil(X |-> U  M) ∧ \ceil(Y |-> V  M) ∧ \ceil(Z |-> W  M)
```

where the pair-wise parts are:

``` k
\ceil(X |-> U  Y |-> V) ∧ \ceil(X |-> U  Z |-> W) ∧ \ceil(Y |-> V  Z |-> W)
```

PR #2354, opened 2021-01-19
Keep set of defined terms in SideCondition

ana-pantilie:
Part of https://github.com/kframework/kore/issues/2351

~Needs https://github.com/kframework/kore/pull/2315~

Implements the first three bullet points from #2351. 

---

###### Review checklist


kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -144,18 +170,18 @@ instance InternalVariable variable => SQL.Column (SideCondition variable) where
     toColumn = SQL.toColumn . Pretty.renderText . Pretty.layoutOneLine . pretty
 
 instance TopBottom (SideCondition variable) where
```
ttuegel, line 172: Why do we need this instance? It doesn't seem right to have one, to me, because we always assume the `SideCondition` is `\top`.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -144,18 +170,18 @@ instance InternalVariable variable => SQL.Column (SideCondition variable) where
...
-    isBottom sideCondition@(SideCondition _ _) =
+    isBottom sideCondition@(SideCondition _ _ _) =
         isBottom assumedTrue
       where
         SideCondition { assumedTrue } = sideCondition
 
 instance Ord variable => HasFreeVariables (SideCondition variable) variable
   where
-    freeVariables sideCondition@(SideCondition _ _) =
+    freeVariables sideCondition@(SideCondition _ _ _) =
```
ttuegel, line 184: I think this should include the free variables of the other fields.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -292,7 +324,7 @@ toPredicate
     :: InternalVariable variable
     => SideCondition variable
     -> Predicate variable
-toPredicate condition@(SideCondition _ _) =
+toPredicate condition@(SideCondition _ _ _) =
```
ttuegel, line 327: I think the `replacements` always come from the `assumedTrue`, so they can be omitted here, but shouldn't the `definedTerms` be included?

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -577,3 +620,234 @@ retractLocalFunction =
...
+                asSet term'
+                <> foldMap assumeDefinedWorker (internalListChild internalList)
+            TermLike.InternalMap_ internalMap ->
+                let definedElems =
+                        getDefinedElementsOfAc internalMap
+                    definedMaps =
+                        generateNormalizedAcs internalMap
+                        & HashSet.map TermLike.mkInternalMap
+                 in foldMap assumeDefinedWorker definedElems
+                    <> asSet term'
```
ttuegel, line 665: Why use `asSet` here?

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -577,3 +620,234 @@ retractLocalFunction =
...
+                    <> definedMaps
+            TermLike.InternalSet_ internalSet ->
```
ttuegel, line 667: Could the logic for Map and Set be combined with `Kore.Builtin.AssocComm.CeilSimplifier`?

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -577,3 +620,234 @@ retractLocalFunction =
...
+
+    isNonEmptySubset subset set
+      | null subset = False
+      | otherwise = all (`HashSet.member` set) subset
+
+    isFunctionalSymbol (App_ symbol children)
+      | isFunctional symbol =
+          all (isDefined sideCondition) children
+      | otherwise = False
+    isFunctionalSymbol _ = False
```
ttuegel, line 746:
```suggestion
    isFunctionalSymbol _ = False
```

kore/src/Kore/Internal/SideCondition.hs
ana-pantilie: Do you think this should include the free variables of the replacements as well? I assume you mean the free variables of the defined terms here, right?

kore/src/Kore/Internal/SideCondition.hs
ana-pantilie: `generateNormalizedAcs` will not add the collection itself to the resulting `HashSet`. I thought it was clearer to separate the two, especially since we should check if the input collection is always defined or not before adding it. Actually, do you think we should do the same for the subcollections as well?

kore/src/Kore/Internal/SideCondition.hs
ana-pantilie: I've thought about this, and I think it would only make sense if we wanted to also generate subcollections there and evaluate their ceils. Do we want to do this? At the moment, it looks like we only check the definedness of the elements and if they are unique.

kore/src/Kore/Internal/SideCondition.hs
ttuegel: Well, the free variables of the term replacements should be the same as from the `assumedTrue` right? Yes, we should definitely include the defined terms.

kore/src/Kore/Internal/SideCondition.hs
ttuegel: Sorry, I mean: Why does the collection itself need to be in the set at all? The set should have all the sub-terms required to decide that the whole collection is defined.

ana-pantilie: Done. We don't add the collection itself anymore, and I added a special case for singleton collections in `isDefined`.

ttuegel (approved):

PR #2369, opened 2021-01-26
kore-parser: Parse and validate multiple pattern files

MirceaS:
Required for https://github.com/runtimeverification/c-semantics/pull/213

---

###### Review checklist

### DIFF
--- a/kore/app/parser/Main.hs
+++ b/kore/app/parser/Main.hs
@@ -91,7 +91,8 @@ main = handleTop $ do
                 else pure Nothing
             let KoreParserOptions { willPrintDefinition } = koreParserOptions
             let KoreParserOptions { appKore } = koreParserOptions
-            when willPrintDefinition $ putDebug parsedDefinition
+            when (willPrintDefinition && not appKore)
+                $ putDebug parsedDefinition
             when appKore $ for_ indexedModules printAppKore
             pure indexedModules
 
--- a/kore/app/parser/Main.hs
+++ b/kore/app/parser/Main.hs
@@ -8,6 +8,7 @@ import Control.Monad.Catch
     , handle
     )
 import qualified Data.Map.Strict as Map
+import Data.Maybe
 
 import Kore.AST.ApplicativeKore
 import Kore.ASTVerifier.DefinitionVerifier
@@ -100,16 +101,17 @@ main = handleTop $ do
             let KoreParserOptions { patternOpt } = koreParserOptions
             for_ patternOpt $ \patternOptions -> do
                 let PatternOptions { mainModuleName } = patternOptions
+                indexedModule <- if willVerify then Just <$>
+                    ( lookupMainModule
+                        (ModuleName mainModuleName)
+                        indexedModules
+                    & lift )
+                    else return Nothing
                 let PatternOptions { patternFileNames } = patternOptions
                 for_ patternFileNames $ \patternFileName -> do
                     parsedPattern <- mainPatternParse patternFileName
                     when willVerify $ do
-                        indexedModule <-
-                            lookupMainModule
-                                (ModuleName mainModuleName)
-                                indexedModules
-                            & lift
-                        _ <- mainPatternVerify indexedModule parsedPattern
+                        _ <- mainPatternVerify (fromJust indexedModule) parsedPattern
                         return ()
                     let KoreParserOptions { willPrintPattern } =
                             koreParserOptions
--- a/kore/app/parser/Main.hs
+++ b/kore/app/parser/Main.hs
@@ -100,15 +100,15 @@ main = handleTop $ do
             let KoreParserOptions { patternOpt } = koreParserOptions
             for_ patternOpt $ \patternOptions -> do
                 let PatternOptions { mainModuleName } = patternOptions
-                indexedModule <-
-                    lookupMainModule
-                        (ModuleName mainModuleName)
-                        indexedModules
-                    & lift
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Options.hs
```diff
@@ -77,10 +77,12 @@ data KoreParserOptions = KoreParserOptions
 
 parsePatternOptions :: Parser PatternOptions
 parsePatternOptions = PatternOptions
-    <$> strOption
-        (  metavar "PATTERN_FILE"
-        <> long "pattern"
-        <> help "Kore pattern source file to parse [and verify]. Needs --module.")
+    <$> some (
```
ttuegel, line 80: Use `many` here; `some` requires at least one `--pattern`, but `kore-parser` can be used to verify a definition and no patterns.

```suggestion
    <$> many (
```

kore/src/Kore/Options.hs
```diff
@@ -77,10 +77,12 @@ data KoreParserOptions = KoreParserOptions
...
+            strOption
+            (  metavar "PATTERN_FILE"
+            <> long "pattern"
+            <> help "Kore pattern source file to parse [and verify]. Needs --module.")
+        )
```
ttuegel, line 85: Let's extract this as `parsePatternFileName` so that the option parser simply reads `many parsePatternFileName`.

kore/app/parser/Main.hs
```diff
@@ -99,21 +99,22 @@ main = handleTop $ do
...
+                let PatternOptions { mainModuleName } = patternOptions
+                let PatternOptions { patternFileNames } = patternOptions
+                for_ patternFileNames $ \patternFileName -> do
+                    parsedPattern <- mainPatternParse patternFileName
+                    when willVerify $ do
+                        indexedModule <-
+                            lookupMainModule
+                                (ModuleName mainModuleName)
+                                indexedModules
+                            & lift
```
ttuegel, line 111: We only need to look up the main module once, so we should move this out of the loop.

After that, let's extract the remaining body of the loop as a top-level function `verifyPatternFile`.

ttuegel (changes requested):

kore/src/Kore/Options.hs
MirceaS: > Use `many` here; `some` requires at least one `--pattern`, but `kore-parser` can be used to verify a definition and no patterns.

This `some` is nested inside the optional `parsePatternOptions`. So what this is basically saying is if you want to have the patterns/module options combo then you must specify at least one pattern.

kore/app/parser/Main.hs
MirceaS: I actually had this snippet outside the loop originally but I had to change it after I realised that an error would be thrown if we ran that code with `willVerify` set to `False` (because `indexedModules` would be empty).
I'll push a workaround, though it won't be very pretty due to the control flow that we're going for...

MirceaS: Also without that `indexedModule...` snippet, the body of the loop is quite small and straightforward...
I feel like if we isolated that as a separate top-level function we would have to pass on to it a lot of local variables that at the moment are implicitly in scope... (`indexedModule`, `willVerify`, `patternFileName`, `koreParserOptions`)
So I am leaning more towards not isolating it, what do you think?

ttuegel (approved):

kore/app/parser/Main.hs
ttuegel: I cleaned up the application logic. Feel free to merge it when you are satisfied.

kore/app/parser/Main.hs
```diff
@@ -80,43 +80,33 @@ main = handleTop $ do
...
-                    parsedPattern <- mainPatternParse patternFileName
-                    when willVerify $ do
-                        _ <- mainPatternVerify (fromJust indexedModule) parsedPattern
-                        return ()
-                    let KoreParserOptions { willPrintPattern } =
-                            koreParserOptions
-                    when willPrintPattern $
-                        lift $ putDoc (debug parsedPattern)
+            when willPrintDefinition $ putDebug parsedDefinition
+            when appKore $ for_ indexedModules printAppKore
```
MirceaS, line 95: @ttuegel 
I see that the logic changed here, in the sense that we may print both the applicative Kore of the modules AND the definition.
Is this what we want?
Isn't the idea of the `appKore` flag to change what is printed if `willPrintDefinition` is also set?

ttuegel: Yes, the new behavior is nonsensical, but the old behavior was nonsensical also. If you want to change `when willPrintDefinition` to `when (willPrintDefinition && not appKore)`, that's fine with me.

kore/app/parser/Main.hs
```diff
@@ -79,58 +80,50 @@ main = handleTop $ do
...
-                        $ unparseToString2
-                        $ completeDefinition
-                        $ toVerifiedDefinition indexedModules
-                else putDoc (debug parsedDefinition)
-
-            let KoreParserOptions { patternOpt } = koreParserOptions
-            for_ patternOpt $ \patternOptions -> do
-                let PatternOptions { patternFileName } = patternOptions
+            when willPrintDefinition $ putDebug parsedDefinition
+            when appKore $ for_ indexedModules printAppKore
```
MirceaS, line 95:
```suggestion
            when (willPrintDefinition && not appKore)
                $ putDebug parsedDefinition
            when appKore $ for_ indexedModules printAppKore
```

PR #2372, opened 2021-01-26
Remove Defined markers in simplifyConjunctionByAssumption

ana-pantilie:
~Needs #2366~

Until #2351 is done, this will fix those cases where a conjunction isn't simplified correctly because the terms differ by `Defined` markers.

---

###### Review checklist

### DIFF
--- a/kore/test/Test/Kore/Step/Simplification/Condition.hs
+++ b/kore/test/Test/Kore/Step/Simplification/Condition.hs
@@ -76,13 +76,18 @@ test_simplify_local_functions =
     , test "contradiction: \"one\" = f(x) ∧ f(x) = \"two\"" fString (Left  strOne) (Right strTwo)
     , test "contradiction: \"one\" = f(x) ∧ \"two\" = f(x)" fString (Left  strOne) (Left  strTwo)
     , test "contradiction: f(x) = \"one\" ∧ \"two\" = f(x)" fString (Right strOne) (Left  strTwo)
+    -- Ignore Defined marker
+    , testDefined "contradiction: f(g(x)) = a ∧ f(g(x)) = b" (Right fg) (Right a) (Right b)
+    , testDefined "contradiction: f(g(x)) = a ∧ f(g(x)) = b" (Left fg) (Right a) (Right b)
     ]
   where
     f = Mock.f (mkElemVar Mock.x)
+    fg = Mock.f (Mock.g (mkElemVar Mock.x))
     fInt = Mock.fInt (mkElemVar Mock.xInt)
     fBool = Mock.fBool (mkElemVar Mock.xBool)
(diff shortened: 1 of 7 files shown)


kore/test/Test/Kore/Step/Simplification/Condition.hs
```diff
@@ -76,13 +76,18 @@ test_simplify_local_functions =
     , test "contradiction: \"one\" = f(x) ∧ f(x) = \"two\"" fString (Left  strOne) (Right strTwo)
     , test "contradiction: \"one\" = f(x) ∧ \"two\" = f(x)" fString (Left  strOne) (Left  strTwo)
     , test "contradiction: f(x) = \"one\" ∧ \"two\" = f(x)" fString (Right strOne) (Left  strTwo)
+    -- Ignore Defined marker
+    , testDefined "contradiction: f(g(x)) = a ∧ f(g(x)) = b" (Right fg) (Right a) (Right b)
+    , testDefined "contradiction: f(g(x)) = a ∧ f(g(x)) = b" (Left fg) (Right a) (Right b)
```
ttuegel, line 81: What is the difference between `Right fg` and `Left fg`?

kore/test/Test/Kore/Step/Simplification/Condition.hs
```diff
@@ -110,6 +120,18 @@ test_simplify_local_functions =
...
 
+    testDefined name eitherFunc eitherC1 eitherC2 =
+        testCase name $ do
+            let equals1 =
+                    mkLocalDefn (applyDefined1 eitherFunc) eitherC1
+                        & Condition.fromPredicate
+                equals2 =
+                    mkLocalDefn (applyDefined2 eitherFunc) eitherC2
+                        & Condition.fromPredicate
+                condition = definedFg <> equals1 <> definedFg <> equals2
```
ttuegel, line 131: Using `definedFg` here actually requires that `eitherFunc` is `Right fg` or `Left fg`, so I would rather we not allow passing in any value there.

ttuegel (changes requested):

kore/test/Test/Kore/Step/Simplification/Condition.hs
ana-pantilie: `Right fg` will apply `mkDefined` to the `f(g(x)` in the second equals, and `Left fg` to the one in the first.

kore/test/Test/Kore/Step/Simplification/Condition.hs
ana-pantilie: Do you mean we should remove `definedFg` from `condition`?

ttuegel (approved):

PR #2395, opened 2021-02-09
Regenerate kevm and kwasm regression tests

### LINKED ISSUE #2409: Re-generate regression tests
The KEVM and KWASM regression tests need to be updated; this is blocking multiple tasks.

ttuegel:
Fixes an error about an unbound variable when the `kollect.sh` script fails. This isn't really a critical fix, because it only occurs when the script fails anyway.

Fixes #2409.

---

###### Review checklist


emarzion:
This does fix the issue in that the script is now exiting gracefully.

However, when running `generate.sh` it only completes a single test (`pop1`) before throwing `Did not collect any Kore files!`. Is this an issue with `kollect.sh` or with `generate.sh`, or am I doing something wrong?

scripts/kollect.sh
```diff
@@ -77,7 +77,7 @@ done
 
 echo '"$@"'
 
-if [[ -z "$kollected" ]]
+if [[ -z "${kollected:-}" ]]
 then
     echo >&2 "Did not collect any Kore files!"
     exit 1
```
ttuegel, line 83: This is what we do when it fails. We should add an `else` branch where we remove the temporary directory. We will need to add a variable where we save the name of the temporary directory.

scripts/kollect.sh
```diff
@@ -77,8 +49,10 @@ done
...
 fi
+
+#TODO Mircea: delete temporary folder at the end?
```
ttuegel, line 58: TODO?

MirceaS, line 58:
```suggestion
```
We agreed to no longer remove the run folder at the end of kollection

MirceaS (approved):

PR #2435, opened 2021-03-03
adding match for And

### LINKED ISSUE #2425: Matching \and patterns
Currently, the matcher treats `\and` patterns literally. This does not impact `simplification` rules because we simplify the left side of equations so that the `\and` is simplified. For #2341, we should stop simplifying the left side of equations. The frontend uses `\and` patterns for `#as` matching, so we need to support it properly.

We should add a new [matching case](https://github.com/kframework/kore/blob/c5cc46bee426516de975e21ba0338a1401737a54/kore/src/Kore/Step/Axiom/Matcher.hs#L173) which finds `\and` patterns and decomposes them into two constraints. The matching cases have the form

``` hs
matchOne (Pair φ₁ φ₂) = ...
```

where `φᵢ` are patterns. `matchOne` solves the implication

```
φ₂ → φ₁    -- Note the direction of the arrow!
```

We have an `\and` pattern when `φ₁ := t₁ ∧ t₂`,

```
φ₂ → t₁ ∧ t₂
-- or, decomposed into two constraints:
(φ₂ → t₁) ∧ (φ₂ → t₂)
```

So, we should detect `φ₁ := t₁ ∧ t₂`, and emit two new constraints with the same `φ₂`.

We also need to add some unit tests for this.

emarzion:
Fixes #2425

- Adding match for `And` such that a match on `And(x,y), z` pushes two matches, `x,z` and `y,z`.
- Adds unit tests.

---

###### Review checklist


andreiburdusa (approved):

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -185,6 +185,7 @@ matchOne pair =
     <|> matchInj         pair
     <|> matchOverload    pair
     <|> matchDefined     pair
+    <|> matchAnd         pair
```
MirceaS, line 188: Shouldn't this be moved further up or it doesn't matter?

ttuegel: I don't think it makes any difference because none of these cases can match the same terms, but myself I would move it up under `matchEqualHeads` because it's kind of a "primitive" operation.

MirceaS (approved):

PR #2344, opened 2021-01-11
Remove the Target type from Kore.Equation

andreiburdusa:
As https://github.com/kframework/kore/issues/2120#issuecomment-756852690 describes, but with an extra case for when swapping a pair produces a collision of keys.

---

###### Review checklist

The author performs the actions on the checklist. The reviewer evaluates the work and checks the boxes as they are completed.

-   [x] **Summary.** Write a summary of the changes. Explain what you did to fix the issue, and why you did it. Present the changes in a logical order. Instead of writing a summary in the pull request, you may push a clean Git history.
-   [ ] **Documentation.** Write documentation for new functions. Update documentation for functions that changed, or complete documentation where it is missing.
-   [ ] **Tests.** Write unit tests for every change. Write the unit tests that were missing before the changes. Include any examples from the reported issue as integration tests.
-   [ ] **Clean up.** The changes are already clean. Clean up anything near the changes that you noticed while working. This does not mean only spatially near the changes, but logically near: any code that interacts with the changes!

PR #2438, opened 2021-03-04
Inefficient Equation renaming

### LINKED ISSUE #2120: Inefficient Equation renaming
Renaming Equations with [`Kore.Equation.Application.targetEquationVariables`](https://github.com/kframework/kore/blob/8ee3470041bd6b5e6a9460fc775bda29036b01b5/kore/src/Kore/Equation/Application.hs#L401) is a significant inefficiency because it traverses the entire Equation every time it is attempted, even if there is nothing to rename. We cannot use the same approach as with [`Kore.Rewriting.RewritingVariable.RewritingVariableName`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Rewriting/RewritingVariable.hs#L64) because the variable parameter of the equations may vary during execution.

The `Target` variable type is used by the matcher so that it prefers to instantiate variables from the equation rather than from the term. It is not strictly necessary to have a distinct type for this, and that would eliminate the overhead of `targetEquationVariables`, but it leaves the overhead of the first `Equation.mapVariables` call, which is paid at every attempt of an equation.

My best suggestion is that we consider _fixing_ the variable type parameter so that we can use the `RewritingVariableName` approach (map all the equation variable names at initialization).

---

#### Plan

- [x] Fix all the `variable` type parameters to `RewritingVariableName` in `Kore.Equation.Application`. Do not remove the type parameter from any data types, but only fix the parameter where it appears. Follow the **Mikado Method** described below. (appetite: M)
- [x] Add an `EquationVariableName` to `RewritingVariableName`. It should be the "smallest" constructor according to `compare`. (appetite: S)
- [ ] Initialize all `Equation` to have `EquationVariableName` variables. (appetite: S)
- [ ] Don't change the variable type to `Target` in `Kore.Equation.Application`. Check before matching that the term has no free `EquationVariableName`s. Check after matching and unification that there are no free `EquationVariableName`s.

#### Mikado Method

The Mikado Method is a technique for making large changes to complex interdependent systems incrementally. The result should be a clean Git history showing incremental changes while refactoring the code to be more malleable. We will use `git stash` as a stack of pending changes.

1. **Goal:** Make the change we want to see (the "goal"). At the first iteration, this will be: fixing the type parameters of [`Kore.Equation.Application.attemptEquation`](https://github.com/kframework/kore/blob/3fd34f99a05745bbc792538042c0513724c3cf37/kore/src/Kore/Equation/Application.hs#L152).
2. **Test:** Build and run tests.
    1. **Pass:** If everything passes, we are done with this goal. Make a commit (`git commit ...`). Get the next pending goal (`git stash pop`) and **Test** again. If there are no more stashed changes, we are done.
    2. **Fail:** If the build or tests fail, identify the source of the failure. This will be the new goal. Stash your changes so far (`git stash`) and begin again at **Goal**.

Description: http://www.methodsandtools.com/archive/mikado.php

andreiburdusa:
Fixes #2120 
This pull request addresses the last two checkboxes, but `Target` was already removed in #2344.

---

###### Review checklist


kore/src/Kore/Equation/Application.hs
```diff
@@ -155,7 +156,11 @@ attemptEquation
     -> TermLike RewritingVariableName
     -> Equation RewritingVariableName
     -> simplifier (AttemptEquationResult RewritingVariableName)
-attemptEquation sideCondition termLike equation =
+attemptEquation
+    sideCondition
+    (not . freeEquationVariableName >>= assert -> termLike)
+    equation
+  =
```
ttuegel, line 163:
```suggestion
attemptEquation sideCondition termLike equation =
    assert  (not $ freeEquationVariableName termLike) $
```

kore/src/Kore/Rewriting/RewritingVariable.hs
```diff
@@ -302,3 +312,10 @@ isElementRuleVariable = isElementRuleVariableName . variableName
 
 isElementRuleVariableName :: ElementVariableName RewritingVariableName -> Bool
 isElementRuleVariableName = any isRuleVariable
+
+freeEquationVariableName
+    :: HasFreeVariables pat RewritingVariableName => pat -> Bool
+freeEquationVariableName pat =
+    any
+        isSomeEquationVariableName
+        (from @_ @(Set.Set _) . freeVariables @_ @RewritingVariableName $ pat)
```
ttuegel, line 321: I suggest naming this function `hasFreeEquationVariables`.

kore/src/Kore/Unification/Procedure.hs
```diff
@@ -64,8 +69,24 @@ unificationProcedure sideCondition p1 p2
...
+    assertNoFreeEquationVariableName (Conditional () predicate substitution) =
+        let
+            predicate' =
+                assert (not . freeEquationVariableName $ predicate) predicate
+            substitution' =
+                mapTerms
+                    (\t -> assert (not . freeEquationVariableName $ t) t)
+                    substitution
+        in
+            Conditional () predicate' substitution'
```
ttuegel, line 92: This assertion should appear in `Kore.Equation.Application` after matching and unification on function arguments.

ttuegel (changes requested):

kore/src/Kore/Equation/Application.hs
```diff
@@ -192,10 +194,13 @@ attemptEquation sideCondition termLike equation =
               & whileMatch
           (equation', predicate) <-
               applyAndSelectMatchResult matchResults
-          return
-              ( equation'
-              , makeAndPredicate predicate matchPredicate
-              )
+
+          assert (not . hasFreeEquationVariableName $ equation')
+              $ assert (not . hasFreeEquationVariableName $ predicate)
```
ana-pantilie, line 199: Could this be factored out? I think it's a little too verbose and it takes longer to read than necessary.

kore/src/Kore/Equation/Application.hs
```diff
@@ -156,6 +157,7 @@ attemptEquation
     -> Equation RewritingVariableName
     -> simplifier (AttemptEquationResult RewritingVariableName)
 attemptEquation sideCondition termLike equation =
+    assert (not $ hasFreeEquationVariableName termLike) $
```
ana-pantilie, line 160: Do we ever use `hasFreeEquationVariableName` on its own (without negating it)? Could we just turn `hasFreeEquationVariableName` into `withoutFreeEquationVariables`?

ana-pantilie (approved):

PR #2460, opened 2021-03-11
Fixed the names of Symbol Attribute fields

### LINKED ISSUE #2222: Dereference `unit` and `element` symbol attributes
We should use the `unit` and `element` symbol attributes so that we don't need `MetadataTools` in `Kore.Builtin.internalize`. The `MetadataTools` are used to construct the internal representations of builtin Map, Set, and List sorts; we need to look up the `unit` and `element` symbols of the sort's `concat` symbol. If this was in the symbol attributes already, there would be no need no look it up.

Needs kframework/k#1614

MirceaS:
Fixes #2222 

---

###### Review checklist


kore/src/Kore/Attribute/Symbol.hs
```diff
@@ -108,29 +108,29 @@ view the effective attributes, use the functions defined in this module.
...
+    , smtlib         :: !Smtlib
+    , smthook        :: !Smthook
+    , memo           :: !Memo
+    , klabel         :: !Klabel
+    , symbolKywd     :: !SymbolKywd
+    , noEvaluators   :: !NoEvaluators
+    , builtinUnit    :: !(Unit SymbolOrAlias)
+    , builtinElement :: !(Element SymbolOrAlias)
+    , builtinConcat  :: !(Concat SymbolOrAlias)
+    -- ^ The above three are only populated if the symbol is a builtinAc symbol
```
ttuegel, line 133: I don't understand what is "built-in" about these symbols. They are user-defined symbols only used for unparsing. (Now that I think of it, the comment should mention these are only used for unparsing.) If simply `unit`, `element`, and `concat` cause too many collisions, the next best names would be `unitSymbol`, etc. As for the documentation, what is a "builtinAc symbol"?

MirceaS: The fields are "built-in" in the sense that they represent the built-in symbol labels but I don't mind changing them if we find a better name for them. `unitSymbol`  and the rest are also unavailable due to import conflicts from `Unit`, `Element` and `Concat`.
And a built-in Ac symbol is a symbol of a built-in associative-commutative type. I tried to use the same naming convention that we use for `InternalAc`, NormalizedAc` etc.

MirceaS:
@ttuegel I addressed your comments with my latest commit but I wanna use this branch to also fix some errors that Virgil noticed related to the original PR, so I'll keep it open a little longer.

MirceaS:
#2222

PR #2535, opened 2021-04-07
Unification refactor

### LINKED ISSUE #2396: Optimize unification inner loop
The inner loop of the unification code has poor performance because using `MaybeT` involves sequencing many binds (`>>=`) in the underlying monad. We will separate the code which identifies unification problems from the code which solves the unification problem. A simplified schematic of the current code is:

``` hs
unify
    :: TermLike variable
    -> TermLike variable
    -> MaybeT unifier (Pattern variable)
unify term1 term2 = 
    matchAndUnify1 term1 term2
    <|> matchAndUnify2 term1 term2
    <|> matchAndUnify3 term1 term2
    ...
```

`unify` could mean `maybeTermAnd` or `maybeTermEquals`.

We will split each `matchAndUnifyN` into a pair:

``` hs
matchN :: TermLike variable -> TermLike variable -> Maybe problem
matchN = ...
{-# INLINE matchN #-}

unifyN :: problem -> unifier (Pattern variable)
```

`problem` stands for a data type representing each type of unification problem. It will be different for each pair of new functions. It is important that the `matchN` functions be inlined to make the pattern matching code more efficient, but they should be very small functions anyway. `unify` will take this simplified form:

``` hs
unify term1 term2
  | Just problem1 <- match1 term1 term2 = unify1 problem1 & lift
  | Just problem2 <- match2 term1 term2 = unify2 problem2 & lift
  | Just problem3 <- match3 term1 term2 = unify3 problem3 & lift
  ...
  | otherwise = empty
```

emarzion:
Fixes #2396 

---

###### Review checklist


kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -807,254 +800,276 @@ unifyEqualsNormalized
...
+    ( TermLike RewritingVariableName ->
+      TermLike RewritingVariableName ->
+      unifier (Pattern RewritingVariableName)
+    ) ->
+    [Element normalized (TermLike RewritingVariableName)] ->
+    [Element normalized (TermLike RewritingVariableName)] ->
+    Map Key (Value normalized (TermLike RewritingVariableName)) ->
+    Map Key (Value normalized (TermLike RewritingVariableName)) ->
+    [TermLike RewritingVariableName] ->
+    [TermLike RewritingVariableName] ->
```
ttuegel, line 822: Why are we taking these structures apart, instead of passing around the containers they came from? This is highly likely to introduce errors.

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -140,123 +140,417 @@ maybeTermEquals
...
+            second $
+                Overloading.unifyOverloading6 firstHead injChild
+    --unifyBoolAnd
+  | Just (UnifyBoolAnd term1 term2) <- matchUnifyBoolAnd1 first second
+        = Builtin.Bool.unifyBoolAnd childTransformers first second term1 term2
+  | Just (UnifyBoolAnd term1 term2) <- matchUnifyBoolAnd2 first second
+        = Builtin.Bool.unifyBoolAnd childTransformers second first term1 term2
+    --unifyBoolOr
+  | Just (UnifyBoolOr (UnifyBoolOrArgs term operand1 operand2)) <- matchUnifyBoolOr1 first second
+        = Builtin.Bool.unifyBoolOr childTransformers term operand1 operand2
```
ttuegel, line 491: As we discussed before, we should not be deconstructing the result of `matchUnifyBoolOr` here. The result should go to `unifyBoolOr` and be deconstructed there. It will make this code much clearer.

kore/src/Kore/Step/Simplification/Unify.hs
```diff
@@ -0,0 +1,1006 @@
...
+import Kore.Internal.TermLike
+import Kore.Rewriting.RewritingVariable
+import Kore.Step.Simplification.ExpandAlias
+import Kore.Syntax.PatternF
+    ( Const (..)
+    )
+import Kore.TopBottom
+import Prelude.Kore
+
+data Unification
```
ttuegel, line 117: This shouldn't exist. These types shouldn't be tied together. For each `match...` and `unify...` function, there should be one data type (unique to that pair) that goes between them. The unique data type and the `match...` function should be in the same module as the `unify...` function.

kore/src/Kore/Step/Simplification/Unify.hs
```diff
@@ -0,0 +1,1006 @@
...
+matchInt
+    :: TermLike RewritingVariableName
+    -> TermLike RewritingVariableName
+    -> Maybe Unification
+matchInt first second
+    | InternalInt_ int1 <- first
+    , InternalInt_ int2 <- second
+        = Just $ UnifyInt int1 int2
+    | otherwise = Nothing
+{-# INLINE matchInt #-}
```
ttuegel, line 181: Every one of these should be localized to the module where we implement unification for that type (in this case, `InternalInt`).

ttuegel (commented):
I'm very confused about what's happening in `Kore.Builtin.AssocComm`. It looks to me like there are substantial changes to that module, beyond what I would consider refactoring.

PR #2540, opened 2021-04-09
Fix unparsing built-ins

### LINKED ISSUE #2494: Unexpected internal builtin
For one of my proofs (in simplified form here), `kprove` outputs this:
```
    Unexpected internal builtin
      error, called at src/Kore/Builtin/External.hs:150:31 in kore-0.41.0.0-JMiLCqb6j8vI3LYm8kwv4e:Kore.Builtin.External
      toPatternF, called at src/Kore/Builtin/External.hs:77:18 in kore-0.41.0.0-JMiLCqb6j8vI3LYm8kwv4e:Kore.Builtin.External
Created bug report: kore-exec.tar.gz
  <k>
    f ( 2 |-> 2
    _0 , .Map , _1 ) ~> _DotVar1 ~> .
  </k>
#And
  {
    false
  #Equals
    2 in_keys ( _0 )
  }
```

The Haskell backend is at commit `de80594bad2e2a92e98650a5da34b143ec9180fd`

I think it fails when trying to concatenate a map variable, `M:Map` with `.Map`, but I'm not sure.

To reproduce:
tmp.k:
```k
module TMP
  imports MAP
  imports INT

  syntax KItem ::= lazyF(Map, Map)
  syntax KItem ::= "end"

  syntax KItem ::= f(Map, Map, Map)  [function, functional]

  rule lazyF(M:Map, N:Map) => f(M, .Map, N)

  rule f(K:KItem |-> V:KItem M:Map, N:Map, R:Map) => f(M, K |-> V N, R)
    requires notBool K ==K 2 andBool notBool K in_keys(N)
    [simplification(20)]

  rule f(K:KItem |-> _:KItem M:Map, N:Map, R:Map) => end
    requires K ==K 2 andBool R ==K M N
    [simplification(20)]

  rule f(_:Map, _:Map, _:Map) => end
    [simplification(50)]

endmodule
```

tmp-proof.k:
```k
module TMP-PROOF
  imports TMP

  claim lazyF(2|->2 _:Map, _:Map)
      => end

endmodule
```

bash:
```sh
kompile tmp.k --backend haskell && kprove tmp-proof.k
```

ttuegel:
Fixes #2494.

This pull request changes the structure of `Kore.Builtin.externalize` so that the error in #2494 is impossible.

---

###### Review checklist


kore/src/Kore/Builtin/AssocComm/AssocComm.hs
```diff
@@ -103,3 +114,66 @@ splitLastInit [a] = Just ([], a)
...
+                            addTerms ves $
+                                concatOpaques' opaq1 ve
+                    Just (opaqs, opaq2) -> pure $
+                        addTerms variableElements $
+                            addTermsOpaque opaqs $
+                                twoOpaques opaq1 opaq2
+            Just (concretes, concrete) -> pure $
+                addTerms variableElements $
+                    addTermsOpaque opaque $
+                        addTerms concretes concrete
```
ttuegel, line 158: We could clarify this in this way:

```suggestion
externalize1
    (UnitSymbol unitSymbol)
    (ConcatSymbol concatSymbol) = worker
  where
    worker [] [] [] = unit
    worker [] [] [singleOpaque] = Left singleOpaque
    worker concreteElements symbolicElements opaqueTerms =
        Right $ foldr1 concat' operands
      where
        operands = (map Free concreteElements) <> (map Free symbolicElements) <> (map Pure operands)
```

Another way might be:

```suggestion
externalize1
    (UnitSymbol unitSymbol)
    (ConcatSymbol concatSymbol) = worker
  where
    worker [] [] [] = unit
    worker concreteElements symbolicElements opaqueTerms =
        Right $ foldr concat' unit operands
      where
        operands = (map Free concreteElements) <> (map Free symbolicElements) <> (map Pure operands)
```

kore/src/Kore/Builtin/External.hs
```diff
@@ -44,94 +48,50 @@ externalize ::
...
-                (toPatternF . Recursive.project)
-                    (Map.asTermLike internalMap)
-            InternalSetF internalSet ->
-                (toPatternF . Recursive.project)
-                    (Set.asTermLike internalSet)
-            InjF inj ->
-                (toPatternF . Recursive.project . synthesize . ApplySymbolF)
-                    (Inj.toApplication inj)
-            _ -> toPatternF termLikeBase
+            AndF andF                 -> (Attribute.Null :<) $ Syntax.AndF         $ fmap Pure andF
```
ttuegel, line 58: We should add `From` instances for all of the `Syntax.PatternF` branches.

kore/src/Kore/Builtin/Map/Map.hs
```diff
@@ -247,3 +255,45 @@ asTermLike builtin =
...
+        )
+externalize1 builtin =
+    AssocComm.externalize1
+        (AssocComm.UnitSymbol unitSymbol)
+        (AssocComm.ConcatSymbol concatSymbol)
+        (map concreteElement (Map.toAscList concreteElements))
+        (element . unwrapElement <$> elementsWithVariables)
+        filteredMaps
+  where
+    filteredMaps = filter (not . isEmptyMap) opaque
```
ttuegel, line 278: Why filter the opaque terms? I fear too much modification between the internal and external structure will mask bugs in the future.

MirceaS: That was the behaviour of the `asTermLike` prior to my changes, but I'll remove that if you think that would be better.

kore/src/Kore/Builtin/AssocComm/AssocComm.hs
MirceaS: I think the first suggestion is not type correct.
the second suggestion might work (and we won't even need the base `[] [] []` case) but has the disadvantage that it always prepends a unit at the end, but that's probably fine if all we do is unparse.

kore/src/Kore/Builtin/Map/Map.hs
ttuegel: Yes, for unparsing I think that would be much better.

kore/src/Kore/Builtin/AssocComm/AssocComm.hs
ttuegel: I edited my suggestion above to remove the one that appends the unit (we decided that was wrong) and fixes the version using `foldr1`.

kore/src/Kore/Builtin/External.hs
MirceaS: I found it better to just define a small local function since all of these are effectively the same and we don't really wanna spread this externalization code too far wide. Let me know what you think of this.

MirceaS:
@ttuegel Any idea why the stylizer didn't apply a commit to this PR after my latest changes? It seems to fail due to that.

kore/src/Kore/Builtin/Int/Int.hs
```diff
@@ -46,10 +45,12 @@ import Data.String (
 import Data.Text (
     Text,
  )
-import qualified Data.Text as Text
 import Kore.Internal.InternalInt
 import Kore.Internal.Pattern as Pattern
-import Kore.Internal.TermLike as TermLike
+import Kore.Internal.TermLike as TermLike hiding (
+    DomainValueF,
+    StringLiteralF,
```
ttuegel, line 52: Why are these hidden now?

kore/src/Kore/Step/Axiom/Identifier.hs
```diff
@@ -88,18 +98,19 @@ Returns 'Nothing' if the 'TermLike' does not conform to one of the structures we
...
-                symbolId = Syntax.symbolOrAliasConstructor symbol
+    matchWorker (_ :< termLikeF) =
+        case termLikeF of
+            ApplyAliasF application ->
+                pure $
+                    Application $
+                        Syntax.symbolOrAliasConstructor $
+                            Alias.toSymbolOrAlias $
+                                Syntax.applicationSymbolOrAlias application
+            ApplySymbolF application -> mkAppId application
```
ttuegel, line 113: Why does the case for `ApplyAliasF` get inlined, but not `ApplySymbolF`?

kore/src/Kore/Step/Axiom/Identifier.hs
```diff
@@ -108,4 +119,73 @@ matchAxiomIdentifier = Recursive.fold matchWorker . Builtin.externalize
...
+
+        NormalizedAc{elementsWithVariables} = normalizedAc
+        NormalizedAc{concreteElements} = normalizedAc
+        NormalizedAc{opaque} = normalizedAc
+
+    acToId unitSymbol _ _ [] [] [] = pure $ Application $ symbolToId unitSymbol
+    acToId _ elementSymbol _ [_] [] [] = pure $ Application $ symbolToId elementSymbol
+    acToId _ elementSymbol _ [] [_] [] = pure $ Application $ symbolToId elementSymbol
+    acToId _ _ _ [] [] [opaque] = opaque
+    acToId _ _ concatSymbol _ _ _ = pure $ Application $ symbolToId concatSymbol
```
ttuegel, line 188: This is obviously complex. It should have unit tests.

kore/src/Kore/Syntax/Pattern.hs
```diff
@@ -105,6 +107,11 @@ newtype
     deriving anyclass (SOP.Generic, SOP.HasDatatypeInfo)
     deriving anyclass (Debug, Diff)
 
+type FutuPattern variable t =
```
ttuegel, line 110: We should revert these changes.

kore/test/Test/Kore/Builtin/External.hs
```diff
@@ -0,0 +1,343 @@
+{-# LANGUAGE Strict #-}
```
ttuegel, line 1: This is redundant, it is specified in `kore.cabal`.

kore/src/Kore/Builtin/Int/Int.hs
MirceaS: They conflict with the external `Syntax.Pattern` ones

kore/src/Kore/Step/Axiom/Identifier.hs
MirceaS: I was originally gonna leave them both inlined but then realised that I can use the `mkAppId` that I defined for other stuff to simplify the case for `ApplySymbolF`. But I'll add another function to encapsulate the Alias case as well if you feel like that would be better.

kore/src/Kore/Syntax/Pattern.hs
MirceaS: This is not the data structure that you originally introduced (I think that's gone). This is a new helper data structure that I defined and that is in use in the `External` module.
I suppose it could be moved to that module though, would that be good?

kore/src/Kore/Step/Axiom/Identifier.hs
MirceaS: We are already exercising this piece of code:
```
    , testGroup
        "Map"
        [ test
            "unitMap"
            (Mock.builtinMap [])
            (Application Mock.unitMapId)
        , test
            "elementMap"
            (Mock.builtinMap [(Mock.a, Mock.a)])
            (Application Mock.elementMapId)
        , test
            "concatMap"
            (Mock.builtinMap [(Mock.a, Mock.a), (Mock.b, Mock.b)])
            (Application Mock.concatMapId)
        ]
```
The above is from `Test.Kore.Step.Axiom.Identifier`. It's true that it isn't technically calling `acToId` directly but the functions that it goes through to get there only do some trivial unpacking of records (that's all `mapToId` and `setToId` do).
Do you think we need anything more than this?

ttuegel:
@MirceaS This looks fine to me now, but you need to leave a review yourself before merging it (because technically I opened the pull request).

MirceaS (approved):

PR #2572, opened 2021-04-23
Don't wrap "bare" Map and Set terms

### LINKED ISSUE #2501: Don't wrap "bare" Map and Set terms
"Bare" Map and Set terms (for example, functions) are wrapped [here](https://github.com/kframework/kore/blob/a4b771355cc2464c28a6a4144df7711d852bbbc2/kore/src/Kore/Builtin/AssociativeCommutative.hs#L229) with the internal structure that should be reserved for built-in Map and Set constructors. This should not happen.

andreiburdusa:
Fixes #2501 

---

###### Review checklist


ana-pantilie:
Is [this](https://github.com/kframework/kore/blob/39a42a5a7d2245086d31fa2a0b5fa1a84e40f456/kore/src/Kore/Step/Simplification/InternalMap.hs#L45) case related to this issue?

andreiburdusa:
> Is [this](https://github.com/kframework/kore/blob/39a42a5a7d2245086d31fa2a0b5fa1a84e40f456/kore/src/Kore/Step/Simplification/InternalMap.hs#L45) case related to this issue?

If there is a single opaque in that case, it comes from [normalizeAbstractElements](https://github.com/kframework/kore/blob/39a42a5a7d2245086d31fa2a0b5fa1a84e40f456/kore/src/Kore/Builtin/AssociativeCommutative.hs#L456), not from the place pointed to in the issue. I think it would be good to add a check there. What do you think, @ttuegel ? Should I add checks to other places where we create `NormalizedAc`s this way?

PR #2580, opened 2021-04-28
Use Makefile for generating regression tests

### LINKED ISSUE #2538: Update regression test script to use evm-semantics Makefile
[This script](https://github.com/kframework/kore/blob/master/scripts/generate-regression-tests.sh) should be updated to use the evm-semantics Makefile so that it gets changes that are made upstream.

ana-pantilie:
Fixes #2538

`scripts/generate-regression-tests.sh` now uses each project's Makefile to run the frontend for our regression tests.

---

###### Review checklist


dopamane (approved):

scripts/generate-regression-tests.sh
```diff
@@ -7,7 +7,18 @@ kollect() {
...
-    "$@" --save-temps --dry-run | xargs $KORE/scripts/kollect.sh "$name" >> "$name.sh"
+    "$@" | xargs $KORE/scripts/kollect.sh "$name" >> "$name.sh"
+    chmod +x "$name.sh"
+}
+
+kollect-file() {
+    local name="$1"
+    local path="$2"
+    shift
+    shift
```
andreiburdusa, line 18: Why did you use `shift` here?
Could this two calls be replaced with one `shift 2` call?

ana-pantilie: `kollect-file` works similarly to `kollect`, but it also takes the output file path as an argument, and we need to shift past that argument now as well. You're right that it can be replaced with `shift 2`.

andreiburdusa (approved):

PR #2599, opened 2021-05-05
Overhaul predicate simplifier

### LINKED ISSUE #2570: Overhaul predicate simplifier
This issue is the first part of addressing the issues raised in #2400. See also: #2566.

We will move `Condition.simplifyPredicate` to its own module and implement a `Predicate.simplify` similar to `TermLike.simplify`. It must not call `simplifyCondition` or `simplifyTermLike`. `Predicate.simplify` will simplify a `Predicate` from the bottom up; at each level, it will iterate until the predicate does not change.

Questions:

- What should be the return type of `Predicate.simplify`? Two reasonable-seeming options are a disjunction of `Condition`s, or a disjunction of `MultiAnd (Predicate _)`. In the former case, the `Condition` will not be simplified. What does the condition simplifier expect?

ttuegel:
Fixes #2570

---

###### Review checklist


test/imp/max-inconsistent-prelude-spec.k.out.golden
```diff
@@ -1,6 +1,6 @@
-kore-exec: [233424] Error (ErrorException):
```
ttuegel, line 1: This test must have been silently failing for a long time...

docs/simplify-predicate.md
```diff
@@ -0,0 +1,88 @@
...
+``` kore
+\and(P[1], P[2]) = \and(P[1], P[2](P[1] = \top))
+```
+
+## \or
+
+### Normalization
+
+Assuming the children of `\or` are in disjunctive normal form, then the clause
+is already normalized.
```
andreiburdusa, line 56: Maybe add something like "because `or` is associative", to anticipate `normalizeOr` better.

docs/simplify-predicate.md
```diff
@@ -0,0 +1,88 @@
+# Simplifying predicates
+
+The goals of simplification are:
+
+1. Transform the predicate into [disjunctive normal form] (normalization).
+2. Apply identities to reduce the complexity of the predicate (simplification proper).
+
+Simplification happens recursively, upward from the bottom. Therefore, we
+describe the simplification steps based on the shape of the top-most
+unsimplified clause of the predicate.
```
andreiburdusa, line 10: This phrasing is not clear for me. If we start simplifying from the bottom, wouldn't the top look the same until the end? Maybe my intuition of what is a clause is wrong.

ttuegel:
I tested this locally and found no significant performance difference with master. It's also worth noting that the proofs which show > 5% increased allocation are all very small.

test/map-symbolic/lookup-3-spec.k.out.golden
```diff
@@ -18,7 +18,7 @@
     true
   #Equals
     Z:MyId in_keys ( MAP
-    ( Y:MyId |-> 2 ) )
+    ( Y:MyId |-> 1 ) )
```
ana-pantilie, line 21: This doesn't look right, we have `X ==K Y`, and we update the value `Y` points to, so it should be `2` not `1`, right?

ttuegel: The meaning of the output is unchanged. The Map appears under `in_keys`, so actually the values are irrelevant.

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -0,0 +1,330 @@
...
+            & pure
+    bottom = normalizeBottom Bottom{bottomSort = notSort}
+
+{- |
+ @
+ \\implies(L, R) = \\or(\\not(L), \\and(L, R))
+ @
+
+ Note: @L@ is carried through to the right-hand side of 'Implies' to maximize
+ the information content of that branch.
```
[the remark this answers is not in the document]
ttuegel: We simplify `\implies` sometimes as the result of `\equals` simplification, if either side is not a function-like pattern. This isn't very common. Carrying the left-hand side over improves simplification a bit in these cases; it can rule out some branches which were also carrying `\not(L)` from another part.

test/imp/max-inconsistent-prelude-spec.k.out.golden
ttuegel: The test actually only searches the output for a particular string with `grep`. I will update the test in another pull request so that this is clearer.

ana-pantilie (approved):

PR #2602, opened 2021-05-06
Refactor Unification Loop

### LINKED ISSUE #2396: Optimize unification inner loop

emarzion:
Fixes #2396

Replacing branch #2535 

The code for [maybeTermEquals](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/AndTerms.hs#L130) and [maybeTermAnd](https://github.com/kframework/kore/blob/master/kore/src/Kore/Step/Simplification/AndTerms.hs#L130) are being modified as follows:  Each `asum` is replaced by a list of patterns + function calls, with each invoked function being replaced by two functions, the first corresponding to the pattern match, which returns the necessary data for the second function which performs the unification, now accepting the data from the match as a single argument.   The current conventions for returning the data in this PR are:

1. If there are two or more possible matches corresponding to one of these functions, the data is returned as part of a custom sum type with two or more constructors; otherwise, it is returned as is.
2. If, within a single match, only one piece of data is returned, then it is returned as is (without being wrapped; the exception to this is if the type is too verbose in which case a newtype is used); otherwise, the data is packaged and returned together in a custom record type.
3. If a function has only a single match and returns no data, then `()` is returned (since we are returning `Maybe ()` it may be better to just return a `Bool`, but in the interest of regularity I've chosen the former option).

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Builtin/Bool.hs
+++ b/kore/src/Kore/Builtin/Bool.hs
@@ -169,7 +169,19 @@ data UnifyBool = UnifyBool
     { bool1, bool2 :: !InternalBool
     }
 
--- | Matches two Bool values.
+{- | Matches
+
+@
+\\equals{_, _}(\\dv{Bool}(_), \\dv{Bool}(_))
+@
+
+and
+
+@
+\\and{_}(\\dv{Bool}(_), \\dv{Bool}(_))
+@
+
+-}
 matchBools ::
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName ->
@@ -200,9 +212,18 @@ unifyBool termLike1 termLike2 unifyData
   where
     UnifyBool{bool1, bool2} = unifyData
 
-{- | Matches two terms when first is a true bool term
-    and the second is a function pattern matching
-    the @BOOL.and@ hooked symbol.
+{- | Matches
+
+@
+\\equals{_, _}(\\dv{Bool}("true"), andBool(_,_))
+@
+
+and
+
+@
+\\and{_}(\\dv{Bool}("true"), andBool(_,_))
+@
+
 -}
 matchUnifyBoolAnd ::
     TermLike RewritingVariableName ->
@@ -254,9 +275,18 @@ unifyBothWith unify termLike1 operand1 operand2 = do
     unify' term1 term2 =
         Pattern.withoutTerm <$> unify term1 term2
 
-{- | Matches two terms when first is a false bool term
-    and the second is a function pattern matching
-    the @BOOL.or@ hooked symbol.
+{- | Matches
+
+@
+\\equals{_, _}(\\dv{Bool}("false"), boolOr(_,_))
+@
+
+and
+
+@
+\\and{_}(\\dv{Bool}("false"), boolOr(_,_))
+@
+
 -}
 matchUnifyBoolOr ::
     TermLike RewritingVariableName ->
@@ -287,9 +317,18 @@ data UnifyBoolNot = UnifyBoolNot
     , value :: Bool
     }
 
-{- | Matches two terms when second is a bool term
-    and the first is a function pattern matching
-    the @BOOL.not@ hooked symbol.
+{- | Matches
+
+@
+\\equals{_, _}(notBool(_), \\dv{Bool}(_))
+@
+
+and
+
+@
+\\and{_}(notBool(_), \\dv{Bool}(_))
+@
+
 -}
 matchUnifyBoolNot ::
     TermLike RewritingVariableName ->
--- a/kore/src/Kore/Builtin/Bool.hs
+++ b/kore/src/Kore/Builtin/Bool.hs
@@ -189,13 +189,13 @@ unifyBool ::
     UnifyBool ->
     unifier (Pattern RewritingVariableName)
 unifyBool termLike1 termLike2 unifyData
-    | bool1 == bool2
-    = return (Pattern.fromTermLike termLike1)
-    | otherwise
-    = Unify.explainAndReturnBottom
-        "different Bool domain values"
-        termLike1
-        termLike2
+    | bool1 == bool2 =
+        return (Pattern.fromTermLike termLike1)
+    | otherwise =
+        Unify.explainAndReturnBottom
+            "different Bool domain values"
+            termLike1
+            termLike2
   where
     UnifyBool{bool1, bool2} = unifyData
 
--- a/kore/src/Kore/Builtin/Bool.hs
+++ b/kore/src/Kore/Builtin/Bool.hs
@@ -188,21 +188,21 @@ unifyBool ::
     TermLike RewritingVariableName ->
     UnifyBool ->
     unifier (Pattern RewritingVariableName)
-unifyBool termLike1 termLike2 unifyData =
-    if bool1 == bool2
-        then return (Pattern.fromTermLike termLike1)
-        else
-            Unify.explainAndReturnBottom
-                "different Bool domain values"
-                termLike1
-                termLike2
+unifyBool termLike1 termLike2 unifyData
+    | bool1 == bool2
+    = return (Pattern.fromTermLike termLike1)
+    | otherwise
+    = Unify.explainAndReturnBottom
+        "different Bool domain values"
+        termLike1
+        termLike2
   where
     UnifyBool{bool1, bool2} = unifyData
 
 matchUnifyBoolAnd ::
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName ->
-    Maybe (BoolAnd (TermLike RewritingVariableName))
+    Maybe BoolAnd
 matchUnifyBoolAnd first second
     | Just True <- matchBool first
       , Just boolAnd <- matchBoolAnd second
@@ -216,7 +216,7 @@ unifyBoolAnd ::
     MonadUnify unifier =>
     TermSimplifier RewritingVariableName unifier ->
     TermLike RewritingVariableName ->
-    BoolAnd (TermLike RewritingVariableName) ->
+    BoolAnd ->
     unifier (Pattern RewritingVariableName)
 unifyBoolAnd unifyChildren term boolAnd =
     unifyBothWith unifyChildren term operand1 operand2
@@ -251,7 +251,7 @@ unifyBothWith unify termLike1 operand1 operand2 = do
 matchUnifyBoolOr ::
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName ->
-    Maybe (BoolOr (TermLike RewritingVariableName))
+    Maybe BoolOr
 matchUnifyBoolOr first second
     | Just value1 <- matchBool first
       , not value1
@@ -266,7 +266,7 @@ unifyBoolOr ::
     MonadUnify unifier =>
     TermSimplifier RewritingVariableName unifier ->
     TermLike RewritingVariableName ->
-    BoolOr (TermLike RewritingVariableName) ->
+    BoolOr ->
     unifier (Pattern RewritingVariableName)
 unifyBoolOr unifyChildren termLike boolOr =
     unifyBothWith unifyChildren termLike operand1 operand2
@@ -274,7 +274,7 @@ unifyBoolOr unifyChildren termLike boolOr =
     BoolOr{operand1, operand2} = boolOr
 
 data UnifyBoolNot = UnifyBoolNot
-    { boolNot :: BoolNot (TermLike RewritingVariableName)
+    { boolNot :: BoolNot
     , value :: Bool
     }
 
@@ -310,13 +310,13 @@ matchBool (InternalBool_ InternalBool{internalBoolValue}) =
 matchBool _ = Nothing
 
 -- | The @BOOL.and@ hooked symbol applied to @term@-type arguments.
-data BoolAnd term = BoolAnd
+data BoolAnd = BoolAnd
     { symbol :: !Symbol
-    , operand1, operand2 :: !term
+    , operand1, operand2 :: !(TermLike RewritingVariableName)
     }
 
 -- | Match the @BOOL.and@ hooked symbol.
-matchBoolAnd :: TermLike variable -> Maybe (BoolAnd (TermLike variable))
+matchBoolAnd :: TermLike RewritingVariableName -> Maybe BoolAnd
 matchBoolAnd (App_ symbol [operand1, operand2]) = do
     hook2 <- (getHook . symbolHook) symbol
     Monad.guard (hook2 == andKey)
@@ -324,13 +324,13 @@ matchBoolAnd (App_ symbol [operand1, operand2]) = do
 matchBoolAnd _ = Nothing
 
 -- | The @BOOL.or@ hooked symbol applied to @term@-type arguments.
-data BoolOr term = BoolOr
+data BoolOr = BoolOr
     { symbol :: !Symbol
-    , operand1, operand2 :: !term
+    , operand1, operand2 :: !(TermLike RewritingVariableName)
     }
 
 -- | Match the @BOOL.or@ hooked symbol.
-matchBoolOr :: TermLike variable -> Maybe (BoolOr (TermLike variable))
+matchBoolOr :: TermLike RewritingVariableName -> Maybe BoolOr
 matchBoolOr (App_ symbol [operand1, operand2]) = do
     hook2 <- (getHook . symbolHook) symbol
     Monad.guard (hook2 == orKey)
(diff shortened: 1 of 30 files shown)


kore/src/Kore/Builtin/Bool.hs
```diff
@@ -163,48 +165,63 @@ builtinFunctions =
...
+    }
+
+matchBools ::
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    Maybe UnifyBool
+matchBools first second
+    | InternalBool_ bool1 <- first
+      , InternalBool_ bool2 <- second =
+        Just $ UnifyBool bool1 bool2
```
ttuegel, line 179: If we are going to use name shadowing, let's do it properly with named field puns:

```suggestion
        Just $ UnifyBool{bool1, bool2}
```

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -163,48 +165,63 @@ builtinFunctions =
...
+    UnifyBool ->
+    unifier (Pattern RewritingVariableName)
+unifyBool termLike1 termLike2 unifyData =
+    if bool1 == bool2
+        then return (Pattern.fromTermLike termLike1)
+        else
+            Unify.explainAndReturnBottom
+                "different Bool domain values"
+                termLike1
+                termLike2
```
ttuegel, line 198:
```suggestion
unifyBool termLike1 termLike2 unifyData
  | bool1 == bool2 = return (Pattern.fromTermLike termLike1)
  | otherwise =
    Unify.explainAndReturnBottom
        "different Bool domain values"
        termLike1
        termLike2
```

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -163,48 +165,63 @@ builtinFunctions =
...
-                            termLike1
-                            termLike2
-    worker _ _ = empty
+    UnifyBool{bool1, bool2} = unifyData
+
+matchUnifyBoolAnd ::
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    Maybe (BoolAnd (TermLike RewritingVariableName))
+matchUnifyBoolAnd first second
```
ttuegel, line 206: Where is the `UnifyBoolAnd` type?

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -231,44 +248,60 @@ unifyBothWith unify termLike1 operand1 operand2 = lift $ do
     unify' term1 term2 =
         Pattern.withoutTerm <$> unify term1 term2
 
+matchUnifyBoolOr ::
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    Maybe (BoolOr (TermLike RewritingVariableName))
```
ttuegel, line 254: Do `BoolAnd` and `BoolOr` really need this type parameter? Isn't it always `TermLike RewritingVariableName`?

kore/src/Kore/Builtin/Int.hs
```diff
@@ -424,24 +428,57 @@ matchIntEqual =
...
-    TermLike variable ->
-    MaybeT unifier (Pattern variable)
-unifyInt term1@(InternalInt_ int1) term2@(InternalInt_ int2) =
-    assert (on (==) internalIntSort int1 int2) $ lift worker
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    UnifyInt ->
+    unifier (Pattern RewritingVariableName)
+unifyInt term1 term2 unifyData =
+    assert (on (==) internalIntSort int1 int2) worker -- should this be part of match?
```
ttuegel, line 457: Yes, this assertion should be part of the matching function.

kore/src/Kore/Step/Simplification/ExpandAlias.hs
```diff
@@ -34,25 +27,21 @@ import Kore.Internal.TermLike (
...
+    { term1, term2 :: !(TermLike RewritingVariableName)
+    }
+
+matchExpandAlias ::
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName ->
-    MaybeT unifier (Pattern RewritingVariableName)
-expandAlias recurse t1 t2 =
+    Maybe UnifyExpandAlias
+matchExpandAlias t1 t2 =
```
ttuegel, line 40: What is the corresponding function for this one, that carries out unification?

kore/src/Kore/Step/Simplification/InjSimplifier.hs
```diff
@@ -66,11 +80,12 @@ data InjSimplifier = InjSimplifier
       --
       --        Returns 'Distinct' if the sort injections cannot match, or 'Unknown' if
       --        further simplification could produce matching injections.
-      unifyInj ::
+      unifyInjs ::
         forall variable.
         InternalVariable variable =>
         Inj (TermLike variable) ->
         Inj (TermLike variable) ->
+        Either Distinct InjUnify ->
```
ttuegel, line 88: `InjUnify` should contain all the data which `unifyInjs` needs to do its job. This will minimize the risk that `unifyInjs` is called with the wrong `Inj`s for the supplied match data.

kore/src/Kore/Step/Simplification/NoConfusion.hs
```diff
@@ -27,6 +25,31 @@ import Prelude.Kore hiding (
     concat,
  )
 
+data UnifyEqualInjectiveHeadsAndEquals = UnifyEqualInjectiveHeadsAndEquals
```
ttuegel, line 28:
```suggestion
data UnifyEqualInjectiveHeads = UnifyEqualInjectiveHeads
```

kore/src/Kore/Step/Simplification/NoConfusion.hs
```diff
@@ -27,6 +25,31 @@ import Prelude.Kore hiding (
...
+matchEqualInjectiveHeadsAndEquals ::
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    Maybe UnifyEqualInjectiveHeadsAndEquals
+matchEqualInjectiveHeadsAndEquals first second
+    | App_ firstHead firstChildren <- first
+      , App_ secondHead secondChildren <- second
+      , Symbol.isInjective firstHead
+      , Symbol.isInjective secondHead
+      , firstHead == secondHead --is one of the above redundant in light of this?
```
ttuegel, line 43: You can remove `Symbol.isInjective secondHead` above, and leave a comment explaining.

kore/src/Kore/Step/Simplification/NoConfusion.hs
```diff
@@ -39,29 +62,50 @@ equalInjectiveHeadsAndEquals ::
...
+        UnifyEqualInjectiveHeadsAndEquals
+            { firstHead
+            , firstChildren
+            , secondChildren
+            } = unifyData
+
+-- data ConstructorAndEqualsAssumesDifferentHeads = ConstructorAndEqualsAssumesDifferentHeads
+--     { firstHead, secondHead :: Symbol
+--     }
+
```
ttuegel, line 91:
```suggestion
```

kore/src/Kore/Step/Simplification/NoConfusion.hs
```diff
@@ -39,29 +62,50 @@ equalInjectiveHeadsAndEquals ::
...
+matchConstructorAndEqualsAssumesDifferentHeads ::
```
ttuegel, line 92: This function no longer assumes anything.

```suggestion
matchDifferentConstructors ::
```

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -127,6 +132,158 @@ matchOverloading termPair =
             Simple pair -> return pair
             _ -> notApplicable
 
+-- flipPairBack :: Pair a -> Pair a
+-- flipPairBack (Pair x y) = Pair y x
+
```
ttuegel, line 137:
```suggestion
```

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -1,16 +1,21 @@
+{-# LANGUAGE PartialTypeSignatures #-}
+
```
ttuegel, line 2:
```suggestion
```

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -127,6 +132,158 @@ matchOverloading termPair =
...
+    , injChild :: TermLike RewritingVariableName
+    }
+
+data OverloadedConstructorSortInjectionAndEquals
+    = Overload1 UnifyOverload1
+    | Overload2 UnifyOverload2
+    | Overload3 UnifyOverload3
+    | Overload4 UnifyOverload4
+    | Overload5 UnifyOverload5
+    | Overload6 UnifyOverload6
```
ttuegel, line 180: These all need real names.

ttuegel (changes requested):
In addition, all of the new top-level definitions need documentation. For the unification cases (which is most of it) this could be as simple as showing the Kore terms that are being unified.

kore/src/Kore/Builtin/Bool.hs
emarzion: There is none, here I am following the convention that if the match is returning only a single datum that we just return that without wrapping.

kore/src/Kore/Builtin/Bool.hs
emarzion: From the looks of it, they probably don't anymore, I'm guessing this might be something left over from the `variable` to `RewritingVariableName` change?  In any case I'll go ahead and change those.

kore/src/Kore/Step/Simplification/ExpandAlias.hs
emarzion: The terms are fed back into `maybeTermAnd` [here](https://github.com/kframework/kore/blob/2a9962c6b9b742952e4ccdb134ae42b823ff8b4f/kore/src/Kore/Step/Simplification/AndTerms.hs#L238).

kore/src/Kore/Step/Simplification/Overloading.hs
emarzion: This code isn't currently being used in this PR, should I just remove it for now?

kore/src/Kore/Builtin/Bool.hs
```diff
@@ -163,48 +165,69 @@ builtinFunctions =
...
-    InternalVariable variable =>
+    forall unifier.
     MonadUnify unifier =>
-    TermLike variable ->
-    TermLike variable ->
-    MaybeT unifier (Pattern variable)
-unifyBool a b =
-    worker a b <|> worker b a
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
```
ttuegel, line 189: We should not pass in the original `TermLike`s to any `unify...` function.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -424,24 +428,58 @@ matchIntEqual =
...
+
+-- | Matches two Int values that have equals sorts.
+matchInt ::
+    TermLike RewritingVariableName ->
+    TermLike RewritingVariableName ->
+    Maybe UnifyInt
+matchInt first second
+    | InternalInt_ int1 <- first
+      , InternalInt_ int2 <- second
+      , on (==) internalIntSort int1 int2 =
```
ttuegel, line 443: I think this should be an assertion. Attempting to unify two things with different sorts doesn't make any sense.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -424,24 +428,58 @@ matchIntEqual =
...
+    UnifyInt{int1, int2} = unifyData
+
+data UnifyIntEq = UnifyIntEq
+    { eqTerm :: !(EqTerm (TermLike RewritingVariableName))
+    , value :: !Bool
+    }
+
+{- | Matches two terms when second is a bool term
+    and the first is a function pattern matching
+    the @INT.eq@ hooked symbol.
```
ttuegel, line 470: I think we can make all of this documentation clearer using the Kore notation:

```suggestion
{- | Matches
@
\\equals{_, _}(\\dv{Bool}(_), eqInt{_}(_, _))
@
```

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -112,10 +113,12 @@ termUnification notSimplifier = \term1 term2 ->
         TermLike RewritingVariableName ->
         unifier (Pattern RewritingVariableName)
     termUnificationWorker pat1 pat2 = do
+        injSimplifier <- Simplifier.askInjSimplifier
+        OverloadSimplifier{isOverloaded} <- Simplifier.askOverloadSimplifier
         let maybeTermUnification ::
                 MaybeT unifier (Pattern RewritingVariableName)
             maybeTermUnification =
-                maybeTermAnd notSimplifier termUnificationWorker pat1 pat2
+                maybeTermAnd notSimplifier termUnificationWorker injSimplifier isOverloaded pat1 pat2
```
ttuegel, line 121: Why are we passing the new parameters? `maybeTermAnd` can retrieve these from its `unifier` monad.

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -647,14 +826,21 @@ compareForEquals first second
...
+      , bytesFirst /= bytesSecond =
+        Just ()
+    | otherwise = Nothing
+{-# INLINE matchBytesDifferent #-}
+
+bytesDifferent ::
+    MonadUnify unifier =>
+    unifier (Pattern RewritingVariableName)
+bytesDifferent =
+    return Pattern.bottom
```
ttuegel, line 846: This code would be simplified if we did not assume that the `Bytes` are different, but checked.

kore/src/Kore/Step/Simplification/Overloading.hs
```diff
@@ -465,7 +465,7 @@ mkInj ::
     Inj () ->
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName
-mkInj inj injChild = (synthesize . InjF) inj{injChild}
+mkInj inj injChild = (synthesize . InjF) (inj :: Inj ()){injChild}
```
ttuegel, line 468: Why is this necessary? The type of `inj` should be fixed by the function's signature.

ttuegel (changes requested):

kore/src/Kore/Step/Simplification/AndTerms.hs
emarzion: They are used in some match functions (`matchSortInjectionAndEquals` for instance) so we do not have access at that point to the `unifier` monad.

kore/src/Kore/Step/Simplification/AndTerms.hs
emarzion: I don't understand, is this not being done on line 837?

kore/src/Kore/Builtin/Bool.hs
emarzion: We are putting this off for the time being per our discussion at the standup about the `explainAndReturnBottom` function.

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -647,14 +920,38 @@ compareForEquals first second
...
```
ttuegel, line 957:
```suggestion
matchBytesDifferent first second
    | _ :< InternalBytesF (Const bytesFirst) <- Recursive.project first
      , _ :< InternalBytesF (Const bytesSecond) <- Recursive.project second =
        Just ()
    | otherwise = Nothing
{-# INLINE matchBytesDifferent #-}

bytesDifferent ::
    MonadUnify unifier =>
    unifier (Pattern RewritingVariableName)
bytesDifferent UnifyBytes{ bytesFirst, bytesSecond }
    | bytesFirst == bytesSecond =
        return $ Pattern.fromTermLike $ internalBytes bytesFirst
    | otherwise =
        empty
```

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -647,14 +920,40 @@ compareForEquals first second
...
+    | otherwise = Nothing
+{-# INLINE matchBytes #-}
+
+bytesDifferent ::
+    MonadUnify unifier =>
+    UnifyBytes ->
+    unifier (Pattern RewritingVariableName)
+bytesDifferent UnifyBytes{ bytes1, bytes2 }
+    | bytes1 == bytes2 =
+        return $ Pattern.fromTermLike $ mkInternalBytes' bytes1
```
emarzion, line 957: Is it better to use a `mk` function instead of returning the original term?

ttuegel (approved):

PR #2636, opened 2021-05-18
Add a debug log entry for proof generation hints

rod-lin:
These commits add a log entry `Kore.Log.DebugRewriteSubstitution` that prints the axiom id and substitution used for each rewrite, which is used in https://github.com/kframework/matching-logic-proof-checker as hints to generate proof objects for rewriting.

The log output looks like yaml format
```
kore-exec: [5208286] Debug (DebugRewriteSubstitution):
    - type: rewriting
      from: <initial term>
      rule-id: <axiom unique id>
      substitution:
      - key: <kore-variable>
        value: <kore-pattern>
      ...
```

The log entry is invoked after `applyWithFinalizer` is finished.

I was not particularly familiar with the code base, so I only managed to get the information needed for an instance of `applyWithFinalizer` (when `rule = RulePattern RewritingVariableName`), which I explicitly added as `applyRuleWithFinalizer`.

---

###### Review checklist


kore/src/Kore/Step/RewriteStep.hs
```diff
@@ -292,6 +295,21 @@ applyWithFinalizer finalize rules initial = do
...
+    -- | Rewrite rules
+    [RulePattern RewritingVariableName] ->
+    -- | Configuration being rewritten
+    Pattern RewritingVariableName ->
+    simplifier (Results (RulePattern RewritingVariableName))
+applyRuleWithFinalizer finalize rules initial = do
+    results <- applyWithFinalizer finalize rules initial
+    debugRewriteSubstitution initial results
+    return results
+{-# INLINE applyRuleWithFinalizer #-}
```
ana-pantilie, line 311: My understanding from the PR description is that you also want to log claim applications, right? I think this should be do-able in the following way:
- don't keep whole `UnifiedRule`s in `DebugRewriteSubstitution`, as I said in the comment above
- add a `From rule UniqueId` constraint [here](https://github.com/kframework/kore/blob/cc94a3f594469011d790d6d89e01069caf13011d/kore/src/Kore/Step/RewriteStep.hs#L171) (like the one for `SourceLocation`)
- add a `From rule UniqueId` constraint in `applyWithFinalizer`
- then I think you'll need to implement `From rule UniqueId` instances for each `rule` the compiler complains about

After these changes you should be able to call `debugRewriteSubstitution` (modified accordingly) in `applyWithFinalizer`, so it will work for both axioms and claims. Please let me know if you run into any problems.

ana-pantilie (changes requested):

rod-lin:
@ana-pantilie thanks for the comments!

I have made the corresponding changes.

For the String one though, I was trying to output something like
```
- type: rewriting
  from: <pattern>
```
where the kore pattern <pattern> are encoded in a **single** line to save space.

I'm not sure how to achieve that without unparsing it to string/text first.

kore/src/Kore/Log/DebugRewriteSubstitution.hs
```diff
@@ -0,0 +1,97 @@
...
+
+instance Pretty DebugRewriteSubstitution where
+    pretty DebugRewriteSubstitution{configuration, appliedRules} =
+        Pretty.vsep $ unparseRule <$> appliedRules
+      where
+        ruleInfo :: UniqueId -> [Pretty.Doc ann]
+        ruleInfo uniqueId =
+            [ "- type: rewriting"
+            , "  from: >"
+            , Pretty.indent 4 $ unparse $ Conditional.term configuration
```
ana-pantilie, line 60: Why does this only print the `term`?

ana-pantilie (commented):
> where the kore pattern are encoded in a single line to save space

Hmm, I think a way we could achieve that is if we wrapped the kore patterns in [Pretty.group](https://hackage.haskell.org/package/prettyprinter-1.7.0/docs/Prettyprinter.html#v:group). We'd need to also change the layout to allow unbounded lines, [here](https://github.com/kframework/kore/blob/7a66654d8be69a3500b6f32d08717c0c8168236b/kore/src/Kore/Log.hs#L200), but this is a global change.

@ttuegel is there any reason why we use the current layout settings? I don't think changing to unbounded lines should impact the other log entries, but if I'm wrong or if there's any reason we don't want that, could we maybe add a user option to switch to this type of layout?

rod-lin:
I see. I added `group` without changing anything globally. We can work with that for now.
I also added the constraints. Thanks for reminding!

ttuegel:
Let's add a new command-line option to output the `DebugRewriteSubstitution` entries to their own YAML file. Then, we don't need to worry about how anything else is pretty-printed, and the prover doesn't have to try to extract data from our log. So, we would have an option like `--trace-rewrites FILENAME`, and if that option is given then the backend will open `FILENAME` and redirect all these `DebugRewriteSubstitution` entries there, printing them in YAML format. The log entries should implement the [`ToJSON`](https://hackage.haskell.org/package/yaml-0.11.5.0/docs/Data-Yaml.html) class. (I know that's strange; recall that JSON is a proper subset of YAML.)

rod-lin:
@ttuegel I added the option `--trace-rewrites` in `KoreLogOptions`,
and the special handling of `DebugRewriteSubstitution`, basically duplicating `solverTranscriptLogger`.

A dependency on `yaml` is also added.

rod-lin:
Sorry after a discussion today, we decided to make some changes in the format and add a bit more stuff.
I'll let you know when it's ready for review again.

ana-pantilie:
Hi @rod-lin. Is this PR actively worked on anymore? There doesn't seem to have been any activity for quite a while now.

rod-lin:
Sorry forgot to reply. Yeah I think I'll just maintain it on my own branch for now. Thanks!

amelieled:
This could be very useful for me! \o/

PR #2637, opened 2021-05-19
Reimplement Condition.simplifyPredicates

### LINKED ISSUE #2454: SideCondition.andCondition is unexpectedly quadratic
The regression involves `SideCondition.fromCondition` and `SideCondition.andCondition`.

ana-pantilie:
Fixes #2454

See https://github.com/kframework/kore/issues/2454#issuecomment-842465303

---

###### Review checklist


ana-pantilie:
`tests/specs/benchmarks/storagevar03-spec.k` from `evm-semantics` runs ~25% faster with these changes.

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -200,80 +202,70 @@ instance From (SideCondition variable) (MultiAnd (Predicate variable)) where
...
+addPredicates ::
+    Ord variable =>
+    MultiAnd (Predicate variable) ->
+    SideCondition variable ->
+    SideCondition variable
+addPredicates predicates sideCondition =
+    sideCondition
+        { assumedTrue =
+            predicates <> assumedTrue sideCondition
+        }
```
ttuegel, line 257: I would have expected to see `assumeTrue predicates <> sideCondition` here.

kore/src/Kore/Reachability/Claim.hs
```diff
@@ -530,8 +530,11 @@ checkImplicationWorker (ClaimPattern.refreshExistentials -> claimPattern) =
     -- the claim is fully simplified at every step, that should not be a
     -- problem.
     sideCondition =
-        SideCondition.assumeTrueCondition
-            (Condition.fromPredicate . Condition.toPredicate $ leftCondition)
+        SideCondition.fromConditionWithReplacements
+            ( Condition.fromPredicate
+                . Condition.toPredicate
```
ttuegel, line 535: What does this do?

kore/src/Kore/Step/Simplification/Condition.hs
```diff
@@ -120,49 +123,71 @@ simplify SubstitutionSimplifier{simplifySubstitution} sideCondition =
...
+            (LogicT simplifier)
+            ()
+    simplifyWithAssumptions (predicate, unsimplifiedSideCond) = do
+        simplifiedSideCond <- State.get
+        let otherSideConds =
+                SideCondition.addPredicates
+                    ( from @_ @(MultiAnd (Predicate _)) simplifiedSideCond
+                        <> unsimplifiedSideCond
+                    )
+                    sideCondition
```
ttuegel, line 188: Why not `SideCondition.addPredicates unsimplifiedSideCond (simplifiedSideCond <> sideCondition)`?

kore/src/Kore/Reachability/Claim.hs
ana-pantilie: It moves the `substitution` back into the `predicate`, I think the TODO above explains why we have to do this.

kore/src/Kore/Internal/SideCondition.hs
ana-pantilie: `SideCondition` doesn't have a `Semigroup` instance and I kept it that way because I think adding one now would be pretty confusing, but maybe I'm wrong. I think people would assume the instance combined every field of the `SideCondition`, when it would just combine the `assumedTrue` (maybe also the `definedTerms`?) but just keep _one_ of the `replacements`.

ttuegel: I see, that reason makes sense, for not having the `Semigroup` instance. Could we add a function like this instead of exposing the fields of `SideCondition`?

``` hs
addAssumptions :: MultiAnd (Predicate variable) -> SideCondition variable -> SideCondition variable
```

I think, now more than ever, it might not be a good idea to expose the fields directly.

kore/src/Kore/Reachability/Claim.hs
ttuegel: Oh, I missed that note in GitHub. Thanks!

emarzion (approved):

kore/src/Kore/Step/Simplification/Pattern.hs
```diff
@@ -124,6 +124,7 @@ makeEvaluate sideCondition pattern' =
                 Conditional.splitTerm withSimplifiedCondition
             term' = substitute (toMap $ substitution simplifiedCondition) term
             simplifiedCondition' =
+                -- TODO: is this comment accurate anymore?
```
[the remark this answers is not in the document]
ana-pantilie: I removed the conversion and the comment because I realized they're outdated: since the `assumedTrue` is `MultiAnd Predicate`, we don't have any distinction between substitutions and other predicates.

kore/src/Kore/Step/Simplification/Condition.hs
```diff
@@ -120,49 +123,71 @@ simplify SubstitutionSimplifier{simplifySubstitution} sideCondition =
...
+simplifyPredicatesWithAssumptions sideCondition predicates@(_ : rest) = do
+    let predicatesWithUnsimplified =
+            zip predicates $
+                scanr ((<>) . MultiAnd.singleton) MultiAnd.top rest
+    State.execStateT
+        ( traverse
+            simplifyWithAssumptions
+            predicatesWithUnsimplified
+        )
+        Condition.top
```
ttuegel, line 171: I am still confused why this is `execStateT` instead of `evalStateT`. Can you remind my why the return value is the "state" instead of the `SideCondition`?

ana-pantilie: After giving this some thought, I think that in the end the result will be similar: with `evalStateT` we'd end up with a list of simplified predicates which we can fold into a `Condition` which will be the final result, right? But I don't think we need to do that, since we already accumulate the result in the state (which we also need as a `SideCondition` during simplification). Does this make sense?

ttuegel: Yes, I think that's fine. In that case, we should use `traverse_` here instead of `traverse`.

PR #2647, opened 2021-05-25
Prune unsatisfiable state with unifyStringEq

### LINKED ISSUE #2586: Unsatisfiable state containing `==String` not pruned by backend
I have a proof which generates the following conjunction in its path condition as it executes:

```
\and{SortGeneratedTopCell{}}(
  \not{SortGeneratedTopCell{}}(\equals{SortString{}, SortGeneratedTopCell{}}(VarSender0:SortString{},VarTokenAddress0:SortString{})),
\equals{SortBool{}, SortGeneratedTopCell{}}(\dv{SortBool{}}("true"),
  Lbl'UndsEqlsEqls'String'UndsUnds'STRING-COMMON'Unds'Bool'Unds'String'Unds'String{}(VarSender0:SortString{},VarTokenAddress0:SortString{})))
)
```

I expect that the backend will prune this state, but currently, this functionality is not implemented.

Please note that this check is a blocker for us on our dexter proof.

dopamane:
Fixes #2586 

---

###### Review checklist


ttuegel:
There are similar tests [here](https://github.com/kframework/kore/blob/7a66654d8be69a3500b6f32d08717c0c8168236b/kore/test/Test/Kore/Step/Simplification/AndTerms.hs#L78) and we can add one for `==String`.

kore/test/Test/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -101,6 +102,26 @@ test_andTermsSimplification =
                     )
             actual <- simplifyUnify mkBottom_ fOfA
             assertEqual "" expect actual
+        , testCase "\\and{b}(\\not{b}(\\equals{s, s}(a, b)), \\equals{s, s}(a, b))" $ do
```
ttuegel, line 105: The `AndTerms` simplifier can't handle unification problems of this type, so I would expect this test to fail (and indeed it does). What did you want to happen here?

ttuegel:
@dopamane We are solving the unification problems `\and(true, _==String_(_, _))` and `\and(false, _==String_(_, _))`. In Matching Logic, `\and` has the semantics of intersection, so these terms simplify to `true` and `false` respectively, with the corresponding conditions attached.

dopamane:
@ttuegel Ok, your explanation along with the added tests makes sense to me. I am struggling to see how `maybeTermAnd` and `maybeTermEqual` fail to simplify those cases. In fact, I am not sure that `maybeTermAnd` and `maybeTermEqual` are the source of the most recent test failures. My previous commit was an attempt at cleaning up the code in hope of exposing the issue. While it does not make the new tests pass, I think it at least cleans up the `maybeTermAnd` and `maybeTermEqual` implementations with respect to `unifyStringEq`.

kore/src/Kore/Builtin/String.hs
```diff
@@ -526,14 +547,15 @@ unifyStringEq ::
...
-unifyStringEq unifyChildren notSimplifier a b =
-    worker a b <|> worker b a
+    UnifyStringEq ->
+    unifier (Pattern RewritingVariableName)
+unifyStringEq unifyChildren (NotSimplifier notSimplifier) unifyData =
+    do
+        solution <- unifyChildren operand1 operand2 & OrPattern.gather
+        let solution' = MultiOr.map eraseTerm solution
+        (if value then pure else notSimplifier SideCondition.top) solution'
+            >>= Unify.scatter
```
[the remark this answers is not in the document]
dopamane: Ok, here is where I'm at with your suggestions:
```haskell
    do
        solution <- unifyChildren operand1 operand2 & OrPattern.gather
        let solution' = MultiOr.map eraseTerm solution
        scattered <- (if value then pure else notSimplifier SideCondition.top) solution' >>= Unify.scatter
        return $ addTerm scattered
  where
    UnifyStringEq{eqTerm, value} = unifyData
    EqTerm{operand1, operand2} = eqTerm
    eraseTerm = Pattern.fromCondition_ . Pattern.withoutTerm
    -- Currently working to fill in the hole for asInternal Sort
    addTerm = Pattern.withCondition (Bool.asInternal _ value) . Pattern.withoutTerm
```

kore/src/Kore/Step/Simplification/AndTerms.hs
```diff
@@ -293,6 +290,12 @@ maybeTermAnd notSimplifier childTransformers first second = do
             lift $ Builtin.KEqual.unifyKequalsEq childTransformers notSimplifier unifyData
         | Just unifyData <- Builtin.KEqual.matchUnifyKequalsEq second first =
             lift $ Builtin.KEqual.unifyKequalsEq childTransformers notSimplifier unifyData
+        | Just unifyData <- Builtin.Int.matchUnifyIntEq first second =
+            lift $ Builtin.Int.unifyIntEq childTransformers notSimplifier unifyData
+        | Just unifyData <- Builtin.Int.matchUnifyIntEq second first =
+            lift $ Builtin.Int.unifyIntEq childTransformers notSimplifier unifyData
+        | Just unifyData <- Builtin.String.matchUnifyStringEq first second = lift $ Builtin.String.unifyStringEq childTransformers notSimplifier unifyData
+        | Just unifyData <- Builtin.String.matchUnifyStringEq second first = lift $ Builtin.String.unifyStringEq childTransformers notSimplifier unifyData
```
ttuegel, line 298: `matchUnifyStringEq` and `matchUnifyIntEq` should handle matching their arguments both ways. That is, this should just be:

```suggestion
        | Just unifyData <- Builtin.Int.matchUnifyIntEq first second =
            lift $ Builtin.Int.unifyIntEq childTransformers notSimplifier unifyData
        | Just unifyData <- Builtin.String.matchUnifyStringEq first second = lift $ Builtin.String.unifyStringEq childTransformers notSimplifier unifyData
```

`matchUnifyKequalsEq` should do the same, actually. These functions are the canonical way to match these unification problems and they should be symmetric with respect to their arguments.

kore/src/Kore/Builtin/Int.hs
```diff
@@ -475,7 +475,8 @@ unifyInt term1 term2 unifyData =
 
 data UnifyIntEq = UnifyIntEq
     { eqTerm :: !(EqTerm (TermLike RewritingVariableName))
-    , value :: !Bool
+    , boolValue :: !Bool
+    , boolTerm :: !(TermLike RewritingVariableName)
```
ttuegel, line 479: Why not replace both of these fields with an [`InternalBool`](https://github.com/kframework/kore/blob/221cb82050490c6b711e1e68b42390971e60685a/kore/src/Kore/Internal/InternalBool.hs#L26)? It is _the_ internal representation of Bool.

ttuegel (changes requested):

kore/src/Kore/Builtin/Int.hs
```diff
@@ -510,9 +510,13 @@ unifyIntEq unifyChildren (NotSimplifier notSimplifier) unifyData =
     do
         solution <- unifyChildren operand1 operand2 & OrPattern.gather
         let solution' = MultiOr.map eraseTerm solution
-        (if value then pure else notSimplifier SideCondition.top) solution'
-            >>= Unify.scatter
+        scattered <-
+            Unify.scatter =<< case Bool.matchBool boolTerm of
+                Just False -> notSimplifier SideCondition.top solution'
+                _ -> pure solution'
```
ttuegel, line 516: This is not correct. `boolTerm` could be `False`, or `True`, or it might not even be a `Bool` at all. Please make `boolTerm` be an [`InternalBool`](https://github.com/kframework/kore/blob/221cb82050490c6b711e1e68b42390971e60685a/kore/src/Kore/Internal/InternalBool.hs#L26) type, as I suggested in my [last review](https://github.com/kframework/kore/pull/2647/files/362e825afe15613934de7a4a88dec283e29499be#r656565430). That way, `unifyIntEq` is correctly defined over its inputs.

ttuegel (changes requested):

ttuegel (approved):

ana-pantilie (approved):

PR #2650, opened 2021-05-25
Remove _PREDICATE sort (#2152)

### LINKED ISSUE #2152: Remove the bogus _PREDICATE sort
Follow up on #1442.

We can remove this after we implement a specialized simplifier for `Predicate`.

Boarders:
Remove the predicate sort in favour of using known sorts at all call
sites. This mostly adds sort arguments to various functions that
previously implicitly used _PREDICATE sort. In turn this means removing
the various underbar methods previously in `Kore.Internal.TermLike`.

Fixes #2152

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -266,10 +266,13 @@ exec
                                 <$> finalConfigs
             exitCode <- getExitCode verifiedModule finalConfigs'
             let finalTerm =
-                    sameTermLikeSort initialSort $
-                        OrPattern.toTermLike
-                            _
-                            (MultiOr.map getRewritingPattern finalConfigs')
+                    MultiOr.map getRewritingPattern finalConfigs'
+                        & OrPattern.toTermLike dummySort
+                        & sameTermLikeSort initialSort
+                  where
+                    -- Dummy sort used to unparse configurations.
+                    -- This is only used for unparsing \bottom.
+                    dummySort = SortVariableSort (SortVariable "R")
             return (exitCode, finalTerm)
       where
         dropStrategy = snd
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -266,8 +266,9 @@ exec
                                 <$> finalConfigs
             exitCode <- getExitCode verifiedModule finalConfigs'
             let finalTerm =
-                    forceSort initialSort $
+                    sameTermLikeSort initialSort $
                         OrPattern.toTermLike
+                            _
                             (MultiOr.map getRewritingPattern finalConfigs')
             return (exitCode, finalTerm)
       where
@@ -423,7 +424,7 @@ search
                 orPredicate =
                     makeMultipleOrPredicate (Condition.toPredicate <$> solutions)
             return
-                . forceSort patternSort
+                . sameTermLikeSort patternSort
                 . getRewritingTerm
                 . fromPredicate_
                 $ orPredicate
--- a/kore/src/Kore/Repl/Interpreter.hs
+++ b/kore/src/Kore/Repl/Interpreter.hs
@@ -1394,7 +1394,7 @@ prettyClaimStateComponent transformation omitList =
             }
   where
     prettyComponent =
-        unparseToString . OrPattern.toTermLike
+        unparseToString . OrPattern.toTermLike _
             . MultiOr.map (fmap hide . getRewritingPattern)
             . transformation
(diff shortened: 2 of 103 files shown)


Boarders:
There are two places that are not fixed yet:
https://github.com/kframework/kore/blob/ff9810813c931beeff2067b2d2a4f76532cf44bb/kore/src/Kore/Repl/Interpreter.hs#L1395
https://github.com/kframework/kore/blob/ff9810813c931beeff2067b2d2a4f76532cf44bb/kore/src/Kore/Exec.hs#L270

If someone knows how to fix those issues then I can do so and move on to fixing the tests.

kore/src/Kore/Exec.hs
```diff
@@ -266,8 +266,9 @@ exec
                                 <$> finalConfigs
             exitCode <- getExitCode verifiedModule finalConfigs'
             let finalTerm =
-                    forceSort initialSort $
+                    sameTermLikeSort initialSort $
                         OrPattern.toTermLike
+                            _
```
ttuegel, line 271:
```suggestion
                    OrPattern.toTermLike
                        initialSort
```

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1392,7 +1392,7 @@ prettyClaimStateComponent transformation omitList =
             }
   where
     prettyComponent =
-        unparseToString . OrPattern.toTermLike
+        unparseToString . OrPattern.toTermLike _
```
ttuegel, line 1395: @andreiburdusa Where should we get the sort here?

andreiburdusa: Instead of adding a sort, maybe we can use `unparseAssoc'`?

ttuegel: Oh, this is just pretty-printing for the user, right? The frontend doesn't need to be able to parse the output?

andreiburdusa: I think it's just for the user since `stuckTransformer` contains the string `"Stuck: \n"` and `provenValue` contains `"Proven."`. But I'm not 100% sure.

Boarders: Could you clarify how `unparseAssoc'` should be used here? For example, I am unsure what is the operator in this case?

Boarders:
Currently quite a few things rely on the `pretty` instances in `Data.Internal.Predicate` (see here: https://github.com/kframework/kore/blob/ff9810813c931beeff2067b2d2a4f76532cf44bb/kore/src/Kore/Internal/Predicate.hs#L303). Is there a way to have these instances without known sort (since this should be getting rid of [fromPredicate_](https://github.com/kframework/kore/blob/ff9810813c931beeff2067b2d2a4f76532cf44bb/kore/src/Kore/Internal/Predicate.hs#L496). If anyone knows if this instance and all related instances should not exist or what to put in its place then please let me know.

kore/src/Kore/Repl/Interpreter.hs
ttuegel: I actually just implemented this on another branch because I needed it there, too. Let me push it to its own pull request, and we can just use `show . Pretty.pretty` here instead of `unparseToString . OrPattern.toTermLike _`.

ana-pantilie: > I think it's just for the user since `stuckTransformer` contains the string `"Stuck: \n"` and `provenValue` contains `"Proven."`. But I'm not 100% sure.

Those get sent to different "output streams" in kore-repl (we separate kore output and auxiliary output, so that we can send kore output to the frontend). So we actually need the kore here to be correct kore, this function prints the configuration which is what people usually send to the frontend to get unparsed back to K.

andreiburdusa: Then my first idea is to associate sorts to claims. I see we now discard the sorts in `extractClaim` in places like https://github.com/kframework/kore/blob/a6bf9b795b2bcc106329e0fa8b235e9d2559836b/kore/src/Kore/Reachability/OnePathClaim.hs#L188

ttuegel: We can just use `SortVariableSort (SortVariable "R")` because the sort will only be used if the `OrPattern` is `\bottom`, and the frontend ignores the sort of `\bottom`.

Issue #2566, opened 2021-04-21
Seemingly infinite loop in equation simplification


ehildenb:
This bug report is generated by running a simplified Maker MCD Spec (https://github.com/makerdao/mkr-mcd-spec) through the kore-repl, and killing it after it hangs for a while.

[kore-exec.tar.gz](https://github.com/kframework/kore/files/6353753/kore-exec.tar.gz)

If you run the `kore-repl`, take 6 steps, it works just fine. But then you do `stepf` one more time, and it seems to hang. I turned on `DebugAttemptEquation`, and dumped the results to a file. From what I can tell, it's just repeatedly trying the same three equations on the same input, forever.

Here are the rules which seem to loop forever:

```
    syntax Bool ::= wish(Address, Address, Map) [function, functional] 
 // ------------------------------------------------------------------ 
    rule wish(ADDRFROM, MSGSENDER,        _) => true                                   requires ADDRFROM ==K MSGSENDER 
    rule wish(ADDRFROM, MSGSENDER, CANADDRS) => MSGSENDER in {CANADDRS[ADDRFROM]}:>Set requires ADDRFROM in_keys(CANADDRS) 
    rule wish(       _,         _,        _) => false                                  [owise] 
```

The log contains many repeats of lines like this:

```
kore-repl: [139189290] Debug (DebugAttemptEquation):
    (InfoReachability) while checking the implication:
    applying equation at /home/dev/src/mkr-mcd-spec-pr/vat.md:237:10-237:123 to term:
        /* Fl Fn D Spa */
        Lblwish'LParUndsCommUndsCommUndsRParUnds'VAT'Unds'Bool'Unds'Address'Unds'Address'Unds'Map{}(
            /* Fl Fn D Sfa */ ConfigVarADDRFROM:SortAddress{},
            /* Fl Fn D Sfa Cl */ LblADMIN'Unds'KMCD-DRIVER'Unds'Address{}(),
            /* Fl Fn D Sfa */ ConfigVarVAT'Unds'CANS:SortMap{}
        )
```

This represents the middle rule of the three listed, but all three show up in the log over and over again. Note that I only turned on logging _after_ taking the first 6 steps.

ttuegel:
That being the only log entry is extremely suspicious. `DebugAttemptEquation` should log the individual steps of attempting the equation: matching, instantiation, checking requirements. If we really don't see any of that, then we must never get far enough to even _match_ the equation.

ana-pantilie:
[...]

PR #2653, opened 2021-05-27
Use new Ceil simplifier

### LINKED ISSUE #2686: Incorrect unification result for LIST.concat
The backend incorrectly gives `\bottom` for this unification problem:

```
\and(concatList(A, B), unitList())
```

Instead, the result should be

```
\and(unitList(), \and(\in(A, unitList()), \in(B, unitList()))
```

### LINKED ISSUE #2689: Sort injections marked "conditionally" simplified
Simplified sort injections are incorrectly marked "conditionally" simplified, so that they are re-simplified every time the side condition changes. The side condition is not considered when simplifying sort injections, so these simplifications are redundant.

ttuegel:
Fixes #2689.
Fixes #2686.

---

###### Review checklist


ttuegel:
I am very suspicious of these performance numbers because:

1. Everything was fine until I rebased onto master, and
2. I ran all the integration tests locally and they finished in a reasonable amount of time.

If the integration tests time-out on Jenkins again, I will investigate further.

kore/src/Kore/Step/Simplification/Condition.hs
```diff
@@ -71,22 +71,21 @@ unmodified.
...
     SubstitutionSimplifier simplifier ->
     SideCondition RewritingVariableName ->
     Conditional RewritingVariableName any ->
     LogicT simplifier (Conditional RewritingVariableName any)
 simplify SubstitutionSimplifier{simplifySubstitution} sideCondition =
-    normalize >=> worker
+    normalize >=> worker 0
   where
-    worker Conditional{term, predicate, substitution} = do
+    worker count Conditional{term, predicate, substitution} = do
```
ana-pantilie, line 83: Is `count` used anywhere?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -477,3 +496,149 @@ ensureSimplifiedCondition repr termLike condition
...
+            pure (OrPattern.fromTermLike termLike)
+        | otherwise =
+            case termLikeF of
+                -- Not implemented:
+                ApplyAliasF _ -> doNotSimplify
+                -- Not simplifiable:
+                EndiannessF _ -> doNotSimplify
+                SignednessF _ -> doNotSimplify
+                -- Handled elsewhere, not a proper term:
+                RewritesF _ -> doNotSimplify
```
ana-pantilie, line 546: Shouldn't this also throw an error, like the old term simplifier?

kore/src/Kore/Step/Simplification/TermLike.hs
```diff
@@ -477,3 +496,149 @@ ensureSimplifiedCondition repr termLike condition
             , Pretty.indent 2 "while simplifying:"
             , Pretty.indent 4 (unparse termLike)
             ]
+
+-- | Simplify the given 'TermLike'. Do not simplify any side conditions.
+simplifyOnly ::
```
ana-pantilie, line 501: So, is the plan to replace the old term simplifier with this function?

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -121,10 +144,19 @@ simplify sideCondition =
                 NotF notF -> simplifyNot =<< traverse worker notF
                 ImpliesF impliesF -> simplifyImplies =<< traverse worker impliesF
                 IffF iffF -> simplifyIff =<< traverse worker iffF
+                CeilF ceilF ->
```
MirceaS, line 147: Should we have a symmetric case for Floor here as well or is this fine because we're planning to get rid of Floor anyway?

kore/src/Kore/Step/Simplification/Predicate.hs
```diff
@@ -90,28 +98,43 @@ type NormalForm = MultiOr (MultiAnd (Predicate RewritingVariableName))
...
-    loop . MultiOr.singleton . MultiAnd.singleton
+simplify sideCondition original =
+    loop 0 (mkSingleton original)
   where
-    loop :: NormalForm -> simplifier NormalForm
-    loop input = do
-        output <- MultiAnd.traverseOrAnd worker input
-        (if input == output then pure else loop) output
+    limit :: Int
+    limit = 4
```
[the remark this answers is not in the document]
ttuegel: This addresses the loop in #2566.

kore/src/Kore/Step/Simplification/Condition.hs
ttuegel: This is left-over from debugging. I suppose we should add a warning here, the same as in the `Predicate` simplifier.

kore/src/Kore/Step/Simplification/TermLike.hs
ttuegel: Yes, we should eventually replace the old `TermLike` simplifier with this one. There are still some problems with the particular delegates of this function. For example, they are mutually recursive, instead of letting the loop in `simplifyOnly` re-simplify patterns. Also, they don't always delegate to the `Predicate` simplifier like they should.

kore/src/Kore/Step/Simplification/Condition.hs
```diff
@@ -44,7 +44,7 @@ import qualified Kore.Internal.Substitution as Substitution
 import Kore.Rewriting.RewritingVariable (
     RewritingVariableName,
  )
-import qualified Kore.Step.Simplification.Predicate as Predicate
+import {-# SOURCE #-} qualified Kore.Step.Simplification.Predicate as Predicate
```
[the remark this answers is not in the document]
ttuegel: This resolves a cyclic dependency in the import graph. The cycle will be resolved when the `TermLike` simplifier is finally fixed so that it does not call the `Condition` simplifier.

kore/src/Kore/Step/Simplification/Predicate.hs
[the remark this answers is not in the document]
ttuegel: No, this should not be passed in. It is an internal detail of the algorithm. I ran this over many proofs and never found a value greater than 1, so I am confident that by setting this limit to 4, we will only see the associated warning when there is a true problem.

kore/src/Kore/Step/Simplification/Predicate.hs
ttuegel: We should de-sugar `Floor` to `Not Ceil Not`, but we don't use `Floor` in practice, so I thought it wasn't important for getting this merged.

ana-pantilie (approved):

MirceaS (approved):

PR #2687, opened 2021-06-21
Remove toPattern from Kore.Step.Simplification.Iff

### LINKED ISSUE #2610: Remove Kore.Internal.OrPattern.toPattern
The function `toPattern` should not exist. Start by removing it from the tests, then from `kore-repl`, and finally remove it from the main library. This is necessary for #2152.

dopamane:
Fixes #2610 

---

###### Review checklist


kore/test/Test/Kore/Step/Simplification/Integration.hs
```diff
@@ -558,7 +558,7 @@ test_simplificationIntegration =
                 )
                     [ [fromNot cfCeil, fromNot chCeil]
                     , [chCeil, cgCeil, cfCeil]
-                    , [chCeil, cfCeil]
+                    , [chCeil, cfCeil, fromNot cgCeil]
```
ttuegel, line 561: I made a truth table to check this change, and found that the old output is equivalent to the new output. After that, I found a clear way to show that. Consider the second and third clauses, we have:

```
old: ( \ceil(f) ∧ \ceil(g) ∧ \ceil(h) ) ∨ ( \ceil(f) ∧ \ceil(h) )
```

If we do some factoring, we find:

```
old: \ceil(f) ∧ \ceil(h) ∧ ( \ceil(g) ∨ \top )
```

Look at the last conjunct: `\ceil(g) ∨ \top`. This is just `\top`. In other words, this term completely subsumes the last two clauses:

```
old: \ceil(f) ∧ \ceil(h)
```

In the new output, we are going to do the same factoring trick:

```
new: ( \ceil(f) ∧ \ceil(g) ∧ \ceil(h) ) ∨ ( \ceil(f) ∧ \not \ceil(g) ∧ \ceil(h) )
new: \ceil(f) ∧ \ceil(h) ∧ ( \ceil(g) ∨ \not \ceil(g) )
new: \ceil(f) ∧ \ceil(h)
```

So, the output in either case is equivalent. The output changes because it isn't minimal, in the sense that we don't do this kind of simplification.

ttuegel (approved):

kore/test/Test/Kore/Step/Simplification/Integration.hs
dopamane: Thank you, this is really helpful!

ana-pantilie (approved):

PR #2688, opened 2021-06-22
modified matchEquation to accept new equation format

### LINKED ISSUE #2593: Correct encoding of ensures clause
See kframework/k#1945.

The backend uses the correct interpretation of `ensures`, but the frontend encodes the rules the wrong way. We should:

1. Document the correct encoding.
2. Modify the backend to accept either encoding.
3. Modify the frontend to emit the correct encoding.
4. Forbid the incorrect encoding in the backend.

MirceaS:
Contributes to #2593 

---

###### Review checklist


ana-pantilie (commented):
I think the changes are good, but we should add some comments to remind ourselves to remove the old parsing when the frontend switches to this new encoding. We should also add some unit tests.

I think it would be fine to also modify the design documents about equations in this PR.

kore/src/Kore/Equation/Equation.hs
```diff
@@ -168,9 +168,8 @@ toTermLike sort equation
                         argumentTerm
                     )
                 )
-                ( TermLike.mkAnd
-                    (TermLike.mkEquals sort left right)
-                    ensures'
+                ( TermLike.mkEquals sort left $
```
ana-pantilie, line 171: This is used for unparsing, right? If so, I think the changes to this function are premature: until the frontend changes its equation encoding we shouldn't make changes here. Not sure if this is the best solution, but we could add a `toTermLikeTODO` function with the new encoding which will replace this once the frontend changes.

kore/test/Test/Kore/Equation/Sentence.hs
```diff
@@ -32,12 +32,12 @@ test_fromSentenceAxiom =
...
     , testCase "⌈f(x))⌉ → f(x) = g(x) ∧ ⌈h(x)⌉" $ do
         let requires = mkCeil sortR (Mock.f Mock.a)
-            ensures = mkCeil sortR (Mock.h Mock.b)
+            ensures = mkCeil Mock.testSort (Mock.h Mock.b)
             left = Mock.f (mkElemVar Mock.x)
             right = Mock.g (mkElemVar Mock.x)
             original =
                 mkImplies requires $
-                    mkAnd (mkEquals sortR left right) ensures
+                    mkEquals sortR left (mkAnd right ensures)
```
ana-pantilie, line 40: Related to the comment above: we shouldn't change the old tests, just add new tests for the new encoding.

ana-pantilie (changes requested):

dopamane (approved):

kore/test/Test/Kore/Equation/Sentence.hs
MirceaS: I tried initially to just add new tests but eventually realized that they cannot coexist because of the `toTermLike` function.
We cannot support unparsing of both formats at the same time as we don't know for a particular equation whether it came from the old format or the new one.
I suppose we could add a flag inside the Equation data structure telling us which format was used for it's parsing for I'm not sure if this is the way to go...
What do you think?

ana-pantilie: We don't need to support both at the same time, we should only support one. Right now, the frontend emits the old encoding, that's why I suggested [this](https://github.com/kframework/kore/pull/2688#discussion_r658954577). When the frontend emits the new encoding, we should just switch to using the new `toTermLike` and remove the old tests.

ana-pantilie (approved):

ana-pantilie:
@MirceaS I didn't notice this during the meeting, looks like you have some merge conflicts which need addressing before this can be merged.

Issue #2071, opened 2020-08-15
Halt at first stuck configuration


ttuegel:
When doing a breadth-first proof search, the backend will actually run all branches of a proof until they are complete or stuck, before returning the first stuck configuration. Actually, it should just stop after the first stuck branch. This happens because the execution is halted in [`throwUnproven`](https://github.com/kframework/kore/blob/master/kore/src/Kore/Strategies/Verification.hs#L310), but this is too late; it actually should halt in `transit`.

ttuegel:
In fact, if there is an unproven node in `throwUnproven`, it should be a hard error, as this would indicate an example of #2021.

PR #2699, opened 2021-06-25
Allow specifying max number of counter-examples

### LINKED ISSUE #2648: Allow specifying max number of counter-examples
This is low priority, because it only affects the generation of automatic optimizations, which is not a _needed_ feature. However, there is no workaround that I'm aware of.

Currently, I'm running a proof like this:

```
kevm prove --backend haskell exp-spec.k OPTIMIZE --concrete-rules-file tests/specs/mcd/concrete-rules.txt --depth 10
```

On `exp-spec.k` which looks like this:

```
requires "lemmas/mcd/verification.k"

module OPTIMIZE
    imports INT
    imports LEMMAS-MCD
    imports EVM
endmodule

module EXP-SPEC
    imports OPTIMIZE

    claim <k> (#next [ EXP ] => #halt) ~> #execute ... </k>
          <schedule> SCHED </schedule>
          <program>  PGM   </program>
          <pc>         PCOUNT       => PCOUNT +Int #widthOp(EXP) </pc>
          <wordStack>  W0 : W1 : WS => ?_                             </wordStack>
          <localMem>   LM           => ?_                             </localMem>
          <gas>        #gas(GAVAIL) => ?_                             </gas>
          <memoryUsed> MU           => ?_                             </memoryUsed>
      requires #stackNeeded(EXP) <=Int #sizeWordStack(W0 : W1 : WS)
       andBool #sizeWordStack(W0 : W1 : WS) +Int #stackDelta(EXP) <=Int 1024


endmodule
```

I happen to know (out of band) that there are _two_ failure cases, one for `W0 ==Int 0` and one for `W0 =/=Int 0`.

Currently, the backend only gives me the one counterexample, for `W0 ==Int 0`.

I would like to be able to specify something like `--max-counterexamples 10`, to tell the backend to keep searching for counterexamples and to disjunct them together.

andreiburdusa:
Fixes #2648 

---

###### Review checklist


kore/src/Options/SMT.hs
```diff
@@ -109,20 +110,22 @@ parseKoreSolverOptions =
...
+    (Read t, Integral t) =>
+    (t -> b) ->
+    String ->
+    Options.ReadM b
+readPositiveIntegral ctor optionName = do
+    readInt <- auto
+    when (readInt <= 0) err
+    return . ctor $ readInt
+  where
+    err = readerError . unwords $ [optionName, "must be a positive integer."]
```
ana-pantilie, line 128: We should move this to another module since this module should be specific to SMT.

kore/app/exec/Main.hs
```diff
@@ -698,13 +700,15 @@ koreProve execOptions proveOptions = do
     specModule <- loadModule specMainModule definition
     let KoreProveOptions{saveProofs} = proveOptions
     maybeAlreadyProvenModule <- loadProven definitionFileName saveProofs
+    let KoreProveOptions{maxCounterexamples} = proveOptions
```
ana-pantilie, line 703: I don't remember, `kore-repl` uses a different type for its options, right? I'm just wondering if we'd run into the case where someone tries to run `kore-repl` with `--max-counterexamples`, and that doesn't really make sense.

kore/test/Test/Kore/Reachability/Prove.hs
```diff
@@ -210,6 +210,48 @@ test_proveClaims =
...
+                            []
+                    let stuck =
+                            [ mkSimpleClaim Mock.b Mock.d
+                            , mkSimpleClaim Mock.c Mock.d
+                            ]
+                        expect = MultiAnd.make (fmap StuckClaim stuck)
+                    assertEqual "" expect actual
+         in [ mkTest "OnePath" simpleOnePathClaim
+            , mkTest "AllPath" simpleAllPathClaim
+            ]
```
ana-pantilie, line 254: It might be redundant, but I think it's best to have an explicit test for the default case as well, it shows our intent and serves as documentation.

kore/src/Kore/Reachability/Prove.hs
```diff
@@ -324,19 +348,37 @@ proveClaim
...
                     infoUnprovenDepth proofDepth
-                    throwStuckClaim unproven
+                    stuckClaims <- State.get
+                    let updatedStuck =
+                            MultiAnd.singleton (StuckClaim unproven) <> stuckClaims
+                    when
+                        (MultiAnd.size updatedStuck >= maxCounterexamples)
+                        $ do
+                            Monad.Except.throwError updatedStuck
+                    State.put updatedStuck
```
ana-pantilie, line 360: I think this part can be factored out and used in both this function and `throwStuckClaims`.

kore/src/Kore/Reachability/Prove.hs
```diff
@@ -497,12 +537,19 @@ throwStuckClaims ::
         (VerifierT m)
         rule
         (ProofDepth, ClaimState SomeClaim)
-throwStuckClaims rule prim state = do
+throwStuckClaims maxCounterexamples rule prim state = do
```
ana-pantilie, line 540: Sorry, I can't remember, did you end up finding out the reason why we need both `throwStuckClaims` and `throwUnproven`? Also, does this need `checkLeftUnproven` as well?

andreiburdusa: From what I understand, first we had `throwUnproven`, but then `throwStuckClaims` was added in order to stop the execution earlier in the case of a stuck claim. Otherwise, during a breadth-first proof search, the backend would have run all the branches until proven or stuck (see https://github.com/kframework/kore/issues/2071).
I don't think `throwStuckClaims` needs `checkLeftUnproven` because `throwUnproven` is the last check.

kore/app/exec/Main.hs
andreiburdusa: I moved it to `KoreExecOptions`. Do you think it's better this way?

MirceaS (approved):

ana-pantilie (approved):

PR #2719, opened 2021-06-30
Remove _PREDICATE sort

### LINKED ISSUE #2152: Remove the bogus _PREDICATE sort

ttuegel:
Fixes #2152.

Opening a new pull request to avoid spamming the original author with notifications.

---

###### Review checklist


kore/src/Kore/Internal/Predicate.hs
```diff
@@ -504,7 +501,7 @@ fromPredicate_ ::
     InternalVariable variable =>
     Predicate variable ->
     TermLike variable
-fromPredicate_ = fromPredicate predicateSort
+fromPredicate_ = fromPredicate undefined
```
ana-pantilie, line 504: We should think about how we want to pretty-print `Predicate`s now.

andreiburdusa:
@ana-pantilie These are the places where we still use `fromPredicate_` and don't have a sort in the context to use (except for the cases where we do pretty printing):
https://github.com/kframework/kore/blob/ce423fb1b8760873f5bc6844864b9f94eba7ed04/kore/src/Kore/Internal/Predicate.hs#L1120
https://github.com/kframework/kore/blob/252ff7bfeac9895d8511a4af6b7d33f4902ad749/kore/src/Kore/Step/Simplification/Predicate.hs#L77
https://github.com/kframework/kore/blob/e0614507f2172e8d161fa8e8eddd844dba724eae/kore/src/Kore/Step/SMT/Translate.hs#L137

ana-pantilie:
@andreiburdusa About the `fromPredicate_` usages:
- The second is fine, we will have to remove that function in a new PR anyway.
- I'm a little confused about the third, judging by the function name and how this code seems to work I think we should be matching on a `Predicate` there instead of a `TermLike`, but we also check the definedness there as well which I think only makes sense in the context of a `TermLike`. Maybe we should change it that the argument is a `Predicate` on which we match, and move the definedness check in the `TermLike` translator? Do you think that makes sense or is there anything I'm missing here?

kore/src/Kore/Simplify/Not.hs
```diff
@@ -78,42 +78,16 @@ Right now this uses the following:
...
-to carry around.
-
--}
-simplifyEvaluated ::
-    MonadSimplify simplifier =>
-    SideCondition RewritingVariableName ->
-    OrPattern RewritingVariableName ->
-    simplifier (OrPattern RewritingVariableName)
-simplifyEvaluated sideCondition simplified =
+simplify sideCondition not'@Not{notSort} =
```
ana-pantilie, line 86: I think the caller should decide what to do with the sort here, so basically in the `Predicate` simplifier we should delete the sort there instead of rely on this function being parametric in the sort. I hope this makes sense, please let me know if it doesn't.

andreiburdusa: I don't understand. If this function uses a concrete `Sort` type, wouldn't we need to provide the sort in the predicate simplifier?

ana-pantilie: I mean, we use this because we're simplifying a term (like for the `Floor` case, its child is always a term) so we should be using that term's sort here.

andreiburdusa:
Needs #2610

kore/test/Test/Kore/Builtin/KEqual.hs
```diff
@@ -83,11 +83,12 @@ test_KEqual =
...
         assertEqual' "" expect actual
     , testCaseWithoutSMT "kseq(x, dotk) equals kseq(x, dotk)" $ do
-        let expect = OrPattern.top
+        let expect = OrPattern.topOf kSort
             xConfigElemVarKItemSort =
                 configElementVariableFromId "x" kItemSort
             original =
-                mkEquals_
+                mkEquals
+                    kSort
```
ana-pantilie, line 91: I was thinking about this, and I want to be 100% sure I understand this. I'd expect the `==_` (where `_` is `K`, `Int` etc.) to return a `Bool`. The reason this doesn't return `Bool` here but `kSort` is because `\\equals` is a predicate and we can give it any result sort. If we did `\\and` unification instead, the resulting sort would be `Bool`, right?

kore/test/Test/Kore/Simplify/Or.hs
```diff
@@ -223,7 +223,7 @@ becomes
             ( stateIntention
                 [ prettyOr or1 or2
                 , "to become:"
-                , Unparser.unparse $ OrPattern.toTermLike expected
+                , Unparser.unparse $ OrPattern.toTermLike (mkSortVariable "_") expected
```
ana-pantilie, line 226: If we have a sort here, we should use that. I guess we should use a sort variable only if the `OrPattern` is `\\bottom`.

kore/src/Kore/Simplify/Equals.hs
```diff
@@ -161,11 +161,16 @@ simplify ::
...
+    Equals
+        { equalsFirst = first
+        , equalsSecond = second
+        , equalsOperandSort = sort
+        } = simplifyEvaluated sort sideCondition first' second'
+      where
+        (first', second') =
+            minMaxBy (on compareForEquals (OrPattern.toTermLike sort)) first second
 
 {- TODO (virgil): Preserve pattern sorts under simplification.
```
RaoulSchaffranek, line 175: I think this change resolves the TODO from the comment, so we should remove it.

kore/src/Kore/Simplify/Equals.hs
```diff
@@ -289,27 +295,27 @@ makeEvaluate
     second@Conditional{term = secondTerm}
     sideCondition =
         do
-            let first' = first{term = if termsAreEqual then mkTop_ else firstTerm}
-            firstCeil <- makeEvaluateCeil sideCondition first'
-            let second' = second{term = if termsAreEqual then mkTop_ else secondTerm}
-            secondCeil <- makeEvaluateCeil sideCondition second'
-            firstCeilNegation <- Not.simplifyEvaluated sideCondition firstCeil
-            secondCeilNegation <- Not.simplifyEvaluated sideCondition secondCeil
+            let termSort = termLikeSort firstTerm
```
RaoulSchaffranek, line 298: Is there a difference between `termSort` and `sort` from the where-clause? Otherwise, I'd suggest keeping only one definition.

RaoulSchaffranek (approved):

kore/src/Kore/Internal/Predicate.hs
```diff
@@ -966,7 +956,9 @@ instance
     pretty (NotPredicate termLikeF) =
         Pretty.vsep
             [ "Expected a predicate, but found:"
-            , Pretty.indent 4 (unparse $ fromPredicate_ <$> termLikeF)
+            , Pretty.indent
+                4
+                (unparse $ fromPredicate (mkSortVariable "_") <$> termLikeF)
```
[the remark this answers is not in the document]
andreiburdusa: `termLikeF` is a `TermLikeF variable (Predicate variable)`. `TermLikeF`  doesn't have a `Pretty` instance, and the use of its `Unparse` instance would require `Predicate` to have an `Unparse` instance as well.

kore/src/Kore/Simplify/TermLike.hs
```diff
@@ -574,7 +576,7 @@ simplifyOnly sideCondition =
                 -- Matching Logic:
                 AndF andF -> do
                     let conjuncts = foldMap MultiAnd.fromTermLike andF
-                    And.simplify Not.notSimplifier sideCondition
+                    And.simplify sort Not.notSimplifier sideCondition
```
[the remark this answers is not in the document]
andreiburdusa: `And.makeEvaluate` might receive an empty `MultiAnd`, so this pr gave it a `Sort` parameter in order to be able to return a `\top`. `And.simplify` calls `And.makeEvaluate`, but its `MultiAnd (OrPattern RewritingVariableName)` might not contain any `Pattern` either. That's why now it has a `Sort` parameter.

kore/test/Test/Kore/Builtin/KEqual.hs
andreiburdusa: Yes, if I replace `mkEquals` with `mkAnd`, the result sort is `Bool`

ana-pantilie: Cool. Could you add some similar tests for `and` as well?

kore/src/Kore/Internal/Predicate.hs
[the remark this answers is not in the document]
ana-pantilie: Ah, I see. Thanks for explaning!

kore/src/Kore/Simplify/TermLike.hs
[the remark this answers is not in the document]
ana-pantilie: I understand. Could you add a small comment in each place where the sort argument is needed? For example, here we can say "MultiAnd doesn't preserve the sort so we need to send it as an external argument".

kore/src/Kore/Builtin/AssociativeCommutative.hs
```diff
@@ -1400,7 +1401,10 @@ unifyEqualsElementLists
 
             let remainder2Terms = map fromConcreteOrWithVariable remainder2
 
-            case elementListAsInternal tools (termLikeSort first) remainder2Terms of
+            case elementListAsInternal
+                tools
+                (sameSort (termLikeSort first) (termLikeSort first))
```
ana-pantilie, line 1406: Did you mean

```suggestion
                (sameSort (termLikeSort first) (termLikeSort second))
```

kore/src/Kore/Builtin/EqTerm.hs
```diff
@@ -46,25 +29,3 @@ matchEqTerm selectSymbol (App_ symbol [operand1, operand2]) = do
     Monad.guard (selectSymbol symbol)
     return EqTerm{symbol, operand1, operand2}
 matchEqTerm _ _ = Nothing
-
-{- | Unification for an equality-like symbol.
-
-This function is suitable only for equality simplification.
--}
-unifyEqTerm ::
```
ana-pantilie, line 54: Where did this go?

kore/src/Kore/Simplify/AndTerms.hs
```diff
@@ -161,7 +164,11 @@ maybeTermEquals notSimplifier childTransformers first second = do
         | Just unifyData <- matchBytes first second =
             lift $ unifyBytes unifyData
         | Just unifyData <- matchBottomTermEquals first second =
-            lift $ bottomTermEquals SideCondition.topTODO unifyData
+            lift $
+                bottomTermEquals
+                    (sameSort (termLikeSort first) (termLikeSort first))
```
ana-pantilie, line 169: This is checking `first` and `first` again.

kore/src/Kore/Simplify/AndTerms.hs
```diff
@@ -193,7 +200,12 @@ maybeTermEquals notSimplifier childTransformers first second = do
         | Just unifyData <- Builtin.Map.matchUnifyEquals tools first second =
             lift $ Builtin.Map.unifyEquals childTransformers tools unifyData
         | Just unifyData <- Builtin.Map.matchUnifyNotInKeys first second =
-            lift $ Builtin.Map.unifyNotInKeys childTransformers notSimplifier unifyData
+            lift $
+                Builtin.Map.unifyNotInKeys
+                    (sameSort (termLikeSort first) (termLikeSort first))
```
ana-pantilie, line 205: This is checking `first` and `first` again.

kore/src/Kore/Simplify/TermLike.hs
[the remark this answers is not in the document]
ana-pantilie: Could you add ^?

ana-pantilie (changes requested):

ana-pantilie (approved):

Issue #2700, opened 2021-06-26
Set simplification not working


ehildenb:
On K master commit (72d33ae0e0), the following very simple spec does not work. This affects some KEVM proofs which are being ported.

```
requires "domains.md"

module TEST
    imports INT
    imports SET

    configuration <k> $PGM:Step </k>

    syntax Step ::= run ( Set ) | done ( Set )

    rule <k> run(S) => done(S) ... </k>

    syntax Bool ::= pred ( Int ) [function, no-evaluators]

    rule (SetItem(X) REST) -Set SetItem(Y) => SetItem(X) (REST -Set SetItem(Y)) requires pred(X) andBool notBool pred(Y) [simplification]

    rule pred(X) => false requires 1 <=Int X andBool X <=Int 9 [simplification]
endmodule

module TEST-SPEC
    imports TEST

    claim <k> run(SetItem(1) -Set SetItem(X)) => done(SetItem(1)) ... </k> requires pred(X)

endmodule
```

To reproduce the failure:

```
kompile --backend haskell test.k --main-module TEST --syntax-module TEST
kprove test.k --def-module TEST --spec-module TEST-SPEC
```

geo2a:
I modified introduced names for all rules in the example:
```
requires "domains.md"

module TEST
    imports INT
    imports SET

    configuration <k> $PGM:Step </k>

    syntax Step ::= run ( Set ) | done ( Set )

    rule [step]: <k> run(S) => done(S) ... </k>

    syntax Bool ::= pred ( Int ) [function, no-evaluators]

    rule [set.diff.empty]: .Set -Set _ => .Set [simplification]
    rule [set.diff.concat]: (SetItem(X) REST) -Set SetItem(Y) => SetItem(X) (REST -Set SetItem(Y)) requires pred(X) andBool notBool pred(Y) [simplification]

    rule [pred.false]: pred(X) => false requires 1 <=Int X andBool X <=Int 9 [simplification]
endmodule

module TEST-SPEC
    imports TEST

    claim <k> run(SetItem(1) -Set SetItem(X)) => done(SetItem(1)) ... </k> requires pred(X)

endmodule
```
To see which equations are applied by `kore-exec`, I set the `KORE_EXEC_OPTS` environmental variable:
```
export KORE_EXEC_OPTS="--log ${LOG_FILE} --debug-equation TEST.step --debug-equation TEST.set.diff.empty --debug-equation TEST.set.diff.concat --debug-equation TEST.pred.false
```
[...]

PR #2727, opened 2021-07-02
Investigate Set simplification #2700

geo2a:
* Trace execution of `Kore.Builtin.Set.evalDifference`

Debugging with the example from the description of issue #2700 produces 7 outputs that look like attempted applications of the builtin set difference:
```
\and{SortSet{}}(
        /* term: */
        /* Fl Fn D Spa */
        LblSet'Coln'difference{}(
            /* Fl Fn D Sfa Cl */
            /* InternalSet: */ /* concrete element: */ LblSetItem{}(
                /* Inj: */ inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("1"))
            ),
            /* Fl Fn D Spa */
            /* InternalSet: */ /* element: */ LblSetItem{}(
                /* Fl Fn D Sfa */
                /* Inj: */ inj{SortInt{}, SortKItem{}}(
                    /* Fl Fn D Sfa */ RuleVarX:SortInt{}
                )
            )
        ),
    \and{SortSet{}}(
        /* predicate: */
        /* D Sfa */ \top{SortSet{}}(),
        /* substitution: */
        \top{SortSet{}}()
    ))
```
I'm posting this here to facilitate our discussion with @ana-pantilie.

### DIFF
--- a/georgy/debug.sh
+++ b/georgy/debug.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+K_FILE=test.k
+
+LOG_FILE=debug.log
+LOG_ENTRIES=DebugAttemptEquation,DebugApplyEquation
+# LOG_ENTRIES=DebugAppliedRewriteRules,DebugAttemptEquation,DebugApplyEquation
+export KORE_EXEC_OPTS="--log ${LOG_FILE} --log-entries ${LOG_ENTRIES}"
+# export KORE_EXEC_OPTS="--log ${LOG_FILE} --debug-equation \"LblSet'Coln'difference{}\""
+
+rm -r test-kompiled
+rm $LOG_FILE
+
+kompile --backend haskell $K_FILE --main-module TEST --syntax-module TEST
+kprove test.k --def-module TEST --spec-module TEST-SPEC
--- a/georgy/equation-lhs.1
+++ b/georgy/equation-lhs.1
@@ -1,0 +1,14 @@
+        /* Fn Spa */
+        LblSet'Coln'difference{}(
+            /* Fn Spa */
+            /* InternalSet: */ Lbl'Unds'Set'Unds'{}(
+                /* element: */ LblSetItem{}(
+                    /* Fl Fn D Sfa */ EquationVarX:SortKItem{}
+                ),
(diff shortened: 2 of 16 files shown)


geo2a:
I've had tried removing the symbolic case and it have produced the result we expected: the user-defined rules are now attempted!

However, we still have problems:
* First, it looks like the initial example faulty in the first place (@ehildenb, please correct me if I'm wrong). I have changed the `requires` clause of the `set.diff.concat` to `notBool pred(X) andBool pred(Y) ` because otherwise the term in the claim was prohibited by it. I have seen attempts to apply this rule in the `kore-exec` log though, which is good --- before it was not even attempted.
* However, the proof only goes thorough and outputs `#Top` only if I remove the `[simplification]` attribute of the `set.diff.concat` rule. With the attribute, The rule is attempted, but never applied, and the logs contain the message `equation didn't match term` in relevant entries.

@ana-pantilie what do we do from here?

The current example spec for reference:
```
requires "domains.md"

module TEST
    imports INT
    imports SET

    configuration <k> $PGM:Step </k>

    syntax Step ::= run ( Set ) | done ( Set )

    rule [step]: <k> run(S) => done(S) ... </k>

    syntax Bool ::= pred ( Int ) [function, no-evaluators]

    rule [set.diff.empty]: .Set -Set _ => .Set [simplification]
    rule [set.diff.concat]: (SetItem(X) REST) -Set SetItem(Y) => SetItem(X) (REST -Set SetItem(Y))
      requires notBool pred(X) andBool pred(Y) 
      [simplification] // commenting this out makes the proof go through

    rule [pred.false]: pred(X) => false requires 1 <=Int X andBool X <=Int 9 [simplification]
endmodule

module TEST-SPEC
    imports TEST

    claim <k> run(SetItem(1) -Set SetItem(X)) => done(SetItem(1)) ... </k> requires pred(X)

endmodule
```

geo2a:
The unit tests for the symbolic case of `evalDifference` fail, but it's to be expected, since there is no built-in decision procedure for symbolic set difference anymore.

ana-pantilie:
@geo2a could you add the output of the failing proof here?

PR #2738, opened 2021-07-07
Add new AC matching case

ana-pantilie:
Part of #2700

- documents current AC matching implementation
- adds case for structurally equal ACs

---

###### Review checklist


kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -821,14 +837,26 @@ matchNormalizedAc pushElement pushValue wrapTermLike normalized1 normalized2
...
                         push (Pair frame1 normalized2')
                     _ -> empty
+    -- Case for ACs which are structurally equal:
+    | length excessAbstract1 == length excessAbstract2
+      , length concrete1 == length concrete2
+      , length opaque1 == length opaque2 = lift $ do
+        traverse_ pushValue abstractMerge
+        traverse_
+            (pushElement . uncurry Pair)
+            (zip abstract1 abstract2)
```
ana-pantilie, line 847: I think this should be:
```suggestion
            (zip excessAbstract1 excessAbstract2)
```

geo2a (approved):

kore/src/Kore/Step/Axiom/Matcher.hs
```diff
@@ -790,19 +795,27 @@ matchNormalizedAc pushElement pushValue wrapTermLike normalized1 normalized2
                 _ -> empty
             lift $ traverse_ pushValue concrete12
             lift $ traverse_ pushValue abstractMerge
+    -- Case for AC iteration:
+    -- Normalized1 looks like K |-> V M:Map
     | [element1] <- abstract1
       , [frame1] <- opaque1
       , null concrete1 = do
```
MirceaS, line 802: shouldn't we also have a case for when we have 1 concrete element and 1 opaque element for instance?

ana-pantilie: Hmm, have you encountered cases where this would be useful? The idea behind map iteration is that the equation LHS contains some `K |-> V M:Map` map and when it matches it will match repeatedly on the the term. For example, say the equation is `rule f(K:Int |-> V:Int M:Map) => (K +Int 1 |-> V -Int 1) f(M)`, this will be able to repeatedly match on `f(1 |-> 1 2 |-> 2 3 |-> 3)` and eventually evaluate it to `f(2 |-> 0 3 |-> 1 4 |-> 2)`.

andreiburdusa (approved):

Issue #2713, opened 2021-06-29
Reset `Simplified`attribute on `SideCondition.Representation`


ana-pantilie:
It occurs to me that in this function, we should also strip the `Simplified` attribute off every `Predicate` and `TermLike`, because those can refer back to earlier `SideCondition`s, which would be pretty inefficient.

_Originally posted by @ttuegel in https://github.com/kframework/kore/pull/2662#discussion_r660229203_

ana-pantilie:
This is where we should do this: https://github.com/kframework/kore/blob/78ec7647e9b9eca9045544fe4a8c6c1fead93587/kore/src/Kore/Internal/SideCondition.hs#L450

ana-pantilie:
See https://github.com/kframework/kore/pull/2755#issuecomment-884942852.

Issue #2730, opened 2021-07-02
Intermediate step toward removing the `Simplified` attribute


ana-pantilie:
After #2713.

As an intermediate step toward removing the `Simplified` attribute, we can remove the `SideCondition` argument it stores, in favor of the changes described in #2520. That way, we get a lot of the performance benefit of removing the attribute without all of the behavior changes at once. We would just have to reset the `Simplified` status of any conditionally-simplified attributes after each rewrite step.

_Originally posted by @ttuegel in https://github.com/kframework/kore/issues/2520#issuecomment-869228163_

ana-pantilie:
I've been thinking about this more, and I think we should post-pone this until https://github.com/kframework/kore/issues/2570 and the removal of the old `TermLike` simplifier is done. The reason is that I'm afraid the `TermLike` simplifier still depends too much on the `Simplified` attribute, and changing the `Condition` case of the `Simplified` attribute might have unexpected consequences.

ana-pantilie:
I recall fixing this created a large performance regression. We need to see if this is still the case.

PR #2755, opened 2021-07-20
Reset simplified attribute in sideconditions.

RaoulSchaffranek:
Part of #2713.

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Internal/SideCondition.hs
+++ b/kore/src/Kore/Internal/SideCondition.hs
@@ -458,8 +458,7 @@ forgetSimplified ::
     SideCondition variable ->
     SideCondition variable
 forgetSimplified condition =
-    let
-        assumedTrue' =
+    let assumedTrue' =
             MultiAnd.map Predicate.forgetSimplified assumedTrue
         replacementsTermLike' =
             mapKeysAndValues TermLike.forgetSimplified replacementsTermLike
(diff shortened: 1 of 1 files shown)


ana-pantilie (commented):
Your changes are exactly what I was thinking.

No tests are failing, but this seems to have a very bad impact on performance, which is not what we expected but now that I think about it I think it makes sense.

I think it would be good exercise, though, to profile one of the specs (`sum-to-n` maybe). We have instructions to do that [in the wiki](https://github.com/kframework/kore/wiki/Profiling#haskell-profiling). In the meantime, I'm going to think about how we should continue here and we'll make a decision after looking at the profile as well.

RaoulSchaffranek:
I recorded two profiles ([profiles.zip](https://github.com/kframework/kore/files/6854484/profiles.zip)).

The `toRepresentation`-function now consumes 15% of the total time. Before the patch, it took only 0.04%. The main reason for the regression seems to be that the function is now called roughly 15k times. Before the patch, it was called 22 times.

I am currently investigating which execution is path leads to the explosion of function calls.

For my own memory, here is the command I used for profiling:
~~~bash
env GHCRTS='-p -l-au' kore-exec test/regression-evm/test-sum-to-n-vdefinition.kore\
   --module VERIFICATION\
   --prove test/regression-evm/test-sum-to-n-spec.kore\
   --spec-module SUM-TO-N-SPEC
~~~

ana-pantilie:
Since this performance issue is not that easily fixable and we also want to do https://github.com/kframework/kore/issues/2730, I'll close this PR and issue.

PR #2766, opened 2021-07-26
Overhaul Predicate simplifier: Equals, v2

### LINKED ISSUE #2570: Overhaul predicate simplifier

ana-pantilie:
Fixes #2570

Completes all cases of `Predicate.simplify`.

I didn't add any tests for `simplifyEquals` because it only calls `Equals.simplify`, which we will need to modify for https://github.com/kframework/kore/issues/2610.

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Simplify/Predicate.hs
+++ b/kore/src/Kore/Simplify/Predicate.hs
@@ -506,7 +506,7 @@ simplifyEquals ::
     simplifier NormalForm
 simplifyEquals sideCondition sort equals =
     Equals.simplify sideCondition equals'
-        >>= return . MultiOr.map (from @(Condition _))
+        <&> MultiOr.map (from @(Condition _))
   where
     equals' =
         equals
--- a/kore/src/Kore/Simplify/Predicate.hs
+++ b/kore/src/Kore/Simplify/Predicate.hs
@@ -47,7 +47,10 @@ import Kore.Internal.Substitution (
     pattern UnorderedAssignment,
  )
 import qualified Kore.Internal.Substitution as Substitution
-import Kore.Internal.TermLike (TermLike)
+import Kore.Internal.TermLike (
+    TermLike,
+    termLikeSort,
+ )
 import qualified Kore.Internal.TermLike as TermLike
 import Kore.Log.WarnUnsimplifiedPredicate (
     warnUnsimplifiedPredicate,
@@ -73,6 +76,7 @@ import Kore.Syntax (
     Not (..),
     Or (..),
     SomeVariableName,
+    Sort,
     Top (..),
     variableName,
  )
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -180,9 +148,11 @@ simplify sideCondition original =
...
                         >>= simplifyForall sideCondition
+                EqualsF equalsF@(Equals _ _ term _) ->
+                    simplifyEquals sideCondition (termLikeSort term)
+                        =<< traverse simplifyTerm equalsF
                 InF inF ->
                     simplifyIn sideCondition =<< traverse simplifyTerm inF
-                _ -> simplifyPredicateTODO sideCondition predicate & MultiOr.observeAllT
       where
         _ :< predicateF = Recursive.project predicate
         ~avoid = freeVariableNames sideCondition
```
JKTKops, line 158: Aren't `where` bindings lazy by default? What is this `~` for?

This line wasn't changed, but asking anyway while we're here.

kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -527,6 +497,23 @@ extractFirstAssignment someVariableName predicates =
...
+simplifyEquals ::
+    forall simplifier sort.
+    MonadSimplify simplifier =>
+    SideCondition RewritingVariableName ->
+    Sort ->
+    Equals sort (OrPattern RewritingVariableName) ->
+    simplifier NormalForm
+simplifyEquals sideCondition sort equals =
+    Equals.simplify sideCondition equals'
+        >>= return . MultiOr.map (from @(Condition _))
```
JKTKops, line 509: Why `m >>= return . f` instead of `fmap f m`?

JKTKops (approved):
Couple of minor comments, but looks good to me.

kore/src/Kore/Simplify/Predicate.hs
ana-pantilie: We have `Strict` on by default, maybe this changes the behavior? I don't know for certain, though.

PR #2767, opened 2021-07-26
Remove old TermLike simplifier

ana-pantilie:
Fixes #nnnn

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Simplify/Data.hs
+++ b/kore/src/Kore/Simplify/Data.hs
@@ -141,11 +141,6 @@ instance
         traceProfSimplify termLike (TermLike.simplify sideCondition termLike)
     {-# INLINE simplifyTermLike #-}
 
-    simplifyTermLikeOnly sideCondition termLike =
-        (traceProfSimplify termLike)
-            (TermLike.simplifyOnly sideCondition termLike)
-    {-# INLINE simplifyTermLikeOnly #-}
-
     simplifyCondition topCondition conditional = do
         ConditionSimplifier simplify <- asks simplifierCondition
         simplify topCondition conditional
--- a/kore/src/Kore/Simplify/Equals.hs
+++ b/kore/src/Kore/Simplify/Equals.hs
@@ -159,7 +159,7 @@ Equals(a and b, b and a) will not be evaluated to Top.
 simplify ::
     MonadSimplify simplifier =>
     SideCondition RewritingVariableName ->
-    Equals Sort (OrPattern RewritingVariableName) ->
+    Equals sort (OrPattern RewritingVariableName) ->
     simplifier (OrCondition RewritingVariableName)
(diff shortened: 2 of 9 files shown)


ana-pantilie:
I think the sporadic property test failures are due to switching from term simplification to predicate simplification in those tests which construct predicates to be simplified. I suspect it has something to do with the differences between term and predicate `\\and` simplification.

I also need to investigate if there is any real performance regression on some of the tests.

ana-pantilie:
The performance issue comes from `SideCondition.addConditionWithReplacements`, since we moved the "term + condition simplifier loop" from the `TermLike` simplifier to the `Pattern` simplifier, and the `Pattern` simplifier ends up calling that function more times than before.

ana-pantilie:
I tested the performance locally and `test/regression-wasm/test-loops.sh` is actually faster on this branch compared to master.

andreiburdusa (approved):

kore/src/Kore/Rewrite/Function/Evaluator.hs
```diff
@@ -332,12 +332,7 @@ reevaluateFunctions ::
     -- | Function evaluation result.
     Pattern RewritingVariableName ->
     simplifier (OrPattern RewritingVariableName)
-reevaluateFunctions sideCondition rewriting = do
-    let (rewritingTerm, rewritingCondition) = Pattern.splitTerm rewriting
-    OrPattern.observeAllT $ do
-        simplifiedTerm <- simplifyConditionalTerm sideCondition rewritingTerm
-        simplifyCondition sideCondition $
-            Pattern.andCondition simplifiedTerm rewritingCondition
+reevaluateFunctions = simplifyPattern
```
andreiburdusa, line 335: Do you think it's useful to keep `reevaluateFunctions` since it's just another name for `simplifyPattern`?

andreiburdusa (approved):

kore/src/Kore/Internal/SideCondition.hs
```diff
@@ -281,6 +281,23 @@ addAssumptions predicates sideCondition =
...
         }
 
+areIncludedIn ::
+    Eq variable =>
+    Foldable f =>
+    f (Predicate variable) ->
+    SideCondition variable ->
+    Bool
+areIncludedIn predicates sideCondition =
+    all (flip isIncludedIn sideCondition) predicates
```
emarzion, line 291: If `isIncludedIn` is only used here, does it make sense to flip the arguments here instead of simply flipping the arguments in the definition?  Of course the name would need to be modified.

emarzion (approved):

kore/src/Kore/Internal/SideCondition.hs
ana-pantilie: I wanted it to be similar to the `elem` function, so that the order of the arguments is intuitive.

PR #2776, opened 2021-08-02
Warn when RHS of a claim is \\bottom.

### LINKED ISSUE #2703: Add log and document prover behavior when claim has `\\bottom` RHS
Hello! This is probably an edge case but I think kprove is supposed to validate the `A-SPEC` spec below:
```
module A

syntax Foo ::= "a"

rule a => a

endmodule
```

```
module A-SPEC

imports A

claim a => . ensures false

endmodule
```

But instead it gets stuck at
```
kore-exec: [165966] Warning (WarnStuckClaimState):
    The proof has reached the final configuration, but the claimed implication is not valid. Location: /home/rodlin/Desktop/illinois/research/matching-logic-prover/matching-logic-proof-checker/examples/imp/.kprove-2021-06-26-19-29-45-382-5bb92c7f-1d1b-4273-a67b-df0c336a1d60/././a-spec.k:5:7-5:27
<k>
  a ~> _DotVar1 ~> .
</k>
```

I think perhaps it applied the semantics rule first before applying the claim inductively?
The java backend is able to prove this.

RaoulSchaffranek:
Fixes #2703.

---

###### Review checklist


kore/src/Kore/Log/WarnClaimRHSIsBottom.hs
```diff
@@ -0,0 +1,37 @@
...
+instance Pretty WarnClaimRHSIsBottom where
+    pretty (WarnClaimRHSIsBottom rule) =
+        Pretty.hsep
+            [ "The right-hand side of the claim is bottom:"
+            , Pretty.pretty (from rule :: SourceLocation)
+            ]
+
+instance Entry WarnClaimRHSIsBottom where
+    entrySeverity _ = Warning
+    helpDoc _ = "warn when the right-hand side of a claim is bottom"
```
JKTKops, line 31: I think we should be giving `oneLineDoc` implementations to all new log entries. That feature probably post-dates the entry that you used as a template for this one. The `oneLineDoc` should just be the prettified source location with no other text. See for example `DebugBeginClaim`.

RaoulSchaffranek: I added the function, but I don't know how to test it. I thought I had to pass `--log-format oneline` to `kprove`, but `kprove` doesn't seem to forward it to `kore-exec`.

ana-pantilie: If you run `kprove` with the following option `--haskell-backend-command "/path/to/kore-exec --log-format online"` it should work.

kore/src/Kore/Reachability/Claim.hs
```diff
@@ -569,17 +570,19 @@ checkImplicationWorker (ClaimPattern.refreshExistentials -> claimPattern) =
...
+            pure Implied
         | not didAnyUnify
           , not (isBottom right) =
-            NotImplied claimPattern
-        | otherwise =
-            Lens.set (field @"left") stuck claimPattern
-                & NotImpliedStuck
+            pure $ NotImplied claimPattern
+        | otherwise = do
+            warnClaimRHSIsBottom claimPattern
```
ana-pantilie, line 582: We should check if the RHS is bottom before emitting the warning.

ana-pantilie (changes requested):

kore/src/Kore/Reachability/Claim.hs
RaoulSchaffranek: Obviously, I checked the guards of the preceding cases and falsely concluded that the RHS must always be bottom in the last case. I added the check in 912d7a7

kore/src/Kore/Log/WarnClaimRHSIsBottom.hs
RaoulSchaffranek: Thanks, this worked! Would it be a good idea to add implementations of `oneLineDoc` for all existing entries in this PR, or should I open a fresh issue about that?

ana-pantilie: I think it's better if we do that separately, so please open an issue. Thanks!

ana-pantilie (approved):
Looks good! The performance data seems off, I think we need to merge master to get accurate results.

JKTKops (approved):

PR #2791, opened 2021-08-09
Fixing trim-source-paths.sh for macos

emarzion:
If developers need to invoke `generate-regression-tests.sh` on their own machines when adding new regression tests, it seems that we should fix the portability issue discovered earlier in `trim-source-paths.sh`.  This pull request fixes that issue.

I ran it on my machine (a mac) without issue, it would be nice if someone with a linux machine could run `generate-regression-tests.sh` just to be sure that everything still works.  I'm still a bit fuzzy on issues concerning the travis CI stuff, but it seems that if it works on someone's linux machine, then the travis pipeline should still work.

---

###### Review checklist


ana-pantilie (dismissed):

emarzion:
@ana-pantilie 

I'm looking at what `trim-source-paths.sh` removes [here](https://github.com/kframework/kore/pull/2588/files#diff-76efca24a8e12510256faf583750ac5136a55049f825fac7418c8a7e0b81d5c1L15).  Is it correct to assume that `/home/jenkins-slave/workspace/haskell-backend_PR-xxxx/` will always correspond to the `KORE` environment variable?

ana-pantilie:
> @ana-pantilie
> 
> I'm looking at what `trim-source-paths.sh` removes [here](https://github.com/kframework/kore/pull/2588/files#diff-76efca24a8e12510256faf583750ac5136a55049f825fac7418c8a7e0b81d5c1L15). Is it correct to assume that `/home/jenkins-slave/workspace/haskell-backend_PR-xxxx/` will always correspond to the `KORE` environment variable?

I'm not sure. Is it possible to just use the `KORE` env var? The names of the repositories (`evm-semantics` and `wasm-semantics`) shouldn't change, so maybe we could search for those inside the `KORE` string and remove whatever comes before? Would that work or am I oversimplifying things?

test/regression-wasm/test-locals-spec.kore
```diff
@@ -1,17 +1,17 @@
...
+// claim `<generatedTop>`(`<wasm>`(`<instrs>`(inj{Instr,KItem}(`aLocal.get`(X))~>inj{Instr,KItem}(`aLocal.set`(X))~>_DotVar2),_0,`<curFrame>`(`<locals>`(`_|->_`(inj{Int,KItem}(X),inj{Val,KItem}(`<_>__WASM-DATA_Val_ValType_Number`(_ITYPE,VAL)))),_DotVar3),_1,_2,_3,_4,_5,_6),_DotVar0)=>`<generatedTop>`(`<wasm>`(`<instrs>`(_DotVar2),_0,`<curFrame>`(`<locals>`(`_|->_`(inj{Int,KItem}(X),inj{Val,KItem}(`<_>__WASM-DATA_Val_ValType_Number`(_ITYPE,VAL)))),_DotVar3),_1,_2,_3,_4,_5,_6),_DotVar0) requires #token("true","Bool") ensures #token("true","Bool") [org.kframework.attributes.Location(Location(6,11,9,20)), org.kframework.attributes.Source(Source(/wasm-semantics/tests/proofs/locals-spec.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol])]
```
ana-pantilie, line 9: The path is wrong here, notice the `/` at the beginning.

ana-pantilie (changes requested):

emarzion:
There are still some relative paths present e.g. [here](https://github.com/kframework/kore/pull/2791/files#diff-f69077f519319a703deeedde5be4b3b6f649406178431bc7c248bc24d080c788R146) that for whatever reason this script wasn't set up to handle.  Would it make sense to add something like a `K_FRAMEWORK` environment variable and remove these in a similar manner?

ana-pantilie:
The link isn't loading for me. I looked for an example and found `system-specific-part/include/kframework/builtin/domains.md`, we'd probably want to keep just `include/kframework/builtin/domains.md`. I think there's also `include/kframework/builtin/kast.md`, right? Are there any other cases? If not, I think we should just do whatever is easiest, IMO this isn't something critical.

ana-pantilie (approved):

PR #2800, opened 2021-08-13
`SideCondition.definedTerms`: check if AC terms are always defined before caching them

### LINKED ISSUE #2779: `SideCondition.definedTerms`: check if AC terms are always defined before caching them
In [SideCondition.definedTerms](https://github.com/kframework/kore/blob/726faa4c269a70f0ad8d2045561beebb9350a829/kore/src/Kore/Internal/SideCondition.hs#L168) we keep a set of terms we assume to be defined during execution. See [assumeDefined](https://github.com/kframework/kore/blob/726faa4c269a70f0ad8d2045561beebb9350a829/kore/src/Kore/Internal/SideCondition.hs#L705) and [isDefined](https://github.com/kframework/kore/blob/726faa4c269a70f0ad8d2045561beebb9350a829/kore/src/Kore/Internal/SideCondition.hs#L787).

For AC structures, we build a minimal set of AC collections from which the definedness of any sub-collection can be inferred. In practice, people have noticed that we keep very large numbers of concrete ACs in memory for this purpose, and this should be avoided.

We can fix this by adding two checks:
1. If the initial collection is always defined, do not generate the minimal set.
2. After generating the minimal set, we should remove the ones which we know are always defined.

Unit tests should be modified/added accordingly.

andreiburdusa:
Fixes #2779 

---

###### Review checklist


JKTKops (approved):
Looks good to me.

ana-pantilie (approved):

PR #2818, opened 2021-08-31
Debugging: equation application, issue #2785

ana-pantilie:
Draft PR which illustrates debugging method for https://github.com/kframework/kore/issues/2785

@emarzion found the commit where the proof from the issue started failing. This commit added the call to `simplifyPredicate` immediately before the one to `simplifyPredicates`, in the `Condition` simplifier. This resulted in this proof (and possibly others) to not pass, seemingly because we attempted some equations without all the necessary information in the side condition.

The method I used to debug this issue was the following:

1. Label the problematic equation, in this particular case, the second `#hashedLocation` one.
2. Keep two versions of the backend on hand: one where we call both `simplifyPredicate` and `simplifyPredicates` (this is the _failing_ version) and another where we only call `simplifyPredicates` (this is the _passing_ version).
3. Run the proof with `--debug-equation` for the labeled equation on both versions of the backend and compare logs.
4. Find the corresponding log entries in the two log files: one where the equation is applicable and applied, the other where it's not applied. This showed me that in the case where it's not applied, the `SideCondition` is `#Top`, so some necessary information is missing, and deeper inspection is needed. These log entries also show us the term to which we apply the equations.
5. Add traces to the problematic part of the code, in both backend versions. Search for the term found in the previous step, compare sequence of logs.
6. This is where I noticed that, on the failing version, after `simplifyPredicate` is called, the result is marked as fully simplified! That explains why, in the failing equation log, the side condition doesn't contain enough information: we don't actually get to apply `simplifyPredicates` because the predicate is marked simplified.

See comments on this PR for detailed information. Also, on a more technical note, I've had problems in the past with `traceM` (it would sometimes not evaluate), so the traces I added all use `trace`. 

---

###### Review checklist

The author performs the actions on the checklist. The reviewer evaluates the work and checks the boxes as they are completed.

-   [ ] **Summary.** Write a summary of the changes. Explain what you did to fix the issue, and why you did it. Present the changes in a logical order. Instead of writing a summary in the pull request, you may push a clean Git history.
-   [ ] **Documentation.** Write documentation for new functions. Update documentation for functions that changed, or complete documentation where it is missing.
-   [ ] **Tests.** Write unit tests for every change. Write the unit tests that were missing before the changes. Include any examples from the reported issue as integration tests.
-   [ ] **Clean up.** The changes are already clean. Clean up anything near the changes that you noticed while working. This does not mean only spatially near the changes, but logically near: any code that interacts with the changes!

PR #2819, opened 2021-08-31
Condition simplifier: forget simplified before calling simplifyPredicates

### LINKED ISSUE #2785: Axiom Not Used When Simplifying Side Condition
When working on a proof in evm-semantics, we noticed this strange behavior around a `#hashedLocation` term that appears in a side condition. The requires clause for the appropriate rule can be inferred from the requires clause of the claim, but the term appears unsimplified in the configuration.

Here's a minimized version.
`test.k`:
```k
module TEST
  imports INT

  syntax Int ::= lookup(Hash) [function, no-evaluators]

  // We use this so that the proof takes a step
  syntax KItem ::= makeLookup(Int)
  rule makeLookup(B) => lookup(#hash(B))

  // -------------------------------------------
  syntax IntList ::= List{Int, ""}

  syntax Bool ::= #rangeUInt(Int) [function, no-evaluators]

  syntax Hash ::= #hash(Int)
                | #hashedLocation( IntList ) [function]

  rule #hashedLocation(       .IntList) => #hash(0)
  rule #hashedLocation(OFFSET .IntList) => #hash(OFFSET)
    requires #rangeUInt(OFFSET)

endmodule
```
`test-spec.k`:
```k
module TEST-SPEC
  imports TEST

  claim <k> makeLookup(B) => 1 ... </k>
    requires lookup(#hashedLocation(B .IntList)) ==Int 1
     andBool #rangeUInt(B)
endmodule
```

As long as the `#hashedLocation` in the `requires` clause of the claim is simplified to `#hash(B)`, the claim will become
```
claim <k> lookup(#hash(B)) => 1 ... </k>
  requires lookup(#hash(B)) ==Int 1
```
which obviously holds.

However, if we run the proof, we'll get a failing output:
```
$ versions
K commit b592f5d634d12da34c231b235aa66978f1798116
Kore commit 726faa4c269a70f0ad8d2045561beebb9350a829
$ kompile test.k --backend haskell
[Warning] Compiler: Could not find main syntax module with name TEST-SYNTAX in
definition.  Use --syntax-module to specify one. Using TEST as default.
$ kprove test-spec.k --backend haskell
kore-exec: [479535] Warning (WarnStuckClaimState):
    The proof has reached the final configuration, but the claimed implication is not valid. Location: /home/max/scratch/evm-semantics/axiom-requires/test-spec.k:4:9-6:27
  #Ceil ( lookup ( #hash ( B ) ) )
#And
  #Not ( {
    1
  #Equals
    lookup ( #hash ( B ) )
  } )
#And
  <k>
    lookup ( #hash ( B ) ) ~> _DotVar1 ~> .
  </k>
#And
  {
    1
  #Equals
    lookup ( #hashedLocation ( B  .IntList ) )
  }
#And
  {
    true
  #Equals
    #rangeUInt ( B )
  }
```
Notice the unsimplified `#hashedLocation` in the 4th component of the MultiAnd.

Doing any of the following make the proof pass:
- deleting the first `#hashedLocation` rule, which isn't used in the claim
- marking the second `#hashedLocation` rule as `[simplification]`
- deleting the `requires` clause of the second `#hashedLocation` rule

My understanding is that none of these workarounds are acceptable in KEVM, so this is blocking at least one proof.

My first guess would be that we don't construct the evaluator correctly, but I'm not very familiar with that part of the code.

ana-pantilie:
Fixes #2785
Needs https://github.com/kframework/kore/pull/2837

See also https://github.com/kframework/kore/pull/2818 for debugging method used.

The performance of this is as optimized as it could be, after https://github.com/kframework/kore/pull/2837 got merged.

---

###### Review checklist


ana-pantilie:
The simplification property test has started failing more frequently because we now simplify some intermediary predicates which are too general.

ana-pantilie:
The behavior seems to be the following: some `\\and` unification falls back to returning `\\equals(arg1, arg2)`. This results in a `TermLike` which contains a `Predicate`, which we will ignore during `TermLike` simplification. This used to be fine, since the resulting `Predicate` was marked simplified during unification. Now, this fix forgets simplified status across the whole term, and since the `TermLike` simplifier won't attempt to simplify `Equals`, we will throw the error at the level above.

I think there's something not right in transforming that `\\and` into an `\\equals` there, or maybe it's inevitable but we should always regard it as fully simplified. I'm thinking of changing `TermLike.forgetSimplified` to ignore subterms with an `\\equals` at the top, or we could probably generalize this to _any_ predicate which appears inside a `TermLike`.

andreiburdusa (approved):

PR #2820, opened 2021-08-31
Remove duplicated code between ==K, ==Int, and ==String

### LINKED ISSUE #2701: Remove duplicated code between ==K, ==Int, and ==String
There is a lot of duplicated code between the unification handlers for `==K`, `==Int`, and `==String`. The implementation of these unification handlers should be identical.

andreiburdusa:
Fixes #2701 

---

###### Review checklist


kore/src/Kore/Builtin/Builtin.hs
```diff
@@ -483,3 +503,37 @@ makeDomainValuePattern ::
 makeDomainValuePattern sort stringLiteral =
     Pattern.fromTermLike $
         makeDomainValueTerm sort stringLiteral
+
+-- | Unification of @eq@ symbols
+unifyEq ::
+    forall unifier unifyData.
+    MonadUnify unifier =>
+    (unifyData -> EqTerm (TermLike RewritingVariableName)) ->
+    (unifyData -> InternalBool) ->
```
ana-pantilie, line 512: The three `Unify..Eq` types are basically the same, would it make sense to just have a single `UnifyEq` type in these cases?

RaoulSchaffranek (approved):
The changes look good to me.  Indeed, they even revealed some more potential for stripping duplicated code. In particular, the `match*Equal`, `matchUnify*Eq`, `unify*Eq`, and `Unify*Eq` definitions look very similar to me. Would it make sense to address them in this PR, or should we open a follow-up issue?

ana-pantilie:
> Would it make sense to address them in this PR, or should we open a follow-up issue?

I think we should try to address them now.

kore/test/Test/Kore/Builtin/String.hs
```diff
@@ -35,31 +32,18 @@ import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 import qualified Kore.Builtin.Builtin as Builtin
 import qualified Kore.Builtin.String as String
+import qualified Kore.Builtin.String.String as String
```
RaoulSchaffranek, line 35: Learned something new today. I didn't know that Haskell lets us reuse a qualifier. Just for my understanding: Are all exports from `Kore.Builtin.String` and `Kore.Builtin.String.String` made available under the `String` qualifier? What if there is a naming collision? Will GHC complain, or will one definition shadow the other?

RaoulSchaffranek (approved):
Good work, the code looks much better now 👍

kore/test/Test/Kore/Builtin/String.hs
andreiburdusa: Yes, in case of a naming collision, GHC will give an `Ambiguous occurrence` error, but only when the ambiguous function name is actually used.

ana-pantilie: I'm not sure, I think I remember GHC complaining in such a situation.

ana-pantilie (approved):
:+1:

PR #2830, opened 2021-09-08
Add oneLineDoc implementations to all existing Entry-instances

### LINKED ISSUE #2781: Add `oneLineDoc` implementations to all existing `Entry`-instances
The Haskell backend supports a short logging format that only logs the source locations of log entries, but no descriptive messages.

Log entries which relate to specific source locations should implement the following method:
https://github.com/kframework/kore/blob/29888c3567cb479fc5ee619f21a1545d88ff66e8/kore/src/Log/Entry.hs#L61-L62

However, not all `Entry`-instances implement this function yet.

Here is an example of an instance which implements the method:
https://github.com/kframework/kore/blob/29888c3567cb479fc5ee619f21a1545d88ff66e8/kore/src/Kore/Log/DebugBeginClaim.hs#L33-L34

Here is an example of instance which doesn't implement the method:
https://github.com/kframework/kore/blob/29888c3567cb479fc5ee619f21a1545d88ff66e8/kore/src/Kore/Log/WarnDepthLimitExceeded.hs#L35-L38

The first step to solve this issue would be to identify all `Entry`-instances which lack an appropriate implementation. The second step is implementing these functions.

The short logging format is enabled by passing `--log-format oneline` command-line argument to `kore-exec`.

andreiburdusa:
Fixes #2781 

---

###### Review checklist


dopamane (approved):
Approved, looks good! However, I think the `Entry` typeclass is overkill and it could be replaced with a simpler sum type. I'm envisioning something along these lines:
```haskell
data Entry variable
  = EntryWarnDepthLimitExceeded WarnDepthLimitExceeded
  | EntryWarnNotImplemented (WarnNotImplemented variable)
  | EntryErrorParse ErrorParse
  | ...

entrySeverity :: Entry variable -> Severity
entrySeverity = \case
  EntryWarnDepthLimitExceeded _ -> Warning
  EntryWarnNotImplemented _ -> Warning
  EntryErrorParse _ -> Error
  ...

helpDoc :: Entry variable -> Pretty.Doc ann
helpDoc = \case
  EntryWarnDepthLimitExceeded _ -> "warn when depth limit is exceeded"
  EntryWarnNotImplemented _ -> "warn ..."
  EntryErrorParse _ -> Pretty.emptyDoc
  ...

oneLineDoc :: Entry variable -> Pretty.Doc ann
oneLineDoc = \case
  ...
```
Thoughts? Any reasons why a sum type would not suffice?

kore/src/Kore/Log/WarnStuckClaimState.hs
```diff
@@ -44,6 +44,13 @@ instance Pretty WarnStuckClaimState where
 
 instance Entry WarnStuckClaimState where
     entrySeverity _ = Warning
+    oneLineDoc (TermsUnifiableStuck claim) =
+        Pretty.vsep
+            ["TermsUnifiableStuck", Pretty.pretty @SourceLocation $ from claim]
+    oneLineDoc (TermsNotUnifiableStuck claim) =
+        Pretty.vsep
+            ["TermsNotUnifiableStuck", Pretty.pretty @SourceLocation $ from claim]
+
```
RaoulSchaffranek, line 53: Is there a reason to use `vsep` instead of `hsep` here? As a user, I would expect that each entry corresponds to exactly one line in the log.

kore/src/Kore/Log/DebugUnifyBottom.hs
```diff
@@ -55,6 +55,7 @@ instance Pretty DebugUnifyBottom where
 
 instance Entry DebugUnifyBottom where
     entrySeverity _ = Debug
+    oneLineDoc _ = mempty
```
RaoulSchaffranek, line 58: Do you think it would be useful to default to `show` if there is no detailed information available? Or is it better to omit the message? This also affects other instances that default to `mempty`.

If we stay with `mempty`, does it produce an empty line or no line at all?

andreiburdusa: I used `mempty` because that's what [Max suggested](https://github.com/kframework/kore/issues/2781#issuecomment-893590259). The one line log looks like this:
```
kore-exec: [361804] Debug (DebugUnifyBottom):
```
Perhaps it is confusing to have nothing after the colon mark, but using `oneLineDoc = Pretty.pretty . show` would print everything in Haskell syntax, like
```
kore-exec: [357145] Debug (DebugUnifyBottom): DebugUnifyBottom {info = "...", first = TermLike {getTermLike = TermAttributes {termSort = SortActualSort (SortActual {sortActualName ...
```
Do you think `DebugUnifyBottom _` would be better in this case, and similarly for the others cases that use `mempty`?

RaoulSchaffranek: Thanks for the clarification. Yes, I think having something after the colon mark would be a good thing for the reason you mentioned. The type-name seems to be a reasonable choice; even if it doesn't provide the user with extra knowledge, it can avoid confusion.

RaoulSchaffranek (approved):

PR #2837, opened 2021-09-13
Add equation application caching to the simplifier

### LINKED ISSUE #2321: Cache the results of applying equations at the level of individual equations
This would improve performance significantly. The advantage of caching the result of applying individual equations is that it would allow us to skip equations which do not match. Otherwise, we would get cache hits only when a term is simplifiable. A significant performance problem for us is large symbolic terms which are not simplifiable because we must try to re-simplify them frequently.

ana-pantilie:
Fixes #2321

This uses a naive caching algorithm, meaning it doesn't attempt to trim the cache in certain situations. I found that a more sophisticated algorithm didn't bring much benefits in practice, but this can change once we gather more data and figure out how to properly optimize things.

See https://github.com/kframework/kore/pull/2846#issuecomment-924231145 for performance data with these changes + the bug fix which isn't present in this PR.

---

###### Review checklist


ana-pantilie:
I'm going to try to use https://hackage.haskell.org/package/psqueues-0.2.7.2 to implement a less naive cache.

ana-pantilie:
With the changes from https://github.com/kframework/kore/pull/2819, it seems like the naive cache is actually performing better than the LFU cache (tested on `regression-evm/test-storagevar03.sh`, which showed a large performance drop with the bug fix).

kore/src/Kore/Simplify/Simplify.hs
```diff
@@ -300,6 +324,45 @@ liftConditionSimplifier (ConditionSimplifier simplifier) =
         scatter results
 -- * Builtin and axiom simplifiers
 
+{- | Used for keeping track of already attempted equations which failed to
+ apply. This uses a naive caching algorithm, meaning it doesn't attempt
+ to trim the cache in certain situations. We found that a more sophisticated
+ algorithm didn't bring much benefits in practice,
```
ana-pantilie, line 330:
```suggestion
 apply.
```

andreiburdusa (approved):

emarzion (approved):

Issue #2472, opened 2021-03-16
Create a tool to check function definitions


ttuegel:
We will create a tool tentatively called `kore-check-functions` which will check the following properties of function definitions:

1. For every equation in a function definition, the right-hand side of the equation is a function pattern.
2. For every pair of equations in a function definition, the equations cannot match the same term.

Recall the [interpretation](https://github.com/kframework/kore/blob/master/design-decisions/2020-05-02-function-rules.md#solution) of function rules in Kore. We will use that notation in the following.

To check the first property is very simple, so we will use that task as an opportunity to set up infrastructure: taking the definition as input, validating it, getting the list of equations, etc.

To check the second property, we will instantiate the pair of equations `E₁` and `E₂` at the same term e.g. `fun({X})` and evaluate

```
Pre₁({Y}) ∧ Args₁({X}, {Y}) ∧ Prio₁({X}) ∧ Pre₂({Y}) ∧ Args₂({X}, {Y}) ∧ Prio₂({X})
```

If this predicate is unsatisfiable, then the equations can never match the same term.

ana-pantilie:
@dopamane For the second part, remember that equations are grouped according to the symbol at the top. Let `[E1, E2, ..., En]` be one of these groups, we should generate all unordered pairs `(Ei, Ej)` and evaluate the predicate in the description. We need to check that all of the results are `\\bottom`, for every group of equations.

PR #2840, opened 2021-09-15
Kore check functions task2

emarzion:
Addresses Task 2 of #2472 

---

###### Review checklist


kore/src/Kore/Exec.hs
```diff
@@ -638,6 +655,52 @@ checkFunctions verifiedModule =
...
+    -- | The main module
+    VerifiedModule StepperAttributes ->
+    m ExitCode
+checkBothMatch verifiedModule =
+    filterM (uncurry bothMatch) equations
+        >>= checkResults
+  where
+    equations = join $ map mkPairs $ Map.elems $ extractEquations verifiedModule
+    -- produces all 'in-order' pairs in a list
+    mkPairs xs =
```
ana-pantilie, line 696: I remember adding some code used to generate unordered pairs (see the type's [smart constructor](https://github.com/kframework/kore/blob/98e241ede419e7a9e1266393104fcbd84fb796e4/kore/src/Kore/Internal/NormalizedAc.hs#L540)). I think we can use this type here to simplify things a little, let me know if this doesn't make sense for some reason.

kore/test/Test/Kore/Exec.hs
```diff
@@ -356,6 +362,105 @@ test_checkFunctions =
...
+        "checkBothMatch"
+        [ testCase "Function patterns do not both match." $ do
+            let verifiedModule =
+                    verifiedMyModule
+                        Module
+                            { moduleName = ModuleName "MY-MODULE"
+                            , moduleSentences =
+                                [ asSentence mySortDecl
+                                , asSentence $ constructorDecl "a"
+                                , --                                , asSentence $ constructorDecl "b"
```
ana-pantilie, line 377: Should this comment be removed?

kore/test/Test/Kore/Exec.hs
```diff
@@ -356,6 +362,105 @@ test_checkFunctions =
...
+                                  asSentence mySymbDecl
+                                , mkEq "a" makeTruePredicate
+                                , mkEq "a" makeFalsePredicate
```
ana-pantilie, line 380: I think this `mkEq` function is misleading, what do its arguments represent?

kore/app/check-functions/Main.hs
```diff
@@ -104,14 +110,26 @@ main = do
...
-koreCheckFunctions opts =
+getLoadedModule :: KoreCheckerOptions -> Main LoadedModule
+getLoadedModule opts =
     loadDefinitions [fileName opts]
         >>= loadModule (mainModuleName opts)
-        >>= checkFunctions
+
+koreCheckBothMatch :: LoadedModule -> Main ExitCode
+koreCheckBothMatch loadedMod =
+    SMT.runNoSMT $ evalSimplifier loadedMod $ checkBothMatch loadedMod
```
ana-pantilie, line 132: I think we might need the SMT solver here.

kore/src/Kore/Exec.hs
emarzion: I'm not sure if that is relevant to this situation, by in-order I mean the order in which they are included in the list, so e.g. `[e0,e1,e2]` should yield `[(e0,e1), (e0,e2), (e1,e2)]` (although not necessarily in that order)

kore/test/Test/Kore/Exec.hs
emarzion: This function takes arguments `name` and `pred` and produces the equation `f(x) = name, assuming pred`.  I think the name can be improved but I'm not sure what a good name for it would be.

kore/src/Kore/Exec.hs
dopamane: From [Stackoverflow](https://stackoverflow.com/questions/34044366/how-to-extract-all-unique-pairs-of-a-list-in-haskell/34045121)
```
import Data.List

pairs :: [a] -> [(a, a)]
pairs l = [(x,y) | (x:ys) <- tails l, y <- ys]
```

kore/test/Test/Kore/Exec.hs
```diff
@@ -356,6 +362,104 @@ test_checkFunctions =
...
+                        , requires = pr
+                        , argument = Nothing
+                        , antiLeft = Nothing
+                        , right = wrap name
+                        , ensures = makeTruePredicate
+                        , attributes = def
+                        }
+      where
+        wrap = applyToNoArgs mySort
+        v = mkElementVariable (testId "V") mySort
```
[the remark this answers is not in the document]
emarzion: The variable stuff came out of an earlier failed approach to write this test using constructor matching as patterns, so I decided to just simplify things by making the function take zero arguments.  I think the test is correct now but let me know if I'm still missing something.

Although not necessarily relevant now I'm wondering if there's any good documentation of what the various fields of `Equation` are supposed to represent, and what functions like `fromIn_` are supposed to do?  I was looking into your proposed fix and couldn't really understand what was to be done.

kore/src/Kore/Exec.hs
```diff
@@ -638,6 +654,51 @@ checkFunctions verifiedModule =
...
+checkBothMatch ::
+    MonadSimplify m =>
+    -- | The main module
+    VerifiedModule StepperAttributes ->
+    m ExitCode
+checkBothMatch verifiedModule =
+    filterM (uncurry bothMatch) equations
+        >>= checkResults
+  where
+    equations = join $ map mkPairs $ Map.elems $ extractEquations verifiedModule
```
ana-pantilie, line 695: I just realized something: I think `equations` will contain simplification equations as well! We want to only run these checks (the one you're working on and the one David worked on) for _function definition_ equations.

First, to make sure that this is indeed the case, I'd add a test for each check where we make sure we ignore simplification equations. Remember, simplification equations are those which have the `simplification` attribute. Then, if those tests fail, we need to extract only those equations which _don't_ have the `simplification` attribute, here and for the first check as well. The tests should pass after this.

kore/test/Test/Kore/Exec.hs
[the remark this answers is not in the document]
ana-pantilie: I think it's fine now.

I'll open a PR which documents these, thanks!

kore/src/Kore/Exec.hs
```diff
@@ -631,13 +648,64 @@ checkFunctions ::
...
+                    Predicate.makeAndPredicate arg1 $
+                        Predicate.makeAndPredicate arg2 $
+                            Predicate.makeAndPredicate prio1 prio2
+        check' = Predicate.mapVariables (pure mkConfigVariable) check
+        sort = termLikeSort $ right eq1
+        patt = Pattern.fromPredicateSorted sort check'
+     in (not . isBottom) <$> Pattern.simplify patt
+
+{- | Checks if any function definition in the module carries two equations that both match
+ - same term.
```
dopamane, line 687:
```suggestion
same term.
```
Haddock will display the `-`. I mistakenly thought it would ignore the dashes when I wrote lines 631-641 for `checkFunctions`. See also line 661.

dopamane:
@ana-pantilie @emarzion When I run `make all`, the only binaries that are copied to `.build/kore/bin` are `kore-exec`, `kore-parser`, and `kore-repl`. Should we also be copying `kore-check-functions`, `kore-format`, `kore-match-disjunction`, and `kore-prof` from `.stack-work/install/.../bin/`? If yes, then for `kore-check-functions` we should add in `include.mk`:
```makefile
KORE_CHECK_FUNCTIONS = $(BUILD_DIR)/kore/bin/kore-check-functions
KORE_CHECK_FUNCTIONS_OPTS = --no-bug-report
export KORE_CHECK_FUNCTIONS
export KORE_CHECK_FUNCTIONS_OPTS
```
and
```makefile
$(BUILD_DIR)/kore/bin/kore-check-functions:
        $(STACK) $(STACK_BUILD) $(STACK_NO_PROFILE) --copy-bins kore:exe:kore-check-functions
```

Additionally in `Makefile`:
```makefile
.PHONY: all clean docs haddock \
        test test-kore test-k \
        kore-exec kore-repl kore-check-functions
```
and
```makefile
kore-check-functions: $(KORE_CHECK_FUNCTIONS)
```

Keep in mind Make uses tabs but these code blocks use spaces.

ana-pantilie (approved):

dopamane (approved):
Looks good!

PR #2852, opened 2021-09-27
Catch exception produced by graph command on mac

### LINKED ISSUE #2817: kore-repl: don't crash when system doesn't support graph visualizer
`kore-repl` shouldn't crash when the system doesn't support the graph visualizer for some reason. We should try to reproduce the crash on a macOS machine by running the `graph` command.

emarzion:
Fixes #2817

Catches the `GVProgramExc` produced by the graph visualizer without terminating the repl session.

---

###### Review checklist


emarzion:
I have tested this on my mac machine.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1448,6 +1455,22 @@ showDotGraph gr =
...
+    Gr CommonClaimState (Maybe (Seq axiom)) ->
+    IO ()
+showDotGraphCatchException gr =
+    catch (showDotGraph gr) $ \(e :: SomeException) ->
+        hPrint stderr $
+            Pretty.vsep
+                [ "Encountered the following exception:"
+                , Pretty.indent 4 $ fromString $ displayException e
+                , "Please note that the graph command is not\
+                  \ currently supported on MacOS."
```
ana-pantilie, line 1471: Let's recommend using `graph` with a file name argument, which works on MacOS as well. This saves the graph image to the file.

kore/src/Kore/Repl/Interpreter.hs
```diff
@@ -1448,6 +1455,22 @@ showDotGraph gr =
...
         $ gr
 
+-- | A version of @showDotGraph@ that catches any exceptions.
+showDotGraphCatchException ::
+    From axiom AttrLabel.Label =>
+    From axiom RuleIndex =>
+    Gr CommonClaimState (Maybe (Seq axiom)) ->
+    IO ()
+showDotGraphCatchException gr =
+    catch (showDotGraph gr) $ \(e :: SomeException) ->
```
ana-pantilie, line 1465: Let's not catch any exception here, we should just catch the exception thrown in this case.

emarzion: Ok, I fixed this so that we are now only catching exceptions coming from from the `GVProgramExc` constructor.  If we wanted to narrow this down more we would need to inspect the message string, which seems a bit brittle to me, but let me know what you think.

ana-pantilie: Looks good, thanks!

ana-pantilie (approved):

PR #2856, opened 2021-09-30
Switching out kprove to kprovex

### LINKED ISSUE #2824: Use new `kprovex` with our integration tests

emarzion:
Fixes #2824 

Estimate: ~~11/3~~ Encountered more issues dealing with kompiled artifacts needing to be placed into separate directories.

11/8

---

###### Review checklist


ehildenb:
So with kprovex, you need to call `kompile` ahead of time. So you'll need to modify your makefile scripts in a way similar to this: https://github.com/kframework/evm-semantics/pull/1116/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R380

Basically, you make it so the `kprove` target depends on the correct `timestamp` (or kompile) target, so that it makes sure that kompile is called ahead of item. If you have multiple proofs that use the same kompiled directory, then this will make sure it only gets kompiled once.

emarzion:
> If you have multiple proofs that use the same kompiled directory, then this will make sure it only gets kompiled once.

I'm still looking through this but I'm a bit unclear on this wording, the issue isn't multiple tests needing the same kompiled directory, it's that within a single directory there now might be multiple kompiled directories, so `kprovex` doesn't know which to use.

ehildenb:
Ah yes, you'll have to put the `*-kompiled` directory in a unique subdirectory then. There isn't really a way around this right now, if you want to be able to run tests in parallel.

scripts/integration-k.sh
```diff
@@ -18,7 +18,7 @@ if make --version | grep -q 'GNU Make 4' 2>/dev/null
 then
     MAKE="make --output-sync --jobs ${JOBS:?} --directory $TOP"
 else
-    MAKE="make -j ${JOBS:?} -C $TOP"
+    MAKE="make -C $TOP"
```
radumereuta, line 21: Why remove the parallelism?

radumereuta (approved):
Just one question, otherwise lgtm.

scripts/integration-k.sh
ana-pantilie: Thank you for taking a look!

I didn't notice this, good catch. Unfortunately the commit message for this change is not descriptive at all, so I'll try to figure out why the original author of the PR decided to remove this.

ana-pantilie: I reverted the change. There was one strange failure regarding the integration tests, but I couldn't reproduce it again on Jenkins or locally.

PR #2895, opened 2021-10-28
Z3 automatic restart

### LINKED ISSUE #2816: Restart Z3 when it crashes
Since we already reinitialize the solver after a certain number of queries I think it should be feasible to restart Z3 when it crashes. We probably should do something similar to when we retry queries which returned `unknown`: when Z3 crashes, restart the process only one time and emit a warning.

We should investigate first to see if restarting is indeed an option, I am not 100% sure that we can actually restart without losing any information. If restarting and reinitializing are the same in terms of what happens to Z3's inner state, then it's do-able.

dopamane:
Fixes #2816 

### Scope:

* Retry running SMT action once if it throws a `SolverException`. Note we now wrap an `AsyncException` inside a `SolverException` in `throwSolverException` to ensure we are only catching exceptions in `runSMT` that are relevant to SMT interaction.
* Add warning log entry `WarnZ3Crash` for SMT crash. I've created a new module `Kore.Log.Warning` and placed `WarnZ3Crash` there. Moving forward, when creating a new warning (or debug, error, info) please group them in `Kore.Log.Warning` (`Kore.Log.Debug` etc.). That way we will never have to create a new module, add imports, and modify our cabal file for each new entry type.
* By default, we will run an SMT action twice if it throws a `SolverException`. To disable this behavior and only run SMT action once, use `--smt-never-auto-restart` flag. 

#### Error Handling

In `SMT.SimpleSMT` `throwSolverException` takes as arguments a `ProcessHandle` and `SomeException`. It examines the error and re-throws it as a `SolverException`. `trySolver` uses `throwSolverException` as a handler and "wraps" all SMT queries, actions, commands etc. so any errors related to SMT interaction must be a `SolverException`.

### Estimate:

Wednesday, December 15th

---

###### Review checklist


kore/src/SMT.hs
```diff
@@ -517,11 +529,18 @@ stopSolver mvar = do
...
-    Exception.bracket
-        (newSolver config)
-        stopSolver
-        (\mvar -> runSMT' config userInit mvar smt)
+runSMT config userInit smt = case autoRestart config of
+    Never -> smtBracket
+    Once ->
+        smtBracket `catch` \(e :: SimpleSMT.SolverException) -> do
+            warnZ3Crash e
+            smtBracket
```
ana-pantilie, line 537: I'm confused, won't this basically re-run the whole prover?

I mean, what we want is if during some query we find that Z3 has crashed, we restart _just_ Z3 and retry that specific query, resuming execution from that point.

ana-pantilie (changes requested):

PR #2915, opened 2021-11-12
Add limited loop to Condition simplifier

### LINKED ISSUE #2878: Regression: Kprove hangs instead of crashing
Test file:
------------
```
module TEST
    imports MAP
    imports BOOL

    syntax Foo ::= "a" | "b" | "c"
 // ------------------------------
    rule <k> a => b ... </k>

endmodule

module TEST-SPEC
    imports TEST

    claim <k> (a ~> _ => .) ... </k>
 
endmodule
```

Commit https://github.com/kframework/kore/commit/ed422630f9383b2774f1ec8ea04eee86ea0afe1e changed the behavior of running kprove on this test case.
In order to reproduce this behavior I create a makefile with the following lines and run `make`:
```
all:
	rm -rf *-kompiled .kprove*
	kompile --main-module TEST --backend haskell test.k
	kprovex test.k --spec-module TEST-SPEC --debug 
```
Prior to this commit, the kfrontend code crashed on a pattern match error shown in https://github.com/kframework/k/issues/2245 and I'm expecting this crash to continue to happen. The actual behavior seems to hang sometime before kore is printed. I also found that adding `--depth 0` to the last command causes the pattern match crash to happen.

ana-pantilie:
Fixes #2878

Note: none of our current integration tests emit the newly added warning, so I believe the `limit = 4` is appropriate.

### Scope:
- Disallows infinite looping of the `Condition` simplifier
- Workaround for problematic cyclic dependency in the `Condition` simplifier

### Estimate:
- Monday, 15 November

---

###### Review checklist


kore/src/Kore/Simplify/Condition.hs
```diff
@@ -78,9 +81,26 @@ simplify ::
...
+        LogicT simplifier (Conditional RewritingVariableName any)
+    loop count input
+        | count >= limit = do
+            warnUnsimplifiedCondition limit original input
+            pure input
+        | otherwise = do
+            output <- worker input
+            if fullySimplified output
+                then return (extract output)
+                else loop (count + 1) (extract output)
```
dopamane, line 102: Instead of explicit recursion can we use `foldM`?:
```suggestion
simplify SubstitutionSimplifier{simplifySubstitution} sideCondition original = do
    normalizedOriginal <- normalize original
    (isFullySimplified, result) <- foldM simplifyingCondition (False, normalizedOriginal) [0..limit]
    unless isFullySimplified $ warnUnsimplifiedCondition limit original result
    return result
  where
    limit :: Int
    limit = 4
    
    simplifyingCondition ::
        (Bool, Conditional RewritingVariableName any) ->
        Int ->
        LogicT simplifier (Bool, Conditional RewritingVariableName any)
    simplifyingCondition result@(True, _) _ = return result
    simplifyingCondition (_, input) _ = do
        output <- worker input
        return (fullySimplified output, extract output)
```

dopamane (changes requested):

kore/src/Kore/Simplify/Condition.hs
andreiburdusa: This also looks fine, although I'd probably have spent more time trying to understand it if I didn't know it's meant to replace the initial recursion.

emarzion (approved):

kore/src/Kore/Simplify/Condition.hs
ana-pantilie: This was my initial thought as well, the code isn't as straightforward with `foldM`. It also seems harder to debug, and we'll inevitably have lots of occasions to debug this I'm afraid, because of the issues with the simplifier. @dopamane, what do you think?

ana-pantilie: TODO: add comment linking to old version

andreiburdusa (approved):

dopamane (approved):

Issue #2923, opened 2021-11-23
Discussion: overhaul the simplifier


ana-pantilie:
# NOTE: This issue description is a draft, WIP

## Introduction

Right now, the simplifier has reached a point where the old design clashes with newer features and it also no longer reflects the practical needs of the simplification expected to be done by the Haskell backend. Basically, the old design is too general and has acquired a lot of technical debt along the years.

What I mean by too general: initially, all Kore patterns were represented in the same way, and the simplifier strove to be able to simplify any arbitrary Kore pattern. It was also decided that each pattern connective has its own simplifier, which would attempt to simplify it as much as possible. These two approaches have introduced cyclic dependencies throughout the simplifier code, making it hard to understand and maintain. Newer features have broken many of the simplifier's initial assumptions, introducing even more brittle behavior into the code.

The following new design starts from the current design, together with additions which follow from previous refactorings and restructurings of parts of the simplifier code.

## Representing Kore patterns

The simplifier simplifies _configurations_. A configuration is a Kore pattern which has the following structure: it is a conjunction between a _term_ and a _condition_. I have noticed K users refer to the _condition_ part as the _side condition_, because it usually comes from the side condition (the `requires`) of the claim which is being proven.

Recall that in Matching Logic, a _predicate_ is a pattern which is semantically either `\\top` or `\\bottom`. We know that some connectives will necessarily construct predicates (`\\ceil`, `\\equals` etc.).

The backend distinguishes between arbitrary predicates and the more specialized kind of predicate, _substitutions_. Therefore, we consider _conditions_ to be _conjunctions of predicates and substitutions_.

A _term_ is a pattern which, in theory, can be any arbitrary Kore pattern. In practice, however, I believe this pattern can be constrained to a subset of Kore patterns. I am basing this on the observation that the code which does term simplification does not attempt to handle any cases where the term is constructed using predicate-making connectives (this is a fairly recent change). Therefore, it seems like these _terms_ should be restricted as described in the following issue: https://github.com/kframework/kore/issues/2640.

TODO: change above will also result in issues with current `Predicate` representation, where there is a "predicate connective" over a `Predicate` (currently it only admits `TermLike`).

ehildenb:
Some ideas about the simplifier:

- Many small and simple passes which only do one thing (eg. "push `#Or` down" or "pull `#And` up")
- Call many of these passes in a specific sequence to achieve a larger goal.
[...]

PR #2937, opened 2021-12-08
Add simplifierx switch

dopamane:
Fixes https://github.com/kframework/kore/issues/2939

See #2923 for simplifier discussion.

### Scope:
* Add `--simplifierx` switch

---

###### Review checklist

### DIFF
--- a/kore/app/share/GlobalMain.hs
+++ b/kore/app/share/GlobalMain.hs
@@ -269,8 +269,8 @@ the project
 data GlobalOptions = GlobalOptions
     { -- | Version flag [default=false]
       willVersion :: !Bool
-      -- | Experimental simplifier flag [default=false]
-    , simplifierx :: !Bool
+    , -- | Experimental simplifier flag [default=false]
+      simplifierx :: !Bool
     }
 
 -- | Record type to store all state and options for the subMain operations
--- a/kore/app/share/GlobalMain.hs
+++ b/kore/app/share/GlobalMain.hs
@@ -113,7 +113,6 @@ import Options.Applicative (
     ParserHelp (..),
     defaultPrefs,
     execParserPure,
-    flag,
     handleParseResult,
     help,
     helper,
@@ -270,6 +269,8 @@ the project
 data GlobalOptions = GlobalOptions
     { -- | Version flag [default=false]
       willVersion :: !Bool
+      -- | Experimental simplifier flag [default=false]
+    , simplifierx :: !Bool
     }
 
 -- | Record type to store all state and options for the subMain operations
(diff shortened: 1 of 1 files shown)


kore/app/share/GlobalMain.hs
```diff
@@ -270,6 +269,8 @@ the project
 data GlobalOptions = GlobalOptions
     { -- | Version flag [default=false]
       willVersion :: !Bool
+    , -- | Experimental simplifier flag [default=false]
+      simplifierx :: !Bool
```
ana-pantilie, line 273: I'd rather create a new datatype instead of using `Bool`, `data ExperimentalSimplifier = Enabled | Disabled`.

ana-pantilie (changes requested):
This only parses an additional CL option, but doesn't do anything with it. This needs to be threaded to the simplifier.

A field of the `ExperimentalSimplifier` type needs to be added to the [simplifier's environment](https://github.com/kframework/kore/blob/dd3d4d9da6d197215e7e6cef7ca235ecfcedc709/kore/src/Kore/Simplify/Data.hs#L84). When the simplifier is ran, it needs to be set to the option received from the command line.

You should also add a method for extracting the `ExperimentalSimplifier` to the [Simplifier class](https://github.com/kframework/kore/blob/dd3d4d9da6d197215e7e6cef7ca235ecfcedc709/kore/src/Kore/Simplify/Simplify.hs#L148). This way, it's available throughout all the simplification code.

That's just the first part. The second part is about making changes in our test suites to accommodate the new option.

The integration tests require changes made to the Makefiles, we should probably add a new target like `make testx` which runs the tests with the `simplifierx` option turned on. This also needs to be added on CI.

In the case of the unit tests, the changes to the simplifier will trigger some compilation errors wherever some form of `run/eval/...simplifier` is defined in the tests. We should see if it's possible to change these so that it runs the simplifier twice: once with the switch turned off and once with the switch turned on.

It remains to be discussed if we think it's necessary to run all of our tests twice.

In the case of the unit tests, things will most likely break on the experimental branch. This is because many unit tests are old and very general. This is good because we can explicitly see what features we will be breaking and make decisions based on concrete examples, but also bad because it will be more time consuming.

In the case of the integration tests, we definitely need to get everything working on the experimental branch. The problem is that we will suffer from much longer CI times. Therefore, I think this PR should not enable CI with the switch turned on just yet, it should add the necessary infrastructure so that when we add the first experimental feature we can turn it on easily.

ana-pantilie:
Since I need new work, I picked this up. I tried to work with having the new option added to `GlobalOptions`, but it seems to me like these options were added to only be used in `GlobalMain`, before calling the specific executable. Since I can't really use your changes, I'm going to open a new PR. That will also mean that you can review/approve it when it's ready.

dopamane:
Ok, thank you for the explanation!

PR #2945, opened 2021-12-16
Add NonExecutable attribute

### LINKED ISSUE #2933: Allow users to specify `non-executable` on rule/simplification rules
We need a way to feed lemmas to only the smt-solver, but _not_ use them as K reductions. To achieve this, I'd like to write:

```
rule LHS => RHS requires REQ [smt-lemma, non-executable]
```

This just means that the rule should not be considered for rewriting (whether it's a simplification rule or not).

This needs to also interact correctly with `no-evaluators`, that is when you put `no-evaluators` on a symbol it should only have `non-executable` rules present for that symbol.

This may also require a small update to the frontend, to make it aware of this new attribute.

ana-pantilie:
Fixes #2933

### Estimate:

---

###### Review checklist


Baltoli:
@ana-pantilie Is it possible to have the attribute apply to rewrite rules as well as functions in this change? For example, given these rules:

```
rule g(X, Y) => X +Int Y [non-executable, priority(1)]
rule g(X, Y) => X requires X >=Int Y
rule g(_, _) => 4567 [owise]
```

The LLVM backend will rewrite `g(3, 100)` to `4567` as expected, while the Haskell backend as tested at this PR will rewrite to `103` using the `non-executable` rule.

ana-pantilie:
@Baltoli We discussed the attribute being available to only equation (function) rules, AFAIK. Has someone requested this feature for rewrite rules as well?

Baltoli:
> @Baltoli We discussed the attribute being available to only equation (function) rules, AFAIK. Has someone requested this feature for rewrite rules as well?

Apologies if so; I may be misremembering - let me try and track down the original conversation again.

Baltoli:
@ana-pantilie Let me know if there is anything else I can do to help you with this issue - it would be good to get it merged, as the frontend support tests passing will depend on this PR.

In particular, you note above that it's necessary to test what's being sent to Z3 - is this from the perspective of the frontend, or the haskell backend?

Thanks so much for your help on this!

ana-pantilie:
@Baltoli Thank you for the help!

> In particular, you note above that it's necessary to test what's being sent to Z3 - is this from the perspective of the frontend, or the haskell backend?

This is just from the perspective of the backend. I will add some tests for this in the backend in a follow-up PR, since I need the updated frontend to do that.

andreiburdusa (approved):

PR #2953, opened 2022-01-12
Add identity Pattern simplifier: Exists.hs

### LINKED ISSUE #2946: Add identity `Pattern` simplifier
We should add an identity `Pattern` simplifier (which doesn't do any actual simplification) attempt to replace the old `Pattern` simplifier with this new one in the code. By running the tests, we can identify any places which don't actually need to call the simplifier, or any places where we only need a small subset of the simplifier's cases implemented.

andreiburdusa:
Part of #2946 

### Scope:
- Replace the current simplifier with an identity simplifier

### Estimate:
- ~~January 14~~
- ~~January 24~~
- February 8
---

###### Review checklist


ana-pantilie (changes requested):
Running the integration tests twice, once normally and once with `--simplifierx` isn't currently enabled, right? If it's not, this PR needs to enable it. We discussed this is necessary once we merge the first experimental feature.

andreiburdusa:
> Running the integration tests twice, once normally and once with `--simplifierx` isn't currently enabled, right? If it's not, this PR needs to enable it. We discussed this is necessary once we merge the first experimental feature.

All the integration tests? I thought we don't want to do that because it would take too much time. I probably misunderstood

ana-pantilie:
> All the integration tests? I thought we don't want to do that because it would take too much time.

It's necessary, otherwise we might break features which we actually use in practice along the way.

ana-pantilie:
Also, this will probably make Jenkins time out. You'll need to increase the time out period on Jenkins for our CI/integration tests.

andreiburdusa:
> Also, this will probably make Jenkins time out. You'll need to increase the time out period on Jenkins for our CI/integration tests.

Shoud I try making nix run the integration tests twice as well. or just Jenkins?

ana-pantilie:
> Shoud I try making nix run the integration tests twice as well, or just Jenkins?

Hmm, let's just run them on Jenkins for now.

kore/src/Kore/Simplify/Exists.hs
```diff
@@ -70,6 +70,9 @@ import Kore.Rewrite.RewritingVariable (
 import qualified Kore.Simplify.AndPredicates as And (
     simplifyEvaluatedMultiPredicate,
  )
+import Kore.Simplify.Pattern (
+    simplifyPatternId,
```
ana-pantilie, line 74: I think it's best if we move this into `MonadSimplify` as well. This way, we have access to it throughout the simplification code and we won't introduce new module dependencies.

ana-pantilie (approved):

JKTKops (approved):

PR #2960, opened 2022-01-19
Parse `non-executable` attribute

Baltoli:
There may be other fixes needed to the code, but this PR seems to address the issues I noted on Slack.


ana-pantilie (approved):
Thank you! I'll make sure that the attribute gets parsed correctly, sorry about this.

PR #2963, opened 2022-02-01
Catch Z3 exception, attempt to restart process

### LINKED ISSUE #2816: Restart Z3 when it crashes

ana-pantilie:
Fixes #2816 
### Scope:
- restarts solver if backend detects that solver crashed
- it only attempts to restart on `push` or `pop` commands, to ensure the safety of retrying the command in the correct context
- see comments for examples of how I manually tested this

### Estimate:
- 7 feb

---

###### Review checklist


ana-pantilie:
Tested with `kore-repl`. After the first `step` completed, I killed the Z3 process. I then ran `step`, and instead of crashing, this PR makes the backend emit a warning, restarts Z3 and allows execution to continue. I then ran `stepf 100` to test the outcome of the proof, and it passes successfully.

```
$ make max-spec.k.repl                                                                                                                                                                                                                                               [15:20:53]
kprovex --haskell-backend-command /Users/anapantilie/RV/kore/.build/kore/bin/kore-repl max-spec.k
Welcome to the Kore Repl! Use 'help' to get started.

Kore (0)> step
Kore (1)> step
kore-repl: [16012643] Warning (WarnRestartSolver):
    (InfoReachability) while simplifying the claim:
    (DebugEvaluateCondition) while evaluating predicate:
    The SMT solver crashed with the following exception:
        Error while communicating with the solver:
            fd:89: hFlush: resource vanished (Broken pipe)
        Solver exit code: -9
    Will restart and reinitialise the solver, attempting to continue execution.
Kore (2)> step
Kore (3)> stepf 100
Proof completed on all branches.
Kore (3)> exit
#Top
```

ana-pantilie:
I was unsure what `initSolver` does, if it only loads the prelude or if it also declares the definition.

https://github.com/kframework/kore/blob/b4315a7454f9d4708a86c30515f1339f32a1d100/kore/src/SMT.hs#L479

That `userInit` there is actually the definition declarations, so it seems like `initSolver` does what we need it to do.

ana-pantilie:
Testing on `kore-exec`. I killed Z3 while the proof was running.

```
$ make test-wrc20.sh.out                                                                                                                                                                                                                                             [16:40:50]
>>> /Users/anapantilie/RV/kore/test/regression-wasm test-wrc20.sh
rm -f test-wrc20.sh.out
./test-wrc20.sh > test-wrc20.sh.out || true
kore-exec: [6507068] Warning (WarnRestartSolver):
    (InfoReachability) while simplifying the claim:
    (DebugEvaluateCondition) while evaluating predicate:
    The SMT solver crashed with the following exception:
        Error while communicating with the solver:
            fd:89: hFlush: resource vanished (Broken pipe)
        Solver exit code: -9
    Will restart and reinitialise the solver, attempting to continue execution.
diff -u test-wrc20.sh.out.golden test-wrc20.sh.out || ( mv test-wrc20.sh.out test-wrc20.sh.out.fail && false )
```

kore/src/Kore/Log/WarnRestartSolver.hs
```diff
@@ -0,0 +1,45 @@
...
+newtype WarnRestartSolver = WarnRestartSolver
+    {solverException :: SolverException}
+    deriving stock (Show)
+
+instance Pretty WarnRestartSolver where
+    pretty WarnRestartSolver{solverException} =
+        Pretty.vsep
+            [ "The SMT solver crashed with the following exception:"
+            , Pretty.indent 4 (pretty $ Exception.displayException solverException)
+            , "Will restart and reinitialise the solver\
```
JKTKops, line 28: Is this similar to other log messages? "Will restart ..." reads a bit strangely; more natural would be "I will restart ..." or "Restarting and reinitialising ..." If this is similar to other messages though, it should stay.

JKTKops (approved):
Looks good; just one piece of feedback.

kore/src/Kore/Log/WarnRestartSolver.hs
ana-pantilie: Done, thanks!

andreiburdusa (approved):

PR #2967, opened 2022-02-09
Parser rewrite

### LINKED ISSUE #2942: Reach performance parity of parser with LLVM backend
Right now, when I run speedscope on the invocation of `kore-match-disjunction` generated by running `krun tests/free-variable-capture.lambda --search --pattern '<k> V:K </k>'` on the k frontend definition `k-distribution/pl-tutorial/1_k/3_lambda++/lesson_2`, it reports that 70% of execution time is spent in `loadDefinitions`. Of that time, 30% of the overall total is spent parsing the kore definition. This amounts to roughly 270 milliseconds since when run without profiling enabled, the application takes 900ms total. This is rather abnormally slow, since the llvm backend is capable of parsing this definition in only 50ms.

There is no reason to imagine that it should take substantially longer to parse something in efficiently-written Haskell than it should take to parse the same code with flex and a recursive descent parser like the LLVM backend uses. I would like us to try to aim for a parser of KORE in the Haskell backend that can parse the same definition in roughly the same amount of time as the LLVM Backend.

This is going to be very necessary if we intend to reach decent performance when running `krun --search --pattern` in the llvm backend. It's not urgent, but fixing it will substantially improve the user experience for such commands, so it would be good to fix as soon as we have the time.

dwightguth:
Fixes #2942

### Scope:

In this PR, we replace the existing Megaparsec-based implementation for parsing KORE files with one based on the lexical analyzer Alex and the parser generator Happy. The exact same AST is generated as before, thus no changes are needed to any code that consumes KORE ASTs. Some changes, however, are required to portions of the unit test suite which relied on the previous interface for parsing and lexing. One minor change is also required in GlobalMain since the new lexical analyzer does not have the same interface as the old one.

### A Note

This is my first PR into this repo. I may have a functional programming background, but I have no experience with writing Haskell code in an industry setting. My code works, and seems to be fast, but is probably stylistically poor in a number of places. Please be patient with me if there are obvious stylistic problems present in the code; I'm happy to address anything of this nature as long as it doesn't have a substantial negative impact on performance.

---

###### Review checklist


dwightguth:
This is ready for review, but there are definitely still some very long lines in the code. Going to need reviewers to comment about how they suggest they be broken up though, because Haskell's whitespace-sensitivity is confusing enough to me that I don't really know how to do it myself in many cases.

kore/test/Test/Kore/Unparser.hs
```diff
@@ -216,21 +218,21 @@ test_parse =
         ]
 
 roundtrip ::
-    (HasCallStack, Unparse a, Eq a, Show a) => Gen a -> Parser a -> Property
+    (HasCallStack, Unparse a, Eq a, Show a) => Gen a -> (FilePath -> Text -> Either String a) -> Property
```
andreiburdusa, line 221: Can we have something like `type Parser = (FilePath -> Text -> Either String a)` in order to avoid this long type signature?

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,314 @@
...
+                alexPosn = alexMove p b,
+                alexChar = b,
+                alexStr =  rest,
+                alexBytePos = n+1})
+
+alexMove :: AlexPosn -> Char -> AlexPosn
+alexMove (AlexPn fp a l c) '\t' =
+    AlexPn fp (a+1) l (c+alex_tab_size-((c-1) `mod` alex_tab_size))
+alexMove (AlexPn fp a l _) '\n' = AlexPn fp (a+1) (l+1) 1
+alexMove (AlexPn fp a l c) _    = AlexPn fp (a+1) l (c+1)
```
ana-pantilie, line 264: All these 1-letter or 2-letter variable names really make things hard to read. We prefer longer, more suggestive variable names in order to make code as readable as possible.

ana-pantilie (changes requested):
I have some questions and some suggestions. Most suggestions are style-based.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,314 @@
...
+  | TokenEOF
+  deriving stock (Eq, Show)
+
+
+alexScanUser :: () -> AlexInput -> Int -> AlexReturn (AlexInput -> Int -> Alex Token)
+
+data AlexInput = AlexInput { alexPosn :: {-# UNPACK #-} !AlexPosn,
+                             alexChar :: {-# UNPACK #-} !Char,
+                             alexStr :: {-# UNPACK #-} !ByteString.ByteString,
+                             alexBytePos :: {-# UNPACK #-} !Int}
```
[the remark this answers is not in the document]
dwightguth: This is how it was implemented in the wrapper code that this particular declaration was copied from.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,314 @@
...
+  deriving stock (Show)
+
+{- | Get the Text argument of a Token whose TokenClass contains such an 
+argument
+-}
+getTokenBody :: Token -> Text
+getTokenBody (Token _ (TokenIdent t)) = t
+getTokenBody (Token _ (TokenSetIdent t)) = t
+getTokenBody (Token _ (TokenString t)) = t
+getTokenBody _ = error "getTokenBody can only be called on tokens which contain a Text field"
```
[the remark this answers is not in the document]
dwightguth: I mean, what's the alternative you're suggesting? There isn't a text field associated with any of the other types of Tokens, so what would it return in those cases? Also, it's only ever called on these cases.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,314 @@
...
+
+data AlexState = AlexState {
+        alex_pos :: !AlexPosn,  -- position at current input location
+        alex_bpos:: !Int,     -- bytes consumed so far
+        alex_inp :: ByteString.ByteString,      -- the current input
+        alex_chr :: !Char,      -- the character before the input
+        alex_scd :: !Int        -- the current startcode
+    }
+
+newtype Alex a = Alex { unAlex :: AlexState -> Either String (AlexState, a) }
```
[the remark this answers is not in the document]
dwightguth: Essentially, yes. I tinkered with using that implementation and a deriving via declaration, but it turned out to be slower than this code, so I left it like this. This is also code that was copied from the Alex wrapper.

kore/src/Kore/Parser/Lexer.x
dwightguth: Same deal here.

kore/src/Kore/Parser/Lexer.x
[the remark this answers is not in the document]
JKTKops: Hmm. I think I told you it was `ExceptT String (State AlexState)` but in fact Ana is correct - it's the other way around. These transformers do not commute with each other and it probably makes a performance difference.

test/parser/test-dv-2.kore.golden
```diff
@@ -1,7 +1,2 @@
-test-dv-2.kore:3:22:
-  |
-3 |     axiom {s} \dv{s}(\or{#String{}}("0", "1")) []
-  |                      ^
-unexpected '\'
-expecting '"'
+test-dv-2.kore:3:22: unexpected token TokenOr
```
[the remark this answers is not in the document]
JKTKops: Happy is not well known for good parser error messages. Just look at GHC's.

One approach I've seen is to parse a "bigger language" than the one you actually care about and then have a pass that validates the parse tree was well-formed, producing better messages there. But we are parsing a language that is intended to be machine-generated so I think the current form in the PR should be okay.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,328 @@
...
+                            Left msg -> Left msg
+                            Right (s', f) -> case unAlex a s' of
+                                               Left msg -> Left msg
+                                               Right (s'', b) -> Right (s'', f b)
+
+instance Monad Alex where
+  m >>= k  = Alex $ \s -> case unAlex m s of
+                                Left msg -> Left msg
+                                Right (s',a) -> unAlex (k a) s'
+  return = App.pure
```
JKTKops, line 255:
```suggestion
```
I believe we only support GHC versions that no longer require a definition of `return`.

If I'm wrong about that, this definition should also have an `INLINE` pragma.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,328 @@
...
+    AlexPn fp (a+1) l (c+alex_tab_size-((c-1) `mod` alex_tab_size))
+alexMove (AlexPn fp a l _) '\n' = AlexPn fp (a+1) (l+1) 1
+alexMove (AlexPn fp a l c) _    = AlexPn fp (a+1) l (c+1)
+
+alexError :: FilePath -> Int -> Int -> String -> Alex a
+alexError fp line column msg = do
+    Alex
+    $ const
+    $ Left
+    $ fp ++ ":" ++ show line ++ ":" ++ show column ++ ": " ++ msg ++ "\n"
```
JKTKops, line 280: This one is a bit hard to read, how about:
```suggestion
alexError fp line column msg = do
    Alex $ const $ Left $ unwords [header, msg] ++ "\n"
    where
      header = List.intercalate ":" [fp, show line, show column]
```
Probably requires an `import qualified Data.List as List`.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,328 @@
...
+      go ('\\' : 'U' : first : second : third : fourth
+                     : fifth : sixth : seventh : eighth : rest) =
+          validate rest $ (((((((digitToInt first * 16 + 
+                                 digitToInt second) * 16 + 
+                                 digitToInt third) * 16 + 
+                                 digitToInt fourth) * 16 + 
+                                 digitToInt fifth) * 16 +
+                                 digitToInt sixth) * 16 +
+                                 digitToInt seventh) * 16 + 
+                                 digitToInt eighth)
```
JKTKops, line 321: In an effort to merge cases and simplify logic here, how about something like this:
```suggestion
      go ('\\' : pre : rest0)
        | Just size <- lookup pre $ zip "xuU" [2,4,8]
        , (body, rest) <- splitAt size rest0
        = runReadS Numeric.readHex body
            >>= validate rest
      where
          runReadS reader str = case reader str of
              [(num, "")] -> pure num
              _other -> Left "invalid hex literal"
              -- probably can be improved by moving function into Alex
```
This will require a `import qualified Data.Numeric as Numeric`. Additionally note that it calls `validate` in the `\xHH` case, which seems like that should be OK. If this causes a big slowdown then that case can be factored back out.

kore/src/Kore/Parser/LexerWrapper.hs
```diff
@@ -0,0 +1,166 @@
...
+
+alexGetStartCode :: Alex Int
+alexGetStartCode = Alex $ \s@AlexState{alex_scd = sc} -> Right (s, sc)
+
+alexSetStartCode :: Int -> Alex ()
+alexSetStartCode sc = Alex $ \s -> Right (s{alex_scd = sc}, ())
+
+alexEOF :: Alex Token
+alexEOF = do
+    AlexInput{alexPosn = p} <- alexGetInput
```
JKTKops, line 98: These should probably all be `INLINE`'d. Not sure how much difference that will make if any.
(Github is being a bit weird about the region I highlighted: I am referring to `alexGetInput` through `alexEOF`.)

kore/src/Kore/Parser/LexerWrapper.hs
```diff
@@ -0,0 +1,166 @@
...
+    alexSetStartCode code
+    action input__ len
+
+token :: (AlexInput -> Int64 -> token) -> AlexAction token
+token t input__ len = return (t input__ len)
+
+-- -----------------------------------------------------------------------------
+-- Basic wrapper, ByteString version
+
+-- | Monad that repeats an operation until a boolean predicate returns True.
```
JKTKops, line 145: Mention that this function is (transitively) only used for testing. If it should be more generally useful, it probably doesn't belong in this module and it could be optimized a bit.

kore/src/Kore/Parser/ParserUtils.hs
```diff
@@ -31,42 +25,8 @@ import qualified Options.Applicative.Types as Options
...
-{- |'parseOnly' is similar to Attoparsec's 'parseOnly'. It takes a parser,
-a FilePath that is used for generating error messages and an input string
-and produces either a parsed object, or an error message.
--}
-parseOnly :: Parser a -> FilePath -> Text -> Either String a
-parseOnly parser filePathForErrors input =
-    case parse parser filePathForErrors input of
-        Left err -> Left (errorBundlePretty err)
-        Right v -> Right v
+newtype Parser a = Parser (FilePath -> Text -> Either String a)
```
JKTKops, line 29:
```suggestion
newtype Parser a
    = Parser { runParser :: FilePath -> Text -> Either String a }
```

kore/test/Test/Kore/Unparser.hs
```diff
@@ -216,21 +218,27 @@ test_parse =
         ]
 
 roundtrip ::
-    (HasCallStack, Unparse a, Eq a, Show a) => Gen a -> Parser a -> Property
+    (HasCallStack, Unparse a, Eq a, Show a) =>
+    Gen a ->
+    (FilePath -> Text -> Either String a) ->
```
JKTKops, line 223: For compatibility with the old parser in this test, import your `Parser` newtype qualified and import the `Parser` type alias from `Kore.Parser.Parser`. Use that type synonym here, and qualify the reference to the `Parser` constructor below.

kore/test/Test/Kore/Unparser.hs
```diff
@@ -216,21 +218,27 @@ test_parse =
...
 roundtrip generator parser =
     Hedgehog.property $ do
         generated <- Hedgehog.forAll generator
-        parse' parser (unparseToText generated) === Right generated
+        parse' (Parser parser) (unparseToText generated) === Right generated
 
 unparseParseTest ::
-    (HasCallStack, Unparse a, Debug a, Diff a) => Parser a -> a -> TestTree
+    (HasCallStack, Unparse a, Debug a, Diff a) =>
+    (FilePath -> Text -> Either String a) ->
```
JKTKops, line 232: Use the `Parser a` type synonym here too.

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,452 @@
...
+    equals               { Token _ TokenEquals }
+    in                   { Token _ TokenIn }
+    next                 { Token _ TokenNext }
+    rewrites             { Token _ TokenRewrites }
+    dv                   { Token _ TokenDomainValue }
+    leftAssoc            { Token _ TokenLeftAssoc }
+    rightAssoc           { Token _ TokenRightAssoc }
+    ident                { Token _ (TokenIdent _) }
+    setIdent             { Token _ (TokenSetIdent _) }
+    string               { Token _ (TokenString _) }
```
JKTKops, line 124:
```suggestion
    ident                { Token _ (TokenIdent $$) }
    identLoc             { Token _ (TokenIdent _) } -- variation which keeps AlexPn
    setIdent             { Token _ (TokenSetIdent $$) }
    string               { Token _ (TokenString $$) }
```
Now using `$n` in an action to refer to these tokens refers to the text contents. I think this allows you to delete `getTokenBody` entirely.

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,452 @@
...
+      : Id { SortVariableSort (SortVariable $1) }
+      | Id SortsBrace { SortActualSort (SortActual $1 $2 ) }
+
+Pattern :: {ParsedPattern}
+	 : SomeVariable { embedParsedPattern $ VariableF $ Const $1 }
+         | ApplicationPattern { $1 }
+         | StringLiteralPattern { $1 }
+
+ElementVariable :: {Variable (ElementVariableName VariableName)}
+	         : ident ':' Sort { Variable{variableName = ElementVariableName{unElementVariableName = mkVariableName $1}, variableSort = $3} }
```
JKTKops, line 213: @ana-pantilie is there a better way to construct these `Variable`s?

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,452 @@
...
+
+NePatterns :: {[ParsedPattern]}
+	    : '(' PatternList ')' { reverse $2 }
+
+PatternList :: {[ParsedPattern]}
+             : Pattern { [$1] }
+	     | PatternList ',' Pattern { $3 : $1 }
+
+{
+
```
JKTKops, line 309:
```suggestion
type Parser a = FilePath -> Text -> Either String a
```
Export this type alias and use it elsewhere as well (including tests). It can also clean up the following functions, see below.

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,452 @@
...
+{- | Helper function for parsing a particular NonTerminal in the grammar. The
+function specified by the %name directive for that NonTerminal should be passed
+as the first argument to the function. -}
+parseNonTerminal :: Alex a -> FilePath -> Text -> Either String a
```
JKTKops, line 313:
```suggestion
parseNonTerminal :: Alex a -> Parser a
```

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,452 @@
...
+parseNonTerminal a fp s = runAlex fp (Text.encodeUtf8 s) a
+
+-- Functions for parsing specific NonTerminals.
+
+parseAliasHead :: FilePath -> Text -> Either String Alias
```
JKTKops, line 318:
```suggestion
parseAliasHead :: Parser Alias
```
etc.

JKTKops (changes requested):
I spent a couple hours looking over this, I will take another look again later to see if I notice anything else.

kore/src/Kore/Parser/Lexer.x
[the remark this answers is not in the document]
dwightguth: well, we could do that, of course. But it's worth noting that the Text associated with a TokenString isn't the same thing as the image of the token. It gets unescaped first. So if we wanted a function that converted a TokenClass to a textual representation of that token in KORE, this function cannot be it.

kore/src/Kore/Parser/Lexer.x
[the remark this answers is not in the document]
dwightguth: I mean, the code that I tried out was `StateT AlexState (Either String) a` because what you suggested did not compile so I assumed that you had made that mistake. But as I said, it was slower than this code.

kore/src/Kore/Parser/Lexer.x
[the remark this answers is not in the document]
ana-pantilie: Let's just leave it as-is. We haven't had many cases in the past where we had to modify the parser, so it should be safe enough to keep it.

ana-pantilie (approved):
I'm approving so that I don't block you from merging it when the other reviewers are also happy with it.

I only have one thing to ask, please add a comment for each piece of code which was taken from another source, and any reasoning behind it.

kore/test/Test/Kore/Unparser.hs
dwightguth: I don't understand what you're suggesting here. Can you be more explicit? Specifically, you mention a Parser type alias in Kore.Parser.Parser, but no such type alias exists currently.

kore/src/Kore/Parser/Parser.y
```diff
@@ -0,0 +1,369 @@
...
+          | '(' ')' { [] }
+
+NePatterns :: {[ParsedPattern]}
+	    : '(' PatternList ')' { reverse $2 }
+
+PatternList :: {[ParsedPattern]}
+             : Pattern { [$1] }
+	     | PatternList ',' Pattern { $3 : $1 }
+
+{
```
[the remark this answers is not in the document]
dwightguth: I mean, I guess in theory we could create function helpers for each action in the grammar, but it seems to me that would really not make the code much clearer at all.

test/parser/test-string-4.kore.golden
```diff
@@ -1,6 +1,2 @@
-test-string-4.kore:1:11:
-  |
-1 | [key{}("\07777")]
-  |           ^
-expecting escape sequence
+test-string-4.kore:1:10: unexpected character '0'
```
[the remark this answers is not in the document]
dwightguth: I mean, it's actually correct where it's pointing now; 0 is not a valid character following \ inside a string. It was wrong before.

kore/src/Kore/Parser/Lexer.x
```diff
@@ -0,0 +1,325 @@
...
+-- Alex wrapper code.
+--
+-- Taken and modified from public domain code that is part of Alex:
+-- https://github.com/haskell/alex/blob/3.2.6/templates/wrappers.hs
+--
+-- Unlike the code in LexerWrapper.hs, this code must live in the same module
+-- as the scanner itself, so it was moved from that file to this one.
+
+
+alexScanUser :: () -> AlexInput -> Int -> AlexReturn (AlexInput -> Int -> Alex Token)
```
JKTKops, line 222: Should be commented that this function will be generated by Alex and the type signature is needed to fix the `()` type.

kore/test/Test/Kore/Repl/ParserTest.hs
```diff
@@ -0,0 +1,179 @@
+module Test.Kore.Repl.ParserTest (
```
JKTKops, line 1: What is this module? Looks like old Megaparsec things.

JKTKops (commented):
Nothing major, just a couple of questions. I'm good on the changes.

kore/test/Test/Kore/Repl/ParserTest.hs
dwightguth: The repl tests used the same test harness as the parser tests, but they tested something completely different from terms, so it wasn't really possible to use the new version of the parser test harness to test the repl. Hence, we copied the old parser test harness and moved it to this directory.

JKTKops (approved):

ana-pantilie:
@dwightguth Is this ready to be merged? Feel free to add automerge whenever that is.

PR #2981, opened 2022-02-18
Enable All Additional GHC2021 Extensions

JKTKops:
### Scope:
The recent acceptance and implementation of `-XGHC2021` in GHC 9.2 automatically enables many of the extensions we are already using. We have agreed that now is not the time to switch to GHC 9 as we are working on the simplifier. However, in preparation to make that switch later, this PR enables all of the GHC2021 extensions which were not already enabled.

The vast majority of our work should see no difference. `TupleSections` is now enabled, which can be convenient in some places. `PolyKinds` will very rarely require a kind annotation - two such instances are fixed by this PR in the `Debug` module.

A main motivation for this preparation is to enable `ImportQualifiedPost`. For example, instead of
```hs
import qualified Kore.Simplify.Simplify as Simplifier
```
you may now write
```hs
import Kore.Simplify.Simplify qualified as Simplifier
```

This vertically aligns module names in imports, making them much easier to scan quickly. If everyone agrees, we should add a note to the style guide to prefer this syntax. A PR modifying all existing files would be very big and I'm not sure that's a good idea, however if we want to take that route we could also enable the warning `-Wprepositive-qualified-module`, which warns on the old syntax.

### Estimate:

Feb 18

---

###### Review checklist


ana-pantilie (commented):
Thank you for looking into this.

I'd imagine, though, that we could automatically change all the files in the repo with this new import syntax. It shouldn't be too hard to just match on the qualified imports and switch things around.

What do you think? Also, @andreiburdusa, do you think this syntax is preferable?

PR #2988, opened 2022-03-02
Add note on LLVM 12

Baltoli:
This documents a workaround for some recent issues experienced on macOS.


ana-pantilie (approved):
Thank you!

PR #3002, opened 2022-03-17
Serialize initialized definition during compilation and deserialize it when running kore-exec

### LINKED ISSUE #2943: Separate out the definition loading and definition verifying routines
Right now, when I run speedscope on the invocation of `kore-match-disjunction` generated by running `krun tests/free-variable-capture.lambda --search --pattern '<k> V:K </k>'` on the k frontend definition `k-distribution/pl-tutorial/1_k/3_lambda++/lesson_2`, it reports that 70% of execution time is spent in `loadDefinitions`. Of that time, 40% of the overall total is spent loading and verifying the kore definition. This amounts to roughly 360 milliseconds since when run without profiling enabled, the application takes 900ms total.

Unfortunately, much of this time is likely being spent doing validation and verification of the kore definition (I assume; it seems to me that indexing the definition ought to be very fast and take less than 100ms on a definition of this size, but I don't fully understand the data structures the haskell backend uses internally). Assuming my assumption is correct, we ought to be able to do away with 100% of that time. This is because kompile already invokes the Haskell backend to validate and verify the kore definition, so it doesn't need to happen again each time that we invoke `kore-exec` or `kore-match-disjunction`.

However, currently these two functions (verifying and loading) are very tightly intercoupled. We ought to instead have separate routines for loading and verifying definitions. The verification ought to only happen at kompile time, whereas the loading ought to be extremely quick.

dwightguth:
Fixes #2943 

### Estimate:

---

###### Review checklist


dwightguth:
I am closing this as we have seen much better performance data with Data.Compact and the change is sufficiently different to warrant its own PR.

PR #3005, opened 2022-03-21
All-path reachability: change wAF from [w] to <w>

ana-pantilie:
As per @traiansf 's suggestion, this PR switches `[w]` to `<w>`. The reason for this is that `[]` suggests a global property, whereas weak always finally means that at some point, on all finite paths, a property holds.

---

###### Review checklist


andreiburdusa (approved):

traiansf:
One of the reasons we used `[w]` was because we were using `◇w` for the one[-path reachability proofs](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2018-11-08-One-Path-Reachability-Proofs.md).

Hence, although I fully support using `<w>`, you could consider adding a note saying that although both `<w>` and  `◇w` refer to a `finally` property and hence the LTL-like diamond operator intuition, they differ by quantifier on traces associated to them.

On the other hand, one could also have an intuition for using `[w]` by noting that in modal logic, as Laurentiu reminded us, diamond corresponds to posibility (it would thus be our one-path next) while box corresponds to necessity (it would thus be our all-path next).  And in dynamic logic, one would use the operators like:

* `<tau> phi` to say that  there exists a path using the sequence of actions `tau` to reach a state in which `phi` holds
* `[tau] phi` to say that on all paths using the sequence of actions `tau` one would reach a state in which `phi` holds.

So... I don't know anymore :-D

PR #3017, opened 2022-03-29
Use Data.Compact to serialize K definition at compilation time

### LINKED ISSUE #2943: Separate out the definition loading and definition verifying routines

dwightguth:
Fixes #2943 

### Scope:

This change makes use of the Data.Compact library to serialize the initialized KORE definition ahead of time and then load it into memory at the start of kore-exec for symbolic execution and search. Nothing necessarily prevents us from using it for other execution modes as well, but I figured we would start with the basic case and we can add on more cases in follow-up changes if we so desire.

Note: the integration tests do not currently test this code path! It is only exercised if krun passes the path to a binary serialized file instead of the path to definition.kore. this requires frontend change that cannot be merged until this change is in master. Thus, the integration tests passing with this change indicates that the old code path still works. I have manually tested the new code path with the modified frontend and it did pass the integration tests.

**DO NOT MERGE** unless the head commit of this branch as of merge time passes the integration tests manually with the modified k frontend in addition to all checks being green.

### Estimate:

---

###### Review checklist


kore/app/exec/Main.hs
```diff
@@ -620,59 +635,87 @@ mainDispatch = warnProductivity . mainDispatchWorker
                 then koreBmc localOptions proveOptions
                 else koreProve localOptions proveOptions
         | Just searchOptions <- koreSearchOptions =
-            koreSearch localOptions searchOptions
+            koreSearch exeLastModifiedTime localOptions searchOptions
+        | True <- serialize =
+            koreSerialize localOptions
```
ana-pantilie, line 640: So if you give it the serialize flag it will only serialize the definition.

kore/app/exec/Main.hs
```diff
@@ -620,59 +635,87 @@ mainDispatch = warnProductivity . mainDispatchWorker
...
     Main (KFileLocations, ExitCode)
-koreSearch LocalOptions{execOptions, simplifierx} searchOptions = do
+koreSearch exeLastModifiedTime LocalOptions{execOptions, simplifierx} searchOptions = do
     let KoreExecOptions{definitionFileName} = execOptions
-    definition <- loadDefinitions [definitionFileName]
     let KoreExecOptions{mainModuleName} = execOptions
-    mainModule <- loadModule mainModuleName definition
+    let KoreExecOptions{koreSolverOptions} = execOptions
+    SerializedDefinition{serializedModule, lemmas, locations} <-
+        deserializeDefinition simplifierx koreSolverOptions definitionFileName mainModuleName exeLastModifiedTime
```
ana-pantilie, line 658: If the definition wasn't serialized before, it will serialize it by default?

ana-pantilie (commented):
Can you explain how this will integrate with the frontend? I just want an example use case scenario, just so I understand how all parts fit together, this is what I've been having most trouble with when reviewing.

kore/app/share/GlobalMain.hs
```diff
@@ -487,7 +541,112 @@ mainParse parser fileName = do
...
+    solverOptions
+    definitionFilePath
+    mainModuleName
+    exeLastModifiedTime =
+        do
+            bytes <- ByteString.readFile definitionFilePath & liftIO
+            let magicBytes = ByteString.drop 8 $ ByteString.take 16 bytes
+            let magic = Binary.decode @Word64 magicBytes
+            case magic of
+                0x7c155e7a53f094f2 -> do
```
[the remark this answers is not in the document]
dwightguth: This is the magic number that the Data.Compact package emits at the top of files that it serializes to in order to do a basic integrity check. They do not export that symbol so we had to recreate it here. This is used to detect whether the input is a serialized definition or a textual kore definition.

kore/app/exec/Main.hs
dwightguth: Well, a lot of the logic needed to do this was part of the Kore.Exec and kore/app/exec/Main files, so it seemed more appropriate to put it here than to do a huge refactoring of where a whole bunch of code lived. As it was, I already had to move around a fair bit.

kore/app/exec/Main.hs
dwightguth: I'm not entirely sure what you're asking. the `deserializeDefinition` function returns a `SerializedDefinition` object which contains everything needed to do execution or search with the haskell backend. What it does is detect whether the file it's reading is a serialized binary file or a textual kore definition and if it is the former, it simply deserializes it, and if it is the latter, it does all the work done by `kore-exec --serialize`, but does not actually write anything to disk, merely using the result to do execution.

dwightguth:
Here is the changes that will be made to the frontend after this is merged: https://github.com/runtimeverification/k/pull/2551

As you can see in that diff, all the frontend is really doing differently is calling `kore-exec --serialize` instead of `kore-parser`. Since `kore-exec --serialize` already parses and verifies the definition, all we're doing is adding the extra step of compacting and serializing it when that's done.

As far as performance, this takes the time to execute a single step of the haskell backend on the c semantics down from like 40 seconds to 7 seconds. That's a pretty substantial improvement. It also sizeably increases the GC throughput in the backend while executing because it puts the initialized KORE definition in what's called a compact region, which essentially means, among other things, that it's not being traced by the garbage collector.

You are correct that it probably does increase the memory usage somewhat, because one consequence of being a compact region is that any pointer into any part of a compact region keeps the entire region alive and prevents garbage collection. But overall as you can see, the difference is relatively minor, and the actual execution time improves substantially. compilation does take longer, but in doing so it saves substantial amounts of time on the fast path, so overall you benefit greatly unless you are compiling a lot of definitions and then never using them. So overall it is a substantial win.

ana-pantilie (approved):
Looks good, just one small request: where I asked specific questions about added/modified code, please add documentation for that as comments in the code. Thank you.

PR #3020, opened 2022-03-30
fix off by one error in lexer

dwightguth:
### Scope:

The unit tests failed on one of my other PRs because the random generation of inputs in the Unparse tests happened to generate a 0x10ffff unicode character, which was erroneously being forbidden by the lexer due to this off-by-one error. Should be a very simple fix to review and merge.

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Parser/Lexer.x
+++ b/kore/src/Kore/Parser/Lexer.x
@@ -295,7 +295,7 @@ unescape t =
   let dropQuotes str = Text.take (Text.length str - 2) . Text.drop 1 $ str
       escape rest c = either Left (\s -> Right (c : s)) $ go rest
       validate rest c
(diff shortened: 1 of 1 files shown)


JKTKops (approved):

ehildenb:
@ana-pantilie @JKTKops and @dwightguth we should get in the habit of turning property tests, which can be flaky (as demonstrated by this fix here), into unit/regression tests when they uncover bugs.

Property tests are good, but are no substitute for unit or regression tests. They may only exercise the failing case once every few months even, so it's important that when it does exercise bad behavior, we explicitly turn that into a test. Also, new features can add property tests, that's good, but they should also be adding explicit unit tests of each code-path, not just property tests where we _assume_ it will achieve good coverage.

PR #3037, opened 2022-04-18
Warn if translation of an SMT lemma fails

dwightguth:
### Scope:

Currently if the translation to SMT of an SMT lemma fails, it is silently discarded and not used. Obviously users would probably prefer to know that this happened as it usually indicates a bug of some kind. This PR essentially consists of a refactoring of the Translator monad to use ExceptT instead of MaybeT in order to allow such warnings to be reported.

### Estimate:

---

###### Review checklist


dwightguth:
This feature would be nice to have, but it's no longer blocking me, so I'm going to close the PR. I will leave the branch up and someone else can investigate the test failure and take the task for themselves if they want to work on this.

PR #3042, opened 2022-04-21
Add support for user-defined \\equals simplification

### LINKED ISSUE #3040: Apply simplification rules to \\equals
Right now simplification rules only apply on terms with symbol applications at the top or with \\ceil at the top. We need to extend this feature to other ML connectives as well, starting with \\equals which is currently blocking the c semantics.

ana-pantilie:
Fixes #3040

### Estimate:

---

###### Review checklist


dwightguth (approved):
This solves my issue in the C semantics. There was some kind of weird other issue that was interfering with the simplification rule being applied in some cases, but I no longer seem to be able to reproduce whatever it was. I figure we can address it later if we run into it again.

The code looks okay to my admittedly novice eye as well. I would recommend Max reviewing the code as well before merging, but this is good to go as far as I'm concerned.

JKTKops (approved):

PR #3053, opened 2022-05-10
Re-export Control.Monad from Logic

### LINKED ISSUE #3052: Compatibility with logict-0.8.0.0
`logict` has [recently updated to 0.8.0.0](https://github.com/Bodigrim/logict/blob/master/changelog.md) which comes with a breaking change: `Control.Monad` and `Control.Monad.Trans` are no longer re-exported from `Control.Monad.Logic`. We should adjust our `Logic` module for compatibility.

JKTKops:
Fixes #3052

### Scope:

Re-exports `Control.Monad.Trans` and `Control.Monad` from `Logic`.

### Estimate:

Done.

---

###### Review checklist

### DIFF
--- a/kore/src/Logic.hs
+++ b/kore/src/Logic.hs
@@ -13,9 +13,9 @@ module Logic (
 ) where
 
 import Control.Applicative
+import Control.Monad
 import Control.Monad.Logic
 import Control.Monad.Trans
-import Control.Monad
 import Prelude
 
 gather :: MonadLogic m => m a -> m [a]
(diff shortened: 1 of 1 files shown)


goodlyrottenapple (approved):
LGTM, tho not too keen on re-exporting modules in the first place...

JKTKops:
> LGTM, tho not too keen on re-exporting modules in the first place...

Me neither, but I wanted to keep the module's interface the same without using CPP to support multiple versions of the package. The other option would be a version bound, but since `logict` is a critical package I think it's preferable to support the latest version if we can.

Issue #3056, opened 2022-05-13
kore-repl arithmetic underflow


goodlyrottenapple:
The current version of `kore-repl`

```
kompile --version
K version:    5.3.0
Build date:   Tue May 10 09:50:42 BST 2022

kore-repl --version
Kore version 0.60.0.0
```

crashes on the following input:

First, create `foo.k`

```
module FOO

syntax Foo ::= "a" | "b" | "c" | "d" | "e"

configuration <k> $PGM:Foo </k>

rule <k> a => b ... </k>

rule <k> a => c ... </k>

rule <k> b => e ... </k>

rule <k> c => d ... </k>

rule <k> d => b ... </k>

endmodule
```

and `spec.k`

```
requires "foo.k"

module SPEC
  imports FOO

  claim <k> a => e ... </k>
  
endmodule
```

then `kompile --backend haskell foo.k` and launch the repl `kprove --debugger spec.k`. Running `stepf 2` followed by `stepf 1` crashes the REPL with

```
kore-repl: [31022715] Error (ErrorException):
    arithmetic underflow
``` 

See the full printout of the REPL session:

```
Welcome to the Kore Repl! Use 'help' to get started.

Kore (0)> stepf 2
Kore (0)> stepf 1
kore-repl: [8790109] Error (ErrorException):
    arithmetic underflow
Created bug report: kore-repl.tar.gz
```

goodlyrottenapple:
As suspected, the excpetion seems to be thrown from an arithmetic operation on `GHC.Natural`. WIth profiling enabled and using the `+RTS -xc` flag, getting the following stack trace on crash:

```haskell
Kore (0)> stepf 1
*** Exception (reporting due to +RTS -xc): (THUNK_STATIC), stack trace:
  GHC.Natural.CAF
  --> evaluated by: Kore.Repl.State.getInterestingBranchingNode.sucResults,
  called from Kore.Repl.State.getInterestingBranchingNode,
  called from Kore.Repl.Interpreter.proveStepsF.targetNode,
  called from Kore.Repl.Interpreter.proveStepsF,
  called from Kore.Repl.Interpreter.replInterpreter0.command,
  called from Kore.Repl.Interpreter.replInterpreter0.evaluateCommand,
  called from Kore.Repl.Interpreter.replInterpreter0,
  called from Kore.Repl.Interpreter.replInterpreter,
  called from Log.fmap,
  called from SMT.fmap,
[...]

PR #3057, opened 2022-05-13
Fix for an aritmetic underflow bug in kore-repl

goodlyrottenapple:
This fixes an underflow exception thrown when `getInterestingBranchingNode 0 graph node` gets called, as this would trigger the underflow on `n - 1` in the computation of `sucResults`.

Fixes [#3056](https://github.com/runtimeverification/haskell-backend/issues/3056)


ana-pantilie (approved):
I think I've seen this before and haven't been able to reproduce it. GJ on fixing it!

JKTKops (approved):

PR #3075, opened 2022-05-30
Nix flake

goodlyrottenapple:
This is part of https://github.com/runtimeverification/k/issues/2607 creating a nix flake expression for the haskell-backend 
 (based on https://github.com/mlabs-haskell/apropos-tx/blob/main/flake.nix)

The new flake is independent of the original nix derivations and should eventually replace the current nix derivation. Atm, it is lacking the following:

* tests
* hls in the dev shell 

This PR also includes `nixfmt`-ed nix files.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,72 +1,59 @@
-{ profiling ? false
-, release ? false
-, threaded ? !profiling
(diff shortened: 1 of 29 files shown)


calin1304 (approved):
LGTM

PR #3078, opened 2022-05-31
Add option --disable-stuck-check <n>

### LINKED ISSUE #3074: kprove fails to prove claim where circularity can apply at first step (but can still take a semantic step)
I think this claim should succeed.
First, the claim cannot be applied as a circularity, so a semantic step must be taken, reaching `<k> foo(N - 1) </k> #And N > 0`.
This  partially unifies with the RHS of the claim, leaving `<k> foo(N-1) </k> #And N > 0 #And not(N-1 == 0)` to be proved.
This unifies completely with the LHS of the claim, so applying it as a circularity it should succeed.
 

```
$ cat test-helpers.k
module TEST-HELPERS
    imports INT

    syntax KItem ::= foo(Int)
    rule <k> foo(I) => foo(I -Int 1) </k>
endmodule


$ kompile --backend haskell test-helpers.k
[Warning] Compiler: Could not find main syntax module with name
TEST-HELPERS-SYNTAX in definition.  Use --syntax-module to specify one. Using
TEST-HELPERS as default.


$ cat test-spec.k
module TEST-SPEC
    imports TEST-HELPERS

    claim [reaches-zero]:
        <k> foo(N) => foo(0) </k>
      requires N >Int 0

endmodule


$ kprove test-spec.k
kore-exec: [163907] Warning (WarnStuckClaimState):
    The proof has reached the final configuration, but the claimed implication is not valid. Location: /home/user/erc20-verification/haskell-test/test-spec.k:5:9-6:24
  #Not ( {
    N
  #Equals
    0
  } )
#And
  <k>
    foo ( N ) ~> .
  </k>
#And
  {
    true
  #Equals
    N >Int 0
  }
  ```

### PR DESCRIPTION
**Pull request by ana-pantilie**
Fixes #3074

### Estimate:

---

###### Review checklist

### CODE REVIEW

File: kore/app/exec/Main.hs [hunk 1]
```diff
@@ -287,6 +288,7 @@ data KoreExecOptions = KoreExecOptions
     , bugReportOption :: !BugReportOption
     , maxCounterexamples :: Natural
     , serialize :: !Bool
+    , stuckCheck :: !StuckCheck
```
ana-pantilie, line 291: I should add this to `KoreProveOptions` instead.

kore/app/repl/Main.hs
```diff
@@ -324,3 +326,6 @@ mainWithOptions LocalOptions{execOptions, simplifierx} = do
 
     smtPrelude :: SMT.Prelude
     smtPrelude = prelude smtOptions
+
+    replStuckCheck :: StuckCheck
```
jberthold, line 330: (nit) maybe 
```suggestion
    replStuckCheck :: Claim.StuckCheck
```
so no additoinal import is needed?

jberthold (approved):
LGTM

PR #3092, opened 2022-06-09
And commands in repl

goodlyrottenapple:


ana-pantilie:
Reopen if we ever want this feature in master.

PR #3097, opened 2022-06-13
fix simplification indexing for rules with variables under \equals

dwightguth:
### Scope:

Previously, if you had a simplification rule whose left-hand side was \equals and one of the children of the \equals was a variable, it would not correctly find that axiom in the index if the term being simplified was an equality between two application patterns.

I have not been able to minimize the example that I was using for testing; when I did, the simplifier was able to solve it even without this change, and I'm not entirely sure what the issue is. If someone wants to help by writing a test for this, that would be great.

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Rewrite/Axiom/EvaluationStrategy.hs
+++ b/kore/src/Kore/Rewrite/Axiom/EvaluationStrategy.hs
@@ -70,15 +70,6 @@ import Pretty (
  )
 import Pretty qualified
 
--- |Describes whether simplifiers are allowed to return multiple results or not.
-data AcceptsMultipleResults = WithMultipleResults | OnlyOneResult
-    deriving stock (Eq, Ord, Show)
-
--- |Converts 'AcceptsMultipleResults' to Bool.
-acceptsMultipleResults :: AcceptsMultipleResults -> Bool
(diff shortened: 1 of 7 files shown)


JKTKops (approved):

dwightguth:
I put this back in draft form because it turns out this solution is incomplete; it still neglects axioms where you have an equality for which one child is a pattern that is not currently a member of the AxiomIdentifier type.

dwightguth:
This is not a full fix. The issue described in #3098 still exists, but we extend the cases we support to include domain values. This is enough of an improvement that I am able to get all the simplification rules I need to work currently, but we really ought to eventually fix this properly. I would like to get this merged as is though because I have a lot of stuff I need to get merged and I don't want to get bogged down in this one particular thing. This is a strict improvement over what was there before, so I think we should evaluate it as such.

ana-pantilie (commented):
Can you please add some tests which cover the cases added in this PR? I believe https://github.com/runtimeverification/haskell-backend/blob/master/kore/test/Test/Kore/Simplify/Integration.hs is an appropriate place to do so, see how we use `evaluateWithAxioms` to run the simplifier with test equations. Let me know if you have any trouble with this!

PR #3105, opened 2022-06-17
[nix] Added source filtering in flake.nix based on gitignore + removing all nix files

goodlyrottenapple:


ana-pantilie (approved):
The CI failure is unrelated, you should just re-run that.

PR #3118, opened 2022-06-24
Extend new unification algorithm to equals simplification

dwightguth:
### Scope:

This PR follows on from #3115 and modifies the simplifier to use the new unification algorithm when simplifying \equals. It still uses the old algorithm for simplifying \and.

Preliminary tests suggest that this reduced the time spent executing the C semantics smoke test from 16 minutes to 11 minutes. Essentially no meaningful amount of execution time remains attributed to the old unification algorithm. The amount of time spent in the new unification algorithm was approximately 2.9%. This compares to roughly 11.6% in matching and 6.7% in ceil simplification, so time would be better spent there at this point.

### Estimate:

---

###### Review checklist


dwightguth:
Closing; this does not actually make anything faster because the time being spent in equals simplification is actually almost entirely time spent in the \ceil simplifier. We need to target that component instead if we want to see performance improvements.

PR #3128, opened 2022-06-29
Support for functional claims

### LINKED ISSUE #3010: Haskell backend supports functional claims directly

JKTKops:
Fixes #3010

### Scope:

(todo)

### Estimate: Sometime next week, probably (week of july 4)

---

###### Review checklist


JKTKops:
One open question is about the name of these things. They aren't exactly restricted to functional claims - perhaps "non-reachability" or "equational" claims would be a better name.

JKTKops:
Since I will be leaving RV for grad school soon, Ana and I have decided to have me step away from this task as it is turning out to be significantly larger than we expected. My last commit gathered several scattered comments I had left for myself about particular difficulties into 3 main places. These 3 comments describe the major difficulty with implementing equational claims in the existing architecture, the attempts I made to work around them, and my suggested larger-scale refactor to fix them and help prevent them in the future.

These are as follows.

1. The `Claim` [class](https://github.com/runtimeverification/haskell-backend/blob/e416be58dba593fedff4216b077894d32dce6bba/kore/src/Kore/Claim/Claim.hs#L171)

```
The Claim class itself is the source of some of the complication
with EquationalClaims. The 'Claim' class is tightly tied to the concept
of a _rewriting_ claim, which is simply too specific to cover other types
of claims that the backend might want to handle.

I recommend separating the concerns of the 'Claim' class into the concerns
of a general 'Claim' and the concerns of a 'RewritingClaim'. I think general
claims really only need functionality for initial simplification, (incremental)
proving, and displaying stuck claims back to the user.

As a result, the 'equationalX' strategies below should be able to be removed.
They are temporary hacks that were needed to get simplification of EquationalClaims
working with the existing infrastructure.
```

2. [The representation](https://github.com/runtimeverification/haskell-backend/blob/e416be58dba593fedff4216b077894d32dce6bba/kore/src/Kore/Claim/EquationalClaim.hs#L46) of `EquationalClaim`s themselves also posed a difficulty. The internals of an `EquationalClaim` are _expected_ to be different before and after simplification (a `Pattern` before, an `OrPattern` after). Because the current `Claim` class cannot express this, we had to settle to use `OrPattern` in both cases. A refactor that separates the concerns of the `Claim` class should make it much more feasible to use a separate proving pipeline on `EquationalClaim`s where this is not a concern at all.

```
[The representation] is necessary due to the current inflexibility of the
simplifiers in the 'Claim' class. Everything is assumed to always return
an OrPattern and there's no way to indicate that we have a regular
Pattern before simplification, but an OrPattern after. We do expect
equational claims to be able to branch, so the OrPattern is relevant.
```

3. Finally, the big hitter. The code assumes _everywhere_ that `claim ~ ClaimPattern`, and this assumption is reified by the `lensClaimPattern` [lens](https://github.com/runtimeverification/haskell-backend/blob/9ca7cf282b139bd8ca5039164da5a629eaadd568/kore/src/Kore/Claim/SomeClaim.hs#L165). This PR's entire goal is to break the assumption that this lens implements, which is a way to zoom in on a `SomeClaim` and get a `ClaimPattern`. My comment describes the pervasiveness, my attempted workarounds, and the suggested refactor, which culminates in deleting the lens entirely:

```
The 'Equational' case of this lens is by far the biggest problem with
getting equational claims to work properly.

There's been an assumption within the codebase that whenever we have
an instance 'Claim claim', we have (morally) that 'claim ~ ClaimPattern'.
This has generally been true, but it isn't a rule of the 'Claim' class and
'Equational' claims are _not_ represented by 'ClaimPattern's. Indeed,
'SomeClaim' is usually used directly instead of 'Claim claim', which is
a bad sign in terms of generality.

This lens is used all over the place. Simplification generally tries to
take apart reachability claims with this lens and apply functions to the
'ClaimPattern' directly. This bakes the assumption into places it shouldn't
be, such as 'Kore.Exec.simplifySomeClaim'. It's also ocassionally used
just to extract sort information, such as 'app/kore/exec/Main.koreProve'
(see line 803).

I've tried leaving this lens and adjusting the use sites to make everything
work without a major refactor but it became too pervasive. I also speculated
that it might be possible to use a partial 'EquationalClaim' <-> 'ClaimPattern'
conversion to trick the existing code into working, but it sounds brittle
and Ana and I decided against spending additional effort on that at this
time.

Thus, the refactor here will end up being fairly involved. The 'Claim' class
should get new functions for simplifying claims "using only mathcing logic
rules." Likely also useful would be functions for extracting sort information
from a claim, like @Claim claim => claim -> Sort@. To prevent similar information
leaks to the one that causes this problem, 'SomeClaim' should be replaced
with the existential
data SomeClaim where
    MkSomeClaim :: Claim claim => claim -> SomeClaim

Necessary information for reporting stuck claims back to the user (functionality
for displaying claims, mainly) should probably be added as a superclass on
'Claim' or as a secondary constraint in 'MkSomeClaim'.
With all of that, this lens should be able to be deleted entirely, replaced
by the generalized 'Claim' class and 'SomeClaim'.
```

When this PR is eventually taken over by someone else, please feel free to ask me questions by pinging me here. I _should_ see them, but I have my email set up to forward GitHub notifications from RV repositories to my RV address, and I'm not sure what the status of my RV address will be at that time. I may also be accessible by slack or at my student email address which I will provide to Ana.

PR #3131, opened 2022-07-04
Jb/2626 json kore hs backend

### LINKED ISSUE #2626: Update SMT encoding documentation
The SMT encoding documentation needs to be updated:

- [upstream](https://github.com/kframework/k/blob/master/pending-documentation.md#smt-lemma-lemma-and-trusted-attributes), and
- in the [design decisions](https://github.com/kframework/kore/blob/master/design-decisions/2018-11-05-Translating-predicates.md) by making a document which reflects the _current state_ of SMT encoding.

jberthold:
Fixes #2626 

### Scope:
* Defines a json-schema for kore patterns;
* defines simple data types representing the json format for Kore patterns;
* provides conversion functions `KoreJson -> ParsedPattern` and `Pattern VariableName ann -> KoreJson`;
* json parsing and marshaling are automatic, providing
  `ByteString -> ParsedPattern` and `Pattern VariableName ann -> ByteString`

### Estimate:
2 more dev.days 

---

###### Review checklist

### DIFF
--- a/kore/src/KoreJson.hs
+++ b/kore/src/KoreJson.hs
@@ -7,7 +7,6 @@ module KoreJson (
     ) where
 
 --    KorePattern(..)
-
 import Data.Aeson as Json
 import Data.Aeson.Encode.Pretty as Json
 import Data.ByteString.Lazy (ByteString)
--- a/kore/src/KoreJson.hs
+++ b/kore/src/KoreJson.hs
@@ -17,11 +17,11 @@ import GHC.Generics -- FIXME switch to TH-generated Json instances
 import Kore.Attribute.Attributes (ParsedPattern)
 import Kore.Internal.Pattern (Pattern)
 import Kore.Syntax.PatternF (PatternF (..))
-
 -- import Kore.Internal.TermLike.TermLike (TermLikeF(..))
 import Kore.Parser (embedParsedPattern)
 import Kore.Sort qualified as Kore
-import Kore.Syntax.Variable (ElementVariableName (..), SetVariableName (..), SomeVariable (..), SomeVariableName (..), Variable (..), VariableName (..))
+import Kore.Syntax qualified as Kore
+import Kore.Syntax.Variable (ElementVariableName (..), SetVariableName (..), SomeVariableName (..), Variable (..), VariableName (..))
 import Prelude.Kore as Prelude
 
 {- | Json representation of Kore patterns as a Haskell type.
@@ -46,7 +46,7 @@ data KorePattern
       -- | application pattern (symbol, [arg] )
       KJApp
         { name :: Id -- may start by a '\\'
-        , sort :: Sort
+        , sorts :: [Sort]
         , args :: [KorePattern]
         }
     | -- | string literal
@@ -159,7 +159,7 @@ instance FromJSON Sort where
 
 -- TODO could omit the args field if empty (custom instance)
 
--- names and arity of known ML connectives
+-- names of known ML connectives. Arity checked during conversion
 data Connective
     = Top
     | Bottom
@@ -173,16 +173,6 @@ data Connective
 
 -- string-tag encoded in the generated instance. Should they be lower-case?
 
-connArity :: Connective -> Int
-connArity Top = 0
-connArity Bottom = 0
-connArity Not = 1
-connArity And = 2
-connArity Or = 2
-connArity Implies = 2
-connArity Iff = 2
-
---
 data Quant
     = Forall
     | Exists
@@ -223,6 +213,8 @@ decodePattern bs =
 data JsonError
     = -- | Problem reported by json parser
       ParseError String
+    | -- | Wrong arg count for a connective or other construct
+      WrongArgCount String Int
     | -- | Inconsistent data parsed. TODO: refine!
       KoreError String
     | NotImplemented String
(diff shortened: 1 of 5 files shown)


kore/src/KoreJson.hs
```diff
@@ -0,0 +1,603 @@
...
+      KJNext
+        { sort :: Sort
+        , dest :: KorePattern
+        }
+    | -- | source rewrites to dest (same sort)
+      KJRewrites
+        { sort :: Sort
+        , source :: KorePattern
+        , dest :: KorePattern
+        }
```
ana-pantilie, line 114: In theory, these could be handled by the `KJConnective` case.

kore/src/KoreJson.hs
```diff
@@ -0,0 +1,603 @@
...
+    | -- TODO textual parser also understands And/Implies/Iff
+
+      -- | left/right associative app pattern
+      KJMultiApp
+        { assoc :: LeftRight
+        , symbol :: Id -- may start by a '\\'
+        , sorts :: [Sort]
+        , args :: [KorePattern]
+        }
```
ana-pantilie, line 136: I think, in theory at least, this can include any binary operation, so they could be collapsed into a single, general, category.

kore/src/KoreJson.hs
```diff
@@ -0,0 +1,603 @@
...
+        , arg :: KorePattern
+        }
+    | -- | mu, nu
+      KJFixpoint
+        { combinator :: Fix
+        , -- no sort
+          var :: Id
+        , varSort :: Sort
+        , arg :: KorePattern
+        }
```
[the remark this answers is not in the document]
jberthold: Agree to merge them into `KJBinder` , however the `Fixpoint` does not have a result sort (which would anyway always be `varSort` IIUC).
`KJQuantifier` has an argument sort (`s_1`) and a possibly different result sort (`s_2`) : `\forall (@X : s_1) . t : s_2`

jberthold:
Notes from meeting:

* sort variables should be objects (one field)
   now:
```
   "sort": { "name": "X" , "args": [ ... (sorts) ...] }
   "sort": "sortvariablename"
```

```
   "sort": { "tag": "SortVariable", "name": "sortvariablename" }
   "sort": { "tag": "Sort", {"name": "sortvariablename" , "args": [ ] } }
```
* Multi-things (syntax sugar) are not strictly necessary, could be supported only on the input side (of HS-Backend)
* Probably go through the textual kore spec. to correct things if need be
* lexical checks on identifier names should be present (parsing)
* Alternative data model: no grouping
  `{ "tag": "And", "sort" : ... , "first" : ... , "second": ... }`

   Allowed tags follow the textual kore tags but without backslash and capitalised. Exception: DomainValue "dv". Syntactic sugar to be left out for now (decide later).

kore/src/Kore/Syntax/Json/Internal.hs
```diff
@@ -0,0 +1,687 @@
...
+checkIdChars name
+    | T.null name = ["Empty"]
+checkIdChars name =
+    ["Illegal initial character " <> show first | not $ isAlpha first]
+    ++ [ "Contains illegal characters: " <> show (nub $ T.unpack illegalChars)
+       | not $ T.null illegalChars
+       ]
+  where
+    first = T.head name
+    illegalChars = T.filter (not . isIdChar) $ T.tail name
```
jberthold, line 271: BTW the `null` case is separate here because the `T.head` and `T.tail` in the `where` clause will crash the program otherwise. I don't like strict Haskell.

kore/src/Kore/Syntax/Json/Internal.hs
```diff
@@ -0,0 +1,687 @@
...
+    deriving anyclass (ToJSON, FromJSON)
+
+------------------------------------------------------------
+-- reading
+
+-- | Errors relating to the json codec, re-exported from API
+data JsonError
+    = -- | Problem reported by json parser
+      ParseError String
+    | -- | MultiOr/MultiApp require a non-empty argument list
```
goodlyrottenapple, line 376: Maybe worth using [NonEmpty](https://hackage.haskell.org/package/base-4.16.2.0/docs/Data-List-NonEmpty.html#t:NonEmpty)?

goodlyrottenapple: (in MultiOr/MultiApp)

kore/src/Kore/Syntax/Json/Internal.hs
```diff
@@ -0,0 +1,687 @@
...
+     in fromPatternF patF
+
+fromPatternF :: Kore.PatternF VariableName (Kore.Pattern VariableName ann) -> KorePattern
+fromPatternF = \case
+    AndF Kore.And{andSort, andFirst, andSecond} ->
+        KJAnd
+            { sort = fromSort andSort
+            , first = fromPattern andFirst
+            , second = fromPattern andSecond
+            }
```
goodlyrottenapple, line 545: There is a trick you can do with instances of [Recursive](https://hackage.haskell.org/package/recursion-schemes-5.2.2.2/docs/Data-Functor-Foldable.html#t:Recursive), namely you can use the [cata](https://hackage.haskell.org/package/recursion-schemes-5.2.2.2/docs/Data-Functor-Foldable.html#v:cata) function, which for `Kore.Pattern` should be:

```haskell
cata :: (CofreeF (PatternF var) ann x -> x) -> Pattern var ann -> x
```
That means, we just need to write a function 

```haskell
fromPatternF :: CofreeF (Kore.PatternF VariableName) ann KorePattern -> KorePattern
fromPatternF (_ :< patt) = case patt of
  AndF Kore.And{andSort, first, second} ->
        KJAnd
            { sort = fromSort andSort
            , first
            , second
            }
  ...
  ```
  and `fromPattern` will simply be `cata fromPatternF`
  
This way you remove the mutual recursion between  `fromPattern` and `fromPatternF` and let `cata` do the recursion for you

File: kore/src/Kore/Syntax/Json/Internal.hs (hunk 4 above)
Comment (goodlyrottenapple) [reply kore/src/Kore/Syntax/Json/Internal.hs:271]: ah yeah, I've fixed a bug due to this about two months ago 🥲

Comment (ana-pantilie) [reply kore/src/Kore/Syntax/Json/Internal.hs:271]: Probably a silly question: would making the bindings lazy (`~first` and `~illegalChars`) work here?

File: kore/src/Kore/Syntax/Json/Internal.hs [hunk 7]
```diff
@@ -0,0 +1,687 @@
...
+isIdChar :: Char -> Bool
+isIdChar c = isAlpha c || isDigit c || c `elem` ['_', '\'']
+
+-- | Set variable names _have to_ start with `@`, followed by a valid identifier
+checkSVarName :: Text -> [String]
+checkSVarName name
+    | T.null name = ["Empty"]
+checkSVarName name =
+    ["Must start with `@'" | T.head name /= '@']
+    <> checkIdChars (T.tail name)
```
ana-pantilie, line 282: Similar question to the one above.

notes-json-kore.org
```diff
@@ -0,0 +1,96 @@
+#+Title: JSON format to represent Kore terms
```
ana-pantilie, line 1: It's great for the reviewers to see the design process detailed out! But I'm not sure if we want to keep this file forever, I'd probably clean the info up a little so that it's clearer to users/other devs. Maybe it could be extracted into a some `docs/json.md` document, and it could also provide a link to the JSON-schema.

ana-pantilie (commented):
This is exceptionally well-documented!

I had a couple of questions and a suggestion, and I also agree with Sam's remarks.

kore/src/Kore/Syntax/Json/Internal.hs
jberthold: That looks very elegant! Would love to use it (although it will make me forget what actually happens with the `CoFree` here...)
I just tried it, but got stuck on one problem: the `DomainValueF` in that type is more liberal than the textual Kore, so I extracted a string literal on the fly (directly inlined the recursion into the `domainValueChild` with the expectation of a string). I don't see how I would do this with the `cata`.
https://github.com/runtimeverification/haskell-backend/pull/3131/files/74e1d8d72d359291edddc50821fd4faf8f1c504f#diff-ffa0765d097e720acdb8048005535bff42394df5ca200e46d72ffa44e0d6d73fR564-R567

goodlyrottenapple: think you should be able to do something like

```haskell
DomainValueF (Kore.DomainValue domainValueSort (KJString value))
    KJDv
        { sort = fromSort domainValueSort
        , value
        }

DomainValueF _ -> error "Bad domain value"
```

goodlyrottenapple: sorry had that wrong. the edit should work

goodlyrottenapple: oops...nm. the first version was correct after all

```haskell
DomainValueF (Kore.DomainValue domainValueSort (KJString value))
    KJDv
        { sort = fromSort domainValueSort
        , value
        }

DomainValueF _ -> error "Bad domain value"
```

kore/src/Kore/Syntax/Json/Internal.hs
jberthold: The only problem is that I need to call the field a different name then :shrug: (conflict with `App` constructor field type)

kore/src/Kore/Syntax/Json/Internal.hs
jberthold: ...but `KJString` is from the target type `KorePattern`, not from the `Pattern/PatternF` family... :confused:

jberthold:
Example of final json format:
```
{
    "tag": "Equals",
    "argSort": {
      "tag": "SortVariable",
      "name": "TH5pKDmw67htB036"
    },
    "resultSort": {
      "tag": "SortApp",
      "name": "T0o5",
      "args": []
    },
    "first": {
        "tag": "App",
        "name": "\\nhc657806",
        "sorts": [
          {
            "tag": "SortVariable",
            "name": "nSv0490"
          }
        ],
        "args": []
      },
    "second": {
      "tag": "Not",
      "sort": {
        "tag": "SortApp",
        "name": "P36117",
        "args": []
      },
      "arg": {
        "tag": "App",
        "name": "\\nhc657806",
        "sorts": [
          {
            "tag": "SortVariable",
            "name": "nSv0490"
          }
        ],
        "args": []
      }
    }
}
```

ana-pantilie (approved):
👍

Issue #3143, opened 2022-07-12
Overhaul graph traversal/execution strategy


ana-pantilie:
- [x] #3168
- [x] Design work towards JSON-RPC new functionality
- [x] Create a front-end ticket when this is done, so we can update the pyk library

jberthold:
## Analysing the JSON-RPC API vs. `execute` requirements

### JSON RPC API data for `execute`
(Description in docs/2022-07-18-JSON-RPC-Server-API.md)

* Request data (exemplified)
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "execute",
  "params": {
    "max-depth": 1,
    "cut-point-rules": ["rule1"],
    "terminal-rules": ["ruleFoo"],
    "state": { <TERM IN KORE JSON FORMAT> }
  }
}
```

* Responses (for successful execution)
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "states": [ LIST_OF_RESULT_STATES ],
    "reason": REASON
  }
}
```
  with   
  * `REASON`:  
    string, one of { `final-state`, `branching`, `stuck`, `depth-bound`, `cut-point-rule`, `terminal-rule`}
  * LIST_OF_RESULT_STATES:  
    a list of objects with `state` (KORE term) and `depth` (depth reached)
    
  Depending on `reason`, the list will contain the following:
  1. (`final-state`) a single term (the final state)
  2. (`branching`) a list of more than one next states
  3. (`stuck`) a single stuck state
  4. (`depth-bound`) a single (non-final, usually non-stuck) state
  5. (`cut-point-rule`) a list of next states (after applying a cut point rule)
  6. (`terminal-rule`) a single final state (after applying a terminal rule)

### Suggestions

#### Stop-on-branching invariant
The data model chosen for the `execute` RPC call requires it to never execute across a branch point. Otherwise, results of an execution would potentially contain combinations of the above `reason`s, which is not supported.

#### Indicating which rule led to stopping
For cases 5 and 6 above, the particular rule that led to stopping should be indicated in an additional field. This could be done by extending making the `reason` field contain an object in these cases, or by adding an optional `rule` field ignored for any reason other than `cut-point-rule` or `terminal-rule`)

#### Possible merger of `reason`s
There is no essential difference between `branching` and `cut-point-rule` other than the suggested `rule` field. Likewise, `final-state` and `terminal-rule` could be merged by determining from the `rule` field which particular case led to termination.

[...]

PR #3146, opened 2022-07-13
[kore-rpc] Initial working implementation + test harness stub 

goodlyrottenapple:
This initial implementation sends back dummy values for all the routes. We first need to address #3143 and #3144 to provide correct behavior.

This PR also includes a stub test harness, which is a Python script that launches a new instance of `kore-rpc` and tests the output, generated per the given RPC JSON request, against the golden response. It can be run via `nix build .#test-rpc` and should eventually be added to the CI test run.

There is also a document describing the current JSON RPC API:

https://github.com/runtimeverification/haskell-backend/blob/eea0670eac043684ad8d0ad812eb584d0437007a/docs/2022-07-18-JSON-RPC-Server-API.md


kore/src/Kore/Syntax/Json.hs
```diff
@@ -6,11 +6,15 @@ License     : BSD-3-Clause
...
+    KoreJson,
     JsonParseError,
     encodePattern,
     encodeKoreJson,
     decodePattern,
     decodeKoreJson,
+    toParsedPattern,
+    term,
+    fromPattern,
 ) where
```
goodlyrottenapple, line 18: @jberthold I'm re-exporting some of the internals here. Hope that's ok?

jberthold: No problem at all, makes sense to export the conversion functions. I would maybe 
* export `KoreJson(..)` rather than `term` 
* also export `KorePattern` (type only, without constructors)
but fine either way

kore/app/rpc/Main.hs
```diff
@@ -2,8 +2,167 @@
...
+                    & handle handleSomeException
+                    & runKoreLog
+                        tmpDir
+                        koreLogOptions
+            )
+
+koreRpcServerRun :: GlobalMain.LocalOptions KoreRpcServerOptions -> GlobalMain.Main ExitCode
+koreRpcServerRun GlobalMain.LocalOptions{execOptions, simplifierx} = do
+    let KoreRpcServerOptions{definitionFileName, mainModuleName, koreSolverOptions, port} = execOptions
+        KoreSolverOptions{timeOut, rLimit, resetInterval, prelude} = koreSolverOptions
```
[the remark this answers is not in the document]
goodlyrottenapple: sure

kore/app/rpc/Main.hs
```diff
@@ -2,8 +2,167 @@
...
+                }
+
+    GlobalMain.clockSomethingIO "Executing" $
+        bracket
+            (SMT.newSolver smtConfig)
+            SMT.stopSolver
+            $ \mvar -> do
+                let solverSetup = SMT.SolverSetup{userInit = declareSMTLemmas metadataTools lemmas, refSolverHandle = mvar, config = smtConfig}
+                runReaderT (SMT.getSMT SMT.initSolver) solverSetup
+                Log.LoggerT $ ReaderT $ \loggerEnv -> runServer port solverSetup loggerEnv simplifierx serializedModule
```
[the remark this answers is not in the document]
goodlyrottenapple: I've added comments describing whats going on

kore/app/rpc/Main.hs
```diff
@@ -2,8 +2,167 @@
...
+ )
+
+data KoreRpcServerOptions = KoreRpcServerOptions
+    { definitionFileName :: !FilePath
+    , mainModuleName :: !ModuleName
+    , koreSolverOptions :: !KoreSolverOptions
+    , koreLogOptions :: !KoreLogOptions
+    , bugReportOption :: !BugReportOption
+    , port :: !Int
+    -- , version :: _ ??
```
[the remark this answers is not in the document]
goodlyrottenapple: I'm not sure... I've removed it

jberthold (commented):
LGTM, except two small details to adjust

kore/app/rpc/Main.hs
```diff
@@ -2,8 +2,172 @@
...
+            str
+            ( metavar "DEFINITION_FILE"
+                <> help "Kore definition file to verify and use for execution"
+            )
+        <*> parseMainModuleName
+        <*> parseKoreSolverOptions
+        <*> parseKoreLogOptions Main.exeName startTime
+        <*> parseBugReportOption
+        <*> option
+            auto
```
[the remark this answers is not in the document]
goodlyrottenapple: not sure what that range is tbh.

kore/src/Kore/JsonRpc.hs
```diff
@@ -85,79 +115,63 @@ instance Exception ReqException
...
     deriving stock (Generic, Show, Eq)
     deriving
         (ToJSON)
-        via CustomJSON '[OmitNothingFields, FieldLabelModifier '[CamelToKebab]] StepResult
+        via CustomJSON '[OmitNothingFields, FieldLabelModifier '[CamelToKebab]] Condition
 
 data ImpliesResult = ImpliesResult
     { satisfiable :: !Bool
-    , substitution :: !(Maybe Text)
+    , contition :: !(Maybe Condition)
```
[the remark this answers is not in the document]
jberthold: And you are saying it should indeed be `contition` with a `t` in the middle? :shrug:

kore/app/rpc/Main.hs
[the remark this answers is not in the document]
jberthold: `[1 .. 65535]` (but using [1..1024] require `root` privileges). This is very minor, though, illegal port numbers will crash when starting inside `runrerver` otherwise

kore/src/Kore/JsonRpc.hs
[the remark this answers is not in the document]
goodlyrottenapple: sorry didnt notice that :D

kore/app/rpc/Main.hs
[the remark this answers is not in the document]
goodlyrottenapple: yeah, it might also crash if the port is taken

jberthold (approved):
Might need a manual rebase but LGTM
(code should probably be merged to unblock others).

test/rpc-server/runTests.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def recv_all(sock):
+    BUFF_SIZE = 4096 # 4 KiB
+    data = b''
+    while True:
+        part = sock.recv(BUFF_SIZE)
+        data += part
+        if len(part) < BUFF_SIZE:
+            # either 0 or end of data
```
tothtamas28, line 18: This might not be a good solution: https://stackoverflow.com/a/7337584.

Did you consider implementing the server using HTTP? Alternatively, if raw sockets are to be used then perhaps  the client can read the socket until it successfully parses the JSON (or runs into an error), but that's harder to implement.

kore/app/rpc/Main.hs
```diff
@@ -2,8 +2,172 @@
...
+                , SMT.prelude = prelude
+                }
+
+    -- launch the SMT solver, initialize it and then pass the SolverSetup object to the rpc server
+    GlobalMain.clockSomethingIO "Executing" $
+        bracket
+            (SMT.newSolver smtConfig)
+            SMT.stopSolver
+            $ \mvar -> do
+                let solverSetup = SMT.SolverSetup{userInit = SMT.runWithSolver $ declareSMTLemmas metadataTools lemmas, refSolverHandle = mvar, config = smtConfig}
```
JKTKops, line 165: Could this line be split up somehow? e.g. above the call to `clockSomethingIO` add a new binding for `setupWithMVar :: MVar ? -> SolverSetup`?

kore/src/Kore/Log/JsonRpc.hs
```diff
@@ -0,0 +1,43 @@
...
+
+instance Entry LogJsonRpcServer where
+    entrySeverity LogJsonRpcServer{level} = case level of
+        LevelDebug -> Debug
+        LevelInfo -> Info
+        LevelWarn -> Warning
+        LevelError -> Error
+        _ -> Debug
+    helpDoc _ = "log JSON RPC Server messages"
+    oneLineDoc LogJsonRpcServer{msg} = pretty $ Text.decodeUtf8 $ fromLogStr msg
```
JKTKops, line 43: This definition is slightly dangerous - if `msg` is more than one line, then `--log-format=oneline` could end up printing one log across multiple lines, which could break log-trawling scripts.

That said we don't have many log-trawling scripts in use so it might not be a huge deal. But perhaps a warning comment?

test/rpc-server/runTests.py
goodlyrottenapple: This is just a test script, the pyk implementation should obviously make sure these things are correct. Using HTTP seems like an overkill given the RPC server is meant to be used locally only, as far as I understand. I think the vscode RPC server also just uses raw sockets without HTTP.

tothtamas28: I'm not against using TCP without HTTP, but in that case it would be much easier to implement clients if the server defined a super thin application layer at least, e.g.:

- The JSON payload is written in a single line, terminated by newline (my preference); or
- The first fixed number of bytes is an integer indicating the size of the rest of the payload in bytes.

Otherwise, we probably won't be able to use the JSON parser in the Python standard library (but maybe we could us [ijson](https://github.com/ICRAR/ijson) instead).

goodlyrottenapple: The server does indeed send a single line terminated by a newline (or at least should!)

tothtamas28: That's great, it simplifies things a lot. Please document it in `JSON-RPC-Server-API.md` so that it becomes part of the contract.

kore/src/Kore/Log/JsonRpc.hs
goodlyrottenapple: I've added `Text.replace "\n" " "`, reckon that should do the trick?

test/rpc-server/runTests.py
goodlyrottenapple: added to the doc now

PR #3147, opened 2022-07-14
3145 json round trip testing and adjustments

### LINKED ISSUE #3145: Kore <-> JSON: Implement round-trip tests using textual kore and json
Part of [k#2626](runtimeverification/k/issues/2626)

Using tests from [`test/parser`](https://github.com/runtimeverification/haskell-backend/tree/master/test/parser)
* add files with json format corresponding to the textual kore files (separate directory) 
* implement round trip testing between textual and json format

And some small changes to the JSON format
* provide a shell script to generate tests (calling into the test data generator sampling)
* change field names as discussed (@tothtamas28 to provide details)
* change `_` to `-` in identifier lexical checks/generator

jberthold:
Fixes #3145 

### Scope:

* Adds round-trip tests between (Kore patterns in) textual Kore files for parser testing and corresponding JSON files (containing arrays of encoded Kore patterns).
* Adjusts some field names as requested by users
* Fixes lexical check for identifiers (allows `-`, forbids `_`)
* adds a script to generate JSON files with randomised test input
* adds a top-level wrapper with `"format": "KORE"` and `"version": 1` fields, placing the kore pattern inside a `term` field.

### Estimate:

---

###### Review checklist


jberthold:
@tothtamas28  Script Usage:
```
   scripts/generate-json-examples.sh [--help]
   scripts/generate-json-examples.sh [--basename NAME] [--count N] [--directory DIR] [--use-multi]

Generates up to 99 examples of Kore JSON files in a directory.

By default, 20 files "sampleData/KoreJSON_[01..20].json" will be created,
the directory and name prefix can be adjusted.

IMPORTANT: The script will _overwrite_ any existing files with identical names.

Available options:

--basename  NAME  File name prefix to use (default: "KoreJSON")
--count N         How many files to create (default: 20)
--directory DIR   Target directory (default: "sampleData")
--use-multi       Include MultiOr and MultiApp (default: not included)

--help            Print this help and exit
```

goodlyrottenapple:
> Not quite ready... 😞 `nix` has a problem with using test data outside the package directory.

anything i can help with?

goodlyrottenapple (approved):
LGTM

radumereuta (changes requested):
If you look at the JSON format for kast, we have a top node that contains the format and version information: https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/json-input/json-in1.json.kast
Can you please add a heading like this with `{"format": "KORE", "version":1 ...}`
This should be validated every time we load such a file.
It's going to save us some pain in the future.

jberthold:
> Also, looking at the tests, I see we're not handling modules and sentences. Is there a plan to include those?

We are only encoding kore _patterns_ at the moment.
Only axioms will be required to support the functionality for ksummarizer so we did not make any further plans yet.

ana-pantilie:
@radumereuta We don't plan on supporting multiple versions of the JSON format so the `version` field won't be necessary.

We are not planning on adding support for modules or sentence declarations, except axioms because the idea is to be able to receive extra rules for some of the operations we export.

jberthold:
> Can you make the version field an integer, just so it is consistent with the KAST JSON.

I consciously did not make it an Int, since it does not make much sense to interpret it as numeric and only one version will ever be allowed in one software release. Checking the version before any other parse errors would also be tricky if the format allowed any integer number there (the parser would fail on other field/type changes before checking the _value_ of the integer.

jberthold:
OK I made it a number in the json encoding since you are the customer.
Note that both `1` and `1.0` are accepted (that's because it is json).
Also added unit tests to ensure that format and version errors take precedence (in most cases they do).

radumereuta:
> > Can you make the version field an integer, just so it is consistent with the KAST JSON.
> 
> I consciously did not make it an Int, since it does not make much sense to interpret it as numeric and only one version will ever be allowed in one software release. Checking the version before any other parse errors would also be tricky if the format allowed any integer number there (the parser would fail on other field/type changes before checking the _value_ of the integer.

I understand your reasoning, and I agree with it. I had the same thought process when I changed kast json from v1 to v2.
But we are already committed to version being an Int in the KAST format and I'm trying to minimize friction in the future.

PR #3156, opened 2022-07-19
Unify SMT and NoSMT; kill MonadExecute

int-index:
This is the first step towards monomorphisation of the monad transformer stack used in `kore` (for improved performance). Previously there were two separate base monads:

```haskell
newtype NoSMT a = NoSMT {getNoSMT :: LoggerT IO a}
newtype SMT a = SMT {getSMT :: ReaderT SolverSetup (LoggerT IO) a}
```

The choice of the base monad was determined at runtime. Now we have a single base monad:

```haskell
newtype SMT a = SMT (Maybe SolverSetup -> LoggerT IO a)
```

This makes it possible to use a monomorphic monad where previously a polymorphic one was required:

```diff
execute ::
    KoreSolverOptions ->
    SmtMetadataTools StepperAttributes ->
    [SentenceAxiom (TermLike VariableName)] ->
-   (forall exe. MonadExecute exe => exe r) ->
+   SMT r ->
    Main r
```

Note that this change alone is insufficient to improve performance, as the rest of the application still uses an abstract `MonadSMT` context. However, it lifts one of the obstacles towards that goal.

### DIFF
--- a/kore/src/SMT.hs
+++ b/kore/src/SMT.hs
@@ -7,11 +7,21 @@ Maintainer  : thomas.tuegel@runtimeverification.com
 -}
 module SMT (
     SMT,
-    getSMT,
+    runNoSMT,
+    runSMT,
     Solver,
     stopSolver,
-    runSMT,
     MonadSMT (..),
+    declare,
+    declareFun,
+    declareSort,
+    declareDatatype,
+    declareDatatypes,
+    assert,
+    check,
+    ackCommand,
+    loadFile,
+    reinit,
     Config (..),
     defaultConfig,
     TimeOut (..),
@@ -30,8 +40,6 @@ module SMT (
     declareFun_,
     setInfo,
     setOption,
-    NoSMT (..),
-    runNoSMT,
     SimpleSMT.SolverException (..),
 
     -- * Expressions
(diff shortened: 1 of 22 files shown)


kore/src/SMT.hs
```diff
@@ -622,6 +481,126 @@ setOption :: MonadSMT m => Text -> SExpr -> m ()
...
     ackCommand $ List (Atom "set-option" : Atom infoFlag : [expr])
 
+{- | Query an external SMT solver (if available).
+
+The solver may be shared among multiple threads. Individual commands will
+acquire and release the solver as needed, but sequences of commands from
+different threads may be interleaved; use 'withSolver' to acquire exclusive
+access to the solver for a sequence of commands.
+-}
+newtype SMT a = SMT (Maybe SolverSetup -> LoggerT IO a)
```
ana-pantilie, line 491: Just wondering, what's the reasoning behind making this explicit instead of using `ReaderT`?

ana-pantilie (approved):
Looks really good! I just had one question.

kore/src/SMT.hs
int-index: No particular reason, just a stylistic preference.

PR #3158, opened 2022-07-19
TermLike -> Syntax.Pattern

### LINKED ISSUE #3148: Define a `TermLike` -> `KorePattern` function
We need this conversion in the `kore-rpc` server, to send back a JSON representation of a kore pattern, namely `KorePattern`, whereas the executor/simplifier/etc. all return a `TermLike` type.

Probably easiest to go via `Kore.Syntax.Pattern` as we alrady have a function to turn that into a `KorePattern`. As per @JKTKops suggestion, we probably need to define
```haskell
uninternalize :: TermLikeF variable child -> PatternF variable child
```
and then use [hoist](https://hackage.haskell.org/package/recursion-schemes-5.2.2.2/docs/Data-Functor-Foldable.html#v:hoist)

For definitions of the two types, see here: https://github.com/runtimeverification/haskell-backend/blob/b06757e252ee01fdd5ab8f07de2910711997d845/kore/src/Kore/Syntax/PatternF.hs#L59-L87
https://github.com/runtimeverification/haskell-backend/blob/b06757e252ee01fdd5ab8f07de2910711997d845/kore/src/Kore/Internal/TermLike/TermLike.hs#L147-L187

The trouble is the two types aren't obviously isomorphic, however both can be unparsed to textual kore, so there should be no issue defining this function...probably.

goodlyrottenapple:
Fixes #3148


kore/test/Test/Kore/Internal/TermLike.hs
```diff
@@ -515,3 +536,61 @@ test_renaming =
...
+    trmLike <- forAll (runKoreGen Mock.generatorSetup ConsistentKore.termLikeGen)
+    let patt = fmap (const Attribute.Null) (from trmLike :: Pattern.Pattern VariableName (TermAttributes VariableName))
+
+    case PatternVerifier.runPatternVerifier context $
+        PatternVerifier.verifyStandalonePattern Nothing patt of
+        Left Kore.Error.Error{errorError} ->
+            fail $
+                unparseToString patt <> "\n\n"
+                    <> show errorError
+        Right trmLike2 -> patt === fmap (const Attribute.Null) (from trmLike2 :: Pattern.Pattern VariableName (TermAttributes VariableName))
```
goodlyrottenapple, line 551: This test isn't enitrely satisfactory. It only tests that going from `TermLike -> Syntax.Pattern -> TermLike -> Syntax.Pattern` is the same as `TermLike -> Syntax.Pattern`.
The difficulty with showing `TermLike -> Syntax.Pattern -> TermLike` is equivalent to the original `TermLike` lies in the fact that InternalList/Map/Set do not get internalized until simplification. I have tried running the simplifier on both the original and round trip `TermLike` to see if they end up equivalent, but unfortunately, they still differ syntactically. Perhaps all we can do is say that semantically they are the same? However, I'm not sure how to encode this property. @JKTKops @ana-pantilie any ideas? If not, I will leave the test as is. It is at least a good sanity check.

jberthold (commented):
LGTM overall, some suggestions.
Leaving approval to @JKTKops and @ana-pantilie as they know the internal ACs better

kore/test/Test/Kore/Internal/TermLike.hs
```diff
@@ -515,3 +536,61 @@ test_renaming =
...
+  where
+    context =
+        PatternVerifier.verifiedModuleContext
+            IndexedModule.IndexedModuleSyntax
+                { indexedModuleName = ModuleName "dummy"
+                , indexedModuleAliasSentences = mempty
+                , indexedModuleSymbolSentences =
+                    Map.fromList
```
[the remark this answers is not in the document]
goodlyrottenapple: yeah I was going to do this and then forgot. thanks will do so.

kore/src/Kore/Internal/TermLike/TermLike.hs
```diff
@@ -1249,3 +1262,131 @@ depth :: TermLike variable -> Int
...
+            mkAux = \case
+                [] ->
+                    wrap $
+                        PatternF.ApplicationF $
+                            Application (Symbol.toSymbolOrAlias unitSymbol) []
+                [x] -> x
+                (x : xs) ->
+                    wrap $
+                        PatternF.ApplicationF $
+                            Application (Symbol.toSymbolOrAlias concatSymbol) [x, mkAux xs]
```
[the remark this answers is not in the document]
goodlyrottenapple: I've never come across `fold1`. Is there an advantage to using it? Seems the way it is atm is more understandable maybe?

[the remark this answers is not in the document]
goodlyrottenapple: Btw, thw `[x] -> x` case comes from looking at the unparser code here
https://github.com/runtimeverification/haskell-backend/blob/b06757e252ee01fdd5ab8f07de2910711997d845/kore/src/Kore/Unparser.hs#L310-L336

kore/src/Kore/Internal/TermLike/TermLike.hs
```diff
@@ -1249,3 +1262,136 @@ depth :: TermLike variable -> Int
...
+        mkInternalList InternalList{internalListUnit, internalListConcat, internalListElement, internalListChild} =
+            mkTrm internalListUnit internalListConcat mkChildren
+          where
+            mkChildren =
+                map
+                    ( \e ->
+                        wrap $
+                            PatternF.ApplicationF $
+                                Application (Symbol.toSymbolOrAlias internalListElement) [e]
+                    )
```
ana-pantilie, line 1365: Can you factor this out and give it a suggestive name?

kore/src/Kore/Syntax/Application.hs
```diff
@@ -70,6 +71,9 @@ data Application head child = Application
     deriving anyclass (Hashable, NFData)
     deriving anyclass (SOP.Generic, SOP.HasDatatypeInfo)
 
+instance Bifunctor Application where
+    bimap f g (Application alias children) = Application (f alias) (fmap g children)
```
ana-pantilie, line 75: Line length

```suggestion
    bimap f g (Application alias children) =
        Application (f alias) (fmap g children)
```

ana-pantilie (commented):
I'm fine with the property test, but it shouldn't be the only test we're relying on here. I want to see a few unit tests for: the AC uninternalizer, the list uninternalizer, uninternalize itself.

kore/src/Kore/Internal/TermLike/TermLike.hs
```diff
@@ -1249,3 +1262,136 @@ depth :: TermLike variable -> Int
...
+                    ++ map
+                        ( \(k, v) ->
+                            wrap $
+                                PatternF.ApplicationF $
+                                    Application (Symbol.toSymbolOrAlias builtinAcElement) $
+                                        uninternalizeKey k :
+                                        concreteElementToApplicationArgs v
+                        )
+                        (HashMap.toList concreteElements)
+                    ++ opaque
```
ana-pantilie, line 1395: For some reason my review for this part of the code vanished.

This is very hard to read, I'd keep the combining of the uninternalized parts at the top, while the transformations themselves should be moved to a lower level since they're just implementation details.

Also, we need to document what output this produces. I mean, AC collections are, well, associative and commutative therefore they can have multiple, equivalent, uninternalized forms. We need to document that this produces uninternalized patterns of the <insert-here> form.

I also want to see unit tests for this and the list one.

ana-pantilie: Ofc, feel free to move the AC uninternalizer and the list uninternalizer to the global scope in order to be able to test them directly.

ana-pantilie (approved):

PR #3159, opened 2022-07-20
Remove a few unused definitions

int-index:
These definitions were unused:

* `liftConditionSimplifier` is dead code
* `simplifyRewriteRule` is dead code
* `TestSimplifier` is dead code
* `simplifyRulePattern` is used in tests but not the app
* `simplifyPatternId` is a no-op wrapper around fromPattern

### DIFF
--- a/kore/src/Kore/Simplify/Exists.hs
+++ b/kore/src/Kore/Simplify/Exists.hs
@@ -307,7 +307,7 @@ makeEvaluateBoundLeft sideCondition variable boundTerm normalized =
         orPattern <- do
             simplifierX <- askSimplifierXSwitch
             case simplifierX of
-                EnabledSimplifierX -> pure (OrPattern.fromPattern substituted)
+                EnabledSimplifierX -> simplifyPatternId substituted
                 DisabledSimplifierX -> simplifyPattern sideCondition substituted
         Logic.scatter (toList orPattern)
   where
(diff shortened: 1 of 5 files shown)


kore/src/Kore/Simplify/Exists.hs
```diff
@@ -307,7 +307,7 @@ makeEvaluateBoundLeft sideCondition variable boundTerm normalized =
         orPattern <- do
             simplifierX <- askSimplifierXSwitch
             case simplifierX of
-                EnabledSimplifierX -> simplifyPatternId substituted
+                EnabledSimplifierX -> pure (OrPattern.fromPattern substituted)
```
ana-pantilie, line 310: This should be reverted.

kore/src/Kore/Simplify/Simplify.hs
```diff
@@ -323,7 +316,7 @@ simplifyPatternScatter sideCondition patt = do
     simplifierX <- askSimplifierXSwitch
     Logic.scatter
         =<< case simplifierX of
-            EnabledSimplifierX -> simplifyPatternId patt
+            EnabledSimplifierX -> pure (fromPattern patt)
```
ana-pantilie, line 319: This should be reverted.

ana-pantilie (commented):
Thanks for doing this clean-up. There are two places which shouldn't be changed for now, though.

kore/src/Kore/Simplify/Exists.hs
ana-pantilie, line 310:
```suggestion
                EnabledSimplifierX -> simplifyPatternId substituted
```

kore/src/Kore/Simplify/Simplify.hs
ana-pantilie, line 319:
```suggestion
            EnabledSimplifierX -> simplifyPatternId patt
```

ana-pantilie (approved):

PR #3161, opened 2022-07-20
Specialize Kore.Exec and Kore.Repl to Simplifier

int-index:
In those modules the abstract monad was never instantiated to anything but `Simplifier`, so it was straightforward to specialize it.


goodlyrottenapple (approved):
LGTM

PR #3166, opened 2022-07-26
Implement --smt-retry-limit option

### LINKED ISSUE #3162: Option to adjust the number of solver query retries
Currently if a z3 query returns unknown, the haskell backend can rerun it once. We could add a command-line option, something like `--max-solver-retries [number]`, to specify how many times the user is willing to retry these, where on each successive retry the timeout would be doubled. This could give the user more control over the z3 timeout and make it easier to figure out how long an expensive query is taking.

JKTKops:
Fixes #3162

### Scope:

Add a new option, `--smt-retry-limit`, which controls how many times we will retry an SMT query that returned `Unknown`. Previously, we would always retry once.

### Estimate:

I think it's done already, but it's unclear where tests for this feature should go (or even if there should be any, as retries were previously untested anyway).

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Rewrite/SMT/Evaluator.hs
+++ b/kore/src/Kore/Rewrite/SMT/Evaluator.hs
@@ -22,6 +22,7 @@ import Control.Lens qualified as Lens
 import Control.Monad.Counter qualified as Counter
 import Control.Monad.State.Strict qualified as State
 import Data.Generics.Product.Fields
+import Data.Limit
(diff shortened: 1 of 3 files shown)


JKTKops:
Please see [the comment on the issue](https://github.com/runtimeverification/haskell-backend/issues/3162#issuecomment-1195973685).

ana-pantilie (approved):
👍

Issue #3119, opened 2022-06-24
kore-exec should output a count of unexplored leaves when it terminates


jberthold:
When a proof fails, `kore-exec` will output a counterexample (or more than one, with `--max-counterexamples`). However, the output of counterexamples does not provide information about how many branches/leaves remained unexplored and might therefore yield additional counterexamples.

This can be found out in debugger mode by calling `leafs`. 
The CLI could output this information when `kore-exec` terminates. 

Needs #3143

ana-pantilie:
@jberthold Let's timebox this: let me know at the end of your day tomorrow what your progress has been and we'll decide what to do with it.

jberthold:
Status: This feature required modifications to how the execution graph is explored (`leavesM` in `Kore.Rewrite.Strategy`) and instigated a rewrite of the graph traversal functionality that `leavesM` provides. Code is in #3163 (closed for archiving), and will be cleaned up so it can be merged and used for `kore-exec`.

PR #3168, opened 2022-07-27
#3119 rewrite graph traversal + output unexplored leaves

jberthold:
Related to #3119

### Estimate:

---

###### Review checklist


kore/src/GraphTraversal.hs
```diff
@@ -0,0 +1,319 @@
...
+    -- | breadth limit, essentially a natural number
+    Limit Natural ->
+    -- | transition function
+    (([Step prim], c) -> Simplifier (TransitionResult ([Step prim], c))) ->
+    -- again, m is probably only for logging
+
+    -- | max-counterexamples, included in the internal logic
+    Natural ->
+    -- | initial node
+    ([Step prim], c) ->
```
ana-pantilie, line 132: What is the `[Step prim]`? I assume `c` is the configuration state.

kore/src/GraphTraversal.hs
```diff
@@ -0,0 +1,319 @@
...
+    TransitionT (..),
+    runTransitionT,
+ )
+import Kore.Simplify.Data (Simplifier)
+import Prelude.Kore
+import Pretty
+
+data TransitionResult a
+    = -- | straight-line execution
+      StraightLine a
```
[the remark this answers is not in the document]
jberthold: Good point, that name is not well-chosen. Maybe `Continuing`?
Interestingly, this particular "result" will never be output to the caller, it signals to continue the recursion.

kore/src/GraphTraversal.hs
jberthold: `prim` is the "mode" of transition (`Simplify`, `Apply*`, `CheckImplications`), a `Step` is one series of transitions with different modes that are executed together. Will change `c` to `config`.

kore/src/Kore/Exec.hs
```diff
@@ -281,41 +282,77 @@ exec
...
-                                    . Strategy.applyBreadthLimit
-                                        breadthLimit
-                                        dropStrategy
-                        rewriteGroups = groupRewritesByPriority rewriteRules
-                        transit instr config =
-                            Strategy.transitionRule
-                                ( transitionRule rewriteGroups strategy
-                                    & profTransitionRule
-                                    & trackExecDepth
+                do
```
ana-pantilie, line 290: This huge block here can be refactored to be made more readable. I think it's nicer see the main logic that is happening at the top, with helper functions having suggestive names. If I want to see the details of the implementation, I can then look for the definitions of the helpers.

kore/src/Kore/Exec/GraphTraversal.hs
```diff
@@ -0,0 +1,338 @@
...
+    unfoldSearchOrder,
+ )
+import Kore.Rewrite.Transition (
+    TransitionT (..),
+    runTransitionT,
+ )
+import Kore.Simplify.Data (Simplifier)
+import Prelude.Kore
+import Pretty
+
```
ana-pantilie, line 38: I think this module would benefit from some extra documentation. My first thought was, how do `TransitionResult`, `StepResult` and `TraversalResult` all relate to eachother?

kore/src/Kore/Rewrite/Strategy.hs
```diff
@@ -93,40 +84,20 @@ import GHC.Generics qualified as GHC
...
     @Strategy prim@ represents a strategy for execution by applying rewrite
     axioms of type @prim@.
 -}
-
--- TODO (thomas.tuegel): This could be implemented as an algebra so that a
--- strategy is a free monad over the primitive rule type and the result of
--- execution is not a generic tree, but a cofree comonad with exactly the
--- branching structure described by the strategy algebra.
 data Strategy prim where
-    -- The recursive arguments of these constructors are /intentionally/ lazy to
```
ana-pantilie, line 117: This is probably out of scope for this PR, but we should probably keep in mind that we'd like to get rid of this type in the near future. Maybe a subsequent PR could take care of that?

ana-pantilie (commented):
This approach is definitely friendlier than the old one, but I think it's still a bit hard to understand just by reading the code, so my suggestions currently concern readability.

I also noticed that allocation has decreased a bit. Should we expect some performance improvements as well? That would be amazing. 😁

kore/src/Kore/Rewrite/Strategy.hs
jberthold: Yes, this type (and maybe even the whole module `Strategy`) is on the list of things to remove. I would think we lift everything worth keeping into the new `GraphTraversal` module.
Currently the `Prove` module still uses `Strategy` for the single-step function, and the repl uses the computation graph directly. Removing this and consolidating the parts we need from `Strategy` into `GraphTraversal` would be follow-up work in a new ticket and PR.

jberthold:
> I also noticed that allocation has decreased a bit. Should we expect some performance improvements as well? That would be amazing.

The new version is a bit faster, yes. My experiment PR had some KEVM timing comparisons - around 3% faster overall, most of it probably because of using `Simplifier` instead of `MonadSimplify`. Other optimisations landed at the same time, so it is hard to tell...

flake.lock
```diff
@@ -98,6 +98,21 @@
         "type": "github"
       }
     },
+    "flake-utils_2": {
+      "locked": {
+        "lastModified": 1642700792,
+        "narHash": "sha256-XqHrk7hFb+zBvRg6Ghl+AZDq03ov6OshJLiSWOoX5es=",
+        "owner": "numtide",
```
jberthold, line 105: Changes here were auto-generated from running command `nix run .#update-cabal-ghc9` (as instructed by @goodlyrottenapple )

kore/src/Kore/Exec/GraphTraversal.hs
```diff
@@ -0,0 +1,338 @@
...
+    -- | breadth limit, essentially a natural number
+    Limit Natural ->
+    -- | transition function
+    ( ([Step instr], config) ->
+      Simplifier (TransitionResult ([Step instr], config))
+    ) ->
+    -- | max-counterexamples, included in the internal logic
+    Limit Natural ->
+    -- | initial node
+    ([Step instr], config) ->
```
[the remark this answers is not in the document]
jberthold: I added a type `TState` with some documentation https://github.com/runtimeverification/haskell-backend/pull/3168/files#diff-2a407a0007073e405b91841c65f2c8b27617e7a6c3aa97f4f1bbb912938f3683R121-R126
Indeed the `Step`s are what the algorithm will do next.

kore/src/Kore/Exec/GraphTraversal.hs
jberthold: I have added documentation that describes all these types together, in a haddock comment on `graphTraversal`. Is this what you had in mind?

ana-pantilie (approved):
LGTM! 🎉

PR #3174, opened 2022-07-29
[nix] Rematerialize GHC9 in the CI

goodlyrottenapple:
Updated the `Update/ Nix` CI script to re-materialize both the current GHC8.10.* and the GHC9 expressions in the `nix/` folder.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -27,22 +27,26 @@ jobs:
           ref: ${{ steps.config.outputs.ref }}
           submodules: recursive
 
-      - name: Install Nix
-        uses: cachix/install-nix-action@v14.1
+      - name: 'Install Nix'
+        uses: cachix/install-nix-action@v15
         with:
           extra_nix_config: |
-            substituters = http://cache.nixos.org https://hydra.iohk.io
+            substituters = http://cache.nixos.org https://cache.iog.io
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
-          install_url: "https://releases.nixos.org/nix/nix-2.3.16/install"
 
-      - name: Install Cachix
+      - name: 'Install Cachix'
         uses: cachix/cachix-action@v10
         with:
           name: runtimeverification
+          extraPullNames: kore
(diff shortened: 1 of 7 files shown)


.github/workflows/update.yml
```diff
@@ -27,22 +27,26 @@ jobs:
...
         uses: cachix/cachix-action@v10
         with:
           name: runtimeverification
+          extraPullNames: kore
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
 
+
       - name: Materialize
-        run: ./nix/rematerialize.sh
+        run: GC_DONT_GC=1 nix run .#update-cabal
```
jberthold, line 46: Should we delete the old script then?
But maybe it is useful to keep it so one can run the update locally. Then we would want to have the `nix run ...` commands in that script.

flake.lock
```diff
@@ -98,6 +98,21 @@
         "type": "github"
       }
     },
+    "flake-utils_2": {
+      "locked": {
+        "lastModified": 1642700792,
+        "narHash": "sha256-XqHrk7hFb+zBvRg6Ghl+AZDq03ov6OshJLiSWOoX5es=",
```
jberthold, line 104: Changes in this file are the same as on my PR #3143, which might cause conflicts if git is not smart enough.

jberthold (commented):
Changesets of the `kore.nix` files are a bit hard to read because of the different layout, but all makes sense.
Besides maintaining the `nix/rematerialize.sh` script, LGTM

.github/workflows/update.yml
goodlyrottenapple: No, the old script works with nix <2.4 without flakes. It should be removed if/when we get rid of non flake nix stuff in the repo

flake.lock
goodlyrottenapple: Yeah sorry I think I forgot to commit the lock file last time I modified the flake.nix file. I will merge master again if you get your changes in first

goodlyrottenapple:
> Changesets of the `kore.nix` files are a bit hard to read because of the different layout, but all makes sense.
> Besides maintaining the `nix/rematerialize.sh` script, LGTM

Yeah sorry about that. I added a step where I run nixfmt on the generated file because for some bizzare reason, running the rematerialize on my local machine and the ci produced files with different formatting. This way we make doubly sure we only trigger the update if the contents actually change

.github/workflows/update.yml
jberthold: :man_shrugging: OK then we keep it as it is for now.

jberthold (approved):

Issue #3172, opened 2022-07-27
Remove --simplifierx


ana-pantilie:
The CL option, the type and https://github.com/runtimeverification/haskell-backend/blob/e39ec77530f3c032d56c50a4d298f02c3ea28b18/kore/src/Kore/Simplify/Simplify.hs#L229 need to be removed.

PR #3179, opened 2022-08-02
Remove simplifierX switch

goodlyrottenapple:
Fixes [#3172](https://github.com/runtimeverification/haskell-backend/issues/3172)

### DIFF
--- a/kore/test/Test/Kore/Exec.hs
+++ b/kore/test/Test/Kore/Exec.hs
@@ -268,14 +268,14 @@ test_matchDisjunction =
     [ testCase "match disjunction" $
         do
             let actual =
-                    matchDisjunctionTest verifiedModule initial [final1, final2]
+                    matchDisjunction verifiedModule initial [final1, final2]
                         & runNoSMT
             result <- actual
             assertEqual "" (mkBottom mySort) result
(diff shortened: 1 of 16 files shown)


kore/test/Test/Kore/Exec.hs
```diff
@@ -1141,9 +1138,9 @@ matchDisjunctionTest ::
...
     [Pattern RewritingVariableName] ->
     SMT (TermLike VariableName)
-matchDisjunctionTest = matchDisjunction DisabledSimplifierX
+matchDisjunctionTest = matchDisjunction
 
 checkFunctionsTest ::
     VerifiedModule Attribute.StepperAttributes ->
     SMT ()
-checkFunctionsTest = checkFunctions DisabledSimplifierX
+checkFunctionsTest = checkFunctions
```
jberthold, line 1146: We could probably just use the original functions now (`matchDisjunctions`, `checkFunctions`)

jberthold (approved):
LGTM.

kore/test/Test/Kore/Exec.hs
goodlyrottenapple: good point

PR #3184, opened 2022-08-04
[#3111] Remove `builtinEvaluators`

### LINKED ISSUE #3111: Serokell: remove BuiltinAndAxiomSimplifier
### Progress until 07/10/2022
- [x] Specialize BuiltinAndAxiomSimplifier to a concrete monad
- [x] Call builtin evaluator functions statically
- [ ] Remove BuiltinAndAxiomSimplifier by doing https://github.com/runtimeverification/haskell-backend/issues/3111#issuecomment-1210727293

dcastro:
Work in progress.

Fixes #3111

### Estimate:

---

###### Review checklist


dcastro:
We're still collecting benchmark results for this PR, I'll post a comment here once they're done.

kore/src/Kore/Rewrite/Function/Evaluator.hs
```diff
@@ -191,6 +201,88 @@ evaluatePattern
...
+            empty
+    maybe missing return $ do
+        axiomIdentifier <- Axiom.Identifier.matchAxiomIdentifier termLike
+        let exact = Map.lookup axiomIdentifier simplifierMap
+        case axiomIdentifier of
+            Axiom.Identifier.Application appId ->
+                let builtintEvaluator = do
+                        name <- Map.lookup appId hookedSymbols
+                        koreEvaluators name
+                 in combineEvaluatorsWithFallBack (builtintEvaluator, exact)
```
goodlyrottenapple, line 236: There is another optimization we could potentially do. If we match on TermLike here, if it is an `ApplySymbolF`, it already carries a `Hook` and we can potentially bypass a lookup inside `hookedSymbols`? Not sure if this is sound or worth it, just thought I'd mention it:

```haskell
ApplySymbolF (Application Symbol{hook} _) | Just name <- getHook hook -> 
  combineEvaluatorsWithFallBack (koreEvaluators name, exact)
```

dcastro: Hm good point! I see that, in `matchAxiomIdentifier`, we use [`mkAppId`](https://github.com/runtimeverification/haskell-backend/blob/ac0e3c0c403bbca4d812d921f8783f4dfcdbcb29/kore/src/Kore/Rewrite/Axiom/Identifier.hs#L202) to extract an `AxiomIdentifier` from an `Application`. Instead, we could just extract the hook directly from the `Application`.

I see that `mkAppId` is also used in `InjF`, `EndiannessF` and `SignednessF`. Would it make sense to generalize this as below?

```hs
    getHookFromTermLikeF :: TermLikeF RewritingVariableName a -> Maybe Text
    getHookFromTermLikeF = \case
        ApplySymbolF (Application symbol _) -> getHook symbol
        InjF (Inj.toApplication -> Application symbol _) -> getHook symbol
        EndiannessF (getConst -> Endianness.toApplication -> Application symbol _) -> getHook symbol 
        SignednessF (getConst -> Signedness.toApplication -> Application symbol _) -> getHook symbol 
        _ -> Nothing
```

dcastro:
[Benchmark results](https://github.com/runtimeverification/haskell-backend/files/9274938/report-builtinEvaluators_monomorph.pdf)

The first column shows how long it took to run the benchmark using the binary built from this branch (ac82cefcaaff8b033507f420c8d51537955ccc57) and the second column using a binary built from master (be7e56172bd02d32b31d91c20fc7ba1a47766fc0)

At the bottom of the document, we see a 1% overall speedup.

kore/src/Kore/Rewrite/Function/Evaluator.hs
goodlyrottenapple: @ana-pantilie?

dcastro:
> [Benchmark results](https://github.com/runtimeverification/haskell-backend/files/9274938/report-builtinEvaluators_monomorph.pdf)

I should point out that this PR builds on top of https://github.com/runtimeverification/haskell-backend/pull/3178#issuecomment-1207952135, so, taken together, this is a ~6% overall speedup.

kore/src/Kore/Rewrite/Function/Evaluator.hs
ana-pantilie: I'm not sure, you could give it a try, but I'd prefer to get this merged first and then open a follow-up PR which attempts these changes.

dcastro: Sure! @goodlyrottenapple / @jberthold can you please review/approve this PR?

kore/src/Kore/Simplify/Simplify.hs
```diff
@@ -142,10 +142,13 @@ data Env = Env
         SideCondition RewritingVariableName ->
         TermLike RewritingVariableName ->
         Simplifier (OrPattern RewritingVariableName)
-    , simplifierAxioms :: !BuiltinAndAxiomSimplifierMap
+    , -- | A map with the user defined evaluators.
+      -- Builtin evaluators are not stored in this map.
+      simplifierAxioms :: !BuiltinAndAxiomSimplifierMap
```
jberthold, line 147: (nit) Just stumbled over this type name here - no built-in things in the `BuiltInAndSomething` any more... 
Maybe worth renaming if anything else needs to still change (or when updating the branch to master state).

jberthold (approved):
LGTM

goodlyrottenapple (approved):

kore/src/Kore/Simplify/Simplify.hs
dcastro: Good catch, thanks! Since I'm now trying to avoid storing user-defined evaluators in maps too, I'll just delete this type once I'm done.

dcastro:
When moving the code around, I had forgotten to wrap `koreEvaluators` in `builtinEvaluation`.
I just pushed a fixup commit with this change: https://github.com/runtimeverification/haskell-backend/pull/3184/commits/b302b905f08dd9d08787b2ff5af4b64ed18eb4bc

This change made some tests fail, e.g., "test bad encoding" and "test bad decoding":
https://github.com/runtimeverification/haskell-backend/blob/b302b905f08dd9d08787b2ff5af4b64ed18eb4bc/kore/test/Test/Kore/Builtin/InternalBytes.hs#L556-L582


Upon investigation, I noticed what looks to be a problem with the tests in the master branch.
In the master branch,

* in `/src/`: the `mkSimplifierEnv` function constructs the builtin `koreEvaluators`, and wraps them `builtinEvaluation`. This means that if a builtin evaluator is used and it returns `AttemptedAxiom.NotApplicable`, the code [will crash](https://github.com/runtimeverification/haskell-backend/blob/e0580ec3a1ecfb05b0e8e2730cc16c9adfd6a6f5/kore/src/Kore/Rewrite/Axiom/EvaluationStrategy.hs#L192-L201).

* in the "test bad encoding" test: the builtin `koreEvaluators` are [NOT wrapped](https://github.com/runtimeverification/haskell-backend/blob/e0580ec3a1ecfb05b0e8e2730cc16c9adfd6a6f5/kore/test/Test/Kore/Rewrite/Function/Integration.hs#L2033-L2034) in `builtinEvaluation`. This test runs the builtin `evalDecodeBytes`, which returns `NotApplicable`. And, because it is not wrapped in `builtinEvaluation`, it'll not crash like it would in a real world scenario. Instead, it'll fall through to the [`unevaluated` branch](https://github.com/runtimeverification/haskell-backend/blob/e0580ec3a1ecfb05b0e8e2730cc16c9adfd6a6f5/kore/src/Kore/Rewrite/Function/Evaluator.hs#L106-L115), and return an `OrPattern`, which the test expects, and thus it passes.


It looks like the test relies on the `evalDecodeBytes` evaluator to behave like a user-defined evaluator (which doesn't crash on `NotApplicable`), rather than a builtin evaluator (which does).

Since, in my branch, `koreEvaluators` are now _always_ wrapped, `evalDecodeBytes` will always behave like a builtin evaluator, and so this test [crashes](https://github.com/runtimeverification/haskell-backend/blob/e0580ec3a1ecfb05b0e8e2730cc16c9adfd6a6f5/kore/src/Kore/Rewrite/Axiom/EvaluationStrategy.hs#L192-L201) when it hits a `NotApplicable`, like it would in a real-world scenario.

/cc @jberthold @goodlyrottenapple what do you suggest we do about these tests? Should the tests catch the `error` thrown from `builtinEvaluation`, instead of expecting an `OrPattern` to be returned?

dcastro:
Update: I've updated the tests, please merge if the fix seems good.

kore/src/Kore/Rewrite/Function/Evaluator.hs
dcastro: I've opened a separate PR with this change: #3195 

Please see this comment: https://github.com/runtimeverification/haskell-backend/pull/3195#issuecomment-1214425441

ana-pantilie (approved):

PR #3195, opened 2022-08-10
[#3111] Bypass `hookedSymbols` lookup

dcastro:
# Work in progress!

Fixes part of #3111

### Scope:

Problem: In `lookupAxiomSimplifier`, if the axiom identifier for the
given term is an `Axiom.Identifier.Application`, we look for the hook
inside the `hookedSymbols` map.

However, in some cases, the term itself might already carry a hook. If
that's the case, then we don't need to perform a lookup in the
`hookedSymbols` map.

See this comment for context:
https://github.com/runtimeverification/haskell-backend/pull/3184#discussion_r938637613

Solution: check if the term carries a hook first, and only do a lookup
in `hookedSymbols` if it doesn't.

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Rewrite/Function/Evaluator.hs
+++ b/kore/src/Kore/Rewrite/Function/Evaluator.hs
@@ -20,12 +20,17 @@ import Control.Error (
     throwE,
  )
 import Control.Monad qualified as Monad
+import Data.Functor.Const (getConst)
+import Data.Functor.Foldable qualified as Recursive
 import Data.Map.Strict qualified as Map
 import Data.Text (Text)
 import Kore.Attribute.Pattern.Simplified qualified as Attribute.Simplified
 import Kore.Attribute.Symbol qualified as Attribute
 import Kore.Attribute.Synthetic
 import Kore.Builtin (koreEvaluators)
+import Kore.Builtin.Endianness.Endianness qualified as Endianness
+import Kore.Builtin.Signedness.Signedness qualified as Signedness
+import Kore.Internal.Inj qualified as Inj
 import Kore.Internal.MultiOr qualified as MultiOr (
     flatten,
     merge,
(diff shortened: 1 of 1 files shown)


dcastro:
@goodlyrottenapple: It appears the test "key outside of map, symbolic opaque terms" in `Test.Kore.Unification.Unifier` is now failing (all other tests, including integration tests, are passing).

In this test, `lookupAxiomSimplifier` is called with a term such that:
* `matchAxiomIdentifier` returns an `Application` with id `"concatMap"`
* `getHookFromTermLike` returns the hook `"MAP.concat"`

Previously, we'd look up "concatMap" in `hookedSymbols`. This lookup would fail (because `Mock.env.hookedSymbols` is an empty map), and `lookupAxiomSimplifier` would return `Nothing`.

After the changes in this PR, we match the "MAP.concat" hook with the builtin evaluator `Kore.Builtin.Map.evalConcat`, and return it in a `Just`.

----

Is the test's expectation wrong (and therefore needs to be updated), or is this a legit use-case that this PR breaks?

goodlyrottenapple:
> Is the test's expectation wrong (and therefore needs to be updated), or is this a legit use-case that this PR breaks?

I am not sure as still fairly new to the codebase. @ana-pantilie would know. Given the name of the test, perhaps we just want to make a copy of `Mock.concatMap` which has a dummy hook instead?

ana-pantilie:
@dcastro I think the test's output was actually subtly wrong because of what you said, so feel free to update it with the current result! Thanks.

breakerzirconia:
The actual result of the only failing test case `Test.Kore.Unification.Unifier.test_unification: "key outside of map, symbolic opaque terms"` now expects the [Type](https://github.com/runtimeverification/haskell-backend/blob/diogo/%233111-bypass-map-lookup/kore/src/Kore/Attribute/Pattern/Simplified.hs#L67) of [constr20](https://github.com/runtimeverification/haskell-backend/blob/diogo/%233111-bypass-map-lookup/kore/test/Test/Kore/Unification/Unifier.hs#L657) to be `Fully` instead of `Partly`, meaning these changes invalidate the `Simplified` attribute.

PR #3192, opened 2022-08-10
[3143] Modifications to the JSON-RPC API for `execute` request

jberthold:
To accommodate the requirements on the future `execute` implementation, the API description was modified in the following way:

In the response to an `execute` request:
* added current `state` field
* added `rule` field (optional, filled for `cut-point-rule` or `terminal-rule`)
* renamed `states` to `next-states` (optional, filled for `branching` or `cut-point-rule`)
* avoid repetition of the `depth` field

See #3143 for additional information.

Fixes #3143

### Scope:
- [X] Requirements analysis and implementation design for RPC `execute`
- [X] Adaptation of RPC server stub code
- [x] Prototype implementation of the new `execute` back-end

### Estimate:

---

###### Review checklist

The author performs the actions on the checklist. The reviewer evaluates the work and checks the boxes as they are completed.

-   [ ] **Summary.** Write a summary of the changes. Explain what you did to fix the issue, and why you did it. Present the changes in a logical order. Instead of writing a summary in the pull request, you may push a clean Git history.
-   [ ] **Documentation.** Write documentation for new functions. Update documentation for functions that changed, or complete documentation where it is missing.
-   [ ] **Tests.** Write unit tests for every change. Write the unit tests that were missing before the changes. Include any examples from the reported issue as integration tests.
-   [ ] **Clean up.** The changes are already clean. Clean up anything near the changes that you noticed while working. This does not mean only spatially near the changes, but logically near: any code that interacts with the changes!

PR #3199, opened 2022-08-12
[3198] fix stopping prover on branches + logging unproven

### LINKED ISSUE #3198: pyk integration test failure since new graph traversal
The rewrite in #3168 causes a pyk integration test ( `test_get_basic_block (pyk.integration_tests.test_kprove.ImpProofTest)` in the K repository) to fail. It appears to output `#Top` where it should return an unproven configuration.

jberthold:
* The rewritten prover was returning a final instead of a stopped state when stopping at branches
* Proof depth for unproven claims was not logged at info level as before.

This fixes both problems, and adds a unit test for stopping at branches.

Fixes #3198

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Exec/GraphTraversal.hs
+++ b/kore/src/Kore/Exec/GraphTraversal.hs
@@ -19,6 +19,7 @@ import Control.Monad (foldM)
 import Control.Monad.Trans.State
 import Data.Limit
 import Data.List.NonEmpty qualified as NE
+import Data.Maybe (maybeToList)
 import Data.Sequence (Seq (..))
 import Data.Sequence qualified as Seq
 import GHC.Natural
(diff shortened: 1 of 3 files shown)


kore/src/Kore/Exec/GraphTraversal.hs
```diff
@@ -263,8 +267,8 @@ graphTraversal
             Seq (TState instr config) ->
             Simplifier (StepResult (TState instr config))
         step a q = do
-            next <- transit a
-            if (isStuck next || isFinal next || isStopped next || shouldStop next)
```
jberthold, line 267: The problem was that `shouldStop` was used here, but not in `checkLeftUnproven` below where "unproven" (stopped) states are turned into a different result.
A cleaner solution is to turn `Branch` into `Stopped`. The open PR #3192 will refactor this to keep the next states in `Stopped`, too, so no information is lost.

ana-pantilie (approved):

PR #3202, opened 2022-08-15
Eliminate top level existentials in side-condtions

### LINKED ISSUE #3193: [Simplification request] Top-level existential elimination
Simplification
--------------

Existential quantification in top-level positive constraints can be eliminated and the bound variables can be replaced with some fresh variables.

Example: ```#Exists X. { P(X) } => P(X0)```, for some `X0` fresh.

I am assuming that this is a sound simplification when doing matching-logic proofs since it is sound when doing first-order logic proofs, and matching logic captures first-order logic.

This simplification cannot be applied under negation.

goodlyrottenapple:
Eliminate existentials in side-condtions by replacing with a fresh type variable

Fixes #3193


kore/src/Kore/Simplify/Condition.hs
```diff
@@ -168,15 +177,33 @@ simplifyPredicates sideCondition original = do
     let predicates =
             SideCondition.simplifyConjunctionByAssumption original
                 & fst . extract
-    simplifiedPredicates <-
+    simplifiedPredicates <- do
+        let eliminatedExists =
+                map (simplifyPredicateExistElim $ freeVariableNames original <> freeVariableNames sideCondition) $
+                    toList predicates
```
ana-pantilie, line 183: This line is pretty long, it should be reformatted a bit.

I'm not entirely sure if it's necessary to avoid variable names in the sidecondition, but let's leave it like this for now. Could you add a TODO comment there about this?

kore/src/Kore/Simplify/Condition.hs
```diff
@@ -168,15 +177,33 @@ simplifyPredicates sideCondition original = do
...
             sideCondition
-            (toList predicates)
+            eliminatedExists
     let simplified = foldMap mkCondition simplifiedPredicates
     if original == simplifiedPredicates
         then return (Condition.markSimplified simplified)
         else simplifyPredicates sideCondition simplifiedPredicates
 
+{- | Simplify an existential predicate by removing the existential binder
+under the assumption that the others are true.
```
ana-pantilie, line 193: > under the assumption that the others are true

Not sure what this refers to?

kore/test/Test/Kore/Simplify/Pattern.hs
```diff
@@ -54,10 +54,7 @@ test_Pattern_simplify =
                     (Mock.constr10 fOfX)
                     ( makeAndPredicate
                         (makeCeilPredicate fOfX)
-                        ( makeExistsPredicate
-                            Mock.yConfig
-                            (makeCeilPredicate fOfY)
-                        )
+                        (makeCeilPredicate fOfY)
```
ana-pantilie, line 57: Reminder about what we discussed during the meeting, to modify the test names to include what's happening with the exists there.

ana-pantilie (approved):
Looks good! Just a few comments.

kore/src/Kore/Simplify/Condition.hs
goodlyrottenapple: sorry, i think that was a copy-paste from the comment below

PR #3206, opened 2022-08-18
Implement new pattern matcher

dwightguth:
Essentially this PR replaces the `matchIncremental` function which was used for all pattern matching previously with a new function named `patternMatch` with the same signature and behavior, but which is much more efficient.

The code contains a high level description of the algorithm in comments. I don't have much in the way of concrete numbers, but it seems to have reduced the time spent in pattern matching in the example on the C semantics I profiled from roughly 15% of the profiled time to <1%. Once I make sure the integration tests are passing, I will do some more extended profiling.

### Estimate:

---

###### Review checklist


jberthold (commented):
LGTM overall, some stylistic nits.
I let others with more knowledge of the AC structures do the final approval

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+    patternMatch' sideCondition [MatchItem pat subject [] Set.empty] MinQueue.empty MultiAnd.top Map.empty
 
-data TermLikeClass
-    = Variables
-    | ConcreteBuiltins
-    | ConstructorAtTop
-    | OtherTermLike
-    | ListBuiltin
-    | AssocCommBuiltin
+data MatchItem = MatchItem (TermLike RewritingVariableName) (TermLike RewritingVariableName) [(SomeVariable RewritingVariableName, SomeVariable RewritingVariableName)] (Set (SomeVariableName RewritingVariableName))
```
goodlyrottenapple, line 141: Can this be made into a record type? I don't know what any of these fields are at a glance.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
     TermLike RewritingVariableName ->
     TermLike RewritingVariableName ->
-    Simplifier (Maybe (MatchResult RewritingVariableName))
-matchIncremental sideCondition termLike1 termLike2 =
-    evalRWST matcher sideCondition initial
-        & fmap fst
+    Bool
+needs (InternalSet_ s) term = needsAc (builtinAcChild s) term
+needs (InternalMap_ m) term = needsAc (builtinAcChild m) term
+needs _ _ = False
```
goodlyrottenapple, line 159: what if the InternalSet/Map is nested inside the pattern? Or does such case not arise/matter?

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+
+    unwrapMapElement ::
+        Element NormalizedMap ->
+        Element NormalizedMap ->
+        [(TermLike RewritingVariableName, TermLike RewritingVariableName)]
+    unwrapMapElement elem1 elem2 =
+        let (key1, val1) = getMapElement elem1
+            (key2, val2) = getMapElement elem2
+         in [(key1, key2), (val1, val2)]
+patternMatch' sideCondition ((MatchItem pat subject boundVars boundSet) : rest) deferred predicate subst = do
```
ana-pantilie, line 267: Please refactor `patternMatch'` so that there's only one main body of the function. That's where you should case split on the arguments and call different worker functions which should be named suggestively, according to the job they do.

Right now this function is very hard to read because it spans over 500 LOC and the top level part of it is scattered throughout the file.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+    tools <- Simplifier.askMetadataTools
+    injSimplifier <- Simplifier.askInjSimplifier
+    overloadSimplifier <- Simplifier.askOverloadSimplifier
+    let InjSimplifier{matchInjs, evaluateInj} = injSimplifier
+    case (pat, subject) of
+        (Var_ var1, Var_ var2) | isFree var1, var1 == var2 -> discharge
+        (ElemVar_ var1, _) | isFree (inject var1), isFunctionPattern subject -> bind (inject var1) subject
+        (SetVar_ var1, _) | isFree (inject var1) -> bind (inject var1) subject
+        (Var_ var1, Var_ var2) | not $ isFree var1, var1 `isBoundToSameAs` var2 -> discharge
+        (StringLiteral_ str1, StringLiteral_ str2) -> if str1 == str2 then discharge else failMatch "distinct string literals"
```
ana-pantilie, line 277: Can you please format the rest of this function like the following suggestion?

```suggestion
        (ElemVar_ var1, _)
        | isFree (inject var1)
        , isFunctionPattern subject ->
            bind (inject var1) subject
        (SetVar_ var1, _) | isFree (inject var1) ->
            bind (inject var1) subject
        (Var_ var1, Var_ var2)
        | not $ isFree var1
        , var1 `isBoundToSameAs` var2 ->
            discharge
        (StringLiteral_ str1, StringLiteral_ str2) ->
            if str1 == str2
                then discharge
                else failMatch "distinct string literals"
```

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+                (InternalList_ InternalList{internalListChild = l1}, InternalList_ InternalList{internalListChild = l2}) ->
+                    if length l1 == length l2
+                        then decomposeList $ zip (toList l1) (toList l2)
+                        else failMatch "list lengths are not equal"
+                (App_ symbol [InternalList_ InternalList{internalListChild = l1}, var@(ElemVar_ _)], InternalList_ InternalList{internalListChild = l2})
+                    | List.isSymbolConcat symbol ->
+                        if length l1 <= length l2
+                            then
+                                let l2' = Seq.drop (length l1) l2
+                                 in decomposeList $ (var, List.asInternal tools sort l2') : zip (toList l1) (toList (Seq.take (length l1) l2))
```
ana-pantilie, line 341: Please factor this out and name it something suggestive.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+                            else failMatch "subject list is too short"
+                (App_ symbol [var@(ElemVar_ _), InternalList_ InternalList{internalListChild = l1}], InternalList_ InternalList{internalListChild = l2})
+                    | List.isSymbolConcat symbol ->
+                        if length l1 <= length l2
+                            then
+                                let l2' = Seq.take (length l2 - length l1) l2
+                                 in decomposeList $ (var, List.asInternal tools sort l2') : zip (toList l1) (toList (Seq.drop (length l2 - length l1) l2))
```
ana-pantilie, line 348: Same as above.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
-        | otherwise = field @"predicate" <>= definedTermLike
+        Simplifier (Either Text (MatchResult RewritingVariableName))
+    bind var term =
+        let varName = variableName var
+            freeVars = FreeVariables.toNames (freeVariables term)
+         in if not $ Set.disjoint freeVars boundSet
+                then failMatch "bound variable would escape binder"
+                else case Map.lookup varName subst of
+                    Nothing -> patternMatch' sideCondition rest deferred (isTermDefined var term) (Map.insert (variableName var) term subst)
+                    Just binding -> if binding == term then patternMatch' sideCondition rest deferred predicate subst else failMatch "nonlinear matching fails equality test"
```
ana-pantilie, line 374: These calls to `patternMatch'` are recursive, right? Is there no way to avoid this, or to refactor it in such a way so that the recursive calls are more explicit/at the top?

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+
+    isTermDefined ::
+        SomeVariable RewritingVariableName ->
+        TermLike RewritingVariableName ->
+        MultiAnd (Predicate RewritingVariableName)
+    isTermDefined var term
+        | SideCondition.isDefined sideCondition term || isSetVariable var = predicate
```
ana-pantilie, line 381: Why `|| isSetVariable`?

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
 import Pair
 import Prelude.Kore
 
 -- * Matching
 
-newtype Constraint variable = Constraint
-    { getConstraint :: Pair (TermLike variable)
-    }
-    deriving stock (Eq)
+{- | Match a TermLike against a pattern as represented by another TermLike.
```
ana-pantilie, line 99: I appreciate this documentation, but I think it would be more helpful to add comments to the other functions in this file as well. Just a few lines describing what they do, and how they fit into this algorithm.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
jberthold: Very good point, actually. This is in fact a state-monadic computation; each case above ultimately has to `failMatch` or call `patternMatch'` recursively.
The old code used `RWST` to separate the side condition, which is probably not necessary, but it would greatly improve readability if this code was using a state monad. Of course this is a big refactoring.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -81,696 +68,401 @@ import Kore.Internal.Predicate (
...
+        (InternalSet_ set1, InternalSet_ set2) ->
+            matchNormalizedAc decomposeList unwrapSetValue unwrapSetElement (wrapSet set2) (unwrapAc $ builtinAcChild set1) (unwrapAc $ builtinAcChild set2)
+        _ -> error "error in matching algorithm: unexpected deferred term"
   where
-    normalized1 = unwrapAc $ builtinAcChild set1
-    normalized2 = unwrapAc $ builtinAcChild set2
-    pushSetValue _ = return ()
-    pushSetElement = push . fmap getSetElement
-    wrapTermLike unwrapped =
+    (MatchItem pat subject boundVars boundSet, rest) = MinQueue.deleteFindMin queue
```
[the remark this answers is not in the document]
dwightguth: I considered doing this but then MinQueue.deleteFindMin would be evaluated even if the queue is empty, which would generate an error. conversely, if I move deleteFindMin into a let expression, then the bindings it creates can no longer be accessed by the functions defined in the where clause. Is there a way to do what you describe that doesn't run into either of these problems?

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: I can, but I gave it a go and it made the code ridiculously more verbose. Are you sure you don't just want me to add a comment explaining what each of the fields are instead?

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: This is covered by the false case; in such cases, the ordering of elements doesn't really matter and thus we don't need to do anything special.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: This is subject to the same issue I mentioned above; putting everything under a single binding would mean that the functions in the where clause would no longer be able to pattern match on fields within the MatchItem being analyzed, which would make the code even harder to read. If you know how to avoid that issue, I'm game, though.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: I'm not entirely sure what you're asking me to factor out here. As far as I can tell this entire code block is something that doesn't really repeat itself anywhere else. I don't think factoring out a function here would really make it more readable. But if you clarify what you're looking for here, I might be able to do something.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: I'm not hugely convinced it makes the code that much more readable; The recursive calls to patternMatch' are already factored out into methods that are pretty clear what their meaning is. It might help things somewhat, but probably not hugely, and it would be a pretty substantial code change. If it turns out to add overhead (and I don't see how it possibly wouldn't), then it wouldn't even be a change worth making since our goal in this PR is to make pattern matching as fast as possible.

I'm willing to give it a try after all other code suggestions on this PR have been resolved if you still think it's worth trying, though.

kore/src/Kore/Rewrite/Axiom/Matcher.hs
dwightguth: `isTermDefined` is a function intended to add a \ceil predicate for each element variable that gets added to the substitution. If it's a set variable, the ceil predicate is not needed because set variables are allowed to match \bottom.

ana-pantilie (approved):
Approving, despite refactoring concerns. We'll fix those later, since they're not trivial to do.

jberthold (approved):
Concur, let's merge and try the suggested refactorings in a follow-up

kore/src/Kore/Rewrite/Axiom/Matcher.hs
goodlyrottenapple: Yes comments would definitely help

goodlyrottenapple (approved):

goodlyrottenapple:
The flaky CI that you disabled in your branch should be fixed in master now, so best to just let master overwrite your change.

PR #3211, opened 2022-08-21
Use reallyUnsafePtrEquality# for faster Id comparisons

int-index:
WIP, pushing to see CI output.


gromakovsky:
No big changes performance-wise unfortunately.
[report-dedup-text_master-6.pdf](https://github.com/runtimeverification/haskell-backend/files/9416078/report-dedup-text_master-6.pdf)

gromakovsky:
Our solution in #3217 seems to work and be good enough, so this PR is no longer necessary.

Issue #3061, opened 2022-05-18
JSON-RPC for driving symbolic execution


goodlyrottenapple:
This is a proposal to switch to using the JSON RPC protocol to talk with kore-repl/kore-exec. The main proposed change would be to remove the current REPL and replace it with a JSON RPC server which would handle the "core" functionality of the current REPL, namely all the functionality such as step/stepf, config, select, etc. excluding any complex commands like piping or appending to file, which would instead be delegated to a new REPL. This new REPL could be written in Haskell or any potentially any other language, e.g. Python, and would simply be a JSON RPC client together with the REPL parser pretty printer, etc. This REPL could potentially live in the K framework repo and thus have better access to higher level functionality such as the pretty printing of kore ASTs back into K (since `config` in the current REPL returns a kore AST). A potential advantage of this approach is a common interface for both a REPL and the summarizer to interact with the haskell backend. 

Propsed features
* new REPL should mirror the current REPL functionality
* have built in pretty printing of K terms (discussed in #1890, #2797)
* should be interruptible/cancellable  #1691
* have state diffing built in

Questions
* What language to use for the new REPL and which repo to put it in?
* Could the LLVM backend also have a JSON RPC interface? Would this make sense?

### Edit: 12 July 2022
We met and discussed that we need the following:

- [x] #3143
- [x] #3144
- [x] #3148 
- [x] #3146

The step request will just be a particular case of the execute request. The execute request will also return the branching states etc. together with the parent node of the branch.

The check-implication request can just expose the current check implication procedure we have, with some minimal changes.

goodlyrottenapple:
# Implementation details

We could use [servant-jsonrpc](https://hackage.haskell.org/package/servant-jsonrpc-1.1.1/docs/Servant-JsonRpc.html) to pull out the current functionality of the REPL into the following API

```haskell
type Step   = JsonRpc "step"   Int Text [Int] 
```
_The `Step` method expects an Int argument and either returns a `Text` error or a list of new states in case of branching. Maybe instead of `[Int]` we should return a hash of the state as it appears in the summarizer?? Or return both?_

```haskell
type Stepf  = JsonRpc "stepf"  Int Text [Int] 
type Select = JsonRpc "select" Int Text ()
```
_Should `Select` take the hash instead? Or maybe have two selects, one with an int and the other with a hash?_
```haskell
type Config = JsonRpc "config" Int Text Text
```
[...]

PR #3212, opened 2022-08-22
[3208] implement implies endpoint

### LINKED ISSUE #3208: JSON-RPC: implement check-implication

jberthold:
Fixes #3208

Part of #3061 

### Scope:

* [X] New implementation for `checkImplication` which only works with a single consequent (not an `OrPattern`)
* [X] RPC server endpoint `implies` calls this new implementation
* [x] Integration tests for the new endpoint implementation
* [ ] Revision of the algorithm to compute and return a simplified predicate if the implication cannot be proven. 

### Estimate:

---

###### Review checklist


kore/src/Kore/Reachability/Claim.hs
```diff
@@ -655,6 +662,158 @@ checkImplicationWorker (ClaimPattern.refreshExistentials -> claimPattern) =
...
+                        let stuckClaim =
+                                (claimPattern :: ClaimPattern){left = OrPattern.toPattern sort (fst stuck)}
+                         in pure $ NotImpliedStuck (stuckClaim, Just $ snd stuck)
+  where
+    sort = termLikeSort leftTerm
+
+    claimPattern@ClaimPattern{left = renamedLeft, right = orRight, existentials} =
+        ClaimPattern.refreshExistentials $
+            -- ensures existentials of RHS do not occur free in LHS
+            ClaimPattern.mkClaimPattern inLeft (OrPattern.fromPattern inRight) inExistentials
```
jberthold, line 769: Not quite sure here, we should maybe rather _check_ that the input data does not produce name clashes when removing existentials, and return an informative error otherwise. That (new) error type might be appropriate for the `assertFunctionLikeConfiguration` above, too.

ana-pantilie:
1. After parsing the two terms, we need to call https://github.com/runtimeverification/haskell-backend/blob/454211800a577c6b8d13c28457ebcdfffe2a871c/kore/src/Kore/Rewrite/ClaimPattern.hs#L436 and `termToExistentials`
2. That will give us two Patterns of the form (term, predicate, []). Call `Pattern.simplify` on the two Patterns so that the substitutions get put into the right places.
3. We need to expect that the two resulting OrPatterns are either empty or singletons. Throw an error if otherwise.
4. Response should be:
     - Implied: return unification condition and true
     - NotImplied: return false
     - NotImpliedStuck: return left and unification condition and return false (or unknown)

kore/src/Kore/Reachability/Claim.hs
```diff
@@ -655,6 +669,180 @@ checkImplicationWorker (ClaimPattern.refreshExistentials -> claimPattern) =
...
+        let nameCollisions =
+                existentials `intersect` getFreeElementVariables (freeVariables left)
+        withContext ("LHS: " <> showTerm leftTerm)
+            . withContext ("RHS: " <> showTerm rightTerm)
+            . withContext ("existentials: " <> show (map unparse2 existentials))
+            $ koreFailWhen (not $ null nameCollisions) $
+                unwords
+                    ( "Existentials capture free variables of the LHS:" :
+                      map (show . unparse2) nameCollisions
+                    )
```
ana-pantilie, line 792: I remember you had a question about this yesterday. I think this approach is good, they can take care of making sure there aren't collisions instead of delegating this to us. We'll see in the future if requirements change.

test/rpc-server/implies/implied-substitution/response.golden
```diff
@@ -0,0 +1 @@
+{"result":{"implication":{"format":"KORE","term":{"tag":"Implies","first":{"tag":"EVar","name":"a","sort":{"args":[],"tag":"SortApp","name":"SortK"}},"second":{"tag":"Exists","var":"x","arg":{"tag":"EVar","name":"x","sort":{"args":[],"tag":"SortApp","name":"SortK"}},"varSort":{"args":[],"tag":"SortApp","name":"SortK"},"sort":{"args":[],"tag":"SortApp","name":"SortK"}},"sort":{"args":[],"tag":"SortApp","name":"SortK"}},"version":1},"satisfiable":true,"condition":{"substitution":{"format":"KORE","term":{"tag":"Equals","first":{"tag":"EVar","name":"a","sort":{"args":[],"tag":"SortApp","name":"SortK"}},"second":{"tag":"EVar","name":"x","sort":{"args":[],"tag":"SortApp","name":"SortK"}},"sort":{"tag":"SortVar","name":"JSONSortVariable"},"argSort":{"args":[],"tag":"SortApp","name":"SortK"}},"version":1},"predicate":{"format":"KORE","term":{"tag":"Top","sort":{"args":[],"tag":"SortApp","name":"SortK"}},"version":1}}},"jsonrpc":"2.0","id":1}
```
ana-pantilie, line 1: If I'm reading this right, the substitution looks like `a #Equals x`.

This is out of the scope of this PR, but we should make sure that the substitutions' order is preserved. One situation where this matters is when you have `var1 #Equals var2`, the way you order that will give different results.

Please open a subsequent PR when you have a chance which fixes this.

test/rpc-server/implies/not-implied-stuck/README.md
```diff
@@ -0,0 +1 @@
+`a => b`, response `False`, with substitution `[b/a]`
```
ana-pantilie, line 1: Why is this stuck? Also, we usually prefer variables to be have names starting with an upper-case letter (like enforced in K). I initially thought `a` and `b` were symbols (specifically, constructors) until I saw the JSON.

ana-pantilie (commented):
There's one test which seems to not be working (or I might have misunderstood something). I'd also like to see what Xiaohong says about the algorithm, we'll probably have his response today so it won't be blocking.

Great job on this, it's a difficult one, and now we have a much better understanding of the issues here and in the main algorithm.

Also, I really like the integration testing infrastructure!

test/rpc-server/implies/implied-substitution/response.golden
jberthold: > If I'm reading this right, the substitution looks like `a #Equals x`.

That's right.
 
> This is out of the scope of this PR, but we should make sure that the substitutions' order is preserved. One situation where this matters is when you have `Var1 #Equals Var2`, the way you order that will give different results.

I hope the order is indeed stable. The code that translates substitutions to json is `Kore.Syntax.Json.fromSubstitution`, and goes left-to-right (`foldl1`) over `Substitution.unwrap`
https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Syntax/Json.hs#L115
 
> Please open a subsequent PR when you have a chance which fixes this.

test/rpc-server/implies/not-implied-stuck/README.md
jberthold: I first thought this fails because the `b` is all-quantified (instead of existentially) but thinking about it again, you are right, this should actually _pass_. Maybe another case similar to #3218 ? Will rework the test to use a different example for now.

ana-pantilie: We should add a check which makes sure that any extra RHS variables are existentially quantified.

ana-pantilie (approved):
👍

PR #3122, opened 2022-06-27
Intern IDs

develop7:
Fixes #3112 

### Benchmark

[report-intern-ids.md](https://github.com/runtimeverification/haskell-backend/files/9160046/report-intern-ids.md)
[report.pdf](https://github.com/runtimeverification/haskell-backend/files/9160047/report.pdf)

---

###### Review checklist

The author performs the actions on the checklist. The reviewer evaluates the work and checks the boxes as they are completed.

-   [ ] **Summary.** Write a summary of the changes. Explain what you did to fix the issue, and why you did it. Present the changes in a logical order. Instead of writing a summary in the pull request, you may push a clean Git history.
-   [ ] **Documentation.** Write documentation for new functions. Update documentation for functions that changed, or complete documentation where it is missing.
-   [ ] **Tests.** Write unit tests for every change. Write the unit tests that were missing before the changes. Include any examples from the reported issue as integration tests.
-   [ ] **Clean up.** The changes are already clean. Clean up anything near the changes that you noticed while working. This does not mean only spatially near the changes, but logically near: any code that interacts with the changes!

### Currently failing unit tests
https://github.com/runtimeverification/haskell-backend/runs/7230663172?check_suite_focus=true#step:8:27642 
https://github.com/runtimeverification/haskell-backend/runs/7230662258?check_suite_focus=true#step:7:15278 https://github.com/runtimeverification/haskell-backend/runs/7230879449?check_suite_focus=true#step:5:27905

PR #3215, opened 2022-08-23
[WIP] Intern IDs

### LINKED ISSUE #3112: Serokell: finish interning strings
### Progress until 07/10/2022
 
- [x] add string interning to the whole codebase
- [x] fix unit test failures coming from different hashable versions
- [ ] serialize interned table
- [ ] fix the rest of the unit test failures

------

Note: Code for this is currently on the `intern-ids` branch (commit 2574d0d287dfab77ef3e5350ee6ef1d51ad5fa82).

Identifiers are isolated in the `Kore.Syntax.Id` module. The existing work uses Edward Kmett's `intern` package; I did not explore alternatives. Some work was needed to handle interning the identifiers without affecting the module API, which I've already completed.^1 The `Eq Id` instance has been modified to take advantage of interning, and I measured approximately an 8% speedup on our integration test suite.

The remaining work I wanted to do was to get the `Ord` instance to take advantage of interning as well. I do not believe that our algorithm requires alphabetically sorted `Id`s anywhere. However our tests do assume this, so I'm not certain. Also, sorting by intern unique is not obviously sound - some data persists between invocations of the backend via serialization, and the ordering of `Id`s may change between runs if they are interned in a different order. In the meeting it was suggested that maybe interning can be done entirely at parse-time, which I think may make this a non-issue :)

Finally, Kmett's `intern` package is somewhat old and he only maintains it because it has users; I imagine that his `InternedText` works quite well, but it may be good to compare with a custom implementation.

----------------------

^1: During the meeting, it was suggested to make interning contextual (passing a map around to different functions); I think this would require the Id API to change as `field @getId` could no longer be a well-typed lens. I also think it would be slower than the global map used by the `intern` package.

gromakovsky:
Fixes #3112

It's #3122 without fixes for unit tests. Currently the main focus is on integration tests.

### Estimate:

---

###### Review checklist


gromakovsky:
Closing in favor of #3217.

PR #3217, opened 2022-08-25
[#3112] Intern IDs

### LINKED ISSUE #3112: Serokell: finish interning strings

breakerzirconia:
Fixes #3112

### Scope:

Profiling revealed that there are a lot of operations over identifiers (i.e. `Kore.Syntax.Id`). Operations such as `==` and `hash/hashWithSalt` operate over text values which can be slow. 

This PR uses string interning to allow for faster implementations of `Eq`/`Hashable`/`Ord`.

* Each interned string has a unique ID. 
* This ID is generated via an incremental counter. Thus, a string's ID may be different across multiple runs.
* `Eq` simply checks if the two strings' IDs are equal
* `Hashable` hashes the ID. NB: since the ID may be different across multiple runs, hashes are not stable.
* `Ord` uses a "fast path" branch to check if the IDs are equal; if so, it returns `EQ`. Otherwise, it falls back on comparing the strings' contents.

We managed to squeeze a ~10% speedup. The report is available [here](https://github.com/runtimeverification/haskell-backend/files/9487622/report-refact_master-7.pdf), using 454211800a577c6b8d13c28457ebcdfffe2a871c from master as the baseline.

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Syntax/Id.hs
+++ b/kore/src/Kore/Syntax/Id.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE BlockArguments #-}
+
 {- |
 Copyright   : (c) Runtime Verification, 2018-2021
 License     : BSD-3-Clause
@@ -7,10 +9,14 @@ Please refer to <http://github.com/runtimeverification/haskell-backend/blob/mast
 module Kore.Syntax.Id (
     -- * Identifiers
     Id (..),
+    pattern Id,
+    getId,
+    idLocation,
     getIdForError,
     noLocationId,
     implicitId,
     generatedId,
+    globalIdMap,
 
     -- * Locations
     AstLocation (..),
@@ -18,6 +24,10 @@ module Kore.Syntax.Id (
     prettyPrintAstLocation,
 ) where
 
+import Control.Lens (lens)
+import Data.Generics.Product
+import Data.HashMap.Strict as HashMap
+import Data.IORef
 import Data.String (
     IsString (..),
  )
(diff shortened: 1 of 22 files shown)


gromakovsky:
We've got results of benchmarks for `36c7c4f3e4e0887113ddcde4bdfc9c58c1757800` vs `master`:
[report-new-ord_master-7.pdf](https://github.com/runtimeverification/haskell-backend/files/9459191/report-new-ord_master-7.pdf)

The difference is about 4%, which is not 9-10% we had before, but still a measurable improvement. Probably that's because in 36c7c4f3e4e0887113ddcde4bdfc9c58c1757800 `Ord` still compares text values, but `Eq` compares numbers. There are a few more things we should do:
1. Do not use `HashMap.size` (which is `O(n)`), store counter instead.
2. Check whether `        | a == b = EQ` case is needed. If compared identifiers are often equal, it should be a useful optimization, but if they are usually different, it's just a small overhead.
3. Optimize `Hashable` instance. We tried doing that and broke one unit test.

We can also change `Ord` to always compare numeric IDs, but it causes a lot of troubles which might be fixable with some hacks, but I'd rather avoid them.

kore/src/Kore/Syntax/Id.hs
```diff
@@ -47,7 +47,7 @@ import System.IO.Unsafe (unsafePerformIO)
 
 data InternedTextCache = InternedTextCache
     { counter :: {-# UNPACK #-} !Word
-    , internedTexts :: !(HashMap Text Word)
+    , internedTexts :: !(HashMap Text InternedText)
```
gromakovsky, line 50: Does it increase the size of serialized data?

dcastro: Haven't measured it but I think it should only increase the hashmap size by 1 pointer per entry.

The reason being: the `Text` object that the entry's key points to, is the same `Text` object that the `InternedText` value points to:

```hs
internText text =
  ...
                    HashMap.alterF
                        \case
                            ...
                            -- Otherwise, create a new ID for it and intern it.
                            Nothing ->
                                let newIden = counter
                                    newInterned = InternedText text newIden
                                    ^^^^ ***the new value, containing a pointer to the same text object as the key***
                                 in ((newInterned, counter + 1), Just newInterned)
                        text
                        ^^^^ ***the new key***
```

So the old memory layout would have looked like this:

```
Entry: Key           Value
        |              |
        |              |
        ↓              ↓
        Text         Word (with an unpacked #Word)
```

And it should now look something like this (notice the extra pointer `InternedText -> Text`:

```
Entry: Key           Value
        |              |
        |              |
        ↓              ↓
        Text <--- InternedText (with an unpacked #Word)
```


On the other hand, wherever an interned string is used, it'll now reuse the same `InternedString` box, instead of creating a new `InternedString` box every time we intern a string. So we'll likely save some memory (and a negligible amount of time) there.

Regarding the serialization itself, it's done using [compact regions](https://hackage.haskell.org/package/compact-0.2.0.0), so the file's data layout should resemble the in-memory layout.

gromakovsky: Ok, sounds good, thank you for this explanation :sunglasses:

gromakovsky (approved):
Glad to see we managed to make it work 👍

jberthold (approved):
LGTM, :tada:

ana-pantilie (approved):
Excellent work! 🎉

PR #3225, opened 2022-08-30
Fix exec depth off-by-one, add unit tests

### LINKED ISSUE #3210: kore-rpc gives back incorrect depth for stuck states
From: https://github.com/runtimeverification/k/pull/2805#discussion_r949120009

On a transition system:

```
module KORE-RPC-TEST-SYNTAX
    imports INT-SYNTAX
endmodule


module KORE-RPC-TEST
    imports KORE-RPC-TEST-SYNTAX

    rule [r01]: 0 => 1
    rule [r12]: 1 => 2
    rule [r23]: 2 => 3
    rule [r24]: 2 => 4
    rule [r45]: 4 => 5
    rule [r56]: 5 => 6

endmodule
```

We call `execute` from the `kore-rpc`, with input `4`, and get back final result `StuckResult(state=state(6), depth=3)`, but it should be `depth=2` instead.

jberthold:
Fixes #3210

The execution depth information was wrong in cases where a state could not be rewritten further:
In the last execution step, rewriting would yield program state `Remaining` but the execution depth counter was incremented.

### Scope:

~This change uses the prior state in such a case, but only if the state did not actually change. In theory, the remaining state could still have been simplified to something different from the prior. However, in practice each rewriting step ends with a simplifier run so the prior state should already be simplified to the extent possible.~

EDIT The actual problem was buried in the `trackExecDepth` function, so fixing that solved the problem without any complicated logic.

---

###### Review checklist


kore/src/Kore/Exec.hs
```diff
@@ -458,17 +458,16 @@ rpcExec
...
 trackExecDepth transit prim (execDepth, execState) = do
     execState' <- transit prim execState
     let execDepth' = (if didRewrite execState' then succ else id) execDepth
     pure (execDepth', execState')
   where
-    didRewrite _ = isRewrite prim
-
-    isRewrite Rewrite = True
-    isRewrite _ = False
+    didRewrite Kore.Rewrite.Bottom = prim == Rewrite
```
goodlyrottenapple, line 468: this seems like the meat of the issue. could you add an in code comment what this is checking?

goodlyrottenapple (approved):
LGTM

PR #3227, opened 2022-08-31
Hoogle targets in Makefile

jberthold:
This adds targets for using a local `hoogle` to the `Makefile`.
`hoogle` must be installed (e.g., `cabal install hoogle`) for this to work.

* `make hoogle-data` creates a hoogle database from haddock documentation
  (will build and run `haddock` unless it has been run before, uses `.stack-work-haddock` work directory)
* `make hoogle` prompts the user for a `Query` and runs it:
```
$ make hoogle
Hoogle 5.0.18.3, https://hoogle.haskell.org/
Query:a -> [a]
Prelude repeat :: a -> [a]
Data.List repeat :: a -> [a]
GHC.List repeat :: a -> [a]
GHC.OldList repeat :: a -> [a]
Data.List.Compat singleton :: a -> [a]
Test.QuickCheck shrinkNothing :: a -> [a]
Test.QuickCheck.Arbitrary shrinkNothing :: a -> [a]
Distribution.Compat.Prelude.Internal repeat :: a -> [a]
Hedgehog.Internal.Prelude repeat :: a -> [a]
GhcPrelude repeat :: a -> [a]
-- plus more results not shown, pass --count=20 to see more
```
* `make hoogle-server` starts a local (https) server on `https://localhost:8080` for ongoing queries.
   - The server stays in the foreground and needs to be terminated with `^C`.
   - This is a `https` server, but using a bogus certificate, so a security warning will pop up.
   - The server provides links to documentation in local haddock files which won't work remotely.

Fixes #nnnn

### Estimate:

---

###### Review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,8 +79,9 @@ clean-execution:
 
 .PHONY: hoogle-cert hoogle-server
 HOOGLE_WORKDIR = .stack-work-haddock/hoogle
-HOOGLE        := $(shell which hoogle)
+HOOGLE        ?= $(shell which hoogle)
 HOOGLE_DB     := $(HOOGLE_WORKDIR)/*/*/*/database.hoo
+
 
 hoogle-data: $(HOOGLE_DB)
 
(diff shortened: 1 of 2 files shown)


goodlyrottenapple:
would this not be better as a nix package? something like
```
nix run .#hoogle-server
```
don't need to rely on a global install of hoogle then

goodlyrottenapple (approved):

ana-pantilie (approved):
Thanks for bringing this back. Could you also remove the hoogle targets from https://github.com/runtimeverification/haskell-backend/blob/4fda2ba9ddd7d36ef81b3d15a606bbc33b0b7701/kore/Makefile#L65? Leaving them there would become a source of confusion.

jberthold:
> would this not be better as a nix package? something like
> 
> ```
> nix run .#hoogle-server
> ```
> 
> don't need to rely on a global install of hoogle then

Yes, that would probably be even better - if it can reuse previously-built haddock files. Maybe we can rework this in a follow-up.

jberthold:
> Thanks for bringing this back. Could you also remove the hoogle targets from
> https://github.com/runtimeverification/haskell-backend/blob/4fda2ba9ddd7d36ef81b3d15a606bbc33b0b7701/kore/Makefile#L65
> ? Leaving them there would become a source of confusion.

Oh wow... I did not see this `Makefile`. Yes, will remove the `hoogle` parts there.
Main differences are that my version shares data with the `haddock` target, and that I had problems with accessing the server with non-secure `http` from browsers on a different machine.

jberthold:
I have made the `https` connection optional now, it is only interesting for non-local use (where the doc. links won't work either)

PR #3228, opened 2022-08-31
[WIP] Remove BuiltinAndAxiomSimplifier

### LINKED ISSUE #3111: Serokell: remove BuiltinAndAxiomSimplifier

gromakovsky:
Resolves #3111

### Scope:

This should be the last step of refactoring we are making for #3111. Basically the main idea is that instead of carrying around functions with type `TermLike RewritingVariableName -> SideCondition RewritingVariableName -> Simplifier (AttemptedAxiom RewritingVariableName)` (aka `BuiltinAndAxiomSimplifier`) we want to apply such functions as soon as possible and carry `Simplifier` values.

There are several things that still need to be done:
1. There is one test that started to fail as result of this refactoring (`Apply with top configuration`). As a workaround we changed the `expect` value in that test, but this value is not supposed to change because we didn't change any logic. We still need to figure out why it changed. See 2ad2608e8dd6bf4dc9457bd7e762b362a0402f3a
2. `definitionEvaluation` seems to be used only in tests. It should be removed carefully. The function itself can be removed easily, but before removing tests we need to check whether they test any logic that is still present. Also we should remove all other functions that are used only by `definitionEvaluation` (transitively).
3. `firstFullEvaluation` should be deleted because it's equivalent to `applyFirstSimplifierThatWorks`.
4. In this PR we basically made some experiments checking how much code can be simplified/refactored. We see that CI passes, but we are not 100% sure that's enough to conclude that the code works as expected. Someone more familiar with this code should check it more carefully.
5. Maybe something can be further simplified and refactored, we just stopped at this point. For example, `lookupAxiomSimplifier` has a TODO comment in the `missing` case. Maybe if we remove it, more refactoring will be possible.
6. There is one simple TODO comment we added.

###### Review checklist

### DIFF
--- a/kore/src/Kore/Builtin.hs
+++ b/kore/src/Kore/Builtin.hs
@@ -16,7 +16,6 @@ This module is intended to be imported qualified.
 module Kore.Builtin (
     Builtin.Verifiers (..),
     Builtin.ApplicationVerifiers,
-    BuiltinAndAxiomSimplifier,
     Builtin.SymbolVerifier (..),
     Builtin.SortVerifier (..),
     Builtin.ApplicationVerifier (..),
@@ -53,10 +52,14 @@ import Kore.Builtin.String qualified as String
 import Kore.IndexedModule.MetadataTools (
     SmtMetadataTools,
  )
-import Kore.Internal.TermLike
-import Kore.Simplify.Simplify (
-    BuiltinAndAxiomSimplifier,
+import Kore.Internal.SideCondition (
+    SideCondition,
  )
+import Kore.Internal.TermLike
+import Kore.Rewrite.RewritingVariable (
+    RewritingVariableName,
+ )
+import Kore.Simplify.Simplify
 import Prelude.Kore
 
 {- | Verifiers for Kore builtin sorts.
@@ -88,7 +91,13 @@ koreVerifiers =
 
   See also: 'Data.Step.Step.step'
 -}
-koreEvaluators :: Text -> Maybe BuiltinAndAxiomSimplifier
+koreEvaluators ::
+    Text ->
+    Maybe
+        ( TermLike RewritingVariableName ->
+          SideCondition RewritingVariableName ->
+          Simplifier (AttemptedAxiom RewritingVariableName)
+        )
 koreEvaluators key =
     asum
         [ Bool.builtinFunctions key
--- a/kore/src/Kore/Builtin.hs
+++ b/kore/src/Kore/Builtin.hs
@@ -102,6 +102,7 @@ koreEvaluators key termLike sideCondition =
             (\evaluator -> evaluator key termLike sideCondition)
             [ Bool.builtinFunctions
             , Int.builtinFunctions
+            , IO.builtinFunctions
             , KEqual.builtinFunctions
             , List.builtinFunctions
             , Map.builtinFunctions
--- a/kore/src/Kore/Builtin.hs
+++ b/kore/src/Kore/Builtin.hs
@@ -93,24 +93,23 @@ koreVerifiers =
 -}
 koreEvaluators ::
     Text ->
-    Maybe
-        ( TermLike RewritingVariableName ->
-          SideCondition RewritingVariableName ->
-          Simplifier (AttemptedAxiom RewritingVariableName)
-        )
-koreEvaluators key =
-    asum
-        [ Bool.builtinFunctions key
-        , Int.builtinFunctions key
-        , IO.builtinFunctions key
-        , KEqual.builtinFunctions key
-        , List.builtinFunctions key
-        , Map.builtinFunctions key
-        , Set.builtinFunctions key
-        , String.builtinFunctions key
-        , Krypto.builtinFunctions key
-        , InternalBytes.builtinFunctions key
-        ]
+    TermLike RewritingVariableName ->
+    SideCondition RewritingVariableName ->
+    Maybe (Simplifier (AttemptedAxiom RewritingVariableName))
+koreEvaluators key termLike sideCondition =
+    asum $
+        map
+            (\evaluator -> evaluator key termLike sideCondition)
+            [ Bool.builtinFunctions
+            , Int.builtinFunctions
+            , KEqual.builtinFunctions
+            , List.builtinFunctions
+            , Map.builtinFunctions
+            , Set.builtinFunctions
+            , String.builtinFunctions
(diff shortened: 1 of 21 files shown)


goodlyrottenapple:
I've had a look at the failing unit test and the error was introduced as a result of the removal of all the closures. When we look at the original code, we have

```haskell
actual <- 
    evaluateWithPredicate
        ( firstFullEvaluation
            [ axiomEvaluatorWithRequires
                (Mock.functionalConstr10 Mock.a)
                (Mock.g Mock.a)
                requirement
            , ...
            ]
        )
        (Mock.functionalConstr10 Mock.a)
        requirement
```

Looking at the definition of `firstFullEvaluation` we see that it is simply a `BuiltinAndAxiomSimplifier` closure around `applyFirstSimplifierThatWorks [definitionEvaluation ...] OnlyOneResult`, so the above is equivalent to

```haskell
actual <- 
    runSimplifierSMT Mock.env $
        applyFirstSimplifierThatWorks
            [ axiomEvaluatorWithRequires
                (Mock.functionalConstr10 Mock.a)
                (Mock.g Mock.a)
                requirement
            , ...
            ]
            OnlyOneResult
            (Mock.functionalConstr10 Mock.a)
            (SideCondition.fromPredicateWithReplacements requirement)
```

Now, looking at the current definition of `applyFirstSimplifierThatWorks`:

https://github.com/runtimeverification/haskell-backend/blob/454211800a577c6b8d13c28457ebcdfffe2a871c/kore/src/Kore/Simplify/Simplify.hs#L515-L522

we see that the `applicationResult` is obtained by applying the simplifier wrapped in `BuiltinAndAxiomSimplifier` to `patt` and `sideCondition`. In this instance it means we will essentially be computing:

```haskell
...
applicationResult <- axiomEvaluatorWithRequires
    (Mock.functionalConstr10 Mock.a)
    (Mock.g Mock.a)
    requirement
    (Mock.functionalConstr10 Mock.a)
    (SideCondition.fromPredicateWithReplacements requirement)
...
```

Since in the new code, we have:

https://github.com/runtimeverification/haskell-backend/blob/3be474569d8b8a3446f51d471ff7f2eb0354564a/kore/src/Kore/Simplify/Simplify.hs#L460-L466

the updated test should match this function call. However, as you can see, this is not the case:

https://github.com/runtimeverification/haskell-backend/blob/3be474569d8b8a3446f51d471ff7f2eb0354564a/kore/test/Test/Kore/Rewrite/Axiom/EvaluationStrategy.hs#L289-L295
    
 Namely, `(SideCondition.fromPredicateWithReplacements requirement)` has become `trueSideCondition`

alexd1971:
Thank you a lot, @goodlyrottenapple!

> the arguments `patt` and `sideCondition` in the new [`applyFirstSimplifierThatWorksWorker`](https://github.com/runtimeverification/haskell-backend/blob/3be474569d8b8a3446f51d471ff7f2eb0354564a/kore/src/Kore/Simplify/Simplify.hs#L448) are now only used to throw an error if more than one configurations appear in `AttemptedAxiomResults`. This is confusing and perhaps these arguments should be factored out entirely?

Yes, I also noticed this and was about to discuss this with you. But you did it first.

jberthold:
> As a side-note, the arguments `patt` and `sideCondition` in the new [`applyFirstSimplifierThatWorksWorker`](https://github.com/runtimeverification/haskell-backend/blob/3be474569d8b8a3446f51d471ff7f2eb0354564a/kore/src/Kore/Simplify/Simplify.hs#L448) are now only used to throw an error if more than one configurations appear in `AttemptedAxiomResults`. This is confusing and perhaps these arguments should be factored out entirely? 

The `sideCondition` is also used in result `NotApplicableUntilConditionChanges`. Which points at a deficiency of the refactoring we do here, as different items in the list might well have embedded _different conditions_. That was previously not possible, at the price of constructing a list of partially-applied function closures to be passed into this function.

**What I would really like to see at this point is an assessment of the performance on the branch vs. current master** , so we can see what this change actually buys us.

> However, this would mean somehow passing the error as data within `AttemptedAxiomResults`, to be able to throw this error in the same place... @ana-pantilie @jberthold thoughts?

Maybe the returned `AttemptedAxiomResult` could be checked by the caller (hopefully knowing the `patt`) for whether there is more than one result? Anyway this is a hard `error` call in that case, what we is debugging information for something that should never happen. but we could also modify the `AttemptedAxiom` type to include an error indicator.

ehildenb:
@gromakovsky I second @jberthold that any change or refactoring that could affect performance, we should post performance numbers for against `master`. Instructions for collecting performance numbers against KEVM live here: https://github.com/runtimeverification/evm-semantics/issues/1165

I believe there is also some work to nix-ify that: https://github.com/runtimeverification/haskell-backend/issues/3323 (but we should not wait for that to finish to continue with profiling each PR we are trying).

gromakovsky:
Oh, right, there is one thing I forgot. Some time in September, after our previous collaboration phase and before the current one, we ran performance tests for this PR. See results:
[report-remove-builtin_master-8.pdf](https://github.com/runtimeverification/haskell-backend/files/9812363/report-remove-builtin_master-8.pdf)

As you can see, we've got 16% worse performance on average. What I find weird here is that, while in most cases the difference is in range 85-110% (which seems reasonable), there are a few cases with long running times where the difference is 150% or 200% or even 220% in the worst case. These results are a bit outdated, maybe something will change in the new version, but I think we will quite likely get similar results, unless those measurements were made in bad environment (e. g. with something else running concurrently). If that happens, I would investigate worst cases more thoroughly, perhaps would run with profiling in both branches. Since the difference is huge, maybe we'll get some insights.

Anyway, there is a big chance we won't see any performance improvement and maybe will even see performance degradation (even after investigation). In this case I would extract only small changes from this PR that remove unneeded logic (e. g. `TermLike` transformations and `AcceptsMultipleResults`) and test just them. They make the codebase simpler and most likely shouldn't spoil performance.

alexd1971:
Here is the latest result of performance testing.
[report-breakerzircon--_master.pdf](https://github.com/runtimeverification/haskell-backend/files/9908726/report-breakerzircon--_master.pdf)

PR #3234, opened 2022-09-02
3214 support not simplification rules

### LINKED ISSUE #3098: Fix axiom simplifier lookup mechanism for \equals and \ceil
Right now, there is a bug in the way we look up axioms to use to simplify \ceil and \equals terms. In particular, if the simplification rule we wrote has a \equals or a \ceil on the lhs whose child is something for which `matchAxiomIdentifier` returns Nothing, then configurations which would match these rules will skip applying any axioms, even though such an axiom could apply. I have extended the supported cases somewhat in #3097 , but a full fix still ought to be performed.

### LINKED ISSUE #3214: Add support for #Not simplification rules
This should be very similar to https://github.com/runtimeverification/haskell-backend/pull/3042.

jberthold:
Fixes #3214 
Fixes #3098 

### Scope:

* Adds a `Not` constructor to the `AxiomIdentifier` type
* in `Simplify.Predicate.simplifyNot`, tries to apply user-defined simplifications (equations) before applying standard simplification

However, for this to have the desired effect, the `AxiomIdentifier` lookup needs to be fixed (see #3098). Otherwise only simplifications of a certain shape (for which where an identifier exists) will be considered. Therefore, the `AxiomIdentifier` gets a new constructor `Other`, and all terms that are not handled specially will be associated with this constructor (making `matchAxiomIdentifier` total).

### Estimate:

---

###### Review checklist


kore/src/Kore/Rewrite/Axiom/Identifier.hs
```diff
@@ -62,6 +63,9 @@ these can be identified.
 
 The expectation is that an axiom can be applied to a term only if the
 identifier of its left-hand-side is the same as the term's identifier.
+
+In practice, recursion should be cut short at a level that is
+selective enough for good performance (see 'matchAxiomIdentifier').
```
jberthold, line 68: :thinking: well... we may or may not want to add that (cutting the recursion short is not supported in the `recursion-schemes` flavour of `fold` AFAICT). Maybe I should delete this comment
```suggestion
```

kore/src/Kore/Rewrite/Axiom/Identifier.hs
```diff
@@ -125,81 +138,59 @@ matchAxiomIdentifier = Recursive.fold matchWorker
                     Signedness.toApplication $ getConst signednessF
             InjF inj -> mkAppId $ Inj.toApplication inj
             InternalListF internalList -> listToId internalList
-            InternalSetF internalSet -> mapToId internalSet
-            InternalMapF internalMap -> setToId internalMap
```
jberthold, line 129: :smirk:

kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -327,13 +330,46 @@ normalizeTop _ = pure (MultiOr.singleton MultiAnd.top)
...
+    applyUserDefined :: MaybeT simplifier NormalForm
+    applyUserDefined = do
+        -- produce a single termlike that we can use for matching
+        let arg :: TermLike RewritingVariableName
+            arg =
+                Pattern.toTermLike
+                    . OrPattern.toPattern helpSort
+                    . toOrPattern helpSort
+                    $ multiOr
+            -- HAAAACK: sort stripped in NormalForm of predicates
```
goodlyrottenapple, line 351: 🤣

goodlyrottenapple: why cant `notSort` be used here btw?

kore/test/Test/Kore/Rewrite/Axiom/Identifier.hs
```diff
@@ -102,8 +116,36 @@ test_matchAxiomIdentifier =
...
+            DV
+        , test
+            "internal int"
+            (TermLike.mkInternalInt $ InternalInt Mock.testSort 42)
+            DV
+        ]
     ]
   where
+    -- FIXME add tests for DV cases
+
```
goodlyrottenapple, line 148: are you planning on adding these in this PR or leave for later?

kore/src/Kore/Rewrite/Axiom/Identifier.hs
```diff
@@ -74,8 +75,12 @@ data AxiomIdentifier
       Equals !AxiomIdentifier !AxiomIdentifier
     | -- | An @\\exists@ pattern with the given child.
       Exists !AxiomIdentifier
+    | -- | A @\\not@ pattern with the given child.
+      Not !AxiomIdentifier
     | -- | Any variable pattern.
       Variable
+    | -- | Anything else.
+      Other
```
ana-pantilie, line 83: Regarding the discussion here: https://github.com/runtimeverification/haskell-backend/issues/3098#issuecomment-1156671650

Is adding this constructor any different to just removing the `AxiomIdentifier` arguments in this type?

My reasoning is that this function was partial so that would shortcircuit attempts to apply equations whenever we'd try to simplify terms which are unlikely to be simplified by rules provided by the user.

kore/test/Test/Kore/Rewrite/Axiom/Identifier.hs
jberthold: Oh, I actually did add those tests (last commit), should have removed the comment, too.

kore/src/Kore/Rewrite/Axiom/Identifier.hs
jberthold: > Is adding this constructor any different to just removing the `AxiomIdentifier` arguments in this type?

Adding this "catch-all" constructor means any (simplification) equation from the user will get added, under a non-specific `Other` key if nothing more specific matches.
If we removed the `AxiomIdentifier` recursion in the type instead, we would store only what we had stored before but under less specific keys. 
For illustration: If we remove the recursion, _any_ `ceil` simplification equations will be stored under a single `Ceil` key, and rules like (1) `\ceil (\dv ...) => ...` and (2) `\ceil (\not (\dv ...)) => ...` will both be tried on any term that has a `\ceil` at the top (say, (3) `\ceil (\and ...)`. 
In the proposed implementation, the keys would be (1) `Ceil  DV`, (2) `Ceil (Not DV)` and (3) `Ceil Other`, previously (3) would have no key, and rules with LHS like (3) would be dropped entirely.

We could maybe filter out the plain `Other` at the top (after the `fold` returns), that would retain some of the prior partiality. LMK if you would prefer that.

kore/src/Kore/Simplify/Predicate.hs
jberthold: I was trying to use `notSort` before, but its type is _variable_ in `simplifyNot`. 
Predicates do not have sort information: https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Internal/Predicate.hs#L156..L169
Not 100% sure but `notSort` might always be `()` in the call sites of `simplifyNot`.

jberthold: I have refactored the types in this module to avoid confusion. Indeed `simplifyNot` (and many other functions) are only ever called with `sort ~ ()`.

goodlyrottenapple: is this because a predicate always has the sort bool?

kore/src/Kore/Rewrite/Axiom/Identifier.hs
ana-pantilie: 👍

kore/src/Kore/Rewrite/Axiom/Matcher.hs
```diff
@@ -271,6 +271,8 @@ matchEqualHeads (Pair (DV_ sort1 dv1) (DV_ sort2 dv2)) = do
     push (Pair dv1 dv2)
 matchEqualHeads (Pair (Equals_ _ _ term11 term12) (Equals_ _ _ term21 term22)) =
     push (Pair term11 term21) >> push (Pair term12 term22)
+matchEqualHeads (Pair (Not_ _ term1) (Not_ _ term2)) =
+    push (Pair term1 term2)
```
jberthold, line 275: Turned out this was what made the tests fail - the matcher was not ready for a `Not`.
We could complete more missing cases here but since a rewrite is waiting around the corner I did not.

kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -325,22 +328,55 @@ normalizeTop _ = pure (MultiOr.singleton MultiAnd.top)
...
+    mbResult <- runMaybeT applyUserDefined
+    maybe standardSimplification pure mbResult
+  where
+    applyUserDefined :: MaybeT simplifier NormalForm
+    applyUserDefined = do
+        -- produce a single termlike that we can use for matching
+        let arg :: TermLike RewritingVariableName
+            arg =
+                Pattern.toTermLike
+                    . OrPattern.toPattern helpSort
```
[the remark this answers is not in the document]
jberthold: Hopefully the direct `OrPattern.toTermLike` is better?
`OrPattern.toPattern` processes the conditions differently if the terms of the pattern are `Top` , is that a problem?

kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -325,22 +328,51 @@ normalizeTop _ = pure (MultiOr.singleton MultiAnd.top)
...
     simplifier NormalForm
-simplifyNot Not{notChild = multiOr, notSort} = do
-    disjunctiveNormalForms <- Logic.observeAllT $ do
-        multiAnd <- Logic.scatter multiOr
-        normalizeNotAnd Not{notSort, notChild = multiAnd} & lift
-    normalizeMultiAnd (MultiAnd.make disjunctiveNormalForms)
+simplifyNot sideCondition Not{notChild = multiOr} = do
+    -- try user-defined rules first
+    mbResult <- runMaybeT applyUserDefined
+    maybe standardSimplification pure mbResult
```
ana-pantilie, line 340: I'm actually not so sure about this, let me explain:

The `standardSimplification` basically pushes the `#Not` inside the dnf, and redistributes stuff so that the result is back in dnf, so you get `((#Not p1) #And ... #And (#Not pn)) #Or ...`. It might make more sense to try to apply any user-defined simplification rules to this form, because it's probably what the users are going to see. I don't think they'll ever see the `#Not (conj1 #Or ... #Or conjn)` form.

jberthold: Sure, we can swap the order and always do the standard simplification

kore/src/Kore/Simplify/Predicate.hs
```diff
@@ -335,37 +335,51 @@ simplifyNot ::
...
-                (Predicate.toMultiAnd . from . snd . Conditional.splitTerm)
-                applied
-
-    standardSimplification :: simplifier NormalForm
-    standardSimplification = do
-        disjunctiveNormalForms <- Logic.observeAllT $ do
-            multiAnd <- Logic.scatter multiOr
-            lift $ normalizeNotAnd Not{notSort = (), notChild = multiAnd}
-        normalizeMultiAnd (MultiAnd.make disjunctiveNormalForms)
+        pure $ MultiOr.map (from . snd . Conditional.splitTerm) applied
```
ana-pantilie, line 378: We should probably throw an error if any of the results is not a predicate.

jberthold: `traverse_ (guard . Conditional.isPredicate) applied` above takes care of that case. 
We can of course make the run fail hard instead of swallowing the error condition. Will modify accordingly

PR #3235, opened 2022-09-02
Added auto profiling to nix

goodlyrottenapple:
Based on the discussion here https://github.com/runtimeverification/haskell-backend/issues/3221
I've implemented a quick script under
```
nix run github:runtimeverification/haskell-backend#profile smoke.tar.gz (4m)
```
which will profile the provided `smoke.tar.gz` for 4 minutes (defaults to 4m if left blank).
The script will produce a `profile-smoke.tar.gz` folder with the following files:

**Time Profiling**
`kore-exec.prof` - Basic profile
`prof-speedscope.json` - same profile loadable in speedscope.app
`raw/prof.eventlog` - basic profile data captured in an eventlog, used to produce `prof-speedscope.json`

**Heap Profiling**
`heap-cost-centers.svg` - basic heap chart produced from `raw/heap-cost-centers.hp`
`heap-cost-centers.html` - interactive heap profile based on `raw/heap-cost-centers.eventlog`

**Heap Profiling GHC9**
`heap-infotables.html` - interactive heap profile (much more detailed than `heap-cost-centers.html`) based on `raw/heap-infotables.eventlog`

Please let me know if these are all the files we want to produce etc.

### DIFF
--- a/nix/profile.nix
+++ b/nix/profile.nix
@@ -1,0 +1,34 @@
+{ stdenv, writeScriptBin, gnutar, kore-exec-prof, kore-exec-infotable, hp2pretty, eventlog2html }:
+
+writeScriptBin "profile" ''
+  #! ${stdenv.shell}
+  filename=$(basename -- "$1")
+  timeout="''${2:-4m}"
+
+  mkdir -p profile
+  ${gnutar}/bin/tar -xzvf "$1" -C ./profile
+  cd profile
+  # generate .prof and .eventlog profiles
+  timeout --foreground -s INT "$timeout" env GHCRTS='-p -l-au' PATH='$PATH:${kore-exec-prof}/bin' ./kore-exec.sh
+  mkdir -p ../profile-$filename
+  cp kore-exec.prof ../profile-$filename/
+  cp kore-exec.eventlog ../profile-$filename/prof.eventlog
+
+  # generate a heap profile based on cost centers
+  timeout --foreground -s INT "$timeout" env GHCRTS='-l -hc' PATH='$PATH:${kore-exec-prof}/bin' ./kore-exec.sh
+  cp kore-exec.hp ../profile-$filename/heap-cost-centers.hp
+  cp kore-exec.eventlog ../profile-$filename/heap-cost-centers.eventlog
+  ${hp2pretty}/bin/hp2pretty kore-exec.hp
+  ${eventlog2html}/bin/eventlog2html kore-exec.eventlog
+  cp kore-exec.svg ../profile-$filename/heap-cost-centers.svg
+  cp kore-exec.eventlog.html ../profile-$filename/heap-cost-centers.html
+
+  # generate a heap profile based on infotables (GHC9)
+  timeout --foreground -s INT "$timeout" env GHCRTS='-l -hi' PATH='$PATH:${kore-exec-infotable}/bin' ./kore-exec.sh
+  cp kore-exec.eventlog ../profile-$filename/heap-infotables.eventlog
+  ${eventlog2html}/bin/eventlog2html kore-exec.eventlog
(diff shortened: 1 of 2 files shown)


goodlyrottenapple:
Unfortunately we can only upload the SVG file on github, out of the ones mentioned above 
heap-cost-centers

goodlyrottenapple:
[profile-smoke.tar.gz.zip](https://github.com/runtimeverification/haskell-backend/files/9478931/profile-smoke.tar.gz.zip)
 These might potentially get pretty big!

ehildenb:
Big is fine! But we have a server we can store large artifacts on, maybe we don't need to upload to github. I'd ilke if we could store and upload _all_ the artifacts to some temporary storage though. We're only able to upload the 1 to github because of size? Or because the others are less clear/well-defined?

goodlyrottenapple:
> Big is fine! But we have a server we can store large artifacts on, maybe we don't need to upload to github. I'd ilke if we could store and upload _all_ the artifacts to some temporary storage though. We're only able to upload the 1 to github because of size? Or because the others are less clear/well-defined?

No, it's to do with file extensions. Github is picky about which ones i can upload here.

ehildenb:
Ok, that's no problem. We can take all the artifacts, and create a tarball of them (same as we do for bug reprots), and upload the tarball. Can the Nix script take care of tarring things up?

goodlyrottenapple:
sure, tar-ing is not a problem.

dwightguth:
I would make the default 7 minutes. That's about the longest it can run and speedscope can still handle the results.

goodlyrottenapple:
this time may very much depend on what machine it gets executed on, too

dwightguth:
Hmmm, that's true. Is there a way to stop it after the profile reaches a certain size instead? Speedscope seems to be able to handle profiles up to about 100 megabytes.

nix/profile.nix
```diff
@@ -0,0 +1,36 @@
+{ stdenv, writeScriptBin, gnutar, kore-exec-prof, kore-exec-infotable, hp2pretty, hs-speedscope, eventlog2html }:
+
+writeScriptBin "profile" ''
```
jberthold, line 3: Somewhat bike-shedding... but could we rename this file and its output to something like `run-profile` or `profiling-run`?
I was reading `nix/profile.nix` and thought "an entirely new environment profile for nix?" :)

nix/profile.nix
```diff
@@ -0,0 +1,36 @@
...
+  #! ${stdenv.shell}
+  filename=$(basename -- "$1")
+  timeout="''${2:-4m}"
+
+  mkdir -p profile
+  ${gnutar}/bin/tar -xzvf "$1" -C ./profile
+  cd profile
+  # generate .prof and .eventlog profiles
+  timeout --foreground -s INT "$timeout" env GHCRTS='-p -l-au' PATH='$PATH:${kore-exec-prof}/bin' ./kore-exec.sh
+  mkdir -p ../profile-$filename/raw
```
jberthold, line 13: (nit) Maybe factor out `profile-$filename`? (and create it in the beginning, together with `./profile`)

jberthold: Should we also include a timestamp in the directory name? Would make it easier to manage the result files if we store them in a custom place (cleaning up based on time, avoiding overwrites)

nix/profile.nix
goodlyrottenapple: sure, i'll change to `run-profiling` (to split the difference :P)

nix/profile.nix
goodlyrottenapple: good idea, I've added a timestamp

jberthold (approved):
:sparkles:

ana-pantilie (approved):
👍

Issue #3197, opened 2022-08-11
Continue when error is thrown when attempting simplification rules


ana-pantilie:
If, for some reason, we don't know if a simplification rule applies or not, just emit a warning and continue execution. Specifically, when checking the requires of a simplification rule, if ErrorDecidePredicateUnknown is thrown, don't crash.

ana-pantilie:
@jberthold please add some links to the code which you think is useful here

jberthold:
The key insight here is that `decidePredicate` will only ever return `Just False` or `Nothing`, never `Just True`.
https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Rewrite/SMT/Evaluator.hs#L164-L169
`decidePredicate` is only called by `evalPredicate`, which is only called by `evalConditional`, in the same module.

`evalPredicate` can in fact return `Just True` when the predicate only contains a `Top`, https://github.com/runtimeverification/haskell-backend/blob/be7e56172bd02d32b31d91c20fc7ba1a47766fc0/kore/src/Kore/Rewrite/SMT/Evaluator.hs#L99

Call sites for `evalConditional`: 
* Simplifier: `simplifyClaimRule`, to filter things that are known to be false: https://github.com/runtimeverification/haskell-backend/blob/be7e56172bd02d32b31d91c20fc7ba1a47766fc0/kore/src/Kore/Rewrite/Rule/Simplify.hs#L140-L143 
* `checkRequires`, to filter by known false requirements: https://github.com/runtimeverification/haskell-backend/blob/be7e56172bd02d32b31d91c20fc7ba1a47766fc0/kore/src/Kore/Equation/Application.hs#L382-L390 
For both callers, tolerating `Unknown` results and returning `Nothing` seems fine (solver timeouts might cause additional work downstream for re-checking).
* `Search`: https://github.com/runtimeverification/haskell-backend/blob/be7e56172bd02d32b31d91c20fc7ba1a47766fc0/kore/src/Kore/Rewrite/Search.hs#L151-L161
In this code, different conditions are returned depending on the result. The difference between an `Unknown` and  a `Just True` result is minor, though.

PR #3238, opened 2022-09-06
Changed `ErrorDecidePredicateUnknown` into a `WarnDecidePredicateUnknown`

goodlyrottenapple:
Fixes [#3197](https://github.com/runtimeverification/haskell-backend/issues/3197)

This PR changes the runtime error `ErrorDecidePredicateUnknown` into a runtime warning instead.
The added integration test confirms that the execution proceeds with the new warning instead of throwing an error. This is what the output of [`test/warn-decide-predicate-unknown`](https://github.com/runtimeverification/haskell-backend/compare/sam/WarnDecidePredicateUnknown?expand=1#diff-17963e33586042210edfe17e87e5a6f5aecb9a3338f071e4cfe9c32a8ed5324b) returns on current master:

```bash
make: Entering directory '/build/kore/test/warn-decide-predicate-unknown'
>>> /build/kore/test/warn-decide-predicate-unknown kompile verification.k
rm -fr ./verification-kompiled
env -u KORE_EXEC_OPTS kompile --backend haskell --haskell-backend-command kore-exec -d . verification.k
[Warning] Compiler: Could not find main syntax module with name
VERIFICATION-SYNTAX in definition.  Use --syntax-module to specify one. Using
VERIFICATION as default.
make: Leaving directory '/build/kore/test/warn-decide-predicate-unknown'
make: Entering directory '/build/kore/test/warn-decide-predicate-unknown'
>>> /build/kore/test/warn-decide-predicate-unknown kprove lemmas-no-smt-spec.k
KORE_EXEC_OPTS = --no-bug-report --smt none
rm -f lemmas-no-smt-spec.k.out
kprove --haskell-backend-command kore-exec  lemmas-no-smt-spec.k >lemmas-no-smt-spec.k.out || true
kore-exec: [167366] Error (ErrorDecidePredicateUnknown):
    Failed to decide predicate:
        /* Sfa */
        \not{_}(
            /* Spa */
            \equals{SortBool{}, _}(
                /* Fl Fn D Sfa */ VarB:SortBool{},
                /* Fl Fn D Sfa Cl */ \dv{SortBool{}}("false")
            )
        )
    with side condition:
        /* D Sfa */ \top{_}()
[Error] Critical: Haskell Backend execution failed with code 1 and produced no
output.
diff -u lemmas-no-smt-spec.k.out.golden lemmas-no-smt-spec.k.out || ( mv lemmas-no-smt-spec.k.out lemmas-no-smt-spec.k.out.fail && false )
--- lemmas-no-smt-spec.k.out.golden     1970-01-01 00:00:01.000000000 +0000
+++ lemmas-no-smt-spec.k.out    2022-09-06 14:39:32.072299222 +0000
@@ -1,9 +0,0 @@
-  #Not ( {
-    B ==Bool true
-  #Equals
-    bool2Word ( B ) ==Int 1
-  } )
-#And
-  <k>
-    doneLemma ( bool2Word ( B ) ==Int 1 ) ~> _DotVar1 ~> .
-  </k>
make: *** [/build/kore/test/warn-decide-predicate-unknown/../include.mk:107: lemmas-no-smt-spec.k.out] Error 1
make: Leaving directory '/build/kore/test/warn-decide-predicate-unknown'
```


jberthold (approved):
LGTM

ana-pantilie (changes requested):
We need to do the following:
- if the call comes from attempting a simplification rule, do not emit `errorDecidePredicateUnknown`, emit a new warning that simplification rule with sourceloc <srcLoc> was attempted but the SMT returned unknown when attempting to discharge the condition.
- after that that new warning is emitted, the solver needs to be reinitialized because it's possible it's in a buggy state
- if the call doesn't come from a simplification rule attempt, do the same thing we did before these changes

goodlyrottenapple:
> if the call doesn't come from a simplification rule attempt, do the same thing we did before these changes
So this code gets called in these instances:

https://github.com/runtimeverification/haskell-backend/blob/a1959233e210f960d49a91ebe7fcec76da14ebe3/kore/src/Kore/Rewrite/Rule/Simplify.hs#L120-L143

https://github.com/runtimeverification/haskell-backend/blob/a1959233e210f960d49a91ebe7fcec76da14ebe3/kore/src/Kore/Rewrite/Search.hs#L126-L167

https://github.com/runtimeverification/haskell-backend/blob/a1959233e210f960d49a91ebe7fcec76da14ebe3/kore/src/Kore/Equation/Application.hs#L316-L371

we want the warn behaviour only for the first and error for the rest?

ana-pantilie:
Only in [haskell-backend/kore/src/Kore/Equation/Application.hs](https://github.com/runtimeverification/haskell-backend/blob/a1959233e210f960d49a91ebe7fcec76da14ebe3/kore/src/Kore/Equation/Application.hs#L316-L371)

And only when the attempted equation has the `simplification` attribute.

goodlyrottenapple:
This is the log message now:
```
kore-exec: [185473] Error (DecidePredicateUnknown):
    Failed to decide predicate:
        /* Sfc */
        \not{_}(
            /* Sfc */
            \equals{SortBool{}, _}(
                /* Fl Fn D Sfa */
                Lbl'UndsEqlsEqls'Bool'Unds'{}(
                    /* Fl Fn D Sfa */ VarB:SortBool{},
                    /* Fl Fn D Sfa Cl */ \dv{SortBool{}}("true")
                ),
                /* Fl Fn D Sfc */
                Lbl'UndsEqlsEqls'Int'Unds'{}(
                    /* Fl Fn D Sfc */
                    Lblbool2Word'LParUndsRParUnds'LEMMAS'Unds'Int'Unds'Bool{}(
                        /* Fl Fn D Sfa */ VarB:SortBool{}
                    ),
                    /* Fl Fn D Sfa Cl */ \dv{SortInt{}}("1")
                )
            )
        )
    SMT limit set at:
        0
Context:
    (InfoReachability) while checking the implication
        (DebugEvaluateCondition) while evaluating predicate
            (DecidePredicateUnknown) while applying equation
                                     in Kore.Rewrite.SMT.Evaluator.filterMultiOr
```

kore/src/Kore/Log/DecidePredicateUnknown.hs
```diff
@@ -0,0 +1,133 @@
...
+    NonEmpty (Predicate variable) ->
+    log ()
+throwDecidePredicateUnknown action smtLimit predicates' =
+    case action of
+        WarnSimplificationEquationInApplication _ ->
+            logEntry DecidePredicateUnknown{action, smtLimit, predicates}
+        _ ->
+            throw DecidePredicateUnknown{action, smtLimit, predicates}
+  where
+    predicates = Predicate.mapVariables (pure toVariableName) <$> predicates'
```
goodlyrottenapple, line 133: we now either throw or log depending on the action

kore/src/Kore/Log/DecidePredicateUnknown.hs
```diff
@@ -0,0 +1,133 @@
...
+import Pretty qualified
+import SMT qualified
+
+data OnDecidePredicateUnknown
+    = WarnSimplificationEquationInApplication (Maybe SourceLocation)
+    | ErrorInApplication (Maybe SourceLocation)
+    | ErrorInMatchWith
+    | ErrorInSimplifyClaimRule
+    | ErrorInFilterMultiOr
+    deriving stock (Show, Eq)
```
goodlyrottenapple, line 42: A new ADT which keeps track of the action ot take and where the warning/error roughly originated in the codebase.

kore/src/Kore/Rewrite/SMT/Evaluator.hs
```diff
@@ -150,24 +153,32 @@ The predicate is always sent to the external solver, even if it is trivial.
...
-                    errorDecidePredicateUnknown predicates
+                Unknown -> do
+                    limit <- SMT.withSolver SMT.askRetryLimit
+                    -- depending on the value of `onUnknown`, this call will either log a warning
+                    -- or throw an error
+                    throwDecidePredicateUnknown onUnknown limit predicates
+                    case onUnknown of
+                        WarnSimplificationEquationInApplication _ ->
+                            -- the solver may be in an inconsistent state, so we re-initialize
+                            SMT.reinit
```
goodlyrottenapple, line 177: re-init the SMT if working on a simplification equation

kore/src/Kore/Rewrite/SMT/Evaluator.hs
```diff
@@ -139,7 +142,7 @@ filterMultiOr multiOr = do
         Conditional variable term ->
         Simplifier (Maybe (Conditional variable term))
     refute p =
-        evalConditional p Nothing <&> \case
+        evalConditional ErrorInFilterMultiOr p Nothing <&> \case
```
jberthold, line 145: A bit unsure about this case... could `filterMultiOr` be used in the context of a simplification?
But not a blocker, even if this is the case, the new code still improves matters.

jberthold (approved):
IMHO this is OK to merge, although I had a question.

kore/src/Kore/Equation/Application.hs
```diff
@@ -112,8 +116,12 @@ attemptEquation sideCondition termLike equation = do
...
             (equation', predicate) <- matchAndApplyResults left
-            let Equation{requires} = equation'
-            checkRequires sideCondition predicate requires & whileCheckRequires
+            let Equation{requires, attributes = Attribute.Axiom{simplification}} = equation'
+                eqSrc = Equation.srcLoc equation'
+                onDecidePredicateUnknown = case simplification of
+                    Attribute.NotSimplification -> ErrorInApplication eqSrc
+                    Attribute.IsSimplification _ -> WarnSimplificationEquationInApplication eqSrc
+            checkRequires onDecidePredicateUnknown sideCondition predicate requires & whileCheckRequires
             let Equation{right, ensures} = equation'
```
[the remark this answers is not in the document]
goodlyrottenapple: I have also wondered this...

kore/src/Kore/Rewrite/SMT/Evaluator.hs
goodlyrottenapple: Yeah, multi-or gets used in half a dozen other places. Wasn't sure how detailed we want to go. @ana-pantilie

kore/src/Kore/Equation/Application.hs
```diff
@@ -112,8 +116,12 @@ attemptEquation sideCondition termLike equation = do
         whileDebugAttemptEquation' . runExceptT $ do
             let Equation{left} = equationRenamed
             (equation', predicate) <- matchAndApplyResults left
-            let Equation{requires} = equation'
-            checkRequires sideCondition predicate requires & whileCheckRequires
+            let Equation{requires, attributes = Attribute.Axiom{simplification}} = equation'
+                eqSrc = Equation.srcLoc equation'
+                onDecidePredicateUnknown = case simplification of
+                    Attribute.NotSimplification -> ErrorInApplication eqSrc
+                    Attribute.IsSimplification _ -> WarnSimplificationEquationInApplication eqSrc
```
ana-pantilie, line 123: Can we move this to the `where`? It would make the main body of the function less cluttered.

kore/src/Kore/Equation/Application.hs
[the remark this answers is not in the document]
ana-pantilie: Hmm, no particular reason, feel free to refactor if you want.

kore/src/Kore/Log/DecidePredicateUnknown.hs
```diff
@@ -0,0 +1,135 @@
...
+                            WarnSimplificationEquationInApplication _ ->
+                                "Kore.Equation.Application.checkRequires"
+                            ErrorInApplication _ ->
+                                "Kore.Equation.Application.checkRequires"
+                            ErrorInMatchWith ->
+                                "Kore.Rewrite.Search.matchWith"
+                            ErrorInSimplifyClaimRule ->
+                                "Kore.Rewrite.Rule.Simplify.simplifyClaimRule"
+                            ErrorInFilterMultiOr ->
+                                "Kore.Rewrite.SMT.Evaluator.filterMultiOr"
```
ana-pantilie, line 114: We'll have to keep these messages up to date somehow. I wonder if there's some way of making sure this code doesn't compile if the names change.

kore/src/Kore/Rewrite/SMT/Evaluator.hs
ana-pantilie: We can leave it like this for now.

test/warn-decide-predicate-unknown/lemmas-no-smt-spec.k.out.golden
```diff
@@ -0,0 +1,3 @@
+<k>
+  doneLemma ( bool2Word ( B ) ==Int 1 ) ~> _DotVar1 ~> .
```
ana-pantilie, line 2: I don't understand this test. What would have been the output without your changes?

ana-pantilie (commented):
Looking good, but still have some questions/suggestions I'd like taken care of before merging.

test/warn-decide-predicate-unknown/lemmas-no-smt-spec.k.out.golden
goodlyrottenapple: yeah this is why i asked about the integration test above. the final result is the same, but this test now throws some warning for the simplification rules before throwing an error and exiting. We could maybe change the test script to just check it produces a warning?

kore/src/Kore/Log/DecidePredicateUnknown.hs
goodlyrottenapple: I think this can be done with generics. Will investigate

goodlyrottenapple: hmmm... we could make a dummy datatype in the same module as the function calls and then use generics to get the data type's module, but then we would have to remember to move the dummy type together with the functions, so it's not safe from re-factoring. A better way might be to use template haskell to get the module name....i think thats doable, but also, template haskell :/

test/warn-decide-predicate-unknown/lemmas-no-smt-spec.stderr.out.golden
```diff
@@ -0,0 +1,132 @@
+kore-exec:  Warning (DecidePredicateUnknown):
```
goodlyrottenapple, line 1: @ana-pantilie I've modified the integration test, to instead comapre against the stderr output

ana-pantilie (approved):

PR #3241, opened 2022-09-07
monomorphize new unifier

dwightguth:
### Scope:

This PR removes use of `UnifierT` from NewUnifier.hs and its callers, replacing it where needed with `LogicT Simplifier`. This also ends up triggering monomorphization of a large number of functions in the Kore.Rewrite package.

### Estimate:

---

###### Review checklist


jberthold:
I guess `MonadUnify` could also be removed then? (leading to the removal of [Kore.Unification.Unify](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Unification/Unify.hs)  which would only re-export `Logic` functions and `InternalVariable`)

jberthold (commented):
The problems with the test setup stem from the type changes in `unificationProcedure` and `transitionRule`.

About [line 247 in kore/test/Test/Kore/Unification/Unifier.hs](https://github.com/runtimeverification/haskell-backend/pull/3241/files#diff-514df8bc553b95c652b5106aed00e48ffcfb3eeccae12f141b7ef36ffde4d32c), your type change of `unificationProcedure` now requires using `observeAllT` instead of `runUnifierT Not.notSimplifier`
```
-                unificationProcedure
-                    SideCondition.topTODO
-                    term1
-                    term2
-                    & Monad.Unify.runUnifierT Not.notSimplifier
-                    & runSimplifier mockEnv
-                    & runNoSMT
+                runNoSMT
+                    . runSimplifier mockEnv
+                    . Monad.Unify.observeAllT
+                    $ unificationProcedure
+                        SideCondition.topTODO
+                        term1
+                        term2
```
The other two errors about `AllPathIdentity` are more subtle, the test needs to be rewritten (as you mentioned on slack, an _actual_ simplifier needs to be used now) so we cannot use the `Test.Terse` combinators (whose value is ... unclear IMHO). Replacing the combinators in `MockAllPaths` would not be too hard.

How does the performance look for this PR?

dwightguth:
> I guess `MonadUnify` could also be removed then? (leading to the removal of [Kore.Unification.Unify](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Unification/Unify.hs) which would only re-export `Logic` functions and `InternalVariable`)

MonadUnify and UnifierT are still used by the old unifier, which is still used by And and Equals simplification. I tried removing these usages at the time I developed the new unifier, and it proved difficult, so I gave up.

jberthold (approved):
LGTM. Memory profile looks like this might accelerate execution :crossed_fingers:

ana-pantilie (approved):

PR #3252, opened 2022-09-13
Remove `ActualEntry` and refactor logging/throwing pretty printing

goodlyrottenapple:
Pre-requisite for  #3238

This PR refactors the log entry types type by adding the context directly to the `SomeEntry` type, making `ActualEntry` redundant. This change also removes context from the environment and instead bundles context with each `SomeEntry` directly. This is necessary when throwing errors, since we want to re-throw errors with the bundled context at each `logWhile` call, mirroring what we do with the log messages in this method.

Finally, the format of the log and error messages has been re-factored, with the context being moved at the bottom of the message for (hopefully) better readability:

```
kore-exec: [167082] Error (ErrorDecidePredicateUnknown):
    Failed to decide predicate:
        /* Sfa */
        \not{_}(
            /* Spa */
            \equals{SortBool{}, _}(
                /* Fl Fn D Sfa */ VarB:SortBool{},
                /* Fl Fn D Sfa Cl */ \dv{SortBool{}}("false")
            )
        )
    with side condition:
        /* D Sfa */ \top{_}()
Context:
    (DebugEvaluateCondition) while evaluating predicate
        (DebugAttemptEquation) while applying equation at /build/kore/test/warn-decide-predicate-unknown/lemmas.k:8:10-8:53
```

### DIFF
--- a/kore/app/repl/Main.hs
+++ b/kore/app/repl/Main.hs
@@ -297,7 +297,7 @@ mainWithOptions LocalOptions{execOptions} = do
     someExceptionHandler :: SomeException -> Main ExitCode
     someExceptionHandler someException = do
         case fromException someException of
-            Just (SomeEntry entry) -> logEntry entry
+            Just entry@(SomeEntry _ _) -> logEntry entry
             Nothing -> errorException someException
         throwM someException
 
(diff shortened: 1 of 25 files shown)


jberthold (approved):
LGTM

ana-pantilie (approved):
👍

PR #3260, opened 2022-09-16
Add a debug log entry for proof generation hints (second try)

zhengyao-lin:
This is the resurrected version of #2636.

This PR tries to add 4 (debug) log entries:
- `DebugInitialClaim` which records the input claim to be proved
- `DebugInitialPattern` which records the initial pattern when trying to prove each claim
- `DebugFinalPatterns` which records the final patterns when each claim is proven
- `DebugRewriteTrace` which records each rewriting step performed on the current configuration during the proof

They are all defined in `kore/src/Kore/Log/DebugRewriteTrace.hs`.

These entries are disabled by default, and only enabled with the new flag `--trace-rewrites <output>`.
With this flag, kore will write all of these four entries into a YAML-format file `<output>`.

For example, when doing a reachability proof, the output YAML file will look something like this
```
---
claim-id: <UNIQUE'Unds'ID attribute of the claim in kore, if it exists>
claim: <a string with the claim encoded in Kore>
task: reachability
steps:
- initial: { term: ..., constraint: ..., substitution: ... } # the term which the rules are applied to
  applied-rules:
  - rule-id: <UNIQUE'Unds'ID attribute of the rule applied>
    substitution: [ { key: <Kore variable>, value: <Kore term> }, ... ]
    results: [ { term: ..., constraint: ..., substitution: ... }, ... ]
  remainders: <remainder terms when some rule is applied>
- ...


# there could be another claim
---
claim-id: ...
...
```

---

###### Review checklist


goodlyrottenapple:
these changes have been merged into the latest master in this PR:  #3284

Issue #3259, opened 2022-09-16
Revise different test data generators for `TermLike`


jberthold:
We currently have several test data generators that generate similar things, in `Test.ConsistentKore` and in `Test.Kore`.
The `TermLike` generator in `Test.ConsistentKore` has been specialised to actual terms in #3256 but the one in `Test.Kore` continues to be provided to other modules from `Test.Kore.TermLike`.

* [x] Analyse what the different generators do differently. There seems to be a discrepancy wrt. their pool of free variables, and maybe the generators are incomplete (in a different way than the desired incompleteness from #3244 ).
* [x] Audit usage of the different generators in other test modules. Which one should be used where?
* [X] When generating applications with Map/Set symbols at the top, do not generate children terms with connectives. 
   #3262 addresses this more generally by avoiding connectives inside _all_ function symbols.
* [ ] Check whether the mock setup for the `ConsistentKore` generator is complete. One of the requirements is to have nullary constructors for all possible sorts. (maybe related to #3257).
* [x] Analyse why the `ConsistentKore` generator loops when removing the `\top` generator.
* [ ] (after results from the above and #3257 are addressed) Is it possible to always use the `ConsistentKore` generator (maybe using the _pattern_ generator) where previously we were using the `TermLike` generator?
* [ ] Remove code related to generating _internalized_ maps and sets

jberthold:
## `Test.ConsistentKore`
* primarily a generator for `Internal.Pattern`
* generates term and predicate using separate generators
  - predicate can be something containing terms (`\ceil`, `\floor`,
    `\in`, `\equals`)
* Many non-terms (e.g., fix point operators, implication) have been
  removed from the generator in [a recent PR](pulls/3256)
* Operates based on a `Setup` and a `Context`:
  - `Setup` contains "known" things like all sorts, specific sorts for
    built-in types like Bool and Int, symbols assumed to exist, etc.
  - `Context` contains available (element and set) variables as well
    as restriction flags to control term generation.
  - **However** does not modify the context when generating `\forall`
    or `\exists` predicates (which introduce fresh sorted variables)
* Code exists to generate `Map` and `Set` terms but the configuration
  seems to not trigger it.

## `Test.Kore`
* defines a `Gen` type that uses an internal `Context`
  - `Context` contains "objectVariables" (just `SomeVariable`s) and
    "ObjectSortVariables" (`SortVariable`s), as well as optional lists
    of known symbols and sorts
[...]

PR #3262, opened 2022-09-19
Adjustments to test  data generators

jberthold:
- [X] More diverse patterns from Test.Kore.Internal.Pattern.internalPatternGen
  `Pattern.fromTermLike` would just copy a given term into the `term` part of a `Pattern`. `Pattern.parsePatternFromTermLike` decomposes the top-level `\and`s and splits the term into predicate and "actual term" parts.
- [X] Remove the `#Top` generator from the term generation (as well as the `otherTopSort` sort from the generator setup)
- [x] Ensure `Test.ConsistentKore.patternGen` does not loop and is able to generate requested sizes.
- [X] Avoid using term connectives (`\and`, `\or`) inside a function (holds for both map- and set-constructing symbols as well as built-in operations like `<Int`)

This can be merged as-is , and fixes part of #3259

---

###### Review checklist


kore/test/Test/Kore/Rewrite/MockSymbols.hs
```diff
@@ -2325,7 +2331,7 @@ generatorSetup =
     ConsistentKore.Setup
         { allSymbols = filter doesNotHaveArguments symbols
         , allAliases = []
-        , allSorts = map fst sortAttributesMapping
+        , allSorts = filter (/= otherTopSort) $ map fst sortAttributesMapping
```
jberthold, line 2334: Note that we still have this `otherTopSort` around in the tests. 
https://github.com/runtimeverification/haskell-backend/blob/ed3edb3e59eafbe0992b8d8aec458cb79773067b/kore/test/Test/Kore/Rewrite/MockSymbols.hs#L2112..L2128 appears a bit strange to me, not well-ordered at least.

goodlyrottenapple: I like `overTheTopSort` :D

kore/test/Test/ConsistentKore.hs
```diff
@@ -818,14 +794,18 @@ symbolGenerator
...
+                    | BuiltinMap.isSymbolElement symbol =
+                        requestConcrete . requestConstructorLike
+                    | BuiltinSet.isSymbolElement symbol =
+                        requestConcrete . requestConstructorLike
+                    | isFunction =
+                        withoutConnectives
+                    | otherwise = id
+                -- TODO (virgil): also allow constructor-like
+                -- stuff with variables.
+                isFunction = Attribute.Symbol.isFunction symbolAttributes
```
goodlyrottenapple, line 807: why not inline above?

jberthold: Good point... the separation was from before when I hadn't found that `isFunction` is ready-made. Will fix it.

ana-pantilie (approved):

PR #3265, opened 2022-09-19
Feature/profiling performance workflow

F-WRunTime:
This adds a workflow capable of triggering a performance workflow on any issue comment when the issue is labeled with 'performance'. The results are posted to the action  after successful execution. 

Due to some complexity of setup this meets all requirements  of the ask but one due to limitation of trigger filters and comments. **Posting results as a comment to the issue** this causes an infinite loop to when also triggering runs off existing issues when performance would be desired to run for investigation purposes. 

Finally after successful execution a slack message will be sent to the haskell-backend on successful execution with a link to the results for review. [Success Example here.](https://github.com/runtimeverification/devops-actions/actions/runs/3086102081)
[Artifact Example here](https://github.com/runtimeverification/devops-actions/actions/runs/3085755569)


F-WRunTime:
[Full Success Answer here](https://github.com/runtimeverification/devops-actions/actions/runs/3086243617) Artifact and Success run

F-WRunTime:
Success / Failure will result in a message like the following to the haskell-backend channel
image

.github/workflows/performance.yaml
```diff
@@ -0,0 +1,40 @@
...
+          echo ${github.actor}
+          
+
+
+    
+
+
+
+
+
```
ana-pantilie, line 40: A lot of empty lines here, can they be removed?

```suggestion
```

ana-pantilie (dismissed):
LGTM! Thank you @F-WRunTime !

ana-pantilie:
@F-WRunTime , I just dismissed my approval because @goodlyrottenapple and @jberthold pointed out that this PR doesn't add any workflow for triggering the profile generation. We were wondering if it is incomplete?

.github/workflows/profiling.yaml
```diff
@@ -0,0 +1,120 @@
+---
+name: Performance Profiling
+on:
+  # Trigger on Issue created; Run against master branch
+  issues:
+    types: [opened, edited]
+    branches: 
+      - 'master'
+  
+  pull_request:
```
ana-pantilie, line 10: Does this mean it can be triggered on PRs as well?

.github/workflows/profiling.yaml
```diff
@@ -0,0 +1,120 @@
...
+    types: [labeled, edited]
+    branches:
+      - 'master'
+
+  # Trigger when an Issue is created or modified; Run against master branch only
+  issue_comment:
```
ana-pantilie, line 16: Can you explain how to trigger it with a comment?

jberthold: I think currently _any_ comment will trigger it if the issue/PR is labelled `performance` (line 28).
I found a [discussion in github community where someone suggests `if (contains(...))` to match on the comment content](https://github.com/orgs/community/discussions/25389#discussioncomment-3247738), which we could maybe add here.

jberthold (approved):
LGTM
As commented, we could add the guard to only run this on certain keywords in a comment but that can be tweaked later.

.github/workflows/profiling.yaml
ana-pantilie: We should definitely not trigger the job on _any_ comment. Let's implement ^ in a subsequent PR.

.github/workflows/profiling.yaml
F-WRunTime: A PR and an Issue from GH point of view are essentially the same object type. So a comment in a PR and a comment in an Issue both cause this to trigger. And check our conditions

PR #3268, opened 2022-09-20
Nix profiling tweaks

### LINKED ISSUE #3266: Performance runs: Reduce data and make script more robust
Suggestions to tweak the profiling-run script in nix: 
- suppress user events for heap-infotables and heap-cost-centers runs (use `l-u`) to avoid the grey markers in the graph
- make the script fail fast with `set -eu` (except the `timeout` commands, they need `|| echo "Failure .. "`)
- maybe remove the raw event logs for heap-cost-centers and heap-info-tables (saves more than 60% of the artefact size)

goodlyrottenapple:
Fixes #3266


flake.nix
```diff
@@ -211,6 +211,8 @@
           profiling-script = pkgs.callPackage ./nix/run-profiling.nix {
             inherit (pkgs.haskellPackages)
               hp2pretty hs-speedscope eventlog2html;
+            kore-exec =
```
ana-pantilie, line 214: Question about how nix works here: will this create a separate, normal, build of `kore-exec`?

nix/run-profiling.nix
```diff
@@ -10,38 +12,38 @@ writeScriptBin "run-profiling" ''
...
+  ${gnutar}/bin/tar -xzvf "$1" -C ./profile-tmp
+  cd profile-tmp
 
   # dump version of kore-exec for posterity
-  ${kore-exec-prof}/bin/kore-exec --version > ../$out_folder/kore-exec.version
+  ${kore-exec}/bin/kore-exec --version > ../$out_folder/kore-exec.version
+
+  # generate .eventlog from non-profiled kore-exec to be used by kore-prof
+  timeout --foreground -s INT "$timeout" env GHCRTS='-l' PATH='$PATH:${kore-exec}/bin' ./kore-exec.sh || true
+  cp kore-exec.eventlog ../$out_folder/kore-prof.eventlog
```
ana-pantilie, line 25: Why is this called `kore-prof`?

flake.nix
goodlyrottenapple: yes, `self.project` referred to here is defined as:

https://github.com/runtimeverification/haskell-backend/blob/ebfd83c08b353177cc1f6b9168d8f5412ac2dae1/flake.nix#L154-L160

it defaults to `profiling = false;`, unlike `projectProfilingEventlog` underneath.

nix/run-profiling.nix
goodlyrottenapple: it's an event log produced by an unprofiled version of kore-exec, so it's only useful for consumption by the kore-prof tool, thus the name.

ana-pantilie: Should we also run `kore-prof` on it as well? I can't see a place where we do that in the current script.

goodlyrottenapple: As @jberthold pointed out, we may want to filter out certain messages when the eventlog is big, so it is probably better to just add the eventlog and let the user run kore-prof on it? we can of course run kore-prof with default seetings but we will have double the data, as was the case when including raw eventlogs for the other profiles

ana-pantilie (approved):
LGTM

nix/run-profiling.nix
jberthold: > produced by an unprofiled version of kore-exec, so it's only useful for consumption by the kore-prof tool
Small correction, it also contains the markers for state machine tracing (currently only unification is instrumented). But fine to name it like this.

jberthold (approved):
LGTM

PR #3269, opened 2022-09-21
Handle incorrect log path for `--bug-report` option

### LINKED ISSUE #3263: `--bug-report` option does not handle directories for some arguments
Blocking: https://github.com/runtimeverification/haskell-backend/issues/3221

I have the following example K definition in file `test.k` (the definition itself doesn't matter):

```
module TEST
    imports INT
    imports BOOL

    configuration <k> $PGM:Contract </k>

    rule <k> Coo(_I) => Boo ... </k>

    syntax Contract ::= Coo ( Int ) | "Boo"

    syntax Bool ::= foo ( Int ) [function, functional, no-evaluators]

endmodule

module TEST-SPEC
    imports TEST

    claim <k> Coo(X) => Boo ... </k> ensures foo(X) andBool notBool foo(X)
endmodule
```

I run the following commands:

```
+ mkdir foo
+ kompile --backend haskell test.k --output-definition test-kompiled
+ export 'KORE_EXEC_OPTS=--log foo/bar.log --log-entries DebugTransition --bug-report test'
+ KORE_EXEC_OPTS='--log foo/bar.log --log-entries DebugTransition --bug-report test'
+ kprove --definition test-kompiled test.k --spec-module TEST-SPEC
```

Note the `mkdir foo` on the first line, and the `--log foo/bar.log` on the `export` line (3rd line).

The resulting `test.tar.gz` bug report has a file `kore-exec.sh` that looks like this:

```
#!/bin/sh
exec kore-exec \
    vdefinition.kore \
    --output result.kore \
    --module TEST \
    --strategy all \
    --max-counterexamples 1 \
    --smt-timeout 40 \
    --smt-retry-limit 1 \
    --smt-reset-interval 100 \
    --smt z3 \
    --log \
    foo/bar.log \
    --log-level \
    warning \
    --enable-log-timestamps \
    --log-entries \
    DebugTransition \
    --prove spec.kore \
    --spec-module TEST-SPEC \
    --graph-search breadth-first \
     \
     \
     \
     \
    "$@"
```

In particular, it contains `--log foo/bar.log`, which means when it's run on someone else's machine, we'll get the following error (because directory `foo` does not exist):

```
+ kprove --definition test-kompiled test.k --spec-module TEST-SPEC                                                                                                                                                   
foo/bar.log: openFile: does not exist (No such file or directory)
```

### Proposed Solution 1

The `--bug-report` option should be aware of which arguments are paths (like the `--log` argument), and should insert `mkdir -p $(dirname ARGUMENT)` before calling `kore-exec` to make sure those directories are existing. In this example, it would insert:

```
mkdir -p $(dirname foo/bar.log)
```

### Proposed Solution 2

The `--log` option should make sure that the directory it's logging to exists, by making that directory and all parent directories.

goodlyrottenapple:
Fixes #3263

### DIFF
--- a/kore/src/Kore/Log.hs
+++ b/kore/src/Kore/Log.hs
@@ -113,6 +113,9 @@ withMainLogger reportDirectory koreLogOptions = runContT $ do
                 & koreLogTransformer koreLogOptions
     pure logAction
 
+{- | Checks if the user supplied path for logging exists. If it doesn't, or if the file
+    already exists, will return a default logging location at ./[exe-name]-[prefix]-[timestamp].log
+-}
 checkLogFilePath :: ExeName -> String -> FilePath -> IO FilePath
 checkLogFilePath exeName prefix logFile = do
     pathExists <- doesDirectoryExist $ takeDirectory logFile
--- a/kore/src/Kore/Log.hs
+++ b/kore/src/Kore/Log.hs
@@ -57,9 +57,18 @@ import System.Clock (
     getTime,
     toNanoSecs,
  )
+import System.Directory (
+    doesFileExist,
+    doesDirectoryExist,
+ )
 import System.FilePath (
     (<.>),
     (</>),
+    takeDirectory,
+ )
+import System.IO (
+    hPutStrLn,
+    stderr
  )
 
 -- | Internal type used to add timestamps to a 'LogMessage'.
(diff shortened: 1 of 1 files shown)


jberthold (approved):
LGTM

ana-pantilie (approved):

PR #3272, opened 2022-09-22
Hotfix/profiling workflow triggers

F-WRunTime:
Want to fix a few things I don't like. 
1. Concurrency will stop test in progress we don't want that. 
2. Triggers with the branch definitions don't do anything nix will control how the tests are run. Differ to nix. 
3. Move a conditional check per @jberthold recommendation to have it check the comment context info and cause a skip instead of a failure. 

NOTE: I kept the user check in the body of a run as we should make sure we're informed someone is trying to trigger .tar.gz uploads to our servers and we should investigate who / why / what they're uploading. 

Please let me know if you have any questions 

Methods on how to trigger profiling: 
1. Create a PR with label performance -> In the body add the .tar.gz , no other comments required; But can have comments if desired 
2. Existing PR w/ Label performance -> Add comment with .tar.gz; other comments allowed 
3. Create an Issue w/ label performance -> Add .tar.gz in body of text; other comments are allowed
4. Existing Issue -> Add comment with .tar.gz and label w/ performance; other comments are allowed


.github/workflows/profiling.yaml
```diff
@@ -49,8 +41,13 @@ jobs:
           cat ${GITHUB_EVENT_PATH}
 
           echo "Creating Download URL from ISSUE"
-          echo "DOWNLOAD_URL=$(cat ${GITHUB_EVENT_PATH} | jq '.comment.body' | grep -oP '(https:)(.*?)(?=\))' )" >> ${GITHUB_ENV}
-
+          if ! echo "DOWNLOAD_URL=$(cat ${GITHUB_EVENT_PATH} | jq '.comment.body' | grep -oP '(https:)(.*?)(?=\))' )" >> ${GITHUB_ENV}; then
+            echo "Must be an issue body"
+          else
```
jberthold, line 46: Should we maybe also _skip_ instead of failing when there is no download URL to be found?

jberthold (approved):
LGTM

.github/workflows/profiling.yaml
jberthold: Makes sense. I think we can leave it as it is then, unless it makes noise on PRs for which performance tests should _not_ be run at all.

PR #3284, opened 2022-09-28
Add a debug log entry for proof generation hints (merge of #3260)

goodlyrottenapple:
This is a merge of #3260 into current master. From the original PR:

---

This PR tries to add 4 (debug) log entries:
- `DebugInitialClaim` which records the input claim to be proved
- `DebugInitialPattern` which records the initial pattern when trying to prove each claim
- `DebugFinalPatterns` which records the final patterns when each claim is proven
- `DebugRewriteTrace` which records each rewriting step performed on the current configuration during the proof

They are all defined in `kore/src/Kore/Log/DebugRewriteTrace.hs`.

These entries are disabled by default, and only enabled with the new flag `--trace-rewrites <output>`.
With this flag, kore will write all of these four entries into a YAML-format file `<output>`.

For example, when doing a reachability proof, the output YAML file will look something like this
```
---
claim-id: <UNIQUE'Unds'ID attribute of the claim in kore, if it exists>
claim: <a string with the claim encoded in Kore>
task: reachability
steps:
- initial: { term: ..., constraint: ..., substitution: ... } # the term which the rules are applied to
  applied-rules:
  - rule-id: <UNIQUE'Unds'ID attribute of the rule applied>
    substitution: [ { key: <Kore variable>, value: <Kore term> }, ... ]
    results: [ { term: ..., constraint: ..., substitution: ... }, ... ]
  remainders: <remainder terms when some rule is applied>
- ...


# there could be another claim
---
claim-id: ...
...
```

---

I tested the code on the C smoke test and it appears to work fine: 

[rewrites.yaml.zip](https://github.com/runtimeverification/haskell-backend/files/9666294/rewrites.yaml.zip)


kore/src/Kore/Log/DebugRewriteTrace.hs
```diff
@@ -0,0 +1,267 @@
...
+ )
+import Data.Yaml (
+    ToJSON,
+    Value (..),
+    array,
+    encode,
+    object,
+    toJSON,
+    (.=),
+ )
```
jberthold, line 36: I think we should not introduce a dependency on `yaml`. We currently depend on `conduit-extra` because of `json-rpc` but I don't see the need to use the yaml package here.
The logger prints out a newline-separated sequence of (yaml) objects and  any valid json is valid yaml , too. We can maybe just as well use `aeson-pretty` instead and produce valid yaml without any other changes.
(the reason that this is yaml and not json is that the top-level sequence is not enclosed in `[` and `]` and separated by newlines instead of commas).

kore/src/Kore/Log/DebugRewriteTrace.hs
```diff
@@ -0,0 +1,267 @@
...
+
+instance ToJSON (Substitution VariableName) where
+    toJSON substitution =
+        array (encodeKV <$> unwrap substitution)
+      where
+        encodeKV assignment =
+            object
+                [ "key" .= unparseOneLine (assignedVariable assignment)
+                , "value" .= unparseOneLine (assignedTerm assignment)
+                ]
```
jberthold, line 130: Probably we should not define these orphans here, they might cause us trouble later if at any point we want _actual_ `ToJson` instances (which are not just strings) for  internal types like `TermLike` or `Pattern`.
We can instead define stand-alone functions "renderX :: X -> Value"  and use those explicitly in the instances for `Debug[Initial|Final]...` below.

kore/src/Kore/Log/DebugRewriteTrace.hs
```diff
@@ -0,0 +1,267 @@
...
+
+instance ToJSON DebugInitialClaim where
+    toJSON (DebugInitialClaim uniqueId claim) =
+        object
+            [ "task" .= ("reachability" :: Text)
+            , "claim" .= claim
+            , "claim-id" .= maybe Null toJSON (getUniqueId uniqueId)
+            ]
```
jberthold, line 138: (related to the above, using `renderX :: X -> Value` instead of implicit `ToJSON` via `(.=)` and using `aeson` not `yaml`)
Maybe worthwhile to define a direct `toEncoding` here for better performance?

kore/src/Kore/Log/DebugRewriteTrace.hs
```diff
@@ -0,0 +1,267 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
```
jberthold, line 1: Better add `NoStrict` and `NoStrictData` here?
```suggestion
{-# OPTIONS_GHC -Wno-orphans #-}
{-# LANGUAGE NoStrict #-}
{-# LANGUAGE NoStrictData #-}
```

kore/src/Kore/Log/DebugRewriteTrace.hs
```diff
@@ -0,0 +1,267 @@
...
+  where
+    action entry
+        | Just initial <- fromEntry entry =
+            unLogAction textLogger $ "---\n" <> encode (initial :: DebugInitialClaim) <> "steps:"
+        | Just initial <- fromEntry entry =
+            unLogAction textLogger $ encode (initial :: DebugInitialPattern) <> "steps:"
+        | Just final <- fromEntry entry =
+            unLogAction textLogger $ encode (final :: DebugFinalPatterns)
+        | Just rewrite <- fromEntry entry =
+            unLogAction textLogger $ encode [rewrite :: DebugRewriteTrace]
```
jberthold, line 265: (nit) maybe specify the type where `fromEntry` is called instead of later in the body?
Like so:
```suggestion
        | Just initial <- fromEntry @DebugInitialClaim entry =
            unLogAction textLogger $ "---\n" <> encode initial <> "steps:"
        | Just initial <- fromEntry @DebugInitialPattern entry =
            unLogAction textLogger $ encode initial <> "steps:"
...
```

kore/src/Kore/Log/DebugRewriteTrace.hs
goodlyrottenapple: good point! removed now.

kore/src/Kore/Log/DebugRewriteTrace.hs
goodlyrottenapple: I don't feel strongly about this either way. @ana-pantilie ?

jberthold (approved):
I guess we can keep the `yaml` format for now and merge this.

kore/src/Kore/Log/DebugRewriteTrace.hs
ana-pantilie: Let's keep it for now, I wouldn't spend more time on this if it's functional and prints the output as desired.

ana-pantilie (approved):

PR #3288, opened 2022-09-29
Profiling Trigger changes; modify when triggers only comment in issue…

F-WRunTime:
Profiling Trigger changes:
- modify when triggers only on comment containing a .tar.gz file in issues/pr comments -- No longer checks an issue body to refrain from collecting stale data on a new request or when a conversation is occurring in a profiling performance issue/pr  
- Do not save arbitrary file name before download occurs 
- Notify slack of in progress w/ a link to action in progress; Success / Failure / In Progress (In progress is not sent on failure to acquire files or bad user/label, failure is sent in these conditions instead)


F-WRunTime:
Comments of status report is now made to the issue and not in slack. 
@ehildenb This is ready for review.

F-WRunTime:
@ehildenb Ready for final review

.github/workflows/profiling.yaml
```diff
@@ -40,78 +27,58 @@ jobs:
           echo "Context: "
           cat ${GITHUB_EVENT_PATH}
 
-          echo "Creating Download URL from ISSUE"
-          if ! echo "DOWNLOAD_URL=$(cat ${GITHUB_EVENT_PATH} | jq '.comment.body' | grep -oP '(https:)(.*?)(?=\))' )" >> ${GITHUB_ENV}; then
```
jberthold, line 44: Reading through this, I thought we might want to check whether the DOWNLOAD_URL is going to work already here... 
could use `curl --head --fail $DOWNLOAD_URL` to test it without an actual download.

.github/workflows/profiling.yaml
```diff
@@ -40,78 +27,58 @@ jobs:
...
+          rm -rf ./*
       - name: Run Tests
         run: |
           set -euo pipefail
+          ISSUE_NUMBER=$(cat ${GITHUB_EVENT_PATH} | jq '.issue.number')
+          gh issue comment ${ISSUE_NUMBER} --body "IN PROGRESS: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          DOWNLOAD_URL=$(cat ${GITHUB_EVENT_PATH} | jq '.comment.body' | grep -oP '(https:)(.*?)(?=\))' )
+          FILE_NAME=$(basename ${DOWNLOAD_URL})
+          echo "Download Test File"
+          curl -LOsS ${DOWNLOAD_URL}
```
jberthold, line 48: See above, we could check whether this is going to work on the flyweight executor (and not send a message if it wouldn't).

jberthold (approved):
OK to go in, but could tweak it a bit more to avoid message noise

Issue #3113, opened 2022-06-22
Serokell: hashing is a top consumer


ana-pantilie:
We found that the cache in attemptEquation enjoys the hit ratio of over 95%, so it is well-placed. However, the lookupCache is expensive because a lot of time is spent hashing and comparing keys.

Suggestion:
1. compute the hash in advance
2. investigate using a different datastructure for less comparisons (like a trie)

ana-pantilie:
For this issue, please prioritize providing some ideas and documentation for how to tackle this issue.

dcastro:
Here's a summary of what we've tried:

* Replacing `HashMap` with a trie map, see the branch `diogo/#3133-cache-hit-ratio-trie`. It didn't have a noticeable impact on performance, possibly due to the need to "linearize" the key.
* Replacing `HashMap` with a `Map`, resulted in performance degradation.
* Avoiding computing an `Equation`'s hash more than once. Negligible impact, 0.93% overall speedup.
* Avoiding computing a `TermLike`'s hash more than once, see the branch `diogo/#3113-cache-hit-ratio-terms`. Negligible impact, 0.33% overall speedup.

There's also an ongoing attempt to store the hash of each `TermLikeF` node in the corresponding `TermAttributes` in the branch `jb/caching-termlike-hashes`.

dcastro:
A couple more details about `jb/caching-termlike-hashes`:

We ran our benchmarks against ebd826f19869725d887de3a99a727cfd97dad470 (using ac0e3c0c403bbca4d812d921f8783f4dfcdbcb29 from master as a baseline). 47 benchmarks succeeded, 23 crashed (most with an "Invalid map: duplicated keys" error).

Considering only the 47 successful benchmarks: 

* Total runtime using an executable built from the target branch: 9451.694s
* Total runtime using an executable built from the baseline branch: 10065.793s
* 9451.694 / 10065.793 = 0.938991493, so a ~6.1% overall speedup.

The results look promising, but at this point, it's not yet known whether this approach is feasible without breaking correctness.

ana-pantilie:
@goodlyrottenapple points out that there might be other subtle bugs created by not updating attributes (like the `Simplified` attribute) in other situations. This is related to doing optimal hashing (if we continue on the path @jberthold started).

ana-pantilie:
Relevant branch with in progress work: https://github.com/runtimeverification/haskell-backend/commits/jb/breakerzirconia/caching-termlike-hashes

[...]

Issue #3286, opened 2022-09-29
[Haskell-Performance] kevm proof slow with extra side conditions


ehildenb:
### Please Prepare Test Data

I have a specification on KEVM: https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/storagevar03-spec.k

This can be run with: `make tests/specs/benchmarks/storagevar03-spec.k.prove` after building KEVM.

It has some side-conditions for validity on the input state: https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/storagevar03-spec.k#L116. But these side-conditions are actually not needed (the lines `CONTRACT_ID >Int 0 andBool notBool #isPrecompiledAccount(CONTRACT_ID, BYZANTIUM)`), so I removed them.

Removing these 4 conditions (on `CONTRACT_ID` and `CALLEE_ID`) ends up improving the performance by ~3X.

So it seems that having these extra side-conditions is causing a big performance slowdown.

I've attached 2 bug reports: (i) [kevm-bug-storagevar03-with-conditions.tar.gz](https://github.com/runtimeverification/haskell-backend/files/9670919/kevm-bug-storagevar03-with-conditions.tar.gz), which is run with the extra conditions as linked above, which takes ??? on my machine, and (ii) , which is run with those conditions removed and takes ??? on my machine.

ehildenb:
@F-WRunTime it seems that it cannot detect that I'm a member of the org? https://github.com/runtimeverification/haskell-backend/actions/runs/3148634684/jobs/5119382938

ehildenb:
Also, it seems to create too many jobs: 
![image](https://user-images.githubusercontent.com/12721972/192939558-7bceb23b-4caf-439b-96f5-dda6f9a54747.png)

ehildenb:
dummy comment to trigger workflow

ehildenb:
Test me.

ehildenb:
Test me.

jberthold:
Ran the two tarballs in a separate ticket #3303 . Nothing immediately suspicious in the profiles, but the top consumers differ to significant extent:

`with-conditions`

| COST CENTRE |                  MODULE                                 | SRC|                                                            %time | %alloc | 
|------------|------------|------------|------------|------------|
[...]

PR #3306, opened 2022-10-07
Experiment caching TermLike hashes

treeowl:
Cache the hash of the `TermLikeF` in the `TermLike`, and extract it in the `Hashable` instance instead of recursing into the `TermLikeF`.

The precomputed hash in produced using the derived `Hashable` instance of `TermLikeF`, which will call `hash` on the respective sub-structure (like `And`, `Or`, ...), wrapped into one of the `TermLikeF` constructors.  For all such sub-structures, the derived `Hashable` instance will then recurse into the contents (containing a `TermLike` again for the case of `TermLike`), thereby using the cached hashes directly instead of performing a deep recursion.

Partially addresses #3113

Fixes #nnnn

### Estimate:

---

###### Review checklist


treeowl:
There's a problem: `mapVariables` and `traverseVariables` generally need `Hashable variable2` to produce their results. However, when `variable1 ~ Concrete`, there are no variables in the input, which (I think?) means there won't be any in the output either. So it should be possible to write similar functions that don't put any constraints on the result variable type. Unfortunately, I don't understand enough about what's going on in `traverseVariables` to know how to adapt it properly.

It would be good to get this figured out, because `mapVariables` "infects" a bunch of other functions and instances with `Hashable` constraints that probably shouldn't be necessary.

treeowl:
Hmmm... All the Nix jobs are failing. Any guesses?

kore/src/Kore/Internal/TermLike/TermLike.hs
```diff
@@ -647,34 +648,31 @@ instance HasFreeVariables (TermAttributes variable) variable where
...
-    { getTermLike ::
-        CofreeF
-            (TermLikeF variable)
-            (TermAttributes variable)
-            (TermLike variable)
+data TermLike variable = TermLike__
+    { _tlAttributes :: TermAttributes variable -- Should this be strict?
+    -- The hash is stored lazily to avoid computing it unnecessarily. Do we generally
+    -- need the hash? If so, it should probably be strict.
+    , -- | A hash of _tlTermLikeF
```
jberthold, line 657: A quick heads-up about this: 
I would definitely prefer if the code was lazy by default but we currently have `Language Strict, StrictData` on by default and turn it of per module. (There is an experimentation ticket to switch to lazy and try to plug holes, to see the performance difference, but always too many things to do).

treeowl: Oof, that's very confusing!

jberthold:
> Hmmm... All the Nix jobs are failing. Any guesses?

It is probably a GHC-9 specific error. Only nix jobs are testing GHC-9 in our CI setup (main build is with GHC-8.10.7 but we have the GHC-9 setup for profiling and to migrate to it at some point).
You could maybe test it by changing the resolver to what nix GHC-9 uses, see
https://github.com/runtimeverification/haskell-backend/blob/master/nix/kore-ghc923.nix.d/default.nix#L41

goodlyrottenapple:
```
src/Kore/Internal/TermLike/TermLike.hs:695:10: error:
    • No instance for (Eq variable)
        arising from the superclasses of an instance declaration
      Possible fix:
        add (Eq variable) to the context of the instance declaration
    • In the instance declaration for ‘Hashable (TermLike variable)’
    |
695 | instance Hashable (TermLike variable) where
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

you can run `nix build .#kore-exec-ghc9 ` locally to check for any more issues with ghc9

treeowl:
@goodlyrottenapple, I see. That's quite a mess. Has the `Hashable` change been dealt with elsewhere in this project? Is there an accepted approach?

treeowl:
It seems GHC in Haddock mode has some sort of trouble with the CPP I need to support both versions of `hashable` we're currently using. Is there some local fix for that? Or can we bump the older `hashable` version to a suitable minimum?

treeowl:
I have no idea how to interpret the remaining test failures.

treeowl:
`vacuousVariables` really should be a no-op operationally, shouldn't it? `unsafeCoerce` would do? I imagine there are some reasons (formal verification?) not to do that, but it seems a bit sad to walk a whole term to change nothing at all.

jberthold:
Profiling run with file from #3288 [kevm-bug-storagevar03-with-conditions.tar.gz](https://github.com/runtimeverification/haskell-backend/files/9670919/kevm-bug-storagevar03-with-conditions.tar.gz) (test)

jberthold:
Profiling with file from #3286 : [first file, kevm-bug-storagevar03-with-conditions.tar.gz](https://github.com/runtimeverification/haskell-backend/files/9670919/kevm-bug-storagevar03-with-conditions.tar.gz)

goodlyrottenapple:
> @goodlyrottenapple, I see. That's quite a mess. Has the `Hashable` change been dealt with elsewhere in this project? Is there an accepted approach?

Don't think its been an issue so far. We added GHC 9 a few months ago and I remember there were only a few changes necessary at the time to make the code compile for both, so as such, there is no apporach in place. We have an issue open to switch to GHC9, but need to go through our dependencies again and see if support for GHC9 has been up-streamed for the packages which weren't compatible at the time.

goodlyrottenapple:
> It seems GHC in Haddock mode has some sort of trouble with the CPP I need to support both versions of `hashable` we're currently using. Is there some local fix for that? Or can we bump the older `hashable` version to a suitable minimum?

Think we can just bump `hashable` if it makes things easier.

treeowl:
Closing in favor of #3338; this branch was off a fork, which apparently causes issues.

Issue #3298, opened 2022-10-05
Implement the `simplify` endpoint for `kore-rpc`


tothtamas28:

ana-pantilie:
This should just call the pattern simplifier and return the result.

ana-pantilie:
Blocked on https://github.com/runtimeverification/haskell-backend/issues/3324

ana-pantilie:
Fixed by https://github.com/runtimeverification/haskell-backend/pull/3307

PR #3307, opened 2022-10-07
RPC server `simplify` endpoint

goodlyrottenapple:
Implements [#3298](https://github.com/runtimeverification/haskell-backend/issues/3298)

This PR adds the `simplify` endpoint to the JSON RPC server

It also removes re-exports from `Kore.Simplify.API`


kore/src/Kore/Simplify/API.hs
```diff
@@ -12,24 +12,6 @@ Portability : portable
 module Kore.Simplify.API (
     evalSimplifier,
     evalSimplifierProofs,
-
-    -- * Re-exports
-    Env (..),
-    Simplifier,
-    runSimplifier,
```
jberthold, line 19: Maybe we should rename the module (again :thinking: ) then?
An `API` module will probably always have a small number of re-exports...

kore/test/Test/Kore/Rewrite/MockSymbols.hs
```diff
@@ -98,17 +98,14 @@ import Kore.Rewrite.SMT.AST qualified as SMT
 import Kore.Rewrite.SMT.Representation.Resolve qualified as SMT (
     resolve,
  )
-import Kore.Simplify.API (
-    Env (Env),
-    MonadSimplify,
- )
-import Kore.Simplify.API qualified as SimplificationAPI.DoNotUse
```
jberthold, line 105: That one we should _definitely_ delete :man_facepalming:

test/rpc-server/execute/branching/response.golden
```diff
@@ -1 +1 @@
...
...","sorts":[],"args":[{"tag":"App","name":"Lbl'-LT-'k'-GT-'","sorts":[],"args":[{"tag":"App","name":"kseq","sorts":[],"args":[{"tag":"App","name":"inj","sorts":[{"tag":"SortApp","name":"SortState","args":[]},{"tag":"SortApp","name":"SortKItem","args":[]}],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortState","args":[]},"value":"c"}]},{"tag":"App","name":"dotk","sorts":[],"args":[]}]}]},{"tag":"App","name":"Lbl'-LT-'generatedCounter'-GT-'","sorts":[],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"0"}]}]}}},{"term":{"format":"KORE","version":1,"term":{"tag":"App","name":"Lbl'-LT-'generatedTop'-GT-'","sorts":[],"args":[{"tag":"App","name":"Lbl'-LT-'k'-GT-'","sorts":[],"args":[{"tag":"App","name":"kseq","sorts":[],"args":[{"tag":"App","name":"inj","sorts":[{"tag":"SortApp","name":"SortState","args":[]},{"tag":"SortApp","name":"SortKItem","args":[]}],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortState","args":[]},"value":"b"}]},{"tag":"App","name":"dotk","sorts":[],"args":[]}]}]},{"tag":"App","name":"Lbl'-LT-'generatedCounter'-GT-'","sorts":[],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"0"}]}]}}}]}}
```
jberthold, line 1: OH, so the server tests weren't run in CI after your `aeson-pretty` change?
We could add a `Makefile` that runs them on `make test` (would require a `test-<something>.sh` script and its expected output in a file if I read `test/include.mk` correctly)

jberthold (commented):
LGTM overall except the `toPattern`

kore/src/Kore/Simplify/API.hs
goodlyrottenapple: yeah, if you look at the changes, it was always just like 3 `Kore.Simplify.Simplify` types/functions that were being exported...most of the re-exports were redundant anyway... given the functions in this module maybe `Kore.Simplify.Env` would be a better name now?

kore/src/Kore/JsonRpc.hs
```diff
@@ -432,7 +419,36 @@ respond runSMT serializedModule =
...
+
+                    pure $
+                        Right $
+                            Simplify
+                                SimplifyResult
+                                    { state =
+                                        PatternJson.fromTermLike $
+                                            TermLike.mapVariables getRewritingVariable $
+                                                Pattern.toTermLike $
+                                                    OrPattern.toPattern sort result
```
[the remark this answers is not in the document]
goodlyrottenapple: oh thanks for that one. I just stringed together the chain of conversions from the existing code, but wasn't entirely sure whether correct. Will use the suggested function above

test/rpc-server/execute/branching/response.golden
goodlyrottenapple: Should we do that in a separate PR? since the original nix derivation for the integration tests would need to be tweaked to include python

.github/workflows/test.yml
```diff
@@ -94,6 +95,9 @@ jobs:
       - name: Run integration tests
         run: nix-build test.nix
 
+      - name: Run JSON RPC server tests
+        run: nix build .#rpc-tests
```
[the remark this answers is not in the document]
goodlyrottenapple: Maybe we just move the test into the rest of integration tests as suggested below and remove `.#rpc-tests` in favour of the more uniform `nix-build test.nix --argstr test rpc-server`?

test/rpc-server/Makefile
```diff
@@ -0,0 +1,6 @@
+test:
+	python3 runTests.py
```
jberthold, line 2: Unfortunately we need other targets, too: `build-test` and `test-k` are [defined via `include.mk`](https://github.com/runtimeverification/haskell-backend/blob/master/test/include.mk#L157) and used by [scripts/integration-k.sh in Jenkins](https://github.com/runtimeverification/haskell-backend/blob/master/scripts/integration-k.sh#L25)

jberthold (commented):
`test/prof-smoketest/test-prof-speedscope-eventlog` needs to be deleted

jberthold (approved):
Pre-approving the changes, LGTM once the `Makefile` is fixed and the speedscope file removed

goodlyrottenapple:
@F-WRunTime can we add the `jsonrpcclient` library for python3 on the jenkins runner? should just be a case of `pip3 install jsonrpcclient`

F-WRunTime:
Do we want to label this with automerge? 
And jenkins runner is updated with pip package python3-pip wasn't installed.

jberthold:
> Do we want to label this with automerge? And jenkins runner is updated with pip package python3-pip wasn't installed.

Yes we want to (I just did). There is another jenkins problem, we need to decide how we proceeds as we were anyway planning to remove jenkins IIUC.

PR #3315, opened 2022-10-11
Use correct token to post comment, avoid checkout work

jberthold:
Two more small adjustments to the profiling workflow.

Currently the final `on_success` and `on_failure` have the wrong token and fail to post a message, see https://github.com/runtimeverification/haskell-backend/actions/runs/3223960737/jobs/5274887739

### DIFF
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -28,6 +28,7 @@ jobs:
           if ! gh api -H "Accept: application/vnd.github+json" /orgs/runtimeverification/members/$GITHUB_ACTOR; then
             exit 1
           fi
+          echo "Github Ref: '${{ env.GITHUB_REF }}'"
           echo "Context: "
           cat ${GITHUB_EVENT_PATH}
       - name: Check File Downloadable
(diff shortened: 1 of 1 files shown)


jberthold:
Also, we are still checking out master instead of the relevant PR commit, see  https://github.com/orgs/community/discussions/25191

goodlyrottenapple:
@F-WRunTime can we switch to this branch being used for the profiling runs? This way we can keep iterating and once happy can merge into master and switch to using master when a profiling run gets triggered

F-WRunTime:
You mean to run these from this branch? Alright I had to make a 1 line change to master to enable dispatch runs. These can now be run manually. I didn't test this method so it might be a bit of a disaster. I'll kick one off and see how it goes, but we can work with this from here: 
image
You can get to this by going to actions and then selecting which branch version of the workflow you want to run

F-WRunTime:
[kevm-bug-storagevar03-without-conditions.tar.gz](https://github.com/runtimeverification/haskell-backend/files/9760044/kevm-bug-storagevar03-without-conditions.tar.gz)

jberthold:
> You mean to run these from this branch? Alright I had to make a 1 line change to master to enable dispatch runs. These can now be run manually.

Interesting! does it "default" to `master` after a timeout or so? Because the dispatch would definitely be handy right now but not   in the intended use case....

F-WRunTime:
> > You mean to run these from this branch? Alright I had to make a 1 line change to master to enable dispatch runs. These can now be run manually.
> 
> Interesting! does it "default" to `master` after a timeout or so? Because the dispatch would definitely be handy right now but not in the intended use case....

Well with dispatch runs we can run gh from the command line and specify that -R flag to run a specific branch flavor of a test or utilize whatever else is just on that branch. 

Didn't originally intend to use branches to run these because nix was essentially repository agnostic

jberthold:
So, IIUC what it does I think the `workflow_dispatch` is definitely useful.

What I meant to say with this comment: 

> Also, we are still checking out master instead of the relevant PR commit, see https://github.com/orgs/community/discussions/25191

was that we want to check out and run the  _Haskell code_ from the branch for a performance run

F-WRunTime:
@jberthold clarified his statement for me. Think we just need to add this to the checkout commands: 
```
  with:
    ref: ${{ github.event.pull_request.head.sha }}
 ```

F-WRunTime:
Test build using my fake.tar.gz

F-WRunTime:
> @F-WRunTime can we switch to this branch being used for the profiling runs? This way we can keep iterating and once happy can merge into master and switch to using master when a profiling run gets triggered

I agree this would be nice. It looks like to make that happen going forward we'll need to merge these changes. and should be in a closer state to utilizing PR code vs Master code on kick off of profiling.

F-WRunTime (approved):

PR #3329, opened 2022-10-20
Disable Integration and Unit tests in jenkins

### LINKED ISSUE #3324: Remove Jenkins jobs
We already run the integration tests in GitHub workflows, and we also run the unit tests using both Stack and Cabal in separate workflows as well. We should remove the jobs which do the same things on Jenkins.

Also take a look at:
- [x] rename workflows to be more descriptive
- [x] we shouldn't be building twice in the Stack unit test workflow
- [x] ~what does the documentation stage in the Jenkins pipeline do? do we need to add a new workflow which does the same thing?~ add Haddock pass in parallel
- [ ] check if the K release upgrade is defined in the same Jenkinsfile, if yes we need to migrate that as well

goodlyrottenapple:
Partially fixes  #3324, namely:

* Better labels for cabal and stack unit tests
* remove unnecessary build step for unit tests in cabal and stack
* remove unit and integration tests from jenkins

I will leave the integration tests update to a separate job because it will probably involve much more testing and i would like to re-enable the json rpc integration tests in github actions asap.


goodlyrottenapple:
I have left the docs stage in jenkins for now, should we port it over or just discard entirely @ana-pantilie? This is what gets called:
https://github.com/runtimeverification/haskell-backend/blob/7922cf79a16b66356f1874407798beb7f7f26793/Makefile#L29-L41

ana-pantilie:
Is it easy to port?

goodlyrottenapple:
> Is it easy to port?

should be. could be tacked onto the stack unit test run maybe? could test and see how it goes

ana-pantilie (approved):

PR #3339, opened 2022-10-27
Add override option for K to KEVM timing test action

jberthold:
The triggering comment may now contain a version tag (of shape `v[0-9]+\.[0-9]+\.[0-9]+`) after the trigger string `/KEVM-perf-run` (separated by a single space). If no such tag is given, the K version is master.


goodlyrottenapple (commented):
lgtm

ehildenb (approved):

PR #3352, opened 2022-11-02
Document performance test automations, small fix

### LINKED ISSUE #3303: [Haskell-Performance] Testing for issue 3286 (please ignore)
Running a performance test for #3286 

EDITED This is now [the first test tarball (with-conditions)](https://github.com/runtimeverification/haskell-backend/files/9670919/kevm-bug-storagevar03-with-conditions.tar.gz) from #3286

### LINKED ISSUE #3323: On-demand performance (timing) test with KEVM proofs
In the style of the on-demand profiling run using a provided tarball, create a github action which runs KEVM proofs with timing results (see https://github.com/runtimeverification/evm-semantics/issues/1165) using HS backend code from the given PR branch.
The timing run and the comparison script (python) are provided as a nix derivation. In addition to actually running the proofs with the branch, we will need a comparison run with HS backend master  (which we can schedule weekly and would probably just store in a fixed location)

jberthold:
Fixes #3323 
Fixes #3303 

* adds documentation for the two automations
* fixes the `in  progress` message for the KEVM timing test (stating the K version correctly now)

The KEVM timing tests still fail for some reason but quite probably the problem is in the nix derivation, not in this repository.


docs/2022-11-02-perf-test-automation.md
```diff
@@ -0,0 +1,73 @@
...
+
+Actions are implemented as nix derivations, i.e. they run in their own isolated environment, possibly using particular versions of their dependencies.
+
+## 1. Profiling runs
+
+[github action](../.github/workflows/profiling.yaml), 
+[run-profiling.nix](../nix/run-profiling.nix)
+
+### Trigger Event
+- a comment containing the string `".tar.gz)"` (anywhere)
```
ana-pantilie, line 22: Is this supposed to contain the `)`?

ana-pantilie (approved):
Cool, thank you! One small comment

docs/2022-11-02-perf-test-automation.md
jberthold: Yes, that's on purpose. We want this to be a _link_, not just a mention of some.tar.gz like here
(if I put a closing parenthesis behind that, the action will start and then fail to extract a valid download URL)

PR #3357, opened 2022-11-04
Add IO hooks

goodlyrottenapple:
Fixes https://github.com/runtimeverification/haskell-backend/issues/3296

This PR adds the following new hooks:
- IO.open
- IO.close
- IO.getc
- IO.read
- IO.write

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -1033,3 +1033,50 @@ Log string to terminal: The string argument is logged as a InfoUserLog log messa
     hooked-symbol logString{}(String{}) : K{}
         [hook{}("IO.logString")]
 ~~~
+
+
+### IO.open
+
+Open a file and return the file descriptor or an IOError
+
+~~~
+    hooked-symbol open{}(/* path */ String{}, /* mode */ String{}) : IOInt{}
+        [hook{}("IO.open")]
+~~~
+
+
+### IO.close
+
+Close a file descriptor and return `dotk{}()` if no error was thrown, otherwise return an IOError.
+
+~~~
+    hooked-symbol close{}(/* fd */ Int{}) : Item{}
+        [hook{}("IO.close")]
+~~~
+
+### IO.getc
+
+Read a single character from a file. Returns an `#EOF{}()` if no mor characters are available.
+
+~~~
+    hooked-symbol getc{}(/* fd */ Int{}) : IOInt{}
+        [hook{}("IO.getc")]
+~~~
+
+
+### IO.read
+Read from a file into a buffer of size `length` and return a `String{}()` if successfull, otherwise return an `IOError{}()`
+
+~~~
+    hooked-symbol read{}(/* fd */ Int{}, /* length */ Int{}) : IOString{}
+        [hook{}("IO.read")]
+~~~
+
+
+### IO.write
+Write the given string into a file. Returns an `IOError{}()` if the write fails and `dotk{}()` otherwise.
+
+~~~
+    hooked-symbol write{}(/* fd */ Int{}, /* value */ String{}) : Item{}
+        [hook{}("IO.read")]
+~~~
--- a/kore/src/Kore/Builtin/IO.hs
+++ b/kore/src/Kore/Builtin/IO.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE RecordWildCards #-}
+
 {- |
 Module      : Kore.Builtin.IO
 Description : Built-in I/O (limited support)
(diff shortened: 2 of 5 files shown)


kore/src/Kore/Builtin/IO.hs
```diff
@@ -74,29 +113,409 @@ symbolVerifiers =
...
+
+    open_mode = case how of
+        ReadOnly -> o_RDONLY
+        WriteOnly -> o_WRONLY
+        ReadWrite -> o_RDWR
+
+foreign import capi unsafe "HsUnix.h open"
+    c_open :: CString -> CInt -> CMode -> IO CInt
+
+openRaw ::
```
goodlyrottenapple, line 214: most of the `*Raw` functions were taken from the [`System.Posix.IO`](https://hackage.haskell.org/package/unix-2.8.0.0/docs/System-Posix-IO.html) module, but needed to be adapted to return the error as K's `IOError` rather than throwing the error within Haskell.

kore/src/Kore/Builtin/IO.hs
```diff
@@ -74,29 +113,409 @@ symbolVerifiers =
...
+                if err == eINTR
+                    then returnErrnoIfMinus1OrRetry f
+                    else return $ Left err
+            else return $ Right res
+
+parseMode :: Text -> Maybe (OpenMode, OpenFileFlags, Maybe FileMode)
+parseMode = \case
+    "r" -> Just (ReadOnly, defaultFileFlags, Nothing)
+    "r+" -> Just (ReadWrite, defaultFileFlags, Nothing)
+    "w" -> Just (WriteOnly, defaultFileFlags{trunc = True}, Just 0o666)
```
goodlyrottenapple, line 173: `0o666` is a magical constant I found here https://github.com/ghc/ghc/blob/311251543f2e37af4a121e58028bfc46267a7fc9/libraries/base/GHC/IO/FD.hs#L238 :P

docs/hooks.md
```diff
@@ -1033,3 +1033,50 @@ Log string to terminal: The string argument is logged as a InfoUserLog log messa
...
+Close a file descriptor and return `dotk{}()` if no error was thrown, otherwise return an IOError.
+
+~~~
+    hooked-symbol close{}(/* fd */ Int{}) : Item{}
+        [hook{}("IO.close")]
+~~~
+
+### IO.getc
+
+Read a single character from a file. Returns an `#EOF{}()` if no mor characters are available.
```
ana-pantilie, line 1059: `more`* typo?

kore/src/Kore/Builtin/IO.hs
ana-pantilie: Can you add this as a comment in the code? Will probably be useful later

kore/src/Kore/Builtin/IO.hs
ana-pantilie: Same as above

kore/src/Kore/Builtin/IO.hs
```diff
@@ -74,29 +113,409 @@ symbolVerifiers =
...
+    -- "with errno set to [EINTR] [...] the state of fildes is unspecified"
+    -- and on Linux, already the first close() removes the FD from the process's
+    -- FD table so closing a second time is invalid
+    -- (see http://man7.org/linux/man-pages/man2/close.2.html#NOTES).
+    rc <- returnErrnoIfMinus1 $ c_close fd
+    return $ case rc of
+        Left err -> Just err
+        Right _ -> Nothing
+
+foreign import ccall unsafe "HsUnix.h close"
```
ana-pantilie, line 486: 😨

ana-pantilie (commented):
Well, this one was tricky. Thank you for implementing it!

Is it possible to add an integration test for this?

kore/src/Kore/Builtin/IO.hs
goodlyrottenapple: >If the programmer can guarantee that the imported function won't call back into Haskell, the foreign import can be marked as 'unsafe' (see Section [Section 2](https://www.haskell.org/definition/ffi/sec-primitive.html) for details of how to do this.) Unsafe calls are cheaper than safe ones, so distinguishing the two classes of external calls may be worth your while if you're extra conscious about performance.

goodlyrottenapple:
> Well, this one was tricky. Thank you for implementing it!
> 
> Is it possible to add an integration test for this?

yes, will add one.

test/io-hooks/test.k
```diff
@@ -0,0 +1,14 @@
...
+  imports K-IO
+
+  syntax IOString ::= readFileAux(IOInt) | readFile(String) [function, impure]
+
+  rule readFileAux(#ENOENT) => "File not found"
+  rule readFileAux(FD:Int) =>
+    #let TXT = #read(FD, 10) #in
+    #let _ = #close(FD) #in TXT
+
+  rule readFile(F) => readFileAux(#open(F))
```
goodlyrottenapple, line 13: Added a simple read test. Testing write would be a bit more tricky and don't have time to add one now. @ana-pantilie if you think we should add a write test too, let me know here and I can add it when I get back.

ana-pantilie: @goodlyrottenapple maybe add it in a subsequent PR, after you get back. We'll discuss this then.

ana-pantilie (approved):

PR #3369, opened 2022-11-21
Strict hash caching

treeowl:
Strict hash caching

Fixes #nnnn

### Estimate:

---

###### Review checklist

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -235,6 +235,7 @@ newline: native
 # Please keep this up-to-date with default-extensions in package.yaml
 language_extensions:
   - BangPatterns
+  - CPP
   - ConstraintKinds
   - DataKinds
   - DefaultSignatures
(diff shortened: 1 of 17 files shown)


jberthold:
Looked at the measurements... unfortunately this is slower than master . :disappointed: So it is the other PR that we want, I guess.

treeowl:
@jberthold There's also a very mysterious error in the integration tests. This is all very odd. Are there any actually lazy operations other than `fromConcrete`?

treeowl:
Or maybe there's something else lazy and the hash never gets taken? Very odd.

jberthold:
> @jberthold There's also a very mysterious error in the integration tests. This is all very odd. Are there any actually lazy operations other than `fromConcrete`?

That error is actually a flake - sometimes the SMT solver cannot keep up (or something else goes wrong) and it sends this `Push canceled` response, making the test fail. A re-run will show whether or not this is the problem here.

> Or maybe there's something else lazy and the hash never gets taken? Very odd.

I could imagine there are many `TermLike` that never make it into the simplifier cache - in the strict version we will compute their hashes unnecessarily. Just a guess - but whatever the reason,  the numbers speak clearly.

treeowl:
Okay, we won't do it then.

PR #3374, opened 2022-11-22
Stop using msplit

treeowl:
`Control.Monad.Logic.msplit` is well known to be slow. See, most notably, the [Reflection without Remorse](https://okmij.org/ftp/Haskell/zseq.pdf) paper. Use a custom variant of `MonadLogic` with a `gather` method to avoid needing it. The `ReaderT` and `LogicT` instances (at least) can be written to avoid the expensive "reflection" where the rest of the computation is rebuilt after each element is produced.

Fixes #nnnn

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Unification/UnifierT.hs
+++ b/kore/src/Kore/Unification/UnifierT.hs
@@ -12,6 +12,7 @@ module Kore.Unification.UnifierT (
     module Kore.Unification.Unify,
 ) where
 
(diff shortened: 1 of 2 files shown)


treeowl:
Nope, not enough. There are a bunch of `ifte` uses, which are just as bad as using `msplit` directly. I'll have to see what happens if we switch to `logict-sequence`. I'll try to release the second major version of that package tomorrow.

PR #3381, opened 2022-11-28
Gather more efficiently

treeowl:
`Control.Monad.Logic.msplit` is well known to be slow. See, most notably, the
[Reflection without Remorse](https://okmij.org/ftp/Haskell/zseq.pdf) paper. Use a custom variant of `MonadLogic` with a `gather` method to implement `gather` more efficiently.

Fixes #nnnn

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Builtin/Builtin.hs
+++ b/kore/src/Kore/Builtin/Builtin.hs
@@ -118,9 +118,6 @@ import Kore.Simplify.Simplify (
     TermSimplifier,
     applicationAxiomSimplifier,
  )
-import Kore.Unification.Unify (
-    MonadUnify,
- )
 import Kore.Unification.Unify qualified as Unify
 import Kore.Unparser
 import Prelude.Kore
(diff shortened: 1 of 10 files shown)


treeowl:
Since the `logict-sequence` idea seems to be stalled, we might as well see if this much more conservative change is worthwhile.

treeowl:
No difference. I guess `gather` isn't producing long lists in these tests. Are there realistic situations where it might?

ana-pantilie:
@ehildenb suggested crafting an example which branches a lot. We can find out if this PR actually has an impact in such a situation.

ehildenb:
[example.zip](https://github.com/runtimeverification/haskell-backend/files/10115108/example.zip)

This example contains two files: `test.k` and `inp.pgm`. You can run it with:

```
kompile test.k --backend haskell --main-module TEST --syntax-module TEST
krun --depth 1 inp.pgm --output kore --parser cat --term --no-expand-macros
```

And it causes a branch of size 1010 (which should trigger constructing a multior of size 1010).

You can make the branching bigger or larger by adding more/less rules in `test.k`, and rekompiling and rerunning.

Maybe this helps us to understand the performance implications of `gather` changes? Not sure, because I don't know if this example will actually exercise that case.

treeowl:
The example @ehildenb unfortunately only produced `gather`s of size 1. Closing this for now, as we don't know of a test case where it makes a difference.

PR #3392, opened 2022-12-01
Try using hoistLogicT

treeowl:
Do we get the same (or equivalent) results when we `hoistLogicT` instead of running the computation and then scattering? It will depend on what the actions are up to.

Fixes #nnnn

### Estimate:

---

###### Review checklist


treeowl:
Does anyone know if there's a reason the code was written to collect all the results and then scatter them, rather than working in a streaming fashion? Changing that doesn't break any of the tests, but I'd like to be sure I'm not doing something really unexpected here.

treeowl:
The KEVM test saw no perf difference. I think this is cleaner from a code standpoint; I can't speak to what makes more sense algorithmically when it comes to the effect ordering.

ana-pantilie (approved):

kore/src/Logic.hs
```diff
@@ -34,8 +36,18 @@ mapLogicT ::
     (forall x. m x -> n x) ->
     LogicT m a ->
     LogicT n a
-mapLogicT nat acts = (lift . nat) (observeAllT acts) >>= scatter
+mapLogicT nat acts = hoistLogicT nat acts
 {-# INLINE mapLogicT #-}
+#if !MIN_VERSION_logict(0,8,0)
```
ana-pantilie, line 41: From @jberthold : we should remove this once we are definitely on the `0.8` version

PR #3394, opened 2022-12-01
Make make tail recursive

treeowl:
`make` would recursively pattern match on the result of recursive computations. That's not great if there are many of them, so stop doin that. The modified code is also arguably easier to read.

Fixes #nnnn

### Estimate:

---

###### Review checklist


treeowl:
I just realized this could be bad for performance if `MultiOrTop` is pretty common; in that case it's better to wait and see if the `Set` will be needed until the end, rather than building it eagerly. Does anyone know?

goodlyrottenapple:
I made changes to MultiOr last and it was somewhat tricky to get right. From what I remember it is called fairly often? I suggest you profile a run with your change vs master and look at the profiles. I found using speedscope useful for seeing the calls to make.

treeowl:
Here's the basic performance report.
[perf-rep.pdf](https://github.com/runtimeverification/haskell-backend/files/10171414/perf-rep.pdf).
I'll add profiling when I can.

ana-pantilie:
Differences are indeed negligible, I think we can just merge this.

ana-pantilie (approved):

PR #3398, opened 2022-12-07
Transfer not simplified claims to applyClaim

### LINKED ISSUE #3377: Claim-splitting inside original claim, not top-level
If an initial simplification of a given claim yields a split (e.g., into two claims), this split is performed at top level, and there will be two separate claims to be proven. This works well if the proof does not require circularity, but otherwise the claim will become unprovable.

I believe that the claim should not be split top-level, but that the splitting should happen under the umbrella of the original claim, and the only circularity applicable should be on the top-level claim.

alexd1971:
Fixes #3377 

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Exec.hs
+++ b/kore/src/Kore/Exec.hs
@@ -133,7 +133,6 @@ import Kore.Log.WarnDepthLimitExceeded (
 import Kore.Log.WarnTrivialClaim
 import Kore.ModelChecker.Bounded qualified as Bounded
 import Kore.Reachability (
-    AllClaims (AllClaims),
     AlreadyProven (AlreadyProven),
     Axioms (Axioms),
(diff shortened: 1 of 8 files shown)


alexd1971:
Performance test:
[report-alexe--tting_master.pdf](https://github.com/runtimeverification/haskell-backend/files/10206471/report-alexe--tting_master.pdf)

ana-pantilie (approved):
The last two tests report a huge improvement! Can you reproduce this to be sure?

ana-pantilie:
Actually, I now see another one is also reporting a huge improvement. Looking at the specs themselves, the ones affected seem to be the ones which have extra claims in the spec, so it seems like for the proofs which rely on the application of circularities this is also a big performance benefit!

PR #3405, opened 2022-12-09
Add `KRYPTO.sha512_256raw` hook

### LINKED ISSUE #3403: Support `Sha512_256raw` from `krypto.md`
[avm-semantics](https://github.com/runtimeverification/avm-semantics/) needs to evaluate `Sha512_256raw` on concrete data. I see that the hash functions implemented in `Kore.Builtin.Krypto.hs` return the output in `base16`. We, however, need the raw bytes.

goodlyrottenapple:
Fixes #3403

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -887,6 +887,15 @@ Compute the SHA256 hash of the input string.
         [hook{}("KRYPTO.sha256")]
 ~~~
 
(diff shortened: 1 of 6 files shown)


ana-pantilie (approved):
Thanks for taking care of this!

PR #3441, opened 2023-01-12
Removing forced nix env setup in workflow

F-WRunTime:
Fixing Nix env setup by workflow and expect env of runner to container nix before running test.


jberthold (approved):
:+1: :fire:

Issue #3417, opened 2022-12-15
Warn the user if a rewrite step takes longer than average


ana-pantilie:
This warning should contain as much information as possible about the state the backend was in and what it used to take the specific step. We need to figure out how to measure this inside the execution monad and make sure that we have enough context to show to the user. **High priority**

jberthold:
Ideas:
* Easiest solution would be a user-defined threshold duration, and measuring how long each step takes (the operation runs in `IO` anyway). A warning can be emitted if the measured duration is longer than the threshold.
* More complicated, compute the average of "rewrite step" executions so far and report outliers.
* The total duration of the rewrite step could be broken down into different sub-steps if the total exceeds the threshold.
* Additional information should be added to the warning to hint at what exactly is slowing things down (the configuration should be included, possibly at a different log level or into a dedicated file -> see rewrite logging into a yaml file)

ana-pantilie:
First prototype: in the kore-repl, when stepping (with `step n` or `stepf n`) if any rewrite step takes longer than a user-provided threshold (both as CL option and as repl command), stop the execution at that particular step and warn the user.

Next, as @Sereja313 suggested we could try to measure and dump rewrite step lengths for kore-exec and run this on the KEVM benchmarking suite we have. This could give us some statistics on how long rewrite steps usually take.

ana-pantilie:
@Sereja313 One feature which was just requested by our users is adding a repl command/option which switches on printing the timing information for each step. Could you look into adding this as well?

goodlyrottenapple:
we should use moving/rolling average for the threshold. this will mean adding a state/mvar to hold the previous value.

ana-pantilie:
Fixed by #3471

PR #3442, opened 2023-01-13
Add step timeout for repl

Sereja313:
Work on #3417 

### Estimate:

---

###### Review checklist

### DIFF
--- a/kore/src/Kore/Reachability/Prove.hs
+++ b/kore/src/Kore/Reachability/Prove.hs
@@ -14,11 +14,14 @@ module Kore.Reachability.Prove (
     Axioms (..),
     ToProve (..),
     AlreadyProven (..),
+    StepTimeout (..),
     proveClaims,
     proveClaimStep,
     lhsClaimStateTransformer,
 ) where
 
+import Control.Concurrent.Async.Lifted (race)
+import Control.Concurrent (threadDelay)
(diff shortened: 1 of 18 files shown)


kore/src/Kore/Reachability/Prove.hs
```diff
@@ -479,6 +490,16 @@ proveClaimStep _ stuckCheck claims axioms executionGraph node =
         | otherwise =
             transitionRule' stuckCheck claims axioms prim state
 
+    withTimeout execStep = case timeout of
+        Nothing -> Just <$> execStep
+        Just (StepTimeout st) -> do
+            let warnThread = liftIO . threadDelay $ st * 1000000
+            race warnThread execStep >>= \case
```
ana-pantilie, line 497: Hmm, I think it's nice to have this behavior: if a step takes way too long it doesn't make sense for the user to wait until it actually finishes, because it might even be looping.

The only issue I can see here is with the `Simplifier` state, in order for this to be 100% correct we would also need to rollback any changes made to the state layer. I don't think this happens automatically, does it?

Now, that layer is a cache so maybe it's actually not problematic, but if we end up modifying that state in the future to keep other things we might run into very subtle bugs here.

Can you look into ^ in a separate PR?

ana-pantilie (approved):

PR #3461, opened 2023-01-25
Switch to GHC 9.2.5

treeowl:
Fixes #nnnn

### Estimate:

---

###### Review checklist


treeowl:
@goodlyrottenapple Could you possibly help me update the nix stuff? I'm not sure what the conventions are here when going from two supported versions to one.

stack.yaml
```diff
@@ -47,42 +47,23 @@ packages:
...
-    commit: 4ab25e717658fa833aaf5efd76db38ccb1178f96
-    subdirs:
+      - pipes-aeson
       - pipes-ghc-events
-    # nix-sha256: 05mpk1r04zsaiwzksbn0yd3s6fyah739rp6s4ir5nbb17b7j59pi
-  - git: https://github.com/ttuegel/pipes-sqlite-simple.git
-    commit: a3920f053f92e76f2a7ff7c68e34f9e926817702
-    # nix-sha256: 13iag206szfmc8ng77j6qjb6r91ysy9qkvijmf9n9dclvmb1g8an
-  - ghc-events-0.13.0
+      - pipes-sqlite-simple
```
goodlyrottenapple, line 62: perhaps it would be best to fold these into this repo (maybe a separate PR after this). I had to fork because of a nix issue in the original repo. I opened a PR but never had a response. @ana-pantilie what do you think?

ana-pantilie:
You need to do `kup install k`, `stack install` the project and then `make golden` in `haskell-backend/test/`

treeowl:
@ana-pantilie That doesn't seem to work. There's no golden target in the `test/rpc-server` makefile. I tried running

```
PATH=/home/dfeuer/src/haskell-backend/.build/kore/bin:$PATH CREATE_MISSING_GOLDEN=true python3 rpc-server/runTests.py
```

I got an error

```
Traceback (most recent call last):
  File "/home/dfeuer/src/haskell-backend/test/rpc-server/runTests.py", line 2, in <module>
    from jsonrpcclient import request, request_uuid, notification
ModuleNotFoundError: No module named 'jsonrpcclient'
```

Any ideas?

treeowl:
[whatsoever.tar.gz](https://github.com/runtimeverification/haskell-backend/files/10552659/whatsoever.tar.gz)

treeowl:
[whatsoever.tar.gz](https://github.com/runtimeverification/haskell-backend/files/10552659/whatsoever.tar.gz)

ana-pantilie:
@treeowl it looks like you deleted `test/regression-c/test-smoke100.sh.out.golden`

treeowl:
> @treeowl it looks like you deleted `test/regression-c/test-smoke100.sh.out.golden`

Oops. Yeah, I missed the fact there were two in that directory. Hopefully it's all fixed now.

README.md
```diff
@@ -216,18 +216,18 @@ or
...
 
-Then, use stack to build against `stack-nix-ghc9.yaml`:
+Then, use stack to build:
 
 ```
-stack --stack-yaml stack-nix-ghc9.yaml build
+stack build
 ```
 
 If you modified the `kore.cabal` file and want to build with GHC 9, you will have to run
```
jberthold, line 233: I think this line and the lines below are obsolete now

default.nix
```diff
@@ -81,9 +81,9 @@ in let
     };
 
     # Change the compiler when updating our own resolver.
-    compiler-nix-name = "ghc8107";
+    compiler-nix-name = "ghc925";
     ghcVersion = "8.10.7";
```
jberthold, line 85: This seems wrong... Or maybe this line does not do anything, actually???

kore/test/Test/Kore/Builtin/Map.hs
```diff
@@ -1582,7 +1582,6 @@ distinctVariables variables =
 -- | Utility function for mapping over the keys of a 'HashMap'.
 mapKeys ::
     Hashable k2 =>
-    Eq k2 =>
     (k1 -> k2) ->
     HashMap k1 v ->
     HashMap k2 v
```
jberthold, line 1587: Just stumbled over this... (and likewise [`mapKeysAndValues` earlier](https://github.com/runtimeverification/haskell-backend/pull/3461/files#diff-030692d92b2a70d51d900894835f4799ecfd91017914f5bacf8067cd18f4d838L445)). Please create a ticket to investigate how we can get rid of it.
`unordered-containers` has its own [`mapKeys` function, with a big warning sign on it for non-injective key mappings](https://hackage.haskell.org/package/unordered-containers-0.2.19.1/docs/Data-HashMap-Internal.html#v:mapKeys). Every call site for this function should ensure the mapping is injective.

kore/test/Test/Kore/Rewrite/MockSymbols.hs
```diff
@@ -46,6 +46,7 @@ import Kore.Attribute.Pattern.ConstructorLike (
     isConstructorLike,
  )
 import Kore.Attribute.Sort qualified as Attribute
+import Kore.Attribute.Sort qualified as AttributeSort
```
jberthold, line 49: Curious... why do we keep the line above importing as `Attribute`? Maybe we could drop that one?

jberthold (commented):
LGTM overall - small things and a question.
The big changes in the C semantics output expectation are due to the `hashable` upgrade.

kore/test/Test/Kore/Rewrite/MockSymbols.hs
treeowl: This whole thing is a mess. I don't know the whole story, but apparently `DuplicateRecordFields` is changing (or going away?) and we have to deal with that. I have never really used that extension, and have no idea what the right approach may be. I can try to find out what's changing. As for that specific thing, yes, I can just use one, but IIRC a lot more lines will have to change.

jberthold: OK, so better leave it as-is then.
I never mastered `DuplicateRecordFields`  and its extension friends for better records - but we use `OverloadedRecordDot`  in another project, which IMHO is real progress compared to before (and I was working in the shop that was driving that proposal so I am very used to writing code with that extension).

treeowl: OK, so it's changing, not going away, but losing its type-directed disambiguation. I have no idea why the access functions for it weren't just changed to behave like `OverloadedRecordDot`, but whatever. In the access cases, we *could* switch to dots, but that won't do anything for the update cases. Leaving it as is seems okay; if someone wants to do something more stylish later that's fine too.

kore/test/Test/Kore/Builtin/Map.hs
treeowl: I can open a ticket, but figuring out what's actually getting mapped where is going to be pretty hairy, going through all sorts of things like `Injection` instances. `mapKeysAndValues` seems like a higher priority, since it's in library code rather than test code. It's not exactly obvious that the way it's used is safe in the face of the renaming that happens within `mapVariables`. Maybe?

treeowl: Ticket #3476 opened.

README.md
treeowl: How should I check?

default.nix
treeowl: I don't know the Nix stuff, so I'm reluctant to do anything to it that might break something. @goodlyrottenapple , `git grep 8\\.10\\.7` references to ghc 8.10.7 remaining in `default.nix`, `nix/fourmolu.nix.d/default.nix`, and `nix/hlint.nix.d/default.nix`. What if anything should be done with them?

Should the reference in `README.md` (for building on ARM64 Darwin) simply be changed to say `ghc-9.2.5`, or is there something more that needs to be done there? I don't have access to such a system to experiment.

treeowl: @goodlyrottenapple `flake.nix` additionally includes `compiler-nix-name ? "ghc8107", profiling ? false`.

treeowl: Right, but how? I don't know which sections need to be changed and which need to be deleted.

README.md
jberthold: No check required IMHO. This sentence talks about "want[ing] to build with GHC 9" which we now do by default with this PR. No special action is required any more, the default `nix run .#update-cabal` should do the job IIUC

default.nix
jberthold: Maybe we just change all of them for now, that seems safe to do. We simply don't want any ghc-8 in our setup any more.

ana-pantilie (approved):

PR #3465, opened 2023-01-26
[#3410] Validate debug options

### LINKED ISSUE #3410: Validate user options
Suggested by @radumereuta
Radu: This started from an error Alexey ran into: https://runtimeverification.slack.com/archives/C03ER623BS4/p1671002632034109
By using `KORE_EXEC_OPTS='--debug-equation ToDebug'` he didn't get the expected behavior.
It turned out that labels needed the module name prefixed.

The haskell tools should validate all the input given by the user and fail with a descriptive error message.

It saves users a lot of time in the long run and it boosts confidence in the tool.

alexd1971:
Problem: When we debug equational/rewrite rules with options `--debug-equation`, `--debug-attempt-equation`, `--debug-apply-equation`, `--debug-rewrite`, `--debug-attempt-rewrite`, `--debug-apply-rewrite`, if rule labels set with these options does not exists in definition we don't want these labels to be silently ignored. `Kore-exec` should throw error pointing the labels which was not found in the definition.

Solution: After initialization check if there exist labels set by command line options in the definition. And if any of them are not defined in the definition throw the error containing comprehensive description.

Fixes #3410 

### Scope:
I've checked all existing options. And only those mentioned in this PR require validation. Other options either already have implemented validation or don't require additional validation besides their type. 

### Estimate:

---

###### Review checklist


alexd1971:
You are right. We can use labels in two ways. First:
```
rule [LabelName]: ... => ...
```
In this case we must use dot-prefixed module name when set `--debug-*` options
Second, as you mentioned:
```
rule ... => ... [label(LabelName)]
```
In this case we must not use dot-prefixed module name.
And if we have both label declaration for a rule the first one has more priority.
Could you explain why we need this two ways of label declaration? I actually don't understand that. Maybe it would be more practical to leave only one of them?

If we assume that that both cases should work, it indeed makes my solution wrong and doesn't allow to validate somehow options values while parsing command line options at `kore-exec` start. And then what about the descriptions of `--debug-*` options? They contain the following:

> The name of a rule is given with the K module name as a dot-separated prefix: 'MODULE-NAME.rule-name'.

The descriptions became wrong as well, and probably should be changed.

IMO validating command line options after parsing definition is not very good idea. We could just warn a user that the definition does not contain labels which are set by command line options. But throw error in that case I guess is not good. What do you think?

If you agree with that we could generate warning log message while attempting/applying equational/rewrite rules. It seems for me simpler than generate registry of all labels after internalizing definition. As far as I understand to create the label registry I need  additionally go through `TermLike` structure.

ana-pantilie:
@alexd1971 the problem is that it won't be enough to validate that the semantics doesn't actually contain a rule with that specific label. So ok, we can keep some validation at parse-time but we still have to do a lot more validation after internalizing the definition.

ana-pantilie (approved):

PR #3473, opened 2023-01-31
Update README section

goodlyrottenapple:


ana-pantilie:
TODO: saving the substituters to the conf file will overwrite it I think, so after that is set up then add the experimental features part

PR #3497, opened 2023-02-14
Fix check-implication output 3

### LINKED ISSUE #3484: What to return if implication is trivially satisfied?
Attached is a bug report: [bug_report.tar.gz](https://github.com/runtimeverification/haskell-backend/files/10582307/bug_report.tar.gz). Where I'm adding tests for this in pyk: https://github.com/runtimeverification/pyk/pull/184

I have a simple IMP definition, and I am feeding the following term into the Haskell backend simplify endpoint:

```
CT1 = <k> int $n ; $n = X ; </k> <state> .Map </state> #And X <Int 3 #And 3 <Int X
CT2 = <k> int $n ; $n = Y ; </k> <state> .Map </state>
```

The first RPC call in the bug report is `simplify: CT1`, which returns basically the same thing as `CT1`, but I expect it to return `#Bottom`. The second call in the bug report is `implies: CT1 => CT2`, which as expected returns `#Top` (because the antecedent is bottom), with empty substitution and empty predicate. Some differences I would expect:

- The first call to `simplify` should return `#Bottom`, because the implication endpoint is able to detect that the antecedent is bottom.
- The second call to `implies` should not return `#Top` for the predicate, but `#Bottom`. It should be that if implication endpoint returns `condition { "substitution": SUBST, "predicate": PRED }`, then I can construct `SUBST(CT2) #And PRED` and have that be equal to `CT1`. But right now, `Subst(CT2) == CT2` (empty substitution), and `PRED == #Top`, so `SUBST(CT2) #And PRED == CT2`, and we _do not_ have that `CT1 == SUBST(CT2) #And PRED` (because `CT1` is bottom).

There is no way I can find for me to work around this, and it's blocking progress on pyk RPC prover. An immediate fix to the `simplifies` endpoint (so it returns `#Bottom` correctly) would fix this. So would a separate fix to the `implies` endpoint.

ana-pantilie:
Fixes #3484 

---

###### Review checklist

### DIFF
--- a/docs/2022-07-18-JSON-RPC-Server-API.md
+++ b/docs/2022-07-18-JSON-RPC-Server-API.md
@@ -280,9 +280,9 @@ If the implication cannot be shown to hold, `satisfiable` is false.
 In some cases, a unifier `condition` for the implication can still be provided, although the implication cannot be shown to be true.
 
 The endpoint implements the following specification:
-- `antecedent` is `\\bottom` => `satisfiable = True` and `condition = \\bottom`
-- `consequent` is `\\bottom` => `satisfiable = False` and `condition = \\bottom`
-- implication holds => `satisfiable = True` and `condition /= \\bottom`
+- `antecedent` is `\\bottom` (implication holds trivially) => `satisfiable = True` and `condition = \\bottom`
+- `consequent` is `\\bottom` and `antecedent` is not `\\bottom` => `satisfiable = False` and `condition = \\bottom`
+- implication holds (not trivial) => `satisfiable = True` and `condition /= \\bottom`
 - implication doesn't hold and the two terms do not unify,
 indicating that the program configuration can be rewritten further => `satisfiable = False` and `condition` is omitted
 - implication doesn't hold and the two terms unify, indicating that the configuration is stuck => `satisfiable = False` and `condition /= \\bottom`
--- a/docs/2022-07-18-JSON-RPC-Server-API.md
+++ b/docs/2022-07-18-JSON-RPC-Server-API.md
@@ -279,6 +279,14 @@ If the implication cannot be shown to hold, `satisfiable` is false.
 
 In some cases, a unifier `condition` for the implication can still be provided, although the implication cannot be shown to be true.
 
+The endpoint implements the following specification:
+- `antecedent` is `\\bottom` => `satisfiable = True` and `condition = \\bottom`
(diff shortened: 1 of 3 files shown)


jberthold (approved):
LGTM

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -279,6 +279,14 @@ If the implication cannot be shown to hold, `satisfiable` is false.
...
 In some cases, a unifier `condition` for the implication can still be provided, although the implication cannot be shown to be true.
 
+The endpoint implements the following specification:
+- `antecedent` is `\\bottom` (implication holds trivially) => `satisfiable = True` and `condition = \\bottom`
+- `consequent` is `\\bottom` and `antecedent` is not `\\bottom` => `satisfiable = False` and `condition = \\bottom`
+- implication holds (not trivial) => `satisfiable = True` and `condition /= \\bottom`
+- implication doesn't hold and the two terms do not unify,
+indicating that the program configuration can be rewritten further => `satisfiable = False` and `condition` is omitted
+- implication doesn't hold and the two terms unify, indicating that the configuration is stuck => `satisfiable = False` and `condition /= \\bottom`
+
```
ehildenb, line 289: The main property we should have hold is that `SUBST(CONSEQUENT) #And PREDICATE` (where `SUBST` is the returned substitution, and `PREDICATE` is the returned predicate) should be equivalent to `ANTECEDENT` (in the matching logic sense of them representing the same sets).

Also, we should consider this to be `valid`, not `satisfiable`, I believe? Satisfiable usually means there is _some_ substitution that makes the above property true, but we want it to be true for _any_ substitution.

If it's actually just satisfiable, then we only have partial subsumption.

Ping @ana-pantilie .

ana-pantilie: @ehildenb We can keep this in mind when designing the check-implication endpoint for the new backend. Right now this implements a workaround to make it easier for users to discern between the algorithm's output. The algorithm wasn't designed with returning the unification condition in mind.

@jberthold: `satisfiable` might not be 100% correct, we can probably change that, it's just the name of the field
@goodlyrottenapple: `check-implication` is a bad name for this procedure (`check-subsumption` would be better)

ehildenb: Makes sense.

PR #3498, opened 2023-02-14
[#3462] Keep original names in SMT transcript

### LINKED ISSUE #3462: SMT: keep names of variables

alexd1971:
Problem: Variable names and some predicate names are translated as `<n>` where n is a natural number. We need to keep original names in comments.

Solution: Put comment like `origName -> <n>` just before declaration.

Fixes #3462


ana-pantilie (approved):
Thank you!

PR #3509, opened 2023-02-20
Do not return Unknown when solver is inexistent

### LINKED ISSUE #3506: krun --smt none makes all smt queries crash the execution
I'll add an example I can guarantee will reproduce the error when I get home, but I think you can probably reproduce this with any execution of kore-exec which triggers the smt solver to try to decide a predicate.

To summarize, if you pass `--smt none` to krun when using the haskell backend, and it tries to decide a predicate using an smt solver, it appears to currently trigger a "failed to decide predicate" error, when in fact what we'd really like is if it branches on the predicate.

ana-pantilie:
Fixes #3506

This just makes `SMT.check` partial, before my changes it would default to returning `Unknown` if there wasn't any solver.

I've also been thinking about whether the `SMT` type is correct. Now we have `SMT a ~ Maybe Solver -> LoggerT IO a`. I'm wondering if the SMT action itself makes sense when there's no solver, basically we might actually want something like `SMT a ~ Maybe (Solver -> LoggerT IO a)`. @goodlyrottenapple and @jberthold, any thoughts?

--

###### Review checklist


jberthold (commented):
Code LGTM. 
Unfortunately we have an [integration test that tests the old behaviour](https://github.com/runtimeverification/haskell-backend/tree/master/test/warn-decide-predicate-unknown) - I guess that test needs to just go away unless we find a way to make the SMT solver fail to decide a predicate.

ana-pantilie:
@jberthold thanks, I didn't notice the error locally (need to fix that on my end...). I just regenerated the test output, I believe it's ok to just check if the proof gets stuck in the same place.

goodlyrottenapple (approved):

Issue #3475, opened 2023-02-01
Align RPC API description and implementations between the two Haskell backends


ana-pantilie:
Currently, the old backend has a new field in the request: `timeout`, and the new backend has a new field in the response: `aborted`. There is also the question of predicates vs substitutions in the response.

PR #3511, opened 2023-02-21
3475 align rpc protocol

jberthold:
Pert of  #3475 

Most of this code was already present, but resided in `Kore.JsonRpc`. The separate module makes it easier to re-use the code elsewhere.

New: `stop-reason: aborted` (never returned by code in this repository)

Also added some additional `FromJSON` and `ToJSON` instances

### DIFF
--- a/kore/src/Kore/JsonRpc/Base.hs
+++ b/kore/src/Kore/JsonRpc/Base.hs
@@ -1,0 +1,237 @@
+{- |
+Copyright   : (c) Runtime Verification, 2022
+License     : BSD-3-Clause
+-}
+module Kore.JsonRpc.Base (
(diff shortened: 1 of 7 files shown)


goodlyrottenapple (approved):
LGTM

kore/src/Kore/JsonRpc/Base.hs
```diff
@@ -0,0 +1,237 @@
+{- |
+Copyright   : (c) Runtime Verification, 2022
```
ana-pantilie, line 2:
```suggestion
Copyright   : (c) Runtime Verification, 2023
```

ana-pantilie (approved):

PR #3517, opened 2023-02-23
Put JSON-RPC types in a separate package

goodlyrottenapple:
Please do not merge yet as we need to consolidate the definitions across the old  and new backend.


kore-rpc-types/src/Kore/Syntax/Json/Types.hs
```diff
@@ -0,0 +1,344 @@
...
+        KJMu{var = Id name} ->
+            reportErrors name "fixpoint expression variable" checkSVarName
+        KJNu{var = Id name} ->
+            reportErrors name "fixpoint expression variable" checkSVarName
+        -- KJDV{value = txt} ->
+        --     reportErrors txt "domain value string" checkStringChars
+        -- KJString txt ->
+        --     reportErrors txt "string literal" checkStringChars
+        -- Input supports std Unicode (as per json spec). toJSON could
+        -- check that only allowed escape sequences will be generated.
```
goodlyrottenapple, line 265: not sure what we want to do here as this is the point the old and new back-end differ... @ana-pantilie @jberthold?

kore-rpc-types/kore-rpc-types.cabal
```diff
@@ -0,0 +1,127 @@
...
+
+common haskell
+  -- Common Haskell dialect
+  default-extensions:
+    BangPatterns
+    -- GHC2021 prep
+    BinaryLiterals
+    -- end
+    ConstraintKinds
+    DataKinds
```
jberthold, line 31: Should we clean up the cabal file before merging? Especially the extension list with the `GHC2021 prep` notes 
(we might also make a pass over the "main" cabal file and remove stuff that does not belong)

kore-rpc-types/src/Kore/Syntax/Json/Types.hs
jberthold: I would say restrict to `latin1` as we are doing in the new back-end. 
We should probably investigate the old code path to find out why the byte arrays aren't a problem here but we need to use the version with restrictions in the new back-end until then.

kore-rpc-types/kore-rpc-types.cabal
goodlyrottenapple: good point. i will focus on this list in this PR

jberthold (approved):
LGTM

PR #3522, opened 2023-02-27
kore-rpc Catch any runtime exception and send as error

### LINKED ISSUE #3520: `kore-rpc` does not send response on solver crash
## Reproduction

Kompile definition:

[definition.txt](https://github.com/runtimeverification/haskell-backend/files/10831782/definition.txt)

The definition can be produced by cloning https://github.com/runtimeverification/evm-semantics/commit/c9b14f64be620c2ce7cb12b466ba972ac822f990, then building the Foundry test definition:

```
$ make tests/foundry/out/kompiled/timestamp
```

Start `kore-rpc`:

```
$ kore-rpc tests/foundry/out/kompiled/definition.kore --module FOUNDRY-MAIN --server-port 3000 --smt-timeout 1 --smt-retry-limit 1
```

Send request:

[request.txt](https://github.com/runtimeverification/haskell-backend/files/10831187/request.txt)

```
$ netcat localhost 3000 < request.txt
```

Server log:

```
kore-rpc: user error (Unexpected result from the SMT solver:
  Command: (push 1 )
  Expected: success
  Result: (error "line 2038 column 8: canceled" )
)
```

## Actual behavior

`kore-rpc` does not send response, which makes the client hang indefinitely.

## Expected behavior

`kore-rpc` should send error response with an application defined error code.

goodlyrottenapple:
Fixes #3520 

image

### DIFF
--- a/kore/src/Kore/JsonRpc.hs
+++ b/kore/src/Kore/JsonRpc.hs
@@ -11,7 +11,7 @@ module Kore.JsonRpc (
 import Control.Concurrent (forkIO, throwTo)
 import Control.Concurrent.MVar qualified as MVar
 import Control.Concurrent.STM.TChan (newTChan, readTChan, writeTChan)
-import Control.Exception (SomeException, ErrorCall (..), mask)
+import Control.Exception (ErrorCall (..), SomeException, mask)
 import Control.Monad (forever)
 import Control.Monad.Catch (MonadCatch, catch, handle)
 import Control.Monad.Except (runExceptT)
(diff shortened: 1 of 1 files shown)


kore/src/Kore/JsonRpc.hs
```diff
@@ -543,10 +544,11 @@ srv serverState moduleName Log.LoggerEnv{logAction} runSMT = do
 
     cancelError = ErrorObj "Request cancelled" (-32000) Null
 
-    bracketOnReqException before onCancel thing =
+    bracketOnReqException before onError thing =
         mask $ \restore -> do
             a <- before
-            restore (thing a) `catch` \(_ :: ReqException) -> onCancel a
+            (restore (thing a) `catch` \(_ :: ReqException) -> onError cancelError a)
+                `catch` \(err :: SomeException) -> print err >> onError (serverError "crashed" $ toJSON $ show err) a
```
jberthold, line 551: If we catch all exceptions here we won't need the `withErrHandler` inside `respond` any more.

jberthold (approved):
We can merge this and follow-up with more changes to un-block people.

PR #3525, opened 2023-02-28
InfoAttemptUnification to DebugAttemptUnification

goodlyrottenapple:
* Renamed `InfoAttemptUnification` to `DebugAttemptUnification`, changing from Info to Debug level. 
* Added a way of logging to an arbitrary `LogAction`. This allows us to pipe the logger into a different one, e.g. monad-logger and is useful when using the backend as a library.
* Enabled HLS in the nix dev shell


jberthold (approved):
LGTM

PR #3527, opened 2023-03-01
[#3523] Put unserialized definition.kore in bug report

### LINKED ISSUE #3523: kore-exec --bug-report doesn't work as expected
Right now if you run `kompile` followed by `krun --bug-report`, you will end up with a bug report that does not actually contain definition.kore, but instead contains the serialized data used to initialize the haskell backend. This should really be corrected so that if the user passes `--bug-report` to a `kore-exec` invocation that does not have access to the original `definition.kore`, it either reports an error, or simply automatically looks for the correct file in the same directory as the serialized data, and reports an error if it can't find it. This way you won't end up in situations where users will report bugs using the --bug-report tool and they don't contain the data you need to actually debug the isssue.

alexd1971:
Problem: `kore-exec` uses serialized definition to improve performance. But adding serialized definition into bug report archive doesn't allow to run `kore-exec.sh` without recompiling definition with current `kore-exec`. The bug report dosen't contain original definition to be compiled. So we need unserialized kore-definition to be put into bug report.

Solution: Copy unserialized `definition.kore` into temporary bug report directory instead of serialized one.

Fixes #3523


ana-pantilie (commented):
Are you sure this works? As far as I know, the `-kompiled` directory generated by the frontend doesn't even contain the `definition.kore` if it was serialized. Can you please check if this is the case or not?

alexd1971:
Let me double check

alexd1971:
Yes, I removed `-kompiled` directory and compiled definition again. Both `definition.kore` and `haskellDefinition.bin` are there

ana-pantilie (approved):
Sounds good, thanks for checking!

PR #3529, opened 2023-03-02
rpc server refactor

goodlyrottenapple:


kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -231,3 +231,10 @@ rpcJsonConfig =
                 , "moving-average-step-timeout"
                 ]
         }
+
+class FromRequest q => FromRequestCancellable q where
+    isCancel :: q -> Bool
+
+instance FromRequestCancellable (API 'Req) where
+    isCancel Cancel = True
+    isCancel _ = False
```
jberthold, line 240: Should we add an `id` field to `Cancel` while we are at it?

kore-rpc-types/src/Kore/Network/JsonRpc.hs
```diff
@@ -0,0 +1,175 @@
+{-# LANGUAGE TypeApplications #-}
+
+{- |
+Module      : Kore.Network.JSONRPC
+Description : JSON RPC server for kore-rpc
+Copyright   : (c) Runtime Verification, 2022
+License     : BSD-3-Clause
+-}
+module Kore.Network.JsonRpc (jsonRpcServer, JsonRpcHandler (..)) where
```
jberthold, line 9: Maybe we could use a more intuitive name?
And we could re-export types that everyone will need from `Data.Network.JsonRpc`...
```suggestion
module Kore.JsonRpc.Server (
    ErrorObj (..),
    Request (..),
    Respond,
    jsonRpcServer,
    JsonRpcHandler (..),
) where
```

kore-rpc-types/src/Kore/Network/JsonRpc.hs
```diff
@@ -0,0 +1,175 @@
...
+
+{- | Create JSON-RPC session around conduits from transport layer.  When
+ context exits session disappears.
+-}
+runJSONRPCT ::
+    (MonadLoggerIO m, MonadUnliftIO m) =>
+    -- | aeson-pretty Config
+    Json.Config ->
+    -- | JSON-RPC version
+    Ver ->
```
jberthold, line 45: Should we maybe just hard-wire the version to `V2`?
Only `initSession` and `decodeConduit` are using it, and  AFAICT using a different version (`V1`) would probably require a different request type anyway.

jberthold (commented):
Looks very good overall!
I left some suggestions to adjust details.

kore-rpc-types/src/Kore/Network/JsonRpc.hs
```diff
@@ -0,0 +1,175 @@
...
+            let loop =
+                    receiveBatchRequest >>= \case
+                        Nothing -> do
+                            return ()
+                        Just (SingleRequest req) | Right True <- isCancel @q <$> fromRequest req -> do
+                            Log.logInfoN "Cancel Request"
+                            liftIO $ throwTo tid CancelRequest
+                            loop
+                        Just req -> do
+                            Log.logInfoN $ Text.pack (show req)
```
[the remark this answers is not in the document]
goodlyrottenapple: went with the suggested method + id and added a debug log with the full dump of the request

kore/src/Kore/JsonRpc.hs
```diff
@@ -465,14 +446,17 @@ runServer ::
     ) ->
     Log.LoggerEnv IO ->
     IO ()
-runServer port serverState mainModule runSMT loggerEnv@Log.LoggerEnv{logAction} = do
+runServer port serverState mainModule runSMT Log.LoggerEnv{logAction} = do
     flip runLoggingT logFun $
-        jsonrpcTCPServer
+        jsonRpcServer
             Json.defConfig{confCompare}
```
jberthold, line 452: Could this use `jsonRpcConfig`?

jberthold (approved):
Looking great now!
One remaining suggestion (using `rpcJsonConfig`), otherwise ready to go.

PR #3534, opened 2023-03-07
[#3493] Replace `functional` with `total` in docs

### LINKED ISSUE #3493: Rename functional into total
The frontend currently transforms K's `total` attribute into Kore's `functional` attribute. We should also look into renaming our internal attribute to `Total` and start parsing `total` instead of `functional` (will need to be accompanied with a change to the frontend).

alexd1971:
Problem: After removing deprecated attribute `functional` in favour of `total` documentation needs changes.

Solution: Make changes of `functional` attribute with `total` in documentation.

Fixes #3493


alexd1971:
@ana-pantilie , I made changes also in `.tex` files. So if those changes are correct, I guess we need regenerate `.pdf` files as well

ana-pantilie (approved):
Thank you!

ana-pantilie:
> I made changes also in .tex files. So if those changes are correct, I guess we need regenerate .pdf files as well

I don't know if people actually use those. If anyone complains, we can try to regenerate them then. Until then, let's merge this.

PR #3535, opened 2023-03-07
Fix docs for the `add-module` RPC endpoint

tothtamas28:


docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -334,13 +334,13 @@ Same as for execute
...
-    "module": "<plain text>"
+    "module": "[] module MODULE-NAME endmodule []"
   }
 }
 ```
 
-* `name` is the name of the main module.
-* `module` is module represented in KORE.
+* `name` is the name of the module to add.
+* `module` is a definition represented in textual KORE format. The definition should contain a module named `name`.
```
jberthold, line 343:
```suggestion
* `module` is a definition represented in textual KORE format. The definition should contain a single module named `name` (which may import modules that have already been loaded earlier).
```

jberthold (dismissed):
Thanks for the suggestion, good idea!
Maybe we should make it more precise that the _single_ module in the _partial_ definition may import modules that were already present in the server state before?

docs/2022-07-18-JSON-RPC-Server-API.md
tothtamas28: I tried a few inputs, and all of the following requests are accepted by the server (started on `--module IMP`).

### Definition with a single module

```json
{"name": "MODULE", "module": "[] module MODULE endmodule []"}
```

### Definition with multiple modules

```json
{"name": "MODULE1", "module": "[] module MODULE1 endmodule [] module MODULE2 endmodule []"}
```

### Module importing the main module

```json
{"name": "MODULE", "module": "[] module MODULE import IMP [] endmodule []"}
```

### Module importing an accompanying module
```json
{"name": "MODULE1", "module": "[] module MODULE1 import MODULE2 [] endmodule [] module MODULE2 endmodule []"}
```

### Module importing an added module

```json
{"name": "MODULE1", "module": "[] module MODULE1 endmodule []"}
{"name": "MODULE2", "module": "[] module MODULE2 import MODULE1 [] endmodule []"}
```

### Module importing an unadded module

```json
{"name": "MODULE1", "module": "[] module MODULE1 endmodule [] module MODULE2 endmodule []"}
{"name": "MODULE3", "module": "[] module MODULE3 import MODULE2 [] endmodule []"}
```
---

Based on this, I'd go with the following wording:

> The definition should contain a module named `name`, which may import other modules in scope.

jberthold: Sorry, I think this is a bit messy... 
Do you think we should allow several modules?  Then why would we require a `name` at all?
And if I add two modules `MODULE1` and `MODULE2`, but give `name: MODULE1`, am I later allowed to use `MODULE2` in a request? Again, why would we then need `name`?
(What I mean is that the `add-module` implementation in `kore-rpc` might not be the desired end state).

jberthold: The suggestion I made was informed by my attempt to implement `add-module` in the new booster, and I implemented the restrictions I mentioned (exactly one module, correct name). We can of course drop that but the `name` field is kind-of pointless if we do. Maybe we drop the `name` field instead?

tothtamas28: > the `add-module` implementation in `kore-rpc` might not be the desired end state

I agree. I think an MVP implementation would just take a single module (without wrapping it in a definition) and that's it. If that turns out to be insufficient we can still evolve the API later.

>  I implemented the restrictions I mentioned (exactly one module, correct name)

That's how I exposed the endpoint in `pyk`:

```python3
def add_module(self, module: Module) -> None:
    result = self._request('add-module', name=module.name, module=Definition((module,)).text)
    assert result == []
```

tothtamas28:
Fixed in https://github.com/runtimeverification/haskell-backend/pull/3543

PR #3537, opened 2023-03-08
Refactor RPC server error types

### LINKED ISSUE #3532: `kore-rpc` should return a more specific error code on SMT solver crash
On an SMT solver crash implementation currently returns an error response with code `-32032`, which according to the [documentation](https://github.com/runtimeverification/haskell-backend/blob/6b574836fb2af7265cab719ccdae6b960fea6f72/docs/2022-07-18-JSON-RPC-Server-API.md#-32032-internal-server-error) is a blanket error for `Internal Server Error`.

As a solver crash is an error the client has a chance of fixing (by simplifying the input), it should be given its own error code.

Moreover, this, as well as all the other errors defined by the documentation, can be considered application errors, thus should be given an error code from the corresponding range (i.e. outside of `-32768...-32000`).

goodlyrottenapple:
Fixes #3532


kore-rpc-types/src/Kore/JsonRpc/Error.hs
```diff
@@ -0,0 +1,47 @@
...
+    | CouldNotParsePattern
+    | CouldNotVerifyPattern
+    | CouldNotFindModule
+    | ImplicationCheckError
+    | SmtSolverError
+    | Aborted
+    | MultipleStates
+    deriving stock (Enum, Show)
+
+backendError :: ToJSON a => JsonRpcBackendError -> a -> ErrorObj
```
goodlyrottenapple, line 32: @tothtamas28 I've re-factored the rpc server errors into runtime errors caused by the haskell backend which start from 0, e.g. 

```json
{"code":4,"message":"Implication Check Error"}
```

vs server errors like the "Not implemented" error which is used by the booster for functionality currently not supported by it; these errors fall in the -32000 to -32099 range.


Let me know what you think

jberthold, line 32: I'd suggest to not use `0` as an error code (just by convention), by adding a `NotAnError` constructor in the beginning. 
Also, we should have a comment on this datatype to only ever add errors at the end so we don' t accidentally change existing error codes (since we use `fromEnum`).

jberthold (commented):
Integration test output needs to be adapted (once we agree on the error code range).

kore-rpc-types/src/Kore/JsonRpc/Error.hs
```diff
@@ -0,0 +1,47 @@
...
+backendError err detail = ErrorObj (toWords $ fromHumps $ show err) (fromEnum err) (toJSON detail)
+
+-- Common runtime error handlers
+
+handleErrorCall, handleSomeException :: JsonRpcHandler
+handleErrorCall = JsonRpcHandler $
+    \err@(ErrorCallWithLocation msg loc) -> do
+        logInfoN $ Text.pack $ show err
```
jberthold, line 40: Maybe better to format the log entry? 
```suggestion
        logWarnN $ Text.pack $ "Error in " <> loc <> ": " <> msg
```

kore-rpc-types/src/Kore/JsonRpc/Error.hs
tothtamas28: Base on the name, I think it is OK to handle the following as "application defined errors":

```
CouldNotParsePattern
CouldNotVerifyPattern
CouldNotFindModule
ImplicationCheckError
SmtSolverError
MultipleStates
```

so in these cases, the error code indeed should be something outside of  the range `-32768..-32000`.

I'm not sure about `Aborted` and `RuntimeError`. When are these errors raised?

> server errors like the "Not implemented" error which is used by the booster for functionality currently not supported by it; these errors fall in the -32000 to -32099 range

I agree, this can qualify as an "implementation-defined server-error". (Unfortunately, the documentation is not very detailed about these labels.)

goodlyrottenapple: `Aborted` is a bit of a blanket error when something has gone wrong sepcifically with the execution. `RuntimeError` is a blanket catch-all for any error raised in the backend (not the rpc server), this could be in execution/implication check/simplification/etc. This used to include the SMT error, and the idea is, if we find we need to pull out a runtime error/exception into a separate error-code we change that error from the blanket `RuntimeError` to a more specific one.

tothtamas28: > `Aborted` is a bit of a blanket error when something has gone wrong sepcifically with the execution

I agree this is an "application defined error" then, as I guess the client can debug the input / definition to work around the exception. Consider renaming it to `ExecutionError` or something similar.

> `RuntimeError` is a blanket catch-all for any error raised in the backend

So for example, if the server runs out of memory, it will raise `RutimeError`? Because that I'd classify as "implementation-defined server-error", as it's not the client's concern. For the same reason, I think having a single error code (from the range `-32000..-32099`) for all these conditions is fine.

goodlyrottenapple: Hmm the line is hazy on `RuntimeError` because it can include stuff like IO errors when e.g. some resource wasn't found/is broken but also includes asserts statements in the code that have been violated, e.g. the recent `Cannot convert a denormalized substitution to a map!` from the slack channel would be thrown via this error too. I would thus suggest keeping it under application errors, because these errors will be coming in specifically from the application logic portion of the code

tothtamas28: My view is that "application error" is something that is meaningful to the client, i.e. the user of the application (e.g. "the SMT solver failed on your input"), whereas a "server error" is an internal error that should be kept hidden from the client (e.g. "OOM"). This is just an educated guess on my part based on analogies though, the JSON-RPC spec (or the XML-RPC spec it is derived from) does not elaborate on these terms.

But according to this logic, an assertion failure falls into the second category, as it indicates a bug in the code, and fixing that is beyond the client's control.

Also, if `RuntimeError` potentially stands for both application errors and server errors, I think the right approach is to be more restrictive, i.e. treat them as server errors by default.

But this is just my view of the issue, and I'm not strictly opposed to alterntives.

goodlyrottenapple: Fair enough, I will move `RuntimeError` into the server error range and we can pull out any other errors into the application error scope as we see fit.

tothtamas28 (commented):
Don't forget to update `docs/2022-07-18-JSON-RPC-Server-API.md`.

jberthold (approved):
LGTM

PR #3546, opened 2023-03-14
[#3533] Refuse RPC input if symbols has macro or alias attribute

### LINKED ISSUE #3533: Refuse RPC input that uses symbols with `macro` or `alias` attribute
In https://github.com/runtimeverification/pyk/issues/248 , an input term was sent for `check-implication` which contained `#rangeUInt`, a symbol that the K definition marks as an `alias` (i.e., resolved and inlined by the front-end).
These symbols (also the ones tagged as `macro`) are present in `definition.kore` but do not have any associated rules or equations - the back-end cannot use them at all. As a result, the implication check erroneously failed.

* When internalising RPC input terms, refuse to use symbols tagged as `alias` or `macro`
* If such a symbol is found, report it in an input error to the RPC caller

alexd1971:
Problem: The symbols taged as `alias` or `macro` are present in definition.kore but do not have any associated rules or equations - the back-end cannot use them at all. So we get erroneous result.

Solution: Add parsing of `macro` and `alias'Kywd'` attributes and if during pattern verification a symbol has one of these attributes then fail.

Fixes #3533


jberthold:
Code LGTM. Would be good to add a small smoke test (maybe extend `test/rpc-server/execute`  - but this will require a definition with an alias and a macro)

jberthold (approved):

PR #3550, opened 2023-03-17
3322 - upgrade github actions

### LINKED ISSUE #3322: Upgrade github action helpers before support for node12 ends (Summer 2023)
Github actions in our workflows need to be revised to use actions in their latest version, to avoid using an outdated `node-js`. See
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/, we started getting these warnings in our github actions.

The open question is whether haskell-specific actions will be available in node16 based versions, as that infrastructure is 3rd party.

jberthold:
* replaces `::set-output` use in `update.yml` by github environment use
* Upgrades a number of github actions that would otherwise lead to warnings (`actions/checkout#v3`, `actions/cache@v3`, `haskell/actions/setup@v2`
* specifies the GHC and `stack` versions for stack build and haddock check
* rotates cache keys and updates which files are hashed for the cache key
* aligns details in `release.yml` to `test.yml` 

Fixes #3322 

### Estimate:

---

###### Review checklist


.github/workflows/release.yml
```diff
@@ -56,7 +60,7 @@ jobs:
           install_url: https://releases.nixos.org/nix/nix-2.13.3/install
           extra_nix_config: |
             access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
-            substituters = http://cache.nixos.org https://hydra.iohk.io
+            substituters = http://cache.nixos.org https://cache.iog.io
```
F-WRunTime, line 63: Worth asking the question of what should be standard here K uses the following. https://github.com/runtimeverification/k/blob/develop/.github/workflows/release.yml#L37 
A couple weeks ago anywhere I found the nix install I tried to update these to use the same references. 
Is cache.iog.io faster? Have something haskell needs? Will this be an issue in 'k' if there is a discrepancy building K?

jberthold: Good question - maybe @goodlyrottenapple can confirm .
The trusted key is for `hydra.iohk.io` but AFAIK this `iog.io` is the newer name for the same thing.

jberthold: From googling around, it seems `cache.iog.io` is indeed the better choice - [the manual](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started.html) got [updated about 11 months ago](https://github.com/input-output-hk/haskell.nix/commit/b1661f08e3d6337b11bf7152bc02e314a76cc557).

F-WRunTime (approved):

PR #3553, opened 2023-03-21
[#3538] Fix converting substitution to `Map`

### LINKED ISSUE #3538: Kore-exec fails with "Cannot convert a denormalized substitution to a map!"
Kore-exec unification fails with the message below. I have attached the automatically generated bug report. As far as I can tell:

1. The failure happens when unifying with a trusted claim's LHS.
2. The failure may be related to the claim having something like this in the configuration: `((K |-> V) M:Map) #as N:Map`.
3. To my eye, the unification is straightforward and unambiguous.

```
kore-exec: [12961366] Error (ErrorException):
    Cannot convert a denormalized substitution to a map!
      error, called at src/Kore/Internal/Substitution.hs:399:5 in kore-0.60.0.0-IjrQEodamAa3SN5XE7f3Yd:Kore.Internal.Substitution
      toMap, called at src/Kore/Unification/NewUnifier.hs:356:22 in kore-0.60.0.0-IjrQEodamAa3SN5XE7f3Yd:Kore.Unification.NewUnifier
```

Other information:
```
$ kore-exec --version
Kore version 0.60.0.0
Git:
  revision:     559424aa4841449a0e83d7c555264d68b03cf475
  branch:       HEAD
  last commit:  Fri Mar 3 05:02:43 2023 -0700
```

In kore-repl this fails when taking step 28.

This is the claim being proven: https://github.com/runtimeverification/elrond-multisig/blob/master/protocol-correctness/proof/functions/proof-perform-action-remove-user-BoardMember.k
This is the trusted claim, with the `#as` I mentioned before highlighted: https://github.com/runtimeverification/elrond-multisig/blob/master/protocol-correctness/proof/functions/proof-change-user-role-BoardMember.k#L16

[kore-exec.tar.gz](https://github.com/runtimeverification/haskell-backend/files/10925709/kore-exec.tar.gz)

alexd1971:
Problem: Kore-exec fails with "Cannot convert a denormalized substitution to a map!".

Solution: When appending one condition to another we can get condition with denormalized substitution which leads to the error. Therefore if we get condition with denormalized substitution we need to simplify this condition before converting substitution to `Map`.

Fixes #3538

### DIFF
--- a/kore/src/Kore/Unification/NewUnifier.hs
+++ b/kore/src/Kore/Unification/NewUnifier.hs
@@ -150,6 +150,7 @@ import Kore.Rewrite.RewritingVariable (
     RewritingVariableName,
     mkConfigVariable,
  )
+import Kore.Simplify.Condition (simplifyCondition)
 import Kore.Simplify.ExpandAlias (
     UnifyExpandAlias (..),
     matchExpandAlias,
@@ -379,7 +380,7 @@ unifyTerms' ::
     Condition RewritingVariableName ->
     Map Sort [AcEquation] ->
     NewUnifier (Condition RewritingVariableName)
-unifyTerms' rootSort _ origVars _ [] bindings constraints acEquations
+unifyTerms' rootSort sideCondition origVars _ [] bindings constraints acEquations
     | Map.null acEquations = do
         let freeBindings = Map.mapMaybe fromFree bindings
             (origBindings, acVarBindings) = Map.partitionWithKey isOrigVar freeBindings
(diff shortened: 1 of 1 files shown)


alexd1971:
Performance test results: [report-master_alexe--o-map.pdf](https://github.com/runtimeverification/haskell-backend/files/11033475/report-master_alexe--o-map.pdf)

kore/src/Kore/Internal/Substitution.hs
```diff
@@ -3,7 +3,7 @@ Copyright   : (c) Runtime Verification, 2018-2021
 License     : BSD-3-Clause
 -}
 module Kore.Internal.Substitution (
-    Substitution,
+    Substitution (..),
```
ana-pantilie, line 6: Please don't export the constructors, they're not exported on purpose.

I'd suggest implementing the logic you need in a separate function in this module, and export that to be used in the desired place.

ana-pantilie (changes requested):

ana-pantilie (approved):

PR #3559, opened 2023-03-29
[#3170] Detect if equational rules are unsupported

### LINKED ISSUE #3170: Detect if equational rules are unsupported
See https://github.com/runtimeverification/k/issues/2750#issuecomment-1196618502

Right now we support equational rules where the LHS has a function(al) symbol, `\\equals` or `\\ceil` at the top. The backend should abort execution if it detects anything else in this spot and provide an appropriate error message containing the source location of the rule.

alexd1971:
Problem: Right now we support equational rules where the LHS has a function(al) symbol, \\equals or \\ceil at the top. The backend should abort execution if it detects anything else in this spot and provide an appropriate error message containing the source location of the rule.

Solution: Match LHS to supported terms. And if it don't match, throw the `UnsupportedLHS` error.

Fixes #3170

### DIFF
--- a/kore/src/Kore/Equation/Sentence.hs
+++ b/kore/src/Kore/Equation/Sentence.hs
@@ -60,6 +60,7 @@ data MatchEquationError variable
     | TotalAxiom
     | ConstructorAxiom
     | SubsortAxiom
+    | UnsupportedLHS !(TermLike variable)
     deriving stock (GHC.Generic)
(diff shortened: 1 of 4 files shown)


ana-pantilie:
Can you add a K test for this?

alexd1971:
> Can you add a K test for this?

Could you give me a clue how a failing rule should look like?

ana-pantilie (approved):

Issue #3561, opened 2023-03-30
`kore-rpc` support logging options directly


ehildenb:
- [ ] Add a new switch to the RPC requests which turns on this logging. Add 3 new optional response fields: `rewrite-rules`, `function-definition-rules`, `simplification-rules` which contain lists of strings. The strings are, for each applied rule, in this order of priority: the label/the source location/the unique id. This change needs to be done in the `kore-rpc-types` package in the Haskell backend repo.
- [ ] Collect the required data in the Haskell backend.
- [ ] Collect the required data in the booster.

----

@geo2a is working on https://github.com/runtimeverification/pyk/issues/300, which is to be able to audit lemma coverage of proofs in a semantics, to determine which simplification rules are actually needed or not.

Currently, we have to launch the RPC server on the side with the correct options, and the scrape the log entries on the side as well.

Can we make each RPC call has an extra optional argument `logs: ["LOG_ID_1", ..., "LOG_ID_N"]`, and the responses come back with the log entries for that given call? That way we can get back exactly the log entries for each call, instead of having to rescrape file logs between calls.

ehildenb:
First log entries that we care to support:

- `rules`: rule names that are used to do rewriting.
- `equations`: rule names that are used from semantics axioms.
- `simplifications`: rule names that are used from simplification axioms.

We could make it more specific too, by having a boolean flag: `log-rules`, `log-equations`, `log-simplifications`.

jberthold:
From discussion in the team:
* It would be good to pin-point exactly what information should be included in the log. That way we will have more flexibility in how we implement it (the information needs to be consistent in both `hs-booster-proxy` and old backend `kore-rpc`). We would then add this information as structured data in the rpc response types.
* Some rules do not have names, do we use the location in this case? (worse than that, some equations do not have location information)
* Experience shows that it often matters more which rules or equations did _not_ apply and why. `kore-rpc` can log this, for specific named equations. Maybe it would be even better to focus on particular _terms_, too (unclear how to specify those).

ehildenb:
The tool in question here is for auditing existing proof-suites and removing lemmas that do not do any work anymore. That usually happens because in PR #NNN, you introduce some `simplification` rules, and then in `#NNN+3`, you introduce more more general rules which make the original rules redundant. So we want a tool that tells us which rules are actually applied, for quickly pruning uneeded simplification rules.

[...]

PR #3574, opened 2023-04-25
RPC log types for the execute endpoint

goodlyrottenapple:
* Implements the log/trace message types described in https://github.com/runtimeverification/pyk/issues/393
* Adds successful simplification and rewrite logging to kore-rpc (implements/subsumes  #3561)
* Fixes https://github.com/runtimeverification/haskell-backend/issues/3575


docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -33,12 +33,16 @@ The server runs over sockets and can be interacted with by sending JSON RPC mess
...
+    "log_successful_rewrites": true,
+    "log_failed_rewrites": true,
+    "log_successful_simplifications": true,
+    "log_failed_simplifications": true
   }
 }
 ```
 
-Optional parameters: `max-depth`, `cut-point-rules`, `terminal-rules`, `moving-average-step-timeout`, `step-timeout` (timeout is in milliseconds), `module` (main module name)
+Optional parameters: `max-depth`, `cut-point-rules`, `terminal-rules`, `moving-average-step-timeout`, `step-timeout` (timeout is in milliseconds), `module` (main module name) and all the `log_*` options, which default to false if unspecified.
```
ana-pantilie, line 45: From @jberthold : please use hyphen instead of underscore.

ana-pantilie:
@ehildenb and @tothtamas28 please review the changes to the documentation. They describe the new design we intend to implement.

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -192,10 +199,65 @@ If `"reason": "cut-point-rule"`, the `next-states` field contains the next state
...
         "substitution": {"format":"KORE", "version":1, "term":{}},
       }
-    ]
+    ],
+    "logs": []
+  }
+}
+```
+
+The `logs` field in the correct responses (if not empty) will contain objects of the following structure:
```
ehildenb, line 208: Can we make mention of any guarantee on teh order we can assume from the log entries?

ehildenb:
The format looks good to me, one ask (as a stretch goal, if it's _easy_), is to specify _where_ in the original kore term the simplification/equation was applied, as a term-index.

Maybe we could also include the _type_ of rule we are applying? For example: `equation`, `ml-simplification`, `simplification`, etc.... But if not, we can just reconstruct that client-side.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -112,6 +117,7 @@ data ExecuteResult = ExecuteResult
     , state :: ExecuteState
     , nextStates :: Maybe [ExecuteState]
     , rule :: Maybe Text
+    , logs :: [LogEntry]
```
jberthold, line 120: We could maybe make this optional for better backwards-compatibility (found it because an integration test was missing the `logs: []` ). The field would be omitted if the list is empty

kore/src/Kore/JsonRpc.hs
```diff
@@ -123,13 +127,34 @@ respond serverState moduleName runSMT =
...
+                    [ [ Simplification
+                        Nothing
+                        (Success Nothing (fromMaybe "UNKNOWN" $ getUniqueId s))
+                        KoreRpc
+                      | fromMaybe False logSuccessfulSimplifications
+                      , s <- toList simps
+                      ]
+                        ++ [Rewrite (Success Nothing (fromMaybe "UNKNOWN" $ getUniqueId r)) KoreRpc | fromMaybe False logSuccessfulRewrites]
+                    | ((r, _), simps) <- toList rules
+                    ]
```
jberthold, line 149: IIUC this will always traverse the list of rules from the result, but omits the 
Maybe better to guard with `logSuccessfulSimplifications, logSuccessFulRewrites` on the outside so we avoid the overhead when no logs at all were requested. Something like
```suggestion
            mkLogs rules
              | logSuccessfulSimplifications || logSuccessfulRewrites =
                Just . concat $
                    [ [ Simplification
                        Nothing
                        (Success Nothing (fromMaybe "UNKNOWN" $ getUniqueId s))
                        KoreRpc
                      | fromMaybe False logSuccessfulSimplifications
                      , s <- toList simps
                      ]
                        ++ [Rewrite (Success Nothing (fromMaybe "UNKNOWN" $ getUniqueId r)) KoreRpc | fromMaybe False logSuccessfulRewrites]
                    | ((r, _), simps) <- toList rules
                    ]
              | otherwise = Nothing
```
(This also changes the type to a `Maybe [LogEntry]` as suggested above)

kore/src/Kore/Exec.hs
```diff
@@ -403,8 +405,8 @@ data StopLabels = StopLabels
 data RpcExecState v = RpcExecState
     { -- | program state
       rpcProgState :: ProgramState (Pattern v)
-    , -- | rule label/id we have stopped on
-      rpcRule :: Maybe Text
+    , -- | rule label/ids we have applied so far
+      rpcRules :: Seq ((UniqueId, Label), Seq UniqueId)
```
ana-pantilie, line 409: Why not create a separate datatype for the `((),)`? I think nesting pairs quickly becomes confusing and makes code harder to read. There's also the issue that pairs are lazy by default, would this be helpful in any way? Recall that this backend uses language-strict.

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -192,10 +199,68 @@ If `"reason": "cut-point-rule"`, the `next-states` field contains the next state
...
+    "logs": []
+  }
+}
+```
+
+If any logging is enabled, the optional `logs` field will be returned containing an array of objects of the following structure:
+
+```json
+{
+  "tag": "simplification",
```
ana-pantilie, line 212: Did you end up adding a separate `function-definition` category? I can't find it in the code, any reason why this is harder than expected?

goodlyrottenapple: no, as @ehildenb  pointed out, you can recover this data in the frontend, so I thought it's best to keep this simple for now.

kore/src/Kore/Exec.hs
goodlyrottenapple: sure can change this to a new data type

docs/2022-07-18-JSON-RPC-Server-API.md
goodlyrottenapple: can you think of use-cases for this? I would like to keep it a flat list for now unless we find a compelling reason to build in such hierarchy

tothtamas28: It's just that if these ordering assumptions are part of the contract, then it's encoded in the schema, and not just in documentation.

But it's not critical, the client can also transform the data to this format if needed.

PR #3576, opened 2023-04-27
Tweak regression test extraction script, update tests

jberthold:
Tweaks to `scripts/generate-regression-tests.sh`:

* If `KORE` is not set, the top directory of `haskell-backend` will be used as the working directory A different directory may be given. The extracted tests will be copied into `$KORE/test/regression-evm`.
* `EVM_SEMANTICS` may be provided, a directory containing an existing check-out of `evm-semantics` to use for execution (will be re-built with dependencies using current `master`).
* `evm-semantics` will always be built with all dependencies (including `k` and `haskell-backend`), to avoid artefacts of the user's K installation ending up in the kore files. The `k` version will be the one indicated by `deps/k_release`, unless the user sets variable `K_VERSION` to override this.
* All source attributes in `*.kore` files are trimmed: `EVM_SEMANTICS` (or `$KORE/evm-semantics`, if no `EVM_SEMANTICS` directory was given) will be replaced by `evm-semantics`  to not contain any user-specific prefix.
* The script now supports extracting only a subset of tests, supplied as arguments (path without the prefix `tests/specs/` and without suffix `-spec.k`).

  Example:

```shell
someone@haskell-backend $ scripts/generate-regression-tests.sh erc20/ds/totalSupply functional/lemmas
```


ana-pantilie (dismissed):
The script worked right out of the box on my Linux machine, which is excellent.

I have one big concern, though, after reviewing I have a feeling the frontend version being used is the one built from KEVM's dependencies. That won't work because we will always be ahead of KEVM in regards to the K version it's using. We need the script to either use whatever frontend is on path or to pull in the version from the backend's deps file.

jberthold:
> I have one big concern, though, after reviewing I have a feeling the frontend version being used is the one built from KEVM's dependencies. That won't work because we will always be ahead of KEVM in regards to the K version it's using. We need the script to either use whatever frontend is on path or to pull in the version from the backend's deps file.

Correct, the current script builds the K dependency that KEVM specifies in the git submodule - we cannot merge this as-is then.
Just using what's on the path is not a great idea IMHO. We can supply the version specified in `deps/` by checking out the respective tag in the K submodule. This might of course create other issues (proofs not working with that version) but I will try to add that change to the script today.

jberthold:
The `deps/k_version` file (`5.6.40`) is lagging a bit behind the evm-semantics dependency (`5.6.67`) at the moment...  https://github.com/runtimeverification/haskell-backend/pull/3566 will fix that. I have changed the script to check out K from the version given in `deps/k_release` before it builds `evm-semantics` (can be overridden by setting `K_VERSION`).

jberthold:
When I use K  version 5.6.40, I am getting lots of warnings like this:
```
[Error] Compiler: Used 'comm' attribute on simplification rule but _==Int_ is
not comm.
	Source(/home/jost/work/RV/code/evm-semantics-1/.build/usr/lib/kevm/include/kframework/lemmas/evm-int-simplification.k)
	Location(46,7,46,36)
	46 |	      X ==Int ((X up/Int Y) *Int Y) => X modInt Y ==Int 0
	   .	      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
However, somehow the proofs still run to completion and kore files get extracted.
I updated the kore files to what we get when using the current value `5.6.40` from `deps/k_release`.

radumereuta:
This should be fixed with this PR: https://github.com/runtimeverification/k/pull/3342/files
K5.6.54 onwards.
Does it disappear if you update to the latest K?

jberthold:
> This should be fixed with this PR: https://github.com/runtimeverification/k/pull/3342/files K5.6.54 onwards. Does it disappear if you update to the latest K?

Yes, version 5.6.67 (submodule version of evm-semantics) works without any issues.

jberthold:
This started failing locally for me when I tried to re-run to update the kore files to the new K version. Investigating...

ana-pantilie:
Note: has to be run with `bash`, doing `$ bash ./scripts/generate-regression-tests.sh` worked for me.

ana-pantilie (approved):

PR #3580, opened 2023-05-04
kore rpc more detailed logging + simplify/implies endpoint

goodlyrottenapple:
This is a followup to the intial rpc logging PR #3574 which adds:
* simplification logging to the simplify and implies endpoints
* Now logs the original and rewritten terms for simplification traces


jberthold (approved):
LGTM

PR #3590, opened 2023-05-23
Lock down optparse-applicative to <0.17.1

goodlyrottenapple:
There seems to be a breaking change in the API for optparse-applicative v0.17 and above, which causes the master cabal cache run to fail

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
           authToken: '${{ secrets.CACHIX_PUBLIC_TOKEN }}'
 
       - name: Build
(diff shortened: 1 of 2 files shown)


jberthold:
Version 0.17.1 must be avoided, it was released between building the upgrade PR and merging it :)
https://github.com/pcapriotti/optparse-applicative/releases/tag/0.17.1
That version allows `ansi-wl-pprint-1.0` which introduces deprecation warnings. I constrained the version to `< 0.17.1` (consistent with stack resolver)

jberthold (approved):

PR #3594, opened 2023-05-25
use K version from deps/k_release for integration tests

### LINKED ISSUE #3583: Upgrade haskell-backend to ghc-9.2.7

jberthold:
Fixes #3583  (follow-up)


.github/workflows/test.yml
```diff
@@ -146,12 +146,13 @@ jobs:
 
       - name: Test paths to Haskell backend binaries
         run: |
-          nix develop github:runtimeverification/k/sam/nix-kore-integration-shell#kore-integration-tests \
+          K_VERSION=v$(cat deps/k_release)
```
goodlyrottenapple, line 149: doing `v$(cat deps/k_release)` twice. can we remove `K_VERSION=v$(cat deps/k_release)`?

jberthold: I might also remove the path checking - that was more for debugging while it did not work properly yet

jberthold:
Merging this now (overriding reviews) because we have a testing gap otherwise.

PR #3604, opened 2023-06-15
3599 prototype get-model endpoint

### LINKED ISSUE #3599: Prototype stub code for get-model end-point (incl. interface draft)
To get familiar with the SMT solver bridge, implement an end-point that sends a constrained term to the SMT solver (and logs the transcript when requested).
- [ ] Document the draft end-point interface for discussion
- [ ] Ask for concrete use case examples if possible
- [ ] Prototype SMT solver bridge extension to support the `get-model` command

jberthold:
Fixes #3599 

### Scope:

* First working version of `get-model` endpoint for preview and discussion
* Documentation and integration tests


docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -438,6 +438,52 @@ requests to the server by passing `"module": "MODULE-NAME"`.
...
+
+### Correct Response:
+
+```json
+{
+  "jsonrpc":"2.0",
+  "id":1,
+  "result":{
+    "satisfiable": "is-satisfiable",
+    "substitution": {"format": "KORE", "version": 1, "term": {...}}
```
tothtamas28, line 473: Can this field be more structured, i.e. a JSON with variable names as keys and terms as values?

Also, I think for the whole API we should drop the

```json
{"format": "KORE", "version": 1, "term": {...}}
```

structure and just use

```json
{...}
```

for terms. For a given version of the API we should never encounter a situation where one KORE term is version `1` and another term is version `2`, thus the KORE format version is implied by the API version. So even if want to be extra prudent about API versioning, we can either put the version in the `result` root once, or simply add a `version` endpoint (but I think we should be fine without either, integration tests are expected to catch these kind of bugs).

jberthold: Well... I agree it seems superfluous, but I added that version and format because it was requested when I first made the json format.
I'd rather stick with it now than change loads of code in various places to remove format and version.

jberthold: As for the substitution structure, we have the equation encoding for substitutions in other places, too, and it has the advantage that field names will always be known statically (which I prefer by a great deal). It is an easy transformation to convert to a dictionary from the equations.

tothtamas28: > I'd rather stick with it now

Sure, it's not something to address in this PR. But if you agree it's a change that would make the API better, we can track it in an issue.

> field names will always be known statically

How about this format then:

```json
[{"var:" "X", "term": {...}}, {"var": "Y", "term": {...}}]
```

(`"var"` can also be an `EVar` JSON.)

I think it's better to have this parsed format as it gives stronger static guarantees on the structure.

If you'd stick with the equation format for now for consistency, that's fine. But if this is something that you'd also consider an improvement, let's make an issue for it.

kore/src/SMT.hs
```diff
@@ -208,6 +210,11 @@ check :: MonadSMT m => m (Maybe Result)
 check = liftSMT checkSMT
 {-# INLINE check #-}
 
+-- | Check if the current set of assertions is satisfiable.
```
geo2a, line 213: I think this should read as something like "get values from the model of a satisfiable set of assertions", since it does not check satisfiability, but rather assumes it has already been checked.

test/rpc-server/get-model/satisfiable/README.md
```diff
@@ -0,0 +1,4 @@
+# `get-model` applied to satisfiable predicates
+
+* Input kore term: `X <= 10 && X < 5`
```
geo2a, line 3:
```suggestion
* Input: `X <= 10 && X < 5`
```

test/rpc-server/get-model/no-predicate/README.md
```diff
@@ -0,0 +1,8 @@
+# `get-model` applied to an input without a predicate
+
+* Input kore term: `\and(\not(X:SortBool), X:SortBool)`
+* Result: `satisfiability-unknown`
+
+The input is interpreted as a _term_, not as a predicate. Therefore,
```
geo2a, line 6: Why is input of this test is interpreted as a term (not predicate), and the input of [test/rpc-server/get-model/not-satisfiable-vars](https://github.com/runtimeverification/haskell-backend/pull/3604/files#diff-198d373c2409878031412be16e04863d5e02c8a2dea0dff09410c2759cc224e2) *is* interpreted as a predicate? Is that the `\equals` that makes the second input a predicate?

I'd actually expect a good deal of inputs to this endpoint to be conjunctions of equalities, akin to further tests. Why are these predicates and this one is not?

geo2a (changes requested):
I left a couple of questions and some minor suggestions.

test/rpc-server/get-model/no-predicate/README.md
jberthold: The fact that you can `\and` two terms obfuscates the matter here, but the input is an `\and` of two non-predicates because a variable is not a predicate (predicates being defined as ML constructs whose value is either `\top` or `\bottom`.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -157,13 +167,32 @@ data SimplifyResult = SimplifyResult
...
+data GetModelResult = GetModelResult
+    { satisfiable :: SatResult
+    , substitution :: Maybe KoreJson
+    }
+    deriving stock (Generic, Show, Eq)
+    deriving
+        (FromJSON, ToJSON)
+        via CustomJSON '[OmitNothingFields, FieldLabelModifier '[CamelToKebab]] GetModelResult
+
+data SatResult
```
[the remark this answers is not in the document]
jberthold: That's correct. 
In this case we could just use the `Result` type of the SMT bindings, except that it goes against the package dependencies, so I won't.

kore/src/Kore/JsonRpc.hs
```diff
@@ -441,6 +435,47 @@ respond serverState moduleName runSMT =
                                                     }
 
                                     pure . Right $ AddModule ()
+        GetModel GetModelRequest{state, _module} ->
+            withMainModule (coerce _module) $ \serializedModule lemmas ->
+                case verifyIn serializedModule state of
+                    Left err ->
+                        pure $ Left $ backendError CouldNotVerifyPattern err
+                    Right stateVerified -> do
```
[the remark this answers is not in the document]
jberthold: That is right - `verifyIn` does this checking internalisation of the input. `parsePatternFromTermLike` on 446 "distributes" the `\and` ed components to a "term" and a "predicate" (of `\and`ed connectives), or `\top` if there are none.

> Perhaps we should even return an error if receive a configuration? Can we do that?

That was one of the questions I asked on slack. We can do that, OTOH the configuration term does not actually contribute to the result (at least that is what I am thinking).

kore/src/Kore/Rewrite/SMT/Evaluator.hs
```diff
@@ -195,35 +196,101 @@ decidePredicate onUnknown sideCondition predicates =
...
+
+{- | Check if the given combination of predicates is satisfiable, and
+  provide a satisfying substitution if so. Otherwise indicate with a
+  boolean whether the solver was able to determine 'unsat' or returned
+  an 'unknown' result.
+
+  This means that if no SMT solver is configured, this function will
+  always return 'Left False'.
+
+  The solver query is retried with configurable scaled timeout. TODO
```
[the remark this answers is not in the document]
jberthold: Actual TODO, we are not repeating the query at the moment.

kore/src/Kore/Rewrite/SMT/Evaluator.hs
```diff
@@ -195,35 +196,101 @@ decidePredicate onUnknown sideCondition predicates =
...
+        satResult <- SMT.check
+        case satResult of
+            Nothing -> pure $ Left Unknown
+            Just Unsat -> pure $ Left Unsat
+            Just Unknown -> pure $ Left Unknown
+            Just Sat ->
+                if null vars -- no free variables, trivial case
+                    then pure $ Right Map.empty
+                    else do
+                        mbMapping <- SMT.getValue vars
```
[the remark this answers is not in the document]
jberthold: `get-value` is already available in our code so I am using that one. `get-model` is basically the same as `get-value` called with a complete list of all bindings made before.

kore/src/SMT.hs
jberthold, line 213:
```suggestion
-- | Retrieve the given variables from the model (only valid if satisfiable)
```
:spaghetti: copy pasta

jberthold:
If `sat, unsat, unknown` is consensus, I'll rename the result type (on Monday).
Now, should it indeed be an error if a non-predicate (term, config) is sent?

test/rpc-server/get-model/no-predicate/README.md
geo2a: Thanks! that makes sense now.

kore/src/Kore/JsonRpc.hs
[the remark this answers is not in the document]
geo2a: I think it is good to leave it as it is for now. Thanks!

kore/src/Kore/Rewrite/SMT/Evaluator.hs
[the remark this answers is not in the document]
geo2a: Great, I just wanted to confirm my understanding. Thanks!

geo2a:
I think this is OK to allow sending a non-predicate and only consider the predicate part. We should explicitly document that the non-predicate part of the term will be ignored.

geo2a (approved):

PR #3610, opened 2023-06-21
3577 automate test update

### LINKED ISSUE #3577: Automate regression test updates using Github Actions
- we'd like to be able to run the job at the click of a button
- would also be good to run it on a schedule
- the job should run the `scripts/generate-regression-tests.sh` (recently updated in https://github.com/runtimeverification/haskell-backend/pull/3576)
- for now, focus on KEVM

jberthold:
Fixes #3577 

* adapts the test update script to latest requirements ([also see the companion PR in evm-semantics](https://github.com/runtimeverification/evm-semantics/pull/1918))
* draft version of an automatic run - pending testing


stack.yaml
```diff
@@ -54,6 +54,9 @@ extra-deps:
   - typerep-map-0.6.0.0
   - monad-validate-1.2.0.1
 
+# TODO remove when co-log is updated
+allow-newer: true
+
```
jberthold, line 59: This was removed by #3612 but is apparently required for `co-log`  (for the `typerep-map` dependency, to be precise). The new `co-log` release should come soon, then we can delete it.

Q is, why did the CI build not pick this up in the other PR? Probably due to caching.

PR #3613, opened 2023-06-29
HOTFIX Freeze library dependencies for cabal + script to do so

jberthold:
* File `cabal.project.freeze` contains a configuration computed from constraints given by `stack --test ls dependencies`

* `scripts/freeze-cabal-to-stack-resolver.sh` automates this process

* `scripts/check-cabal-stack-sync.sh` outputs differences between what `stack` outputs and the contents of the `cabal.project.freeze` file. This might show differences but they should all be for build tools (`alex`, `hsc2hs`, `tasty-discover`)


stack.yaml
```diff
@@ -51,7 +51,7 @@ extra-deps:
   - chronos-1.1.5
   - co-log-0.5.0.0
   - decision-diagrams-0.2.0.0
-  - typerep-map-0.6.0.0
+  - typerep-map-0.5.0.0
```
jberthold, line 54: I did not see a good reason to use version 0.6 here and it created problems for cabal.
_Not_ going to delete `allow-newer` for now, though...

cabal.project.freeze
```diff
@@ -0,0 +1,290 @@
...
+             any.Only ==0.1,
+             any.QuickCheck ==2.14.2,
+             QuickCheck -old-random +templatehaskell,
+             any.StateVar ==1.2.2,
+             any.adjunctions ==4.4.2,
+             any.aeson ==2.0.3.0,
+             aeson -cffi +ordered-keymap,
+             any.aeson-pretty ==0.8.9,
+             aeson-pretty -lib-only,
+             any.alex ==3.2.7.4,
```
jberthold, line 14: This was manually down-graded to 3.2.7.4 to fix the `format` derivation build. The reason is two-fold:
1. The constraints from `cabal.project.freeze` are being picked up when building `fourmolu` 
2. constraints on tool versions (`--constraint="alex == 3.2.7.4" in this case`)  are ignored by `cabal freeze`

These is currently manual work (the script will try to bump `alex` to the latest available version). Hopefully next time we upgrade the resolver we won't have trouble with it. (just have to make sure `hackage.nix` index state is not too old for our choice of resolver).

geo2a (approved):

PR #3614, opened 2023-06-29
Types specifying the `simplify-implication` RPC endpoint

geo2a:
This PR modifies `kore-rpc-types` to include types that specify the new `simplify-implication` RPC endpoint. Based on the specification in https://github.com/runtimeverification/haskell-backend/issues/3787.

Note that the CI fails due to Z3 starvation, which is unrelated to these changes.


ehildenb:
We need a testing harness! Let's make it easy from the start to test these functionalities.

goodlyrottenapple:
@geo2a can you add a simple test to `test/rpc-server` which exercises the endpoint as @ehildenb suggests? doesn't have to simplify anything, literally just return the same pattern for now...

geo2a:
@ehildenb @goodlyrottenapple I've added a stub implementation into `kore-rpc` and an integration tests.

I've also extended the RPC API docs. @ehildenb please have a look at the shape of the responses and advise on how we should format them so `pyk` is happy. Probably the Haskell-style CamelCase tags are not OK, but I'm reluctant to invent another format before I get feedback from you.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -177,6 +178,48 @@ data GetModelResult = GetModelResult
...
+    = -- | implication is valid
+      ImplicationValid
+    | -- | implication is invalid, explains why
+      ImplicationInvalid ImplicationInvalidReason
+    | -- | implication is unknown, explains why
+      ImplicationUnknown ImplicationUnknownReason
+    deriving stock (Generic, Show, Eq)
+    deriving
+        (FromJSON, ToJSON)
+        via CustomJSON '[] ImplicationValidityResult
```
goodlyrottenapple, line 201: could use the same `via CustomJSON '[FieldLabelModifier '[CamelToKebab]]` as above. I think this is the naming convention in pyk?

goodlyrottenapple (approved):
LGTM @ehildenb @tothtamas28 please let us know if there are any tweaks to the responses you would like

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -388,6 +388,127 @@ Same as for execute
...
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "validity": {"tag": "implication-valid",
+                 "contents": {"tag": "matching-failed",
+                              "contents": "Shared variables: X:SortWordStack{}"
+                 }},
+  }
+}
+```
```
ehildenb, line 454: The result here should be "unification-failed", and that's what we should be checking. Eventually, if unification succeeds and matching fails, we want ot return an core problem which may or may not be unsat, that the user can then forward to the more powerful implies endpoint.

Just want to make sure that when we get this response back, what it actually means is _there is absolutely no overlap between the terms, we know because we found differing constructors_.

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -388,6 +388,127 @@ Same as for execute
...
+#### Implication **invalid**: terms match, but constraints subsumption failed
+
+Matching between antecedent and consequent configurations is successful, but constraints do not agree. Response contains the matching substitution and the unsatisfiable core of constraints.
+
+```
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "validity": {"tag": "implication-valid",
```
ehildenb, line 465:
```suggestion
    "validity": {"tag": "implication-invalid",
```

I think this should likely actually just be `unknown` result actually, but we return the unsat core (the constarints we arent' sure about).

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -388,6 +388,127 @@ Same as for execute
...
+  "result": {
+    "validity": {"tag": "implication-unknown",
+                 "contents": {"tag": "constraint-subsumption-unknown",
+                              "contents": {"tag": "constraint-subsumption-failed",
+                                           "contents": {"format": "KORE", "version": 1, "term": {}}
+                 }},
+    "substitution": {"format": "KORE", "version": 1, "term": {}},
+  }
+}
+```
```
ehildenb, line 510: How is this different than "Implication **invalid**: terms match, but constraints subsumption failed"

geo2a: I thought, we'd want to have a tri-state constraint subsumption check, just like we have a tri-stat matching routine. But perhaps we can indeed expose the "failed" and "unknown" constraint subsumption as just "failed" and have less user-facing cases.

ehildenb: Part of the confusion for me is that for the failure cases, you're using "terms don't `match`" as the criteria, but that's not how the tri-state matching algorithm works. It either reports back:

- Terms _match_ with substitution _alpha_.
- Terms _definitely don't unify_.
- Unknown.

The key here is that unification and matching are different algorithms, different complexities, and different properties. Saying a term matches another means it's _completely subsumed_, and saying that they _don't unify_ means they have _no overlap_.

So I guess the language here needs to be cleared up, because here for example 
image
 we're saying "terms don't match" because of differing constructor, where I think it should be saying "terms don't unify".

Unification and matching are not interchangeable. See https://github.com/runtimeverification/hs-backend-booster/issues/193 again for the desired interface.

PR #3620, opened 2023-07-12
Revert 'Eliminate top level existentials in side-condtions'

### LINKED ISSUE #3605: Possible implies endpoint not giving correct result
I have a simple configuration with just a k cell and a Map.
```
  configuration <T color="yellow">
                  <k> $PGM:Pgm </k>
                  <state> .Map </state>
                </T>
```
And am passing in something that looks like this (before being converted to kore)
```
antecedent: <generatedTop>
  <T>
    <k>
      K_CELL
    </k>
    <state>
      STATE_CELL
    </state>
  </T>
  <generatedCounter>
    GENERATEDCOUNTER_CELL
  </generatedCounter>
</generatedTop>
consequent: #Exists X . #Exists Y . <generatedTop>
  <T>
    <k>
      K_CELL
    </k>
    <state>
      STATE_CELL
    </state>
  </T>
  <generatedCounter>
    GENERATEDCOUNTER_CELL
  </generatedCounter>
</generatedTop>
#And { true #Equals 0 >Int Y }
#And { true #Equals 0 <=Int X }
```
with the goal of figuring out if `{ true #Equals 0 >Int Y } #And { true #Equals 0 <=Int X }`  is satisfiable, and the endpoint returns satisfiable=false. In #3601 it is mentioned that this satisfiable field should actually be called "valid", but I thought this implication should actually be valid because `{ true #Equals 0 >Int Y } #And { true #Equals 0 <=Int X }`  is satisfiable and the variables X and Y are existentially quantified. It's possible I'm just not using the endpoint correctly or am not understanding something matching-logic related.

[bug_report.tar.gz](https://github.com/runtimeverification/haskell-backend/files/11761058/bug_report.tar.gz)

goodlyrottenapple:
This PR fixes #3605 
The bug was introduced by  #3202, which eliminated top level existentials in side-conditions. However, this was incorrect because when translated to Z3, these variables would get a universal quantification, which is what was happening in #3605

### DIFF
--- a/kore/src/Kore/Reachability/Claim.hs
+++ b/kore/src/Kore/Reachability/Claim.hs
@@ -773,7 +773,7 @@ checkSimpleImplication inLeft inRight existentials =
         rhsBottom <-
             fmap isBottom . liftSimplifier $
                 SMT.Evaluator.filterMultiOr $srcLoc
-                    =<< Pattern.simplify right
+                    =<< Exists.makeEvaluate SideCondition.top existentials right
 
         case (trivial, rhsBottom) of
             (True, _) -> pure (claimToCheck, Implied Nothing)
--- a/kore/src/Kore/Simplify/Condition.hs
+++ b/kore/src/Kore/Simplify/Condition.hs
@@ -16,15 +16,8 @@ import Control.Monad.State.Strict (
     StateT,
  )
 import Control.Monad.State.Strict qualified as State
-import Data.Functor.Foldable qualified as Recursive
 import Data.Generics.Product (
     field,
- )
-import Data.Set (
-    Set,
- )
-import Kore.Attribute.Pattern.FreeVariables (
-    freeVariableNames,
  )
 import Kore.Internal.Condition qualified as Condition
 import Kore.Internal.Conditional qualified as Conditional
@@ -60,8 +53,6 @@ import Kore.Simplify.SubstitutionSimplifier (
     SubstitutionSimplifier (..),
  )
 import Kore.Substitute
-import Kore.Syntax.Exists qualified as Exists
-import Kore.Syntax.Variable (SomeVariableName)
 import Kore.TopBottom qualified as TopBottom
 import Logic
 import Prelude.Kore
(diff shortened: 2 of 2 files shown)


kore/src/Kore/Reachability/Claim.hs
```diff
@@ -773,7 +773,7 @@ checkSimpleImplication inLeft inRight existentials =
         rhsBottom <-
             fmap isBottom . liftSimplifier $
                 SMT.Evaluator.filterMultiOr $srcLoc
-                    =<< Pattern.simplify right
```
goodlyrottenapple, line 776: this didn't end up being the issue, but it's not correct and should be changed nonetheless.

jberthold (dismissed):
LGTM, but the RPC integration tests need to be adapted.

jberthold:
As it turns out, [one of the RPC server `implies` tests](https://github.com/runtimeverification/haskell-backend/tree/master/test/rpc-server/implies/not-implied-stuck/) is using a term that [hits an error call now](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Simplify/Exists.hs#L394-L401). 
We probably just want to use a different test but it would be interesting to understand the restriction in detail.

goodlyrottenapple:
> As it turns out, [one of the RPC server `implies` tests](https://github.com/runtimeverification/haskell-backend/tree/master/test/rpc-server/implies/not-implied-stuck/) is using a term that [hits an error call now](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Simplify/Exists.hs#L394-L401). We probably just want to use a different test but it would be interesting to understand the restriction in detail.

Does the test make sense in the first place? `X` and `Z` are not booleans but rather of sort K, so the back-end infers that `Z` is the term and `#Not (X #Equals Z)` is the predicate, hence the error.

jberthold:
> > As it turns out, [one of the RPC server `implies` tests](https://github.com/runtimeverification/haskell-backend/tree/master/test/rpc-server/implies/not-implied-stuck/) is using a term that [hits an error call now](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Simplify/Exists.hs#L394-L401). We probably just want to use a different test but it would be interesting to understand the restriction in detail.
> 
> Does the test make sense in the first place? `X` and `Z` are not booleans but rather of sort K, so the back-end infers that `Z` is the term and `#Not (X #Equals Z)` is the predicate, hence the error.

Yes, that's what happens. Just not sure why it would be such a bad thing to have an existential quantification over both term and predicate...
But I was meaning to suggest that we modify the test and move on.

jberthold:
Unfortunately, [the "error somewhere else"](https://github.com/virgil-serbanuta/kore/blob/master/kore/src/Kore/Step/Simplification/Exists.hs#L316-L321) is triggered by foundry proofs (I tested `GasTest.testInfiniteGas` which introduces an existential gas value).
[bug-report-stripped.tar.gz](https://github.com/runtimeverification/haskell-backend/files/12045266/bug-report-stripped.tar.gz) is an RPC interaction log for this proof, failing in request 6.

Issue #3624, opened 2023-07-14
backend silently crashing for some proofs


iFrostizz:
### First occurence

This issue first occured when Lucas was running some foundry proofs.
Suddenly, kevm crashed with a `JSONDecodeError` at position `0, 0`. This means that the received response from the backend on pyk is empty, so the json decoding fails because it's not valid json.

### Confirming

Then, kevm has been ran using the `--bug-report` flag, which confirmed our first guess; The response4.json is empty. (not this is not the raw bug report, it's too heavy for uploading here, please ask if needed!)

[rpc.tar.gz](https://github.com/runtimeverification/haskell-backend/files/12053994/rpc.tar.gz)

### POC

This lets us creating a short POC that replays the behavior of pyk. To use, download and extract the [requests.tar.gz](https://github.com/runtimeverification/haskell-backend/files/12053932/requests.tar.gz) and copy this script so that both are at the root of an `evm-semantics` repo. This allows us to setup a foundry env easily.

```python
import os
from pathlib import Path
from pyk.kcfg import KCFGExplore
from pyk.kore.rpc import KoreServer, JsonRpcClient
from kevm_pyk.foundry import Foundry

foundry_root = Path('tests/foundry')
foundry = Foundry(foundry_root)

rpc_dir = Path('requests')

with KCFGExplore(foundry.kevm) as kcfg_explore:
    kore_server = KoreServer(
        kcfg_explore.kprint.definition_dir,
        kcfg_explore.kprint.main_module,
        smt_timeout=125,
        smt_retry_limit=4,
    )
    sock = JsonRpcClient._create_connection('localhost', kore_server.port, None)
    file = sock.makefile('r')
    for rpc_file in sorted(os.listdir(rpc_dir)):
        if rpc_file.endswith('request.json'):
            print(f'reading {rpc_file}')
            with open(rpc_dir / rpc_file) as f:
                payload = f.read()
                print(f'sending {payload[:100]} ...')
                req = payload
                sock.sendall(req.encode())
                resp = file.readline().rstrip()
                rec = resp[:1000]
                if len(rec) == 0:
                    print('received empty response from the backend !')
                else:
                    print(f'received {rec} ...')
    kore_server.close()
    sock.close()
```

### Sanity check

Now, to make sure that none of these are actually related to pyk or kevm, we have sent the raw requests directly to the `kore-rpc` server using `nc` and observed an empty response on request 4.
```sh
# shell1
[...]

PR #3625, opened 2023-07-17
3624 omit trace collection unless required

jberthold:
Part of #3624

Omits tracing simplifications and rewrites when no trace was requested by the caller.
This fixes a memory issue for execute requests with large depth - traces would otherwise take up large amounts of heap and bring the server down.

### DIFF
--- a/kore/src/Kore/Equation/Application.hs
+++ b/kore/src/Kore/Equation/Application.hs
@@ -22,7 +22,9 @@ import Control.Error (
 import Control.Monad (
     (>=>),
  )
+import Control.Monad.Reader (asks)
 import Control.Monad.State (modify)
+import Data.Bifunctor
 import Data.Map.Strict (
     Map,
  )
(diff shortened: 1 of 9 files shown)


geo2a:
The changes look good to me. One of the property tests fails on mac12, and I have been able to reproduce it with the seed hedgehog suggested, i.e.:

```
cabal test kore-test --test-options '--pattern "$NF ~ /convert a valid pattern to a Syntax.Pattern and back/" --hedgehog-replay "Size 84 Seed 1362623070684199242 108609992799031899"'
```

I have not investigated if this failure is resulting from the changes introduced in this PR. My gut tells me that it is not, and we just caught it with random testing. 

I suggest we:
* "fix" this PR by re-running the failing job and merge it;
* track this failure in a separate issue, quoting the reproduction command.

jberthold:
> The changes look good to me. One of the property tests fails on mac12, and I have been able to reproduce it with the seed hedgehog suggested, i.e.:
> 
> ```
> cabal test kore-test --test-options '--pattern "$NF ~ /convert a valid pattern to a Syntax.Pattern and back/" --hedgehog-replay "Size 84 Seed 1362623070684199242 108609992799031899"'
> ```
> 
> I have not investigated if this failure is resulting from the changes introduced in this PR. My gut tells me that it is not, and we just caught it with random testing.

Unfortunately, this is a known issue (part of #3259) . It happens much more rarely since some improvements were made, but we sometimes hit https://github.com/runtimeverification/haskell-backend/blob/2e33122cd354045da4e5f40dea9344e02c673a64/kore/test/Test/ConsistentKore.hs#L230 while trying to generate a "consistent" term (since we don't want to try forever...)

> I suggest we:
> 
> * "fix" this PR by re-running the failing job and merge it;
> * track this failure in a separate issue, quoting the reproduction command.

If you want you can add a comment to #3259 but I guess we have higher priorities. :-|

geo2a (approved):

PR #3628, opened 2023-07-25
Fail when checking definition with Z3 returns Unknown

### LINKED ISSUE #3619: Possible to run with a potentially inconsistent SMT setup
The [initialisation code for the SMT solver](https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Rewrite/SMT/Lemma.hs#L84-L88) checks the prelude of all declarations and `smt-lemma` assertions and will refuse to work when it  is `Unsat`, but it does not prevent running with a setup that yields `Unknown`.

The `Unknown` result may come from a timeout, where it might be OK to proceed, but it can also come from the solver being (algorithmically) unable to decide even the lemmas of the prelude. This will prevent the solver from ever returning a `Sat` result for any subsequent queries, it will always be `Unknown` or `Unsat`.

At least there should be a warning emitted during the setup which points this out to the user.

Case in point: this [bug_report.tar.gz (compressed and some artefacts removed](https://github.com/runtimeverification/haskell-backend/files/11974788/bug_report.tar.gz) was provided with a report of the SMT solver failing with `ErrorDecidePredicateUnknown` for a simplification RPC request. Its SMT prelude contains inconsistent lemmas which the solver cannot decide.

goodlyrottenapple:
Fixes #3619
The definition provided in the above issue now throws the following error:

```
kore-rpc: [4462799] Error (ErrorException):
    The definitions sent to the solver may not be consistent (Z3 timed out).
      error, called at src/Kore/Rewrite/SMT/Lemma.hs:180:9 in kore-0.60.0.0-inplace:Kore.Rewrite.SMT.Lemma
```

### DIFF
--- a/kore/src/Kore/Rewrite/SMT/Lemma.hs
+++ b/kore/src/Kore/Rewrite/SMT/Lemma.hs
@@ -93,10 +93,7 @@ declareSMTLemmas tools lemmas = do
         Just Sat -> pure ()
         Just Unsat -> errorInconsistentDefinitions
         Just Unknown -> do
-            SMT.localTimeOut doubleTimeOut $ do
-                SMT.reinit
-                declareSortsSymbols $ smtData tools
-                mapM_ declareRule lemmas
+            SMT.localTimeOut doubleTimeOut $
                 SMT.check >>= \case
                     Nothing -> pure ()
                     Just Sat -> pure ()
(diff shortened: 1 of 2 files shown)


geo2a (approved):

kore/src/Kore/Rewrite/SMT/Lemma.hs
```diff
@@ -84,8 +84,11 @@ declareSMTLemmas ::
 declareSMTLemmas tools lemmas = do
     declareSortsSymbols $ smtData tools
     mapM_ declareRule lemmas
-    isUnsatisfiable <- fmap (Unsat ==) <$> SMT.check
-    when (fromMaybe False isUnsatisfiable) errorInconsistentDefinitions
+    SMT.check >>= \case
```
jberthold, line 87: CI failures... do we need to retry or bump up the timeout here?

PR #3634, opened 2023-08-01
RPC server should return stuck

goodlyrottenapple:
The server currently returns `depth-bound` as a reason, when no applicable rules were found. The reason should instead be `stuck`


geo2a (approved):

jberthold (approved):
OK as a fix, although there is probably a more subtle problem in `toTransitionResult`

Issue #3630, opened 2023-07-26
Could the backend report vacuous nodes through the `execute` response ?


iFrostizz:
This issue originates from https://github.com/runtimeverification/evm-semantics/issues/1944
We would like if possible the backend to report if the current executed node is vacuous.

A minimal example written in the Foundry integration of KEVM was tested.

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity =0.8.13;

import "forge-std/Test.sol";

contract VacuousTest is Test {
    function test_unreachable(bool flag) public {
        vm.assume(false);
        assert(false);
    }
}
```

It can be stored in `evm-semantics/tests/foundry/test`
Then 
```sh
kevm foundry-kompile
kevm foundry-prove --test VacuousTest.test_unreachable
```

It features a call to `vm.assume(false)` which stucks the execution at this line, when the K cell is roughly:

`<k> #foundry_assume ~> false </k>`

Currently, the [`next_cterms`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kcfg/explore.py#L468C1-L470) that is decoded from pyk is an empty list, so pyk deduces that the execution is stuck at this point. But we would like, if possible, the backend to be able to automatically return a new status code `vacuous` (or related) if the current node is vacuous.

I attached the return of the test ran with the kevm foundry integration and adding the `--debug` to be able to see the requests and response with the haskell backend (split in 2 to circumvent the Github file size limit). We can see that the `depth-bound` status code is returned with a first depth of 2, then of 0 in the last response that contains the `depth-bound` status code. In the two cases the response contains the `#foundry_assume false`.

[dump_part1.txt](https://github.com/runtimeverification/haskell-backend/files/12168718/dump_part1.txt)
[dump_part2.txt](https://github.com/runtimeverification/haskell-backend/files/12168720/dump_part2.txt)

Please ask me for additional information / files if needed.

Thanks !

goodlyrottenapple:
Would you also happen to have the bug report tar file? This would make it easier to test and validate any changes.

iFrostizz:
Thanks for looking into that @goodlyrottenapple !
Please find the bug report attached

[bug_report.tar.gz](https://github.com/runtimeverification/haskell-backend/files/12170342/bug_report.tar.gz)

ehildenb:
[...]

PR #3637, opened 2023-08-04
Return vacuous in `kore-rpc` when state goes to #bottom

goodlyrottenapple:
Another attempt at addressing https://github.com/runtimeverification/evm-semantics/issues/1944. Addresses #3630 (needs a follow-up in booster)

This change mimicks what kore-exe in prover mode returns(as introduced by https://github.com/runtimeverification/haskell-backend/pull/2451), namely the configuration goes to bottom


kore/src/Kore/Exec/GraphTraversal.hs
```diff
@@ -397,7 +411,7 @@ data StepResult a
 data TraversalResult a
     = -- | remaining queue length and stuck results (always at most
       -- maxCounterExamples many).
-      GotStuck Int [a]
+      GotStuck Int [Either a a]
```
jberthold, line 414: Should maybe modify the comment here to explain what `Left` and `Right` in `GotStuck` mean...

kore/src/Kore/Exec/GraphTraversal.hs
```diff
@@ -420,7 +434,7 @@ instance Pretty a => Pretty (TraversalResult a) where
         GotStuck n as ->
             Pretty.hang 4 . Pretty.vsep $
                 ("Got stuck with queue of " <> Pretty.pretty n)
-                    : map Pretty.pretty as
+                    : map (either Pretty.pretty Pretty.pretty) as
```
jberthold, line 437: Maybe better to describe in text whether it was `Left` ("stuck") or `Right` ("vacuous")?

PR #3638, opened 2023-08-09
Run integration tests on `self-hosted` workflow runners

geo2a:
Speed-up CI a little bit
* Run integration tests concurrently with unit tests
* Run integration tests on a self-hosted runner (instead of a public one)


goodlyrottenapple (approved):
Nice one

PR #3650, opened 2023-09-06
Prune #Bottom states when branching

goodlyrottenapple:
Fix for change introduced by #3637, wherein branching could contain bottom states which were not being pruned. (see https://github.com/runtimeverification/pyk/pull/634#issuecomment-1707957228)


cabal.project.freeze
```diff
@@ -11,7 +11,6 @@ constraints: any.Cabal ==3.6.3.0,
              aeson -cffi +ordered-keymap,
              any.aeson-pretty ==0.8.9,
              aeson-pretty -lib-only,
-             any.alex ==3.2.7.4,
```
jberthold, line 14: Curious, why would you like to remove those dependencies?
Maybe we should adapt the scripts that check or re-freeze if we need to remove them.

jberthold (approved):
Code change LGTM.
Re-ran the integration test in CI :crossed_fingers:

jberthold:
> Re-ran the integration test in CI 🤞

That did not work, though... `kore-exec` gets killed (error 137) [in a search command](https://github.com/runtimeverification/haskell-backend/actions/runs/6100066478/job/16564111760#step:6:6514) 
```
krun --haskell-backend-command kore-exec --definition ./test-kompiled --search-final 2.test \
    --output-file 2.test.out --output /tmp/nix-shell.NiDd7o/.krun-2023-09-06-23-59-15-7rLPGoxTwy/result.kore  \
    --searchType FINAL --search 
```

geo2a (approved):

cabal.project.freeze
goodlyrottenapple: I have already removed these from the booster. they should not be fixed because the nix version of cabal comes with a different version of alex and happy and cant figure out a build plan using these

PetarMax:
@goodlyrottenapple @geo2a  I've also addressed this issue by pruning the `#Bottom` branches in `pyk` right now.

PR #3651, opened 2023-09-08
Emit KoreJson \right-assoc for collections

goodlyrottenapple:
This PR modifies the encoding of lists, maps and sets to use `\right-assoc{concatSymbol}(...)` to remove differences in bracketing. THis should also make it easier to write a diffing application which ignores the order of/sorts the list of children.


kore/src/Kore/Syntax/Json/Internal.hs
```diff
@@ -307,25 +330,305 @@ fromPatternF (_ :< patt) = case patt of
...
+                { sort = fromSort domainValueSort
+                , value
+                }
+
+        -- The encoded lists and ac types are all of the following form:
+        -- If the structure has no elements, we simply get the application of the unit symbol with no arguments
+        -- If the structure has exactly one child, we just get the child.
+        -- This child will be an application of internalElement symbol to its arguments
+        -- For multiple children we will get a \right-assoc{concatSymbol}(x1,...x_n)
+        foldApplication Kore.Symbol{symbolConstructor = unitSymbolConstructor, symbolParams = unitSymbolParams} Kore.Symbol{symbolConstructor = concatSymbolConstructor, symbolParams = concatSymbolParams} = foldAux
```
jberthold, line 556: For reference: https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Internal/TermLike/TermLike.hs#L1385-L1440 differs in `foldApplication` only.
Unfortunately there is no good way to avoid the duplication.

test/rpc-server/runTests.py
```diff
@@ -107,20 +107,23 @@ def checkGolden (resp, resp_golden_path):
 
 def runTest(def_path, req, resp_golden_path):
     with subprocess.Popen(f"{SERVER} {def_path} --module TEST --server-port {PORT} --log-level {server_log_level[VERBOSITY]}".split()) as process:
-        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```
jberthold, line 110: Maybe we should rewrite this to run a whole array of tests on the same server? (does not have to be in this PR, obviously)

geo2a (commented):
The integration test looks good to me!

I've skimmed through the implementation, and it aligns with my intuition. Though I have to admit that my expertise here is limited.

I agree with Jost's comment about unit-testing the new `fromTermLike` function

kore/src/Kore/Syntax/Json.hs
```diff
@@ -100,15 +101,14 @@ prettyJsonOpts =
 fromTermLike :: TermLike VariableName -> KoreJson
 fromTermLike =
     addHeader
-        . fromPattern
-        . from @_ @(Kore.Pattern _ (TermLike.TermAttributes VariableName))
+        . Kore.Syntax.Json.Internal.fromTermLike
```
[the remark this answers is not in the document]
goodlyrottenapple: Not sure what the best way to go about this is tbh. `fromPattern` is basically unused, but we use the inverse `toParsedPattern` in a lot of places. The problem is we go from `KorePattern` to `ParsedPattern` which at some point gets turned into a `TermLike` which we then turn back into a `KorePattern`. Perhaps that path should be added and we keep the other unit tests??

jberthold (approved):
Would still be good to come up with unit tests for `Internal.fromTermLike` but pre-approving

PR #3652, opened 2023-09-11
Use (||) directly when enabling tracing

geo2a:
This PR fixes the case of an `ExecuteRequest` with logging parameters `{logSuccessfulRewrites = Just True, logSuccessfulSimplifications = Nothing}` not retuning any logs (except of terminal/cut-point case).

### DIFF
--- a/kore/src/Kore/JsonRpc.hs
+++ b/kore/src/Kore/JsonRpc.hs
@@ -124,6 +124,8 @@ respond serverState moduleName runSMT =
                 case verifyIn serializedModule state of
                     Left err -> pure $ Left $ backendError CouldNotVerifyPattern err
                     Right verifiedPattern -> do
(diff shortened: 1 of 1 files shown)


kore/src/Kore/JsonRpc.hs
```diff
@@ -134,7 +136,7 @@ respond serverState moduleName runSMT =
                                             then EnableMovingAverage
                                             else DisableMovingAverage
                                         )
-                                        (fromMaybe False $ liftA2 (||) logSuccessfulRewrites logSuccessfulSimplifications)
```
geo2a, line 137: This evaluates to `False` if `logSuccessfulSimplifications` is `Nothing`, independent of `logSuccessfulRewrites`.

goodlyrottenapple (approved):

PR #3657, opened 2023-09-13
Add `Fallback` to `LogEntry`

### LINKED ISSUE #3654: Log rewritten term on succesful rewrite
To facilitate generating `kore-rpc-booster` abort K claims (https://github.com/runtimeverification/hs-backend-booster/issues/295), we need to:
* modify the RPC log type in `kore-rpc-types` to allow Booster to log the original term when it aborts 
* `kore-rpc`'s behavior to log `rewrittenTerm` on `Success`, as currently it just returns `Nothing`. We may consider logging `originalTerm` on `Failure` too, but I think this is redundant, as the Booster should have had aborted at the same state. 

## Changing the Log type

I propose adding `originalTerm` to the `Failure` constructor:

```
data LogRewriteResult
    = Success
        { rewrittenTerm :: Maybe KoreJson
        , substitution :: Maybe KoreJson
        , ruleId :: Text
        }
    | Failure
        { reason :: Text
        , originalTerm :: Maybe KoreJson
        , _ruleId :: Maybe Text
        }
```

## Changing `Kore.JsonRpc.respond`

The responses Kore returns when given an `ExecuteRequest` should include the `rewrittenTerm` in the `Rewrite Success` logs. Currently, the `mkLogs` function always sets the rewritten term to Nothing.

I would need help figuring out where to get this term from and which functions to use to externalize it properly.

geo2a:
This PR modifies the `LogEntry` type in `kore-rpc-types` in two ways:
* add `origianlTerm` into `Rewrite` log entry constructor (and emit Nothing for now in Kore to retain the current behavior. This change is not necessary at the moment, but we may need to log these terms in future.
* Add `Fallback` constructor to `LogEntry` to enable `kore-rpc-booster` to emit a trace when it falls back from Booster to Kore.

### DIFF
--- a/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
+++ b/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
@@ -18,7 +18,7 @@ import Deriving.Aeson (
     StripPrefix,
  )
 
-data LogOrigin = KoreRpc | Booster | Llvm
+data LogOrigin = KoreRpc | Booster | Llvm | Proxy
     deriving stock (Generic, Show, Eq)
     deriving
         (FromJSON, ToJSON)
--- a/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
+++ b/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
@@ -46,7 +46,8 @@ data LogRewriteResult
 
 data LogEntry
     = Rewrite
-        { result :: LogRewriteResult
+        { originalTerm :: Maybe KoreJson
+        , result :: LogRewriteResult
         , origin :: LogOrigin
         }
     | Simplification
--- a/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
+++ b/kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
@@ -3,8 +3,10 @@
 module Kore.JsonRpc.Types.Log (module Kore.JsonRpc.Types.Log) where
 
 import Data.Aeson (FromJSON, ToJSON)
+import Data.List.NonEmpty (NonEmpty)
 import Data.Text (Text)
 import GHC.Generics (Generic)
+import Kore.JsonRpc.Types.Depth (Depth (..))
 import Kore.Syntax.Json.Types (KoreJson)
 
 import Deriving.Aeson (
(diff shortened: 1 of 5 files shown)


kore/src/Kore/JsonRpc.hs
```diff
@@ -180,10 +181,13 @@ respond serverState moduleName runSMT =
                                     ++ [ Rewrite
                                         { result =
                                             Success
-                                                { rewrittenTerm = Nothing
+                                                { rewrittenTerm =
+                                                    PatternJson.fromTermLike . Pattern.term
+                                                        <$> successfullyRewrittenTerm
```
jberthold, line 186: That seems wrong... we would emit all the rewrite traces with the same rewritten term (the final result).
The rewritten/original terms must be stored in the `Exec.RuleTrace`  if we want to emit them (i.e., the change needs to go deeper into the rewriting logic).

geo2a: Right. Good catch. I though I could get away with a simple solution, but did not give it enough thought. Thanks!

goodlyrottenapple:
`originalTermIndex` is meant to be a pointer to avoid re-sending large configurations in the logs over and over. not been implemented

kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
```diff
@@ -46,7 +48,8 @@ data LogRewriteResult
 
 data LogEntry
     = Rewrite
-        { result :: LogRewriteResult
+        { originalTerm :: Maybe KoreJson
```
goodlyrottenapple, line 51: do we need the original term? the intention for the rewrite trace was to be used when debugging proofs with single step turned on in pyk, so we already have the `originalTerm` anyway.

kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
```diff
@@ -55,6 +58,21 @@ data LogEntry
...
+        -- ^ state after fallback
+        , reason :: Text
+        -- ^ fallback reason
+        , ruleIds :: NonEmpty Text
+        -- ^ rules applied during fallback, the first rule is the one that caused the fallback
+        , recoveryDepth :: Depth
+        -- ^ depth reached in fallback, must be the same as (length ruleIds)
+        , origin :: LogOrigin
+        -- ^ proxy server the log was emitted from
+        }
```
goodlyrottenapple, line 75: this looks good. should have all the info we need :)

kore-rpc-types/src/Kore/JsonRpc/Types/Log.hs
geo2a: Agreed, let's drop it after all.

jberthold (approved):
LGTM (Changes many of the same places as #3658 ...  let's get it in and move on)

Issue #3660, opened 2023-09-20
Consistent SMT solver failure in a Kontrol proof


geo2a:
@hjorthjort and @PetarMax are stuck on a proof that consistently fails to make progress due to an SMT solver error. Unfortunately, `pyk` does not provide more information than saying `pyk.kore.rpc.JsonRpcError: Smt solver error`. We need to drill deeper into what is being sent to Z3.

The stuck configuration looks like this:
```
<k>
  SDIV chop ( ( ( ( VV0_amount_114b9705:Int *Int VV0_amount_114b9705:Int ) *Int ( VV0_amount_114b9705:Int *Int VV0_amount_114b9705:Int ) ) *Int 115792089237316195423570985008687907853269984665640564039457584007913129639899 ) ) 280000000000000000
  ~> ...
</k>
 ...
#And ( #Not ( { VV0_amount_114b9705:Int #Equals 0 } )
#And ( { 0 <=Int CALLER_ID:Int #Equals true }
#And ( { 0 <=Int ORIGIN_ID:Int #Equals true }
#And ( { 0 <=Int NUMBER_CELL:Int #Equals true }
#And ( { 0 <=Int VV0_amount_114b9705:Int #Equals true }
#And ( { VV0_amount_114b9705:Int <=Int 2499999 #Equals true }
#And ( { CALLER_ID:Int <Int pow160 #Equals true }
#And ( { ORIGIN_ID:Int <Int pow160 #Equals true }
#And ( { NUMBER_CELL:Int <=Int maxSInt256 #Equals true }
#And { VV0_amount_114b9705:Int <Int pow256 #Equals true } ) ) ) ) ) ) ) ) ) )
``` 

And we have a [bug_report.tar.gz](https://files.slack.com/files-pri/T5FKNMQEN-F05T6PK2HK6/download/hjort-bug.tar.gz?origin_team=T5FKNMQEN) (Slack link).

PR #3661, opened 2023-09-21
Enable configuration of Z3 satisfiability tactics

geo2a:
Fixes [#3660](https://github.com/runtimeverification/haskell-backend/issues/3660)

This PR: 
* adds the `--smt-tactic <s-expression>` command line option [Z3 tactic](https://microsoft.github.io/z3guide/docs/strategies/tactics/);
* If the option is not supplied, the default tactic `(check-sat)` will be used;
* The `execute` RPC integration tests are now run with a customized tactic (and also without, as prior) to ensure that it is picked up and the behavior is the same.

If a tactic is provided, it will be used to check satisfiability of rewrite/simplification rules' side conditions. Tactics are specified as SMT-LIB2 s-expression, but are Z3-specific, i.e. other solvers may not support them.


ehildenb:
Let's make sure the integration test change is committed prior to merge, as an example for folks.

ehildenb (approved):

Issue #3542, opened 2023-03-10
Add parser support for multiary `\and` and `\or`


tothtamas28:
Implement the following changes in the KORE parser:

```diff
< "\and" "{" <sort> "}" "(" <pattern> "," <pattern> ")"
> "\and" "{" <sort> "}" "(" <patterns> ")"
```

```diff
< "\or" "{" <sort> "}" "(" <pattern> "," <pattern> ")"
> "\or" "{" <sort> "}" "(" <patterns> ")"
```

As long as output only uses the previous format (i.e. binary connectives), the change is backward compatible.

### Flat representation for constrained terms

This simplifies

```
\and{SortGeneratedTopCell{}}(
    configuration,
    \and{SortGeneratedTopCell{}}(
        constraint1,
        constraint2
    )
)
```

to

```
\and{SortGeneratedTopCell{}}(configuration, constraint1, constraint2)
```

### Prepare removal of syntax sugar

Textual KORE currently has syntax sugar

```
"\left-assoc" "{" "}" "(" <multi-or> ")"
"\right-assoc" "{" "}" "(" <multi-or> ")"
```

which have limited use:

* Syntax `"\left-assoc" "{" "}" "(" <multi-or> ")"` is used in the LLVM backend to output the result of `--search`. The Haskell backend supports this syntax to be compatible with that output.
* Syntax `"\right-assoc" "{" "}" "(" <multi-or> ")"` has no uses.

With support for multiary `\or`,  the syntactic sugar for `\or` could be dropped altogether:

* Changing the LLVM backends output for `--search` is straightforward.
*  No need to adjust the LLVM backend's parser as the backend doesn't support ML connectives in the first place.

(@Baltoli, please check if my understanding is correct.)

ana-pantilie:
@tothtamas28 what is the priority of this? Doing it properly would mean changes in both the frontend and the backends. Is there a compelling reason that would require this change?

tothtamas28:
Priority for this is moderate / low.

> Is there a compelling reason that would require this change?

Outlined above.

> changes in both the frontend and the backends

Changes can be introduced gradually:

1. Add parser support without using the new constructs in the output. This is not a breaking change.
2. Start using new syntax in the output.
3. Drop syntax sugar.

tothtamas28:
> I think such a change needs discussing with

I agree.

> Did you discuss this at last week's K dev meeting?

No, let's discuss it this week.

ehildenb:
[...]

PR #3666, opened 2023-10-02
Support multi-ary \and/\or in Haskell backend

dwightguth:
Part of #3542 

Essentially, we add support for \and and \or with 0 or more children. The parser automatically desugars 0 or 1 children, so the AST will always have at least 2 children. We modify the Internal modules to use a new class, BinaryAnd and BinaryOr, which preserves the old representation. Externalization and internalization then converts between the two formats.

Most of this code is just needed to separate the `And` in `Kore.Internal` from the `And` in `Kore.Syntax`. The actual changes are relatively small.

### Estimate:

---

###### Review checklist


kore/src/Kore/Validate/PatternVerifier.hs
```diff
@@ -230,15 +232,39 @@ verifyOperands operandSort = \operator -> do
...
 
+internalizeAnd ::
+    And Sort Verified.Pattern ->
+    BinaryAnd Sort Verified.Pattern
+internalizeAnd And{andSort, andChildren = [andFirst, andSecond]} =
+    BinaryAnd{andSort, andFirst, andSecond}
+internalizeAnd And{andSort, andChildren} =
+    let first = head andChildren
+     in let second = internalizeAnd And{andSort, andChildren = tail andChildren}
+         in BinaryAnd{andSort, andFirst = first, andSecond = synthesize (Internal.AndF second)}
```
geo2a, line 243: I understand that the case of getting an `And` with 0 or 1 child is unlikely, but I think it would be better to handle this case with an explicit call to `error`, i.e. add another equation
```suggestion
internalizeAnd ::
    And Sort Verified.Pattern ->
    BinaryAnd Sort Verified.Pattern
internalizeAnd And{andSort, andChildren}
    | length andChildren <= 1 = error "Cannot internalize And with less than two children"
    | otherwise = case andChildren of
        [andFirst, andSecond] -> BinaryAnd{andSort, andFirst, andSecond}
        _ ->
            let first = head andChildren
                second = internalizeAnd And{andSort, andChildren = tail andChildren}
             in BinaryAnd{andSort, andFirst = first, andSecond = synthesize (Internal.AndF second)}
```

kore/src/Kore/Validate/PatternVerifier.hs
```diff
@@ -230,15 +232,39 @@ verifyOperands operandSort = \operator -> do
...
+         in BinaryOr{orSort, orFirst = first, orSecond = synthesize (Internal.OrF second)}
+
 verifyAnd ::
     And Sort (PatternVerifier Verified.Pattern) ->
-    PatternVerifier (And Sort Verified.Pattern)
-verifyAnd = verifyOperands andSort
+    PatternVerifier (BinaryAnd Sort Verified.Pattern)
+verifyAnd operands = do
+    verifiedAnd <- verifyOperands And.andSort operands
+    return (internalizeAnd verifiedAnd)
```
geo2a, line 260: Actually, alternatively to the suggestion above, we could verify that the And has at lease two children here. I think I prefer this option. Please go with the one you find most  appropriate. The same will need to be done for Or.
```suggestion
    koreFailWhen (length operands < 2) "Cannot internalize And with less than two children"
    verifiedAnd <- verifyOperands And.andSort operands
    return (internalizeAnd verifiedAnd)
```

geo2a (commented):
Thanks @dwightguth, this looks good. I've left two comments about verification and internilization of And/Or with less that two children.

I'll let @goodlyrottenapple to have a look too.

kore/src/Kore/Validate/PatternVerifier.hs
goodlyrottenapple: definitely agree. using head/tail is usually not a good idea because when the list is empty you get a generic error and don't know where it's coming from. @geo2a's suggestion throws a better error message.

goodlyrottenapple (approved):

PR #3677, opened 2023-10-13
Slimmed down nix distribution of kore

goodlyrottenapple:
The closure of the compiled kore binaries in nix comes to about 3.5gb as it includes links to all the haskell libraries used to compile the executables. The nix infrastructure for haskell includes a nice utility for trimming the size of closures (see `justStaticExecutables`  in  https://nixos.org/manual/nixpkgs/stable/#haskell). However, this is not compatible with the way haskell.nix packages things. Since we don't care about cross-compilation, which haskell.nix is good at, and given the bloated nature of haskell.nix, I have decided to re-write the nix packaging with the much more light-weight [stacklock2nix](https://github.com/cdepillabout/stacklock2nix), which is a thin wrapper around the Haskell packaging infrastructure used by nixpkgs. With this change, the closure of the kore binaries is around 370mb (roughly 120mb per executable)

Furthermore this PR greatly simplifies the nix flake itself, as it removes a lot of packages that are never used in practice, such as all the profiling builds, since we usually build those locally via stack/cabal anyway. In a similar spirit, the unit tests are now being run via the dev shell, calling cabal directly. This eliminates the need for a separate cabal unit test run.


geo2a (approved):
I see there's a problem with calling the Kore integration tests from the K flake, but otherwise looks great, approving. Thanks @goodlyrottenapple!

.github/workflows/test.yml
```diff
@@ -146,46 +159,6 @@ jobs:
                 --override-input haskell-backend . --update-input haskell-backend \
                 --command bash -c "cd test && make -j2 --output-sync test"
 
-  cabal:
-    name: 'Cabal / Unit Tests'
```
jberthold, line 150: No `cabal` build any more?

goodlyrottenapple: the nix job calls cabal directly so no point in having a separate one

goodlyrottenapple:
@jberthold @geo2a I was thinking of disabling integration tests in this PR and merging, then fixing K haskell-backend deps update PR and then re-enabling integration tests in another PR here?

PR #3684, opened 2023-10-23
Update `KRYPTO` hooks to use `Bytes` instead of `String`

### LINKED ISSUE #3683: Update `KRYPTO` hooks to use `Bytes`
All of the `KRYPTO` hooks now use `Bytes` instead of `String` to represent raw byte strings (runtimeverification/blockchain-k-plugin#163). We need to update everything here accordingly.

Scott-Guest:
Fixes #3683 

The `KRYPTO` hooks now use `Bytes` for raw strings instead of `String`, so this PR updates the evaluators here accordingly.

Additionally, I updated all tests that reference `KRYPTO` lest they give errors about hooked symbols having incorrect sorts. Specifically,
- I regenerated the `regression-evm` tests by running `generate-regression-tests.sh` with my branch of KEVM that uses the updated hooks.
- For all the tests that I couldn't regenerate, I manually patched them by changing the existing `KRYPTO` symbols into functions which wrap the new hooks with appropriate calls to `String2Bytes` and `Bytes2String`.

Corresponding PR testing this on the booster: runtimeverification/hs-backend-booster#370


Scott-Guest:
Blocked on #3687 - the `regression-evm` tests need to be updated to account for the `KRYPTO` changes

Scott-Guest:
Holding off on actually merging this until I can locally test integration with KEVM and the booster

yale-vinson:
@Scott-Guest Is this still blocked? This is blocking Semantics [#2107](https://github.com/runtimeverification/evm-semantics/pull/2107) and I would like to get that one closed out as it has been aging in the backlog for quite a while.

goodlyrottenapple (approved):
LGTM. Do we need to modify stuff upstream once this is merged? If so, is everything else ready?

Scott-Guest:
@goodlyrottenapple There are number of changes needed upstream, but they're all trivial and/or ready to go:
- A few KORE files need to be patched similar to the changes made here:
    - A few booster tests which reference KRYPTO. This is ready to go (runtimeverification/hs-backend-booster#370)
    - K's profiling testsuite: https://github.com/runtimeverification/k/tree/master/k-distribution/tests/profiling. This is straightforward, and I'll address it when K bumps its booster version.
- All semantics using the `KRYPTO` hooks need to bump their K and blockchain versions simultaneously, then update to the `Bytes`-based new interface
    - For KEVM, this requires a quick textual find and replace for the wrappers I added in runtimeverification/evm-semantics#2128
    - For AVM, this requires removing calls to `String2Bytes` and `Bytes2String` in 3 rules: https://github.com/runtimeverification/avm-semantics/blob/5cd370b7a14bab2ad64d554908f6b6ac62c027e5/lib/include/kframework/avm/teal/teal-driver.md?plain=1#L54
    - For MultiverseX, this requires removing calls to `Bytes2String` in 2 rules: https://github.com/runtimeverification/mx-semantics/blob/58bc391445aa48676003c4ea67592851a3c1cef9/vmhooks/cryptoei.md#L23

PR #3687, opened 2023-10-25
Update `generate-regression-tests.sh` work with modern evm-semantics

### LINKED ISSUE #3686: Update `generate-regression-tests.sh` to work with current evm-semantics
The script `generate-regression-tests.sh` used to re-generate all of the tests in `test/regression-evm` is out of date and no longer works with recent versions of the evm-semantics. 

Specifically, it relies on the old evm-semantics Makefile-based build system, and needs to be updated to use the `poetry -C kevm-pyk ...` based commands instead.

Scott-Guest:
Fixes #3686

This PR does the following
- Updates `generate-regression-tests.sh` to work with modern versions of evm-semantics which no longer use a Makefile-based build system.
- Fixes all issues reported by `shellcheck`.
- Runs the new script and updates all the `regression-evm` tests.

---

###### Review checklist


scripts/generate-regression-tests.sh
```diff
@@ -72,136 +72,153 @@ which git > /dev/null || err "git not available"
...
 
-    cd ${EVM_SEMANTICS}
+    pushd "${EVM_SEMANTICS}"/"$name"
 
-    mkdir $tmp
-    mv $archive $tmp
-    cd $tmp
-    tar -xf $archive
+    mkdir "$tmp"
+    cp kevm-bug-*.tar.gz "$tmp"
```
Scott-Guest, line 137: The `kevm-bug-*.tar.gz` glob here is to work around some weird behavior where `functional/lemmas` generates a bug report named `kevm-bug-lemma.tar.gz` (without the "s"). I updated the script to run each test in its own directory, so the glob should find a unique file.

test/regression-evm/test-totalSupply.sh
```diff
@@ -5,13 +5,18 @@ exec kore-exec \
...
     --strategy all \
     --max-counterexamples 1 \
-    --smt-timeout 99 \
-    --smt-retry-limit 1 \
+    --smt-timeout 125 \
+    --smt-retry-limit 3 \
     --smt-reset-interval 100 \
     --smt z3 \
+    --log \
+    /home/sguest/work/haskell-backend/evm-semantics/tests/specs/erc20/ds/totalSupply-spec.debug-log \
```
Scott-Guest, line 13: Ah need to `sed` the script as well, will address this

test/regression-evm/test-dsvalue-peek-pass-rough.sh
```diff
@@ -5,13 +5,18 @@ exec kore-exec \
...
     --strategy all \
     --max-counterexamples 1 \
-    --smt-timeout 99 \
-    --smt-retry-limit 1 \
+    --smt-timeout 125 \
+    --smt-retry-limit 3 \
     --smt-reset-interval 100 \
     --smt z3 \
+    --log \
+    /home/sguest/work/haskell-backend/evm-semantics/tests/specs/mcd/dsvalue-peek-pass-rough-spec.debug-log \
```
jberthold, line 13: :raised_eyebrow:

test/regression-evm/test-totalSupply.sh
jberthold: Why are we even producing these logs? This should probably be removed from the `evm-semantics` code.
But yes, a `sed` script to kill the `--log` and its argument (either one or two lines, depending on `\` being present)  is good enough. Or I'd suggest to rather  just fix this manually now, we are not running this  script automatically and the correct fix is to change `evm-semantics`.

jberthold (approved):
Pre-approving, provided the scripts are changed (manually) before merging

Scott-Guest:
@jberthold It looks like Pyk (and correspondingly KEVM) always enables logging in order to detect whether a proof took zero steps: https://github.com/runtimeverification/pyk/blob/fd2cd65b01919ac608e0c4f8a8d21251f5d8a973/src/pyk/ktool/kprove.py#L252

I ended up just sed-ding out the `--log` entry rather than removing this check.

PR #3692, opened 2023-11-01
Catch IOError when a query times out and retry correctly

goodlyrottenapple:
This (hopefully) fixes https://github.com/runtimeverification/k/issues/3729, specifially the error below, thrown [here](https://github.com/runtimeverification/haskell-backend/blob/58a101d279efb64e7042bab2968a22f27a9e2706/kore/src/SMT/SimpleSMT.hs#L444-L450)
```
kore-exec: [41516716] Error (ErrorException):
    user error (Unexpected result from the SMT solver:
      Command: (push 1 )
      Expected: success
      Result: (error "line 178 column 8: push canceled" )
    )
```

This error occurs when smt-limit is low and we send an assert query to the solver. When this happens, we want to re-try instead of letting the error bubble up, as it currently does.

I have tested this and confirmed that at a threshold level of `smt-timeout` where Z3 fails intermittently, with this change, we re-try the specified number of times before failing fully.

### DIFF
--- a/kore/src/Kore/Rewrite/SMT/Evaluator.hs
+++ b/kore/src/Kore/Rewrite/SMT/Evaluator.hs
@@ -188,7 +188,7 @@ decidePredicate onUnknown sideCondition predicates =
             & runMaybeT
   where
     query :: MaybeT Simplifier Result
-    query = SMT.withSolver . evalTranslator $ do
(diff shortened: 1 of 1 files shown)


jberthold (approved):
LGTM. Good job tracking this down!

geo2a (approved):

PR #3697, opened 2023-11-23
Add optional unknown predicate into "execute" response

geo2a:
Add `"unknown-predicate"` as an optional field to `"execute"` JSON-RPC response. The intention is to change servers' response on encountering `ErrorDecidePredicateUnknwon`, i.e. when Z3 returns Unknwon while checking a rewrite rule's side condition.

The current behavior is to return `SmtSolverError` even if progress was made. I.e. we could have made 100 rewrite steps and only get Unknown during step 101.

Instead, we should return the state before the error, and the predicate that Z3 had trouble deciding. This way, the client will be able to make an informed decision about that to do next and/or report the problematic predicate to the user, so that they can add lemmas if appropriate.

Unfortunately, altering `kore-rpc` itself is quite difficult. therefore I will implement a prototype of this in `Proxy.hs` of `kore-rpc-booster`. The type and docs need to be update here though.

### DIFF
--- a/docs/2022-07-18-JSON-RPC-Server-API.md
+++ b/docs/2022-07-18-JSON-RPC-Server-API.md
@@ -156,7 +156,6 @@ The `unknown-predicate` field MAY contain a predicate that could not be decided 
     },
     "depth": 2,
     "reason": "aborted",
-    "rule": "rule1",
     "unknown-predicate": {"format":"KORE", "version":1, "term":{}}
     "logs": []
   }
--- a/docs/2022-07-18-JSON-RPC-Server-API.md
+++ b/docs/2022-07-18-JSON-RPC-Server-API.md
@@ -164,7 +164,7 @@ The `unknown-predicate` field MAY contain a predicate that could not be decided 
 ```
 
 ##### `"reason": "cut-point-rule"`
-Execution was about to perform a rewrite with a rule whose label is one of the `cut-point-rules` labels/IDs of the request.
+Execution was about to perform a rewrite with a rule whose label is one of the `cut-point-rules` labels/IDs of the request.  
 An additional `rule` field indicates which of the `cut-point-rules` labels/IDs led to stopping.
 An additional `next-states` field contains the next state (which stems from the RHS of this rule) in a singleton list.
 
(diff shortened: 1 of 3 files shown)


docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -142,10 +142,29 @@ This response has no additional fields.
...
+    "rule": "rule1",
+    "unknown-predicate": {"format":"KORE", "version":1, "term":{}}
+    "logs": []
+  }
+}
+```
 
 ##### `"reason": "cut-point-rule"`
-Execution was about to perform a rewrite with a rule whose label is one of the `cut-point-rules` labels/IDs of the request.  
+Execution was about to perform a rewrite with a rule whose label is one of the `cut-point-rules` labels/IDs of the request.
```
jberthold, line 167:
```suggestion
Execution was about to perform a rewrite with a rule whose label is one of the `cut-point-rules` labels/IDs of the request.  
```

jberthold: Those two trailing spaces cause a line break when rendering the markdown...

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -142,10 +142,29 @@ This response has no additional fields.
...
+  "id": 1,
+  "result": {
+    "state": {
+      "term": {"format":"KORE", "version":1, "term":{}},
+      "predicate": {"format":"KORE", "version":1, "term":{}},
+      "substitution": {"format":"KORE", "version":1, "term":{}},
+    },
+    "depth": 2,
+    "reason": "aborted",
+    "rule": "rule1",
```
jberthold, line 159:
```suggestion
```

jberthold (approved):
LGTM, just the stray `rule` field needs to go

geo2a:
Thanks @jberthold!

docs/2022-07-18-JSON-RPC-Server-API.md
```diff
@@ -142,7 +142,25 @@ This response has no additional fields.
 
 ##### `"reason": "aborted"`
 Execution reached a state that the server cannot handle.
-This response has no additional fields.
+The `unknown-predicate` field MAY contain a predicate that could not be decided by the server's constraint solver.
```
tothtamas28, line 145: Does this case also cover an SMT solver crash, or just an `unkown` status?

PR #3704, opened 2023-12-12
Concurrent `kore-rpc` server

goodlyrottenapple:
Implements the proposal from https://github.com/runtimeverification/hs-backend-booster/issues/383

Namely:

`add-module`
-------------

* now returns the id of a given module, of the form `m<sha256_of_given_module>` 
* computes the SHA256 hash of the unparsed module string and saves the internalised module under this key
* takes an optional `name-as-id: true` argument which implements the previous behaviour of `add-module` (still returns the new id), i.e. adds the module to the module map under the module name as well as the id. 
* if the same module is added twice with `name-as-id: true`, the second request will fail with a `Duplicate module` error
* if the same module is sent twice with `name-as-id: false` or without `name-as-id`, the second request is idempotent and will succeed

As discussed on the issue, the IDs and original names are not disjoint for ease of implementation. the `m` pre-pended to the hash is necessary to make the name a valid kore identifier. 

I have also not added the `modules` field to other methods as I'm unsure whether this should be replacing the current  `module` parameter, which would be a backwards incompatible change to the API that would break a lot of our integration tests, or if it should be an additional field, @tothtamas28?


kore/src/Kore/JsonRpc.hs
```diff
@@ -455,54 +458,113 @@ respond serverState moduleName runSMT =
...
-
-                                    pure . Right . AddModule $ AddModuleResult (getModuleName name)
+        AddModule AddModuleRequest{_module, nameAsId = nameAsId'} -> runExceptT $ do
+            let nameAsId = fromMaybe False nameAsId'
+            parsedModule@Module{moduleName = name} <-
+                withExceptT (backendError CouldNotParsePattern) $
+                    liftEither $
+                        parseKoreModule "<add-module>" _module
+            LoadedDefinition{indexedModules, definedNames, kFileLocations} <-
+                liftIO $ loadedDefinition <$> MVar.readMVar serverState
```
jberthold, line 468: Should maybe `takeMVar` here and `putMVar` instead of `modifyMVar` at the end (enforcing sequential consistency). Otherwise two concurrent requests could pass the check for duplicate module names but then add the same module name to the map.

test/rpc-server/add-module/add/response.golden
```diff
@@ -1 +1 @@
-{"jsonrpc":"2.0","id":1,"result":{"module":"NEW"}}
+{"jsonrpc":"2.0","id":1,"result":{"module":"m8e723e34b156f187e7cce8658dbb513244788a1ab81c1f496b909b1a2d1cc0c3"}}
```
jberthold, line 1: You need to add the `name-as-id` option to the add-module request, too so you can use the name `NEW` in the next request

geo2a (approved):

jberthold (approved):

tothtamas28:
> if the same module is added twice with `name-as-id: true`, the second request will fail with a `Duplicate module` error

I think the `Duplicate module` error is only necessary if the modules are not the same (but the names are). Otherwise, the request can be idempotent.

> I have also not added the `modules` field to other methods

Does this mean that the modules are still added globally, but now the unique ID's work as well for module imports? If so, that's reasonable, we can tackle the breaking changes in the new year.

goodlyrottenapple:
> I think the `Duplicate module` error is only necessary if the modules are not the same (but the names are). Otherwise, the request can be idempotent.

Ok, will make that change

> Does this mean that the modules are still added globally, but now the unique ID's work for module imports? If so, that's reasonable, we can tackle the breaking changes in the new year.

Yes all modules are added globally and you can send in a module that imports other modules via their ID.

tothtamas28 (approved):

PR #3705, opened 2023-12-13
Assume definedness of current configuration before rewriting

geo2a:
Fixes https://github.com/runtimeverification/haskell-backend/issues/3703

Kore is designed to compute and check definedness of the current configuration before attempting rewrite rules. In cases of large concrete maps and sets with a handful of symbolic variables, the created massive `#Ceil` predicates grind Kore to a halt.

However, when falling back to Kore from Booster, everything must be defined, though it is difficult to convey this information from Booster to Kore.

As an experimental workaround, we assume definiteness of every sub-term of the current configuration before attempting rewrite rules. This information should probably be conveyed from Booster in future.


goodlyrottenapple:
The current implementation will definitely break kore-exec. We should probably thread a flag to this function whether to add the defined terms and maybe add `"preserves-definedness" : true` to the execute request to trigger this addition

geo2a:
Sounds good. Also, I think what we do is safe only if we're rewriting for at most one step, since we lose Booster's definiteness guarantee if we make more than one step in Kore. @goodlyrottenapple @jberthold what do you think? I don't think we need to strictly enforce this contract in Kore, but we do need to document it.

goodlyrottenapple:
> we lose Booster's definiteness guarantee if we make more than one step in Kore

I don't think this is true... i was under the impression the old back-end checks ceils after each rewrite step

geo2a:
> i was under the impression the old back-end checks ceils after each rewrite step

Right. Then it's fine!

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -36,6 +36,7 @@ data ExecuteRequest = ExecuteRequest
     , terminalRules :: !(Maybe [Text])
     , movingAverageStepTimeout :: !(Maybe Bool)
     , stepTimeout :: !(Maybe Int)
+    , enableAssumeInitialDefined :: !(Maybe Bool)
```
goodlyrottenapple, line 39: maybe a bit shorter flag?

```suggestion
    , assumeStateDefined :: !(Maybe Bool)
````

Comment (geo2a) [reply kore-rpc-types/src/Kore/JsonRpc/Types.hs:39]: Yeah, makes sense.

File: kore/src/Kore/Simplify/Ceil.hs [hunk 2]
```diff
@@ -361,3 +366,62 @@ makeSimplifiedCeil
...
+{- | This returns a set of all subterms that need a Ceil
+
+When the term's ceil implies the ceils of its subterms, this also adds
+the subterms' simplified ceils to the result. This is needed because the
+SMT solver can't infer a subterm's ceil from a term's ceil, so we
+have to provide that information.
+
+As an example, if we call @enumerateSubtermsNeedingCeil@ for @f(g(x))@,
+ the return value will be @{ceil(f(g(x))), ceil(g(x)))}@.
+-}
```
goodlyrottenapple, line 379: I think this description should be made a little clearer. we are not returning ceils, we are returning sets of all subterms which are assumed to be defined.

goodlyrottenapple (approved):
LGTM

jberthold (approved):

PR #3708, opened 2023-12-20
Concurrent  `kore-rpc` strikes back

geo2a:
This is a revert of the revert of https://github.com/runtimeverification/haskell-backend/pull/3704. Needs to be merged in sync with https://github.com/runtimeverification/hs-backend-booster/pull/427 and also a change to `pyk` to handle the new modules IDs.


tothtamas28:
Could you also update the [docs](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md) according to these changes, please?

test/rpc-server/add-module/add-again/README.md
```diff
@@ -0,0 +1,32 @@
...
+        |
+        V
+        e
+        |
+        V
+        f
+```
+
+Expected:
+* Module successfully added even though it has been added already, because the content is exactly the same.
```
jberthold, line 32: Probably the module would be added twice then?
How is `add-module/add-again` executed? The [test script](https://github.com/runtimeverification/haskell-backend/blob/32d61ab9da5468c38593a138b079a53fcc0fe158/test/rpc-server/runTests.py) does not mention it.

jberthold (approved):
Code LGTM 
(:+1: new code does what we wanted)

Pre-approving, but let's add `add-module/add-again` to the test script.

PR #3710, opened 2024-01-09
Fix dev shell by including secp256k1

goodlyrottenapple:
The dev shell should have the `libsecp256k1` library included, otherwise `cabal build` fails, when building the `secp256k1-haskell` library:

```bash
[_61] rejecting: secp256k1-haskell-0.6.1 (conflict: pkg-config package libsecp256k1-any, not found in the pkg-config database)
```


jberthold (approved):
Makes sense

PR #3717, opened 2024-02-08
Bump nixpkgs

Baltoli:


Baltoli:
@goodlyrottenapple to summarise, if this CI failure on macOS is workable to fix then we should have a go at adding a single source of Nixpkgs truth to all the repositories such that we do Nixpkgs updates as we do other dependency updates (i.e. a push to that repo triggers an update here, in the LLVM backend and in K itself).

flake.nix
```diff
@@ -1,7 +1,7 @@
 {
   description = "K Kore Language Haskell Backend";
   inputs = {
-    nixpkgs.url = "nixpkgs/nixos-23.05";
+    nixpkgs.url = "github:nixos/nixpkgs";
```
goodlyrottenapple, line 4: this may not be the best idea as its quite easy to accidentally bump nixpkgs on any update to the lock file, which has caused issues in the past.... we should definitely set up the common dependency before merging this (if everything works on this version of nixpkgs)

Baltoli: Noted; will revert my change! Thanks

flake.nix
```diff
@@ -123,7 +123,7 @@
         in pkgs.haskell-backend.pkgSet.shellFor {
           packages = pkgs.haskell-backend.localPkgsSelector;
           nativeBuildInputs =
-            [ pkgs.haskell.packages.ghc928.cabal-install pkgs.z3 ];
+            [ pkgs.haskell.packages.ghc928.cabal-install pkgs.z3 pkgs.libffi ];
```
goodlyrottenapple, line 126: do we still need this change??

Baltoli: Maybe not - have reverted to see if it was just the cache all along (which I suspect will be the case given the CI passes elsewhere)

goodlyrottenapple (approved):

PR #3721, opened 2024-02-15
Add module fix

goodlyrottenapple:
Fixes a corner case tested for in https://github.com/runtimeverification/pyk/pull/871 where

* we send module `M1` with the name `"m<hash of M2>"` and `"name-as-id": true`
* we send `M2`

In this case, the server should reject `M2`, but currently both requests succeed, with `M1` being overridden by `M2`

### DIFF
--- a/kore/src/Kore/JsonRpc.hs
+++ b/kore/src/Kore/JsonRpc.hs
@@ -527,6 +527,7 @@ respond serverState moduleName runSMT =
                         parseKoreModule "<add-module>" _module
             st@ServerState
                 { serializedModules
+                , receivedModules
                 , loadedDefinition = LoadedDefinition{indexedModules, definedNames, kFileLocations}
                 } <-
                 liftIO $ MVar.takeMVar serverState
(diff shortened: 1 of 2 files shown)


kore/src/Kore/JsonRpc.hs
```diff
@@ -734,6 +737,7 @@ respond serverState moduleName runSMT =
 
 data ServerState = ServerState
     { serializedModules :: Map.Map ModuleName SerializedDefinition
+    , receivedModules :: Map.Map ModuleName Text
```
jberthold, line 740: This is of course a very straightforward implementation to address the corner case.

I was thinking we could instead check the prior `indexedModules` for collisions: 
When we look up `newModule` in [lines 583](https://github.com/runtimeverification/haskell-backend/pull/3721/files#diff-e317d89df99d4965b340614d133c23673315aa243ae168373132bc5c9b8812f3R583-R586) we could also look it up in `indexedModules` (before the addition) as well. If we find a module `m<hash>` it must be identical to `newModule`.

However, this requires us to do part of the internalisation work  beforehand, and the comparison on the `IndexedModule`.

kore/src/Kore/JsonRpc.hs
```diff
@@ -561,6 +558,9 @@ respond serverState moduleName runSMT =
                                         ServerState
                                             { serializedModules =
                                                 Map.insert (coerce name) foundSerModule serializedModules
+                                            , receivedModules = case Map.lookup (coerce moduleHash) receivedModules of
+                                                Just recMod -> Map.insert (coerce name) recMod receivedModules
+                                                Nothing -> receivedModules
```
jberthold, line 563: I wonder if this `Nothing` case is actually possible... :thinking: if we have already found the module hash in the indexed modules we must have received it via `add-module` - unless someone loads a maliciously-named module on startup, another corner case :smiley_cat: )

jberthold (approved):
Code LGTM
Does it pass the new tests in pyk? (Maybe we could add 1-2 tests on our own turf for good measure)

kore/src/Kore/JsonRpc.hs
```diff
@@ -527,30 +527,27 @@ respond serverState moduleName runSMT =
...
-                                pure ()
+                    case Map.lookup (coerce name) receivedModules of
+                        -- if a different module was already added, throw error
+                        Just m | _module /= m -> throwError $ backendError DuplicateModuleName name
                         _ -> pure ()
 
+                -- Check for a corner case when we send module M1 with the name "m<hash of M2>"" and name-as-id: true
+                -- followed by adding M2. Should not happen in practice...
+                case Map.lookup (coerce moduleHash) receivedModules of
+                    Just m | _module /= m -> throwError $ backendError DuplicateModuleName name
```
jberthold, line 548: Of course we could return `coerce moduleHash` in the error here...

kore/src/Kore/JsonRpc.hs
goodlyrottenapple: I tried that and couldn't get it to work, because the internalisation code does a bunch of stuff like imports where things get muddy.... basically, I tried internalising the new module and then compare to the one that already exists, but somehow ended up with the previous module instead. Since this is a pretty unrealistic corner case, I really don't want to be spending time understanding/refactoring the internalisation to figure out where this was going wrong...

kore/src/Kore/JsonRpc.hs
goodlyrottenapple: yeah :/

kore/src/Kore/JsonRpc.hs
goodlyrottenapple: Yeah, I dunno. There doesn't seem to be a good answer there... probably better to return the module hash indeed, tho this error would be very confusing if it ever happened...again these are rather philosophical points...

goodlyrottenapple:
> Code LGTM Does it pass the new tests in pyk? (Maybe we could add 1-2 tests on our own turf for good measure)

Yes, the new tests are passing on this branch. I don't think adding more tests here is very useful, since the rpc testing infrastructure here is somewhat cumbersome. Perhaps I can add some more tests to the booster PR instead.

geo2a (approved):

PR #3732, opened 2024-02-27
Pin the hashable library to working version 1.4.2.0

### LINKED ISSUE #3730: Change to `simplify` endpoint behavior w.r.t. generating ceil conditions
In pyk, we use the `simplify` endpoint to generate initial ceil conditions to feed to the backends on proofs (https://github.com/runtimeverification/pyk/blob/b970e0cbc6c0df6872e6552af76239478a831b9b/src/pyk/kcfg/explore.py#L276).

In th elatest K update to pyk (https://github.com/runtimeverification/pyk/pull/913), we have a test failing because the `cterm_assume_defined` ends up computing different definedness constraints on the old backend on the version before vs. version in that PR.

Attached are bug reports of th einteraction collect on `master` (`ceil-gen-simplify-good.tar.gz`), and on the update-deps branch (`ceil-gen-simplify-bad.tar.gz`). The first request/response pair is what differs, in the `master` version we have 2 `#Ceil` conditions generated for the initial state, and in the `update-deps` version we only have 1.

I've also made a PR which isolates a more specific test of this: https://github.com/runtimeverification/pyk/pull/914. The K spec in question is this one: https://github.com/runtimeverification/pyk/blob/master/src/tests/integration/k-files/mini-kevm-spec.k, against this definition: https://github.com/runtimeverification/pyk/blob/master/src/tests/integration/k-files/mini-kevm.k.

### LINKED ISSUE #3731: Adding `symbol(ite)` to `#if_#then_#else_#fi` causes regression
Adding `symbol(ite)` to `#if_#then_#else_#fi` (in runtimeverification/k#4024, partially reverted since) caused regression in the following `pyk` test:

https://github.com/runtimeverification/pyk/blob/e1b9407afc8cce2be00ff83101e29b2dd16844eb/src/tests/integration/proof/test_mini_kevm.py#L51

Bug reports:
* Original behavior: [if_then_else_fi.zip](https://github.com/runtimeverification/haskell-backend/files/14403707/if_then_else_fi.zip)
* Faulty behavior: [ite.zip](https://github.com/runtimeverification/haskell-backend/files/14403708/ite.zip)

Setting the original symbol, i.e.

```
symbol(#if_#then_#else_#fi_K-EQUAL-SYNTAX_Sort_Bool_Sort_Sort)
```

does not produce the issue. This indicates that the symbol name might be hardcoded somewhere.

The issue is very similar to #3730.

jberthold:
Fixes #3730
Fixes #3731

After upgrading GHC and stack resolver to newer versions, which bring in `hashable-1.4.3.0`, downstream upgrade jobs started failing because of missing `#Ceil` conditions.

Systematic testing showed that the upgrade to `hashable` is causing these omissions. A manual downgrade of the dependency to 1.4.2.0 (as before) may fix the problems.


geo2a:
I can confirm that this indeed fixes the `pyk` test reported in #3730.

geo2a (approved):

PR #3733, opened 2024-02-27
Remove `test/regression-c`

geo2a:
I propose to remove this integrations test for several reasons:
* it is inscrutable, as there's no C input program (only Kore) and no source K semantics (only a compiled Kore definition)
* It appears we've de-prioritized supporting symbolic execution for the C semanatics, and once we would like to attempt that again, we will likely use `kore-rpc-booster`, rather than `kore-exec`
* The output of this test seems to be flaky: I had to update the output to merge #3732. The new output looked "OK", but I did not look very carefully, as it is, again, inscrutable Kore
* adding this test in the first place was a temporary arrangement that seems to have overstayed its welcome (see https://github.com/runtimeverification/haskell-backend/pull/3274)
* it is relatively slow to run, taking minutes in my machine


jberthold (approved):
Concur with the arguments stated for the removal.
It would be great if we could work with the C semantics using booster but at the moment this test is not of great value for us.

PR #3735, opened 2024-02-28
Tidy-up integration tests

geo2a:
* add `-Wno unused-var -Wno missing-syntax-module` to calls to `kompile` and `kprove`, silencing the unused variables and missing syntax module warnings
* replace the `functional` attribute with`total`
* Replace `.` with `.K`
* add `no-evaluators` attribute to function symbols with no evaluation rules


jberthold (approved):
Great clean-up  :bucket: :broom: 
I found some remaining warnings about using `latex(..)` in the test output (files `test/smc/smc.k` and `test/imp/imp.k`), maybe we can still remove those. Or just merge as-is.

geo2a:
Yeah, I'll clean these up too!

PR #3737, opened 2024-03-01
Add `get-model` and `simplify` tests that Z3 cannot decide

geo2a:
* add an integration test for when Z3 returns `unknown` during `get-model`
* add an integration test for when Z3 returns `unknown` during `simplify`
* add example error response into the docs

## Summary by CodeRabbit

- **Documentation**
	- Added an example illustrating an error response in the JSON-RPC Server API documentation.
	- Introduced documentation on applying `get-model` to non-linear arithmetic predicates, including example inputs and responses.
	- Documented the simplification process for predicates with non-linear arithmetic, detailing error responses from the SMT solver.

- **Tests**
	- Added new test cases for the RPC server, covering scenarios such as unknown results from `get-model` requests and specific SMT solver errors.


jberthold (approved):
LGTM (what does a rabbit know anyway? :rabbit: )

PR #3741, opened 2024-03-15
Remove uses of `klabel{}(_)` attribute for special-casing symbols

Baltoli:
See #3742 for full context; this PR adds a backwards-compatible check for `symbol'Kywd` as well as `klabel` for the pseudo-builtin sorts `Endianness` and `Signedness`. When this is merged, I'll be able to finish working on https://github.com/runtimeverification/k/pull/4045. Then, when that PR lands, the frontend will no longer emit `klabel(_)` to KORE, and we can safely remove the backwards-compatible part of this PR.

The integration test suite's checked-in KORE files are all updated to use the new attribute rather than the old one, but the tests that use K to compile fresh KORE will still be using `klabel(_)` until the dependency update job induced by the changes above goes through.

I have verified manually that building K with this HB commit allows the failing integration test in #3742 to pass.

Thanks @jberthold for the suggestions!

~~The implementation here would ideally be made backwards-compatible; my version induces a dependency loop (hence failing tests, it would need an update to the K dependency to go through first for the integration test suite to pass).~~


kore/src/Kore/Builtin/Verifiers.hs
```diff
@@ -181,8 +181,8 @@ lookupApplicationVerifier symbol verifiers = do
             . symbolAttributes
     getKlabel =
         fmap KlabelSymbolKey
-            . Attribute.Symbol.getKlabel
-            . Attribute.Symbol.klabel
+            . Attribute.Symbol.getSymbol
+            . Attribute.Symbol.symbolKywd
```
jberthold, line 185: This should probably be renamed to `getSymbol'Kywd` to avoid confusion

jberthold: Probably the line above that says
```
    key <- getHook symbol <|> getKlabel symbol
```
could be changed to 
```
    key <- getHook symbol <|> getKlabel symbol <|> getSymbol'Kywd symbol
```
to achieve backward compatibility.
Needs to be tested but could avoid creating a broken release.

jberthold (approved):
LGTM

PR #3745, opened 2024-03-18
Prune the remainder branch early if its condition is unsatisfiable

### LINKED ISSUE #3738: Invalid function evaluation
The ~Booster~  Haskell backend applies a simplification rule that should not be applied because the `requires` clause is not true. This causes an infinite evaluation loop. ~Actually, it's likely that the Haskell backend does that, but I'm just calling the Booster backend and seeing the results.~

The longer story (reproduction details at the end):

While doing property verification for Elrond's pair contract with the Kasmer tool, the execution reaches a point where it has to evaluate a certain expression (hunting through the equation evaluation logs is time consuming, so I didn't try to identify it). When evaluating this expression, the backend tries to evaluate
```
#getRange(
    SBChunk(#empty(expression-1)) SBChunk(#bytes(b"\x1c\x00\x00\x00")) SBChunk(#bytes(b"\x01")),
    expression-1,
    4
)
```
To do that, it tries to apply this rule ( https://github.com/runtimeverification/mx-backend/blob/pair-lemmas/kmxwasm/k-src/mx-lemmas.md?plain=1#L122-L125 ):
```
  rule #getRange(SBChunk(A:SBItem) B:SparseBytes, Start, Width)
      => #splitGetRange(SBChunk(A) B, Start, size(SBChunk(A)) -Int Start, Start +Int Width -Int size(SBChunk(A)))
    requires Start <Int size(SBChunk(A)) andBool size(SBChunk(A)) <Int Start +Int Width
```
It should not be able to apply this rule because `SBChunk(#empty(expression-1))` evaluates to `expression-1`, and it is `false` that `expression-1 <Int expression-1`.

However, `expression-1` is actually `expression-2 +Int 9974`, so when evaluating this, the backend tries to apply various simplification rules, including this one (https://github.com/runtimeverification/wasm-semantics/blob/master/kwasm-lemmas.md?plain=1#L289):
```
    rule VAL1 +Int VAL2                     <Int MAX => true requires VAL1 <Int MAX andBool VAL2 <Int MAX andBool #distinctBits(VAL1, VAL2) [simplification]
```

The backend quickly evaluates `expression-2 <Int expression-2 +Int 9974` and `9974 <Int expression-2 +Int 9974` to true. However, this rule should not apply because the `#distinctBits(VAL1, VAL2)` function evaluates to true only in a very specific case (see this: #distinctBits(VAL1, VAL2) ). Indeed, the logs show that the backend attempts to evaluate `#distinctBits(expression-2 +Int 9974, expression-2 +Int 9974)` and fails, because the simplification rule does not apply.

Regardless, the backend happily applies the rule above and evaluates `expression-1 <Int expression-1` to true. Then it applies the `#getRange` rule above and produces `#splitGetRange(bytes-stuff, expression-1, 0, 4)` whose evaluation loops because the third arg should never be `0`.

Versions:
```
$ kompile --version
K version:    v6.1.82-0-g8c99ee7740
Build date:   Jo feb 22 23:25:11 EET 2024

$ kore-rpc-booster --version
hs-backend-booster version:
  revision:     6067fee77da53ac357ce5d17ad34e4639070804a
  branch:       HEAD
  last commit:  Tue Jan 16 09:27:16 2024 -0700
```

Bug report:
[weird-function-eval.tar.gz](https://github.com/runtimeverification/hs-backend-booster/files/14380439/weird-function-eval.tar.gz)

geo2a:
Fixes #3738

**TL;DR** prune the rule remainder early with a direct call to the SMT solver on the remainder predicate, rather then constructing the remainder configuration and simplifying it.

----------------------

With this change, upon computing a remainder condition after applying rewrite rules, we immediately ask the SMT solver to decide the satidfiability of this condition. If the solver returns unsat, we prune the remainder branch.

Previously, we would add the remainder predicate to the path condition and construct the remainder branch from the current configuration. We will then call the simplifier to evaluate functions in the configuration. It turns out, we may start evaluating functions under a false assumption, leading to pathological cases of applying simplifications which are invalid (see #3738).


kore/src/Kore/Rewrite/RewriteStep.hs
```diff
@@ -290,16 +294,36 @@ finalizeRulesParallel
...
+                    -- remainder condition is UNSAT: we prune the remainder branch early to avoid
+                    -- jumping into the pit of function evaluation in the configuration under the
+                    -- contradictory condition (the unsatisfiable remainder)
+                    Just False ->
+                        return
+                            Step.Results
+                                { results = Seq.fromList results
+                                , remainders = mempty
+                                }
+                    -- remainder condition is UNKNOWN: TODO: what happens here?
```
jberthold, line 313: With the current code for `evalPredicate`, the result will never be `Just True`.  When the solver returns `Unknown` the result is `Nothing` here, the same as for `Sat` from the solver.
We could instead stop execution by passing `ErrorDecidePredicateUnknown` above. If the precondition is already an `Unknown` problem for the solver we will likely hit that error later anyway.

geo2a:
KEVM again, at the last commit, with 1 worker:

```
| Test                                     | georgy-use-HB-with-remainder-fix time | master-8c26ad4 time | (georgy-use-HB-with-remainder-fix/master-8c26ad4) time
|------------------------------------------|---------------------------------------|---------------------|--------------------------------------------------------
| bihu/forwardToHotWallet-failure-4-spec.k | 96.84                                 | 100.36              | 0.9649262654444002
| erc20/ds/transferFrom-failure-1-b-spec.k | 139.3                                 | 134.0               | 1.0395522388059701
| erc20/ds/transferFrom-failure-1-a-spec.k | 93.46                                 | 89.61               | 1.042963954915746
| erc20/ds/transferFrom-success-1-spec.k   | 97.21                                 | 92.46               | 1.0513735669478694
| erc20/ds/transferFrom-failure-2-c-spec.k | 56.87                                 | 53.8                | 1.0570631970260222
| TOTAL                                    | 483.68                                | 470.23              | 1.0286030240520596
```

test/rpc-server/execute/branching-prune-unsat-remainder/test.k
```diff
@@ -0,0 +1,31 @@
...
+  rule a(I, _) => b requires I <=Int 0
+  rule a(I, _) => b requires notBool (I <=Int 0)
+
+  syntax Bool ::= f(Int)  [function, total]
+  rule f(0) => true
+  rule f(I) => loop(I) requires I =/=Int 0
+
+  syntax Bool ::= g(Int)  [function]
+
+  // this simplification shooed never be applied, as it requires false.
```
virgil-serbanuta, line 16: s/shooed/should/?

jberthold (approved):

goodlyrottenapple (approved):

PR #3748, opened 2024-03-20
Refactor rpc error types and add smt unknown reason

### LINKED ISSUE #3747: Reporting SMT timeouts over JsonRpc responses
In its current state pyk receives a single error response related to SMT Solver errors (see [here](https://github.com/runtimeverification/pyk/blob/baa8f77ef8db366d1d8024c5500b648e9451f79a/src/pyk/kore/rpc.py#L920))

It would be very useful if it is possible to distinguish timeout errors from the rest and communicate it over JsonRpc calls either via a separate response code or as a part of the response message.

goodlyrottenapple:
Fixes #3747


tothtamas28 (commented):
The schema change as documented looks good to me.

kore-rpc-types/src/Kore/JsonRpc/Error.hs
```diff
@@ -28,25 +43,70 @@ unsupportedOption = ErrorObj "Unsupported option" (-32003) . toJSON
...
+        (ToJSON)
+        via CustomJSON '[SumUntaggedValue] JsonRpcBackendError
+
+class CN (f :: Type -> Type) where
+    constructorCodeAndName' :: Int -> f x -> (Int, String)
+    countConstructors :: proxy f -> Int
+
+instance CN f => CN (D1 c f) where
+    constructorCodeAndName' n (M1 x) = constructorCodeAndName' n x
+    countConstructors _ = countConstructors @f undefined
```
goodlyrottenapple, line 92: not sure what the latest consensus is on proxy values...

geo2a (approved):

PR #3756, opened 2024-04-04
Booster-532: align fourmolu and hlint versions

jberthold:
Preparing https://github.com/runtimeverification/hs-backend-booster/issues/532

* Use `fourmolu 0.14.0.0`, reformatting the code as required (large change)
* use `hlint-3.6.1` to check code style  (adjusting code and config as required)
* provide `hlint-3.6.1` in nix derivation `#fourmolu`


jberthold:
The changes to the trailing `$` look horrible! And for some reason this only happens within the context of the project, not with isolated files outside of it (with the same configuration file available)

jberthold:
For posterity: https://github.com/fourmolu/fourmolu#language-extensions-dependencies-and-fixities explains (in brief) that `fourmolu` rearranges and re-indents operator chains if it does not know their fixities (because of reexporting and `NoImplicitPrelude`) - I adjusted this in the `fourmolu.yaml`

geo2a (approved):

flake.nix
```diff
@@ -239,9 +238,9 @@
           mkShell {
             nativeBuildInputs = [
               (haskell.lib.justStaticExecutables
-                # (ghcVersion pkgs).fourmolu
-                (import nixpkgs2305 { inherit system; }).haskellPackages.fourmolu_0_12_0_0
```
goodlyrottenapple, line 243: the import of `nixpkgs2305` should be removed entirely now that we no longer need the old versions

goodlyrottenapple (approved):

Issue #3764, opened 2024-04-10
Branch returned although one of the next states simplifies to #bottom


jberthold:
On slack, [Petar reported  a problem](https://runtimeverification.slack.com/archives/CC360GUTG/p1712247244243789) where  a branch result returned by (both `kore-rpc` and) `kore-rpc-booster` contains two `next-states`, but one of them simplifies to `#Bottom` with an isolated `simplify` request.
This is caused by `kore-rpc` execution not discovering the `#Bottom`-ness of this state, otherwise execution would continue.

Observed behaviour: 
* Execution stops and a `Branching` result is returned. One of the states in `next-states` simplifies to `#Bottom` in a subsequent `simplify` request.
Expected behaviour: 
* Execution should continue with the non-bottom alternative of the branch.

- [ ] to protect against this behaviour in `kore-rpc-booster`, modify the `postExecSimplify` mechanism to detect this case and continue execution.
- [ ] reproduce this result with `kore-rpc` and investigate why the bottom state is not found

jberthold:
There were once known issues with spurious branches at depth 0 in `kore-rpc` earlier, so this behaviour should be tested this with a `noop` step before the branch, too. The reason appears to be that the simplifier is not called for branch results at depth 0. This should be investigated.

jberthold:
I have confirmed with tests that in this case, it makes no difference whether `kore-rpc` has performed an execution step before the bogus `Branching` result is obtained.

PR #3765, opened 2024-04-11
3764 continue execution if post-exec-simplify reveals a bogus branch

jberthold:
Part of #3764 (work-around in booster, not fixing kore)

When `kore-rpc` returns a `Branching` result where all but one of the branches are actually `#Bottom` (when simplified), execution in `kore-rpc-booster` should continue.
Such a case was observed in the wild, see #3764


jberthold:
Going to add an integration test (either the reported case or a small example if I can find one), but the code can already be reviewed

booster/test/rpc-integration/test-issue3764-vacuous-branch/README.md
```diff
@@ -0,0 +1,19 @@
...
+The files for this test are a slight reduction of the original input, but the definition has been left intact for further analysis.
+
+* `state-branch-in-zero` 
+  - The `k` cell contains the `JUMPI` instruction. Booster aborts instantly on the uncertain condition (containing variables).
+  - `kore-rpc` returns `Branching`, splitting on the jump condition
+  - The second branch is then pruned (by post-exec simplification), and execution continues for one more step (booster returns `DepthBound`)
+* `state-branch-after-one`
+  - The `k` cell contains `#gas[_, _] ~> JUMPI ... ~> CONTINUATION`, with the `JUMPI` instruction from before.
+  - In one step, the `#gas[_, _]` instruction is removed since `useGas` contains `false`
+  - After that, the same as in `state-branch-in-zero` happens.
```
geo2a, line 16: Interesting! So it does not matter in this case that Kore is branching at depth 0?

geo2a: Ah no, we rewrite to remove `#gas` in Booster probably.

jberthold: Yes, and yes :)  We rewrite the `#gas` in Booster (and every fall-back to `kore` is only a single step anyway). *but* I ran the same request against `kore-rpc` and got a `branching` result.

jberthold:
Need to bump up the delay in `runDirectoryTest.sh` because the server is starting so slowly in the new test...

booster/tools/booster/Proxy.hs
```diff
@@ -384,43 +424,38 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
...
-                                                                    , boosterResult.logs
-                                                                    , koreResult.logs
-                                                                    , fallbackLog
-                                                                    ]
-                                                            }
+                                            -- perform post-exec simplification.
+                                            -- NB This has been found to make a difference,
+                                            -- even for results produced by legacy-kore.
+                                            postExecResult <-
+                                                simplifyExecResult logSettings r._module def koreResult
```
geo2a, line 438: We *may* want to get away without the simplification if we made more a step in Kore (i.e. `Brandhing` at depth 1, rather than 0). We should consider trying that in case we see a degradation in performance due to the additional simplify requests.

booster/tools/booster/Proxy.hs
```diff
@@ -429,14 +464,28 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
...
+                        simplifyExecResult logSettings r._module def boosterResult
+                    case postExecResult of
+                        Left (nextState, newLogs) ->
+                            executionLoop
+                                logSettings
+                                def
+                                ( currentDepth + boosterResult.depth
+                                , time + bTime
+                                , koreTime
+                                , combineLogs $ rpcLogs : boosterResult.logs : newLogs
```
geo2a, line 478: If Kore gave as new simplification logs here, we would not emit them at `-l SimplifyJson`. If we want to be faithful, we need to replicate the code on lines 406-414:
```
                                    -- Kore may have produced simplification logs during rewriting. If so,
                                    -- output them Kore at "SimplifyJson" level, effectively
                                    -- appending them to Booster's traces. Erase terms from the traces.
                                    when (isJust koreResult.logs) $ do
                                        outputLogsAtLevel (Log.LevelOther "SimplifyJson")
                                            . map RPCLog.logEntryEraseTerms
                                            . filter isSimplificationLogEntry
                                            . fromJust $ koreResult.logs
```

to avoid code duplication, we should factor-out this into a function and put it into `simplifyExecRequest`.

However, I do not think that this is absolutely necessary to include this change in this PR, as here we are more concerned with correctness than the completeness of the simplification logs. As we agreed to strip the RPC simplification logs anyway, we will deal with this as part of https://github.com/runtimeverification/haskell-backend/issues/3801.

geo2a (approved):
We need to run the performance scripts on this one, but we cannot do that just yet as I understand? Considering this, and the fact that this is a correctness change, I'm inclined to approve without seeing the numbers.

booster/tools/booster/Proxy.hs
jberthold: I'll just quickly add those lines to `simplifyExecResult`, that is quick to do.

jberthold:
### KEVM performance test (using unmerged update branch) without any issues

| Test                                                             | 3764-can-continue-execution-after-postExecSimplify time | master-5997ca3b4 time | (3764-can-continue-execution-after-postExecSimplify/master-5997ca3b4) time 
|------------------------------------------------------------------|---------------------------------------------------------|-----------------------|----------------------------------------------------------------------------
| kontrol/test-expectreverttest-test_expectrevert_message-0-spec.k | 55.13                                                   | 60.38                 | 0.9130506790327924                                                         
| bihu/functional-spec.k                                           | 109.07                                                  | 113.84                | 0.9580990864371046                                                         
| kontrol/test-expectreverttest-test_expectrevert_bytes4-0-spec.k  | 78.96                                                   | 73.35                 | 1.076482617586912                                                          
| erc20/ds/allowance-spec.k                                        | 46.75                                                   | 42.85                 | 1.0910151691948657                                                         
| TOTAL                                                            | 289.91                                      | 290.42                | 0.9982439225948625

booster/tools/booster/Proxy.hs
jberthold: I checked the code once again today, and I believe `simplifyExecResult` calls `simplifyExecuteState`, which calls `handleSimplify`, which already outputs the simplification logs from `kore` at `SimplifyJson` level. So unless there is a code path that escaped my attention, I don't think we need to add any logging here.

PR #3807, opened 2024-04-12
Cache the result of building an LLVM lib in `run-with-tarball.sh`

goodlyrottenapple:
This change hashes the LLVM `definition.kore` file, extracted from a tarball and checks if we already have a compiled LLVM library cached from before. This is useful when re-running a full set of tarballs several times, but also speeds up things on a single run as many of the definitions within a set of tests are the same.


scripts/run-with-tarball.sh
```diff
@@ -93,11 +80,33 @@ if [ -z "${LLVM_LIB}" ]; then
...
+        #generate matching data
+        (cd $TEMPD && mkdir -p dt && llvm-kompile-matching llvm-definition.kore qbaL ./dt 1/2)
+        # find library dependencies and source files
+        for lib in libff libcryptopp blake2; do
+            LIBFILE=$(find ${PLUGIN_DIR} -name "${lib}.a" | head -1)
+            [ -z "$LIBFILE" ] && (echo "[Error] Unable to locate ${lib}.a"; exit 1)
+            PLUGIN_LIBS+="$LIBFILE "
+            PLUGIN_INCLUDE+="-I$(dirname $LIBFILE)/../include "
+        done
+        PLUGIN_CPP="${PLUGIN_DIR}/include/plugin-c/blake2.cpp ${PLUGIN_DIR}/include/plugin-c/crypto.cpp ${PLUGIN_DIR}/include/plugin-c/plugin_util.cpp"
```
jberthold, line 98: Just noticed the left-over `blake2` source file here...
```suggestion
        PLUGIN_CPP="${PLUGIN_DIR}/include/plugin-c/crypto.cpp ${PLUGIN_DIR}/include/plugin-c/plugin_util.cpp"
```

jberthold (approved):
:+1: good idea!

Issue #3800, opened 2024-04-11
Clean-up work after merging repositories


jberthold:
- [x] avoid running integration tests for kore-exec when nothing changed https://github.com/runtimeverification/haskell-backend/pull/3808
    - github path filter action for kore and kore-rpc-types (and tests/kore-exec, of course), as well as all general build setup
- [x] test performance scripts, adjust .gitignore to ignore artefacts
- [ ]  remove generate-json and the update-regression-tests scripts ?
- [x]  adjust the `README` files https://github.com/runtimeverification/haskell-backend/pull/3812
    - remove booster README
    - add information about booster to top-level README and update it
- [ ]  merge documentation folders
     - `design-decisions` -> `docs/design-decisions` ?
     - `booster/docs` -> `docs` (with renamed files) ?

- [x] move secondary kore binaries to dev-tools package https://github.com/runtimeverification/haskell-backend/pull/3811
- [ ]  rework directory structure for tests
    - tests/kore-exec
    - tests/rpc-integration
        - merging old and new tests
        - using kore-rpc-client and script
    - tests/interactive (for kore-repl), if we keep those at all

jberthold:
No further work planned because not a priority at all.

PR #3808, opened 2024-04-15
use path filters to determine change set and select tests to run

jberthold:
Part of  #3800

* Determines change sets using simple directory / file filters
* Skips running kore integration tests if no changes were detected


geo2a (approved):
Looks good to me!

PR #3809, opened 2024-04-15
Do not include simplification RPC logs into the Execute/Simplify response

### LINKED ISSUE #3801: Discourage simplification logging in responses
EDIT by @geo2a:

Responses from `kore-rpc-booster` will include simplification logs produced by Kore if `-l SimplifyJson` is given.

This happens because we log the simplifications from Kore at the level of `Proxy.hs`, and we get the information about those simplifications by requesting it over RPC in the fallback requests.

As we've decided to fade-out the RPC logs for simplifications, we should not return the logs in the responses.

geo2a:
Fixes #3801 

Changes to how the internal simplify requests are issued:
- do not ask for simplification RPC logs in proxy's internal requests. For the purposes of `-l SimplifyJson` logging, the request to Kore are still given the flags if this log level is enabled.
- when receiving simplify responses to internal request from Kore, filter-out the simplify logs to not leak them to the client.


booster/test/rpc-integration/runDirectoryTest.sh
```diff
@@ -24,8 +24,8 @@
 #   SERVER_OPTS: additional options to pass to the SERVER
 #                  (default: none)
 #   CONNECTION_ATTEMPTS: attempts to establish a connection with the server
-#                        after sleepeng for 1 second
-#                          (default: 15)
+#                        after sleepeng for 2 seconds
+#                          (default: 20)
```
ehildenb, line 28:
```suggestion
#                        after sleeping for 2 seconds
#                          (default: 20)
```

booster/tools/booster/Proxy.hs
```diff
@@ -376,7 +378,7 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
                                                 [ rpcLogs
                                                 , boosterResult.logs
                                                 , boosterStateSimplificationLogs
```
jberthold, line 380: Should we also filter these two? Or is the goal just to remove the `kore` simplification logs?
`boosterStateSimplificationLogs` should be empty because `simplifyExecuteState` disables those logs, but `boosterResult.logs` will still contain the ones that were produced during the execute request.

jberthold (commented):
LGTM for the purpose of disabling `kore-rpc` simplification logging in internal simplification requests during execute. It is a bit unclear how much of the simplification logging should be removed.

We still need simplification logging options from the `Simplify` request to support `SimplifyJson` so they will probably stay around. At the same time, we don't seem to have documentation for these logs in `Simplify` responses.

booster/tools/booster/Proxy.hs
geo2a: Right, `boosterResult.logs` will actually contain simplification logs, so we should modify the Booster to no emit them in the first place. Let's indeed filter them out here for now...

jberthold (approved):

PR #3811, opened 2024-04-15
Clean-up the executables in package `kore` 

geo2a:
Part of #3800 

- remove `kore-check-functions`, `kore-format` and `kore-simplify`
- Move `kore-parser` and `kore-match-disjunction` to the `hs-backend-booster-dev-tools` package

Before merging, we should run the integration tests of K to make sure nothing is broken.


include.mk
```diff
@@ -32,15 +32,6 @@ STACK = stack --allow-different-user
...
 
-ifeq ($(strip $(NIX)),)
-KORE_PARSER = $(BUILD_DIR)/kore/bin/kore-parser
-else
-KORE_PARSER = kore-parser
-endif
-KORE_PARSER_OPTS =
-export KORE_PARSER
-export KORE_PARSER_OPTS
-
```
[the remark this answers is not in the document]
geo2a: Oops!

dev-tools/package.yaml
```diff
@@ -198,10 +198,6 @@ executables:
       - kore
       - optparse-applicative
       - transformers
-    ghc-options:
-    - -rtsopts
-    - -threaded
-    - -with-rtsopts=-N
   kore-parser:
```
jberthold, line 201: I would always leave the `-rtsopts` active (without it, it is not allowed to pass `+RTS` options at all)
But they can of course be added quickly in a local build if need be...

geo2a: Sounds good, I've brought them back.

goodlyrottenapple (approved):

PR #3814, opened 2024-04-16
Remove `web/` and `docker/`

geo2a:
Part of #3800


jberthold (approved):
:fire: :fire: :fire:  LGTM :put_litter_in_its_place:

By chance, I stumbled over https://github.com/runtimeverification/haskell-backend/deployments today and saw that the pages were last deployed in 2022

PR #3816, opened 2024-04-17
Run integration tests using nix runner matrix

jberthold:
We are currently not running integration tests on Mac OS.
* many tests need the `--no-haskell-binary` option to run reliably on Mac OS
* For the `rpc-integration` tests, we need to increase the connection attempt counter
* We want to run the tests on an ARM based Mac to get acceptable performance.


jberthold:
Integration tests on Mac M1 appear to be hanging. On hold for now, to investigate at a later time.

PR #3818, opened 2024-04-17
Implement logging of applied equations as JSON in Kore

### LINKED ISSUE #3817: Kore throws away simplification traces when processing '"execute"' requests
The `rpcExec` function, which is the entry-point to Kore's rewriter, throws away the simplification traces produced by the simplifier: https://github.com/runtimeverification/haskell-backend/blob/master/kore/src/Kore/Exec.hs#L496, which means we would actually never have simplification logs in an Execute response form Kore.

geo2a:
Fixes #3817 
Fixes https://github.com/runtimeverification/haskell-backend/issues/3815

Changes to Kore:
- change the `Kore.JsonRpc.ServerState` type to contain an optional handle to write the simplification logs into
- change the `Kore.Simplify.SimplifyEnv` type to hold an optional handle as well, instead of a boolean flag.
- in `Kore.Equation.Application.applyEquation`, if the simplified contains the handle, output the trace as JSON into it. Include only the unique ID, not the terms

Changes to `Proxy.hs` and `dev-tools/kore-rpc-dev/Server.hs`:
- if `-l SimplifyJson` is set, thread the handle to `stderr`/log file through to Kore's `ServerState`
- in `Proxy.hs`, do not modify the request parameters when falling back to Kore or making internal simplify requests. Do not output logs received form Kore at `-l SimplifJson` anymore: Kore will now do it itself

Note that the behavior of `kore-rpc-booster -l SimplifyJson` is tested by the `booster/test/rpc-intergration/test-log-simplify-json` integration test, which compares the log to a golden file. The changes in this PR do not break this test.


geo2a:
Subsumed by https://github.com/runtimeverification/haskell-backend/pull/3824

PR #3822, opened 2024-04-21
MX Performance and booster analysis script tweaks

jberthold:


scripts/booster-analysis.sh
```diff
@@ -18,11 +18,11 @@ nix_shell() {
 
 nix_shell "cabal build kore-rpc-booster"
 
-export SERVER=$(nix_shell "cabal exec which kore-rpc-booster")
+export SERVER=$(nix_shell "cabal exec which kore-rpc-booster" | tail -1)
```
jberthold, line 21: Cuts off the output of `hpack` from the shell hook

PR #3824, opened 2024-04-22
Add JSON logging to Kore

geo2a:
Fixes https://github.com/runtimeverification/haskell-backend/issues/3817
Fixes https://github.com/runtimeverification/haskell-backend/issues/3815

* the `Entry` type class gets a new `oneLineJson` method. Log entries will now need to implement this method if they are to be logged as JSON. The default implementation if just the type constructor name.
* The `DebugAttemptEquation` log entry gets an a `oneLineJson` implementation that serialised the successful/failed equation applications in the same form as the Booster, i.e. following the RPC log format. The terms are not included to keep the log size manageable.
* The `LogSomeAction` constructor of the `LogType` (a filed of `KoreLogOptions`) is repurposed. Instead of holding one log action, it now holds two, plus a predicate that specifies which log entries to log using which action. Then, the `makeKoreLogger1` function uses the selector function to filter the logs, render them with the appropriate rendered (json or standard) and log them using the appropriate log action.
* `booster/Server.hs` is modified to initialize `LogSomeAction` with the appropriate log actions. If `-l SimplifyJson` log level is activated, we enable `DebugAttemptEquation` and `DebugApplyEquation` (as before). If `--simplification-log-file` is passed too, then Kore's simplification will be redirected there, otherwise they will end-up on stderr. Booster's simplilfications behave in the same way.
* Finally, `booster/Proxy.hs` does not modify the request parameters when falling back to Kore or making internal simplify requests. We do not output logs received form Kore at `-l SimplifJson` anymore: Kore will now do it itself.

## Example

Put simplification logs into a file
```
kore-rpc-booster def.kore --module DEF --server-port 12341 -l SimplifyJson --simplification-log-file logs.json
```
Or simply to `stderr`:

```
kore-rpc-booster def.kore --module DEF --server-port 12341 -l SimplifyJson
```
where there will be prepended with `[SimplifyJson]`


kore/src/Kore/Log.hs
```diff
@@ -321,6 +374,84 @@ makeKoreLogger exeName startTime timestampSwitch koreLogFormat logActionText =
...
+            \case
+                [] -> []
+                xs -> ("Context" <> Pretty.colon) : (indent <$> mkContext xs)
+
+        mkContext =
+            \case
+                [] -> []
+                [(doc, typeName)] ->
+                    [Pretty.hsep [Pretty.parens typeName, doc]]
+                (doc, typeName) : xs -> (Pretty.hsep [Pretty.parens typeName, doc]) : (indent <$> (mkContext xs))
```
goodlyrottenapple, line 451: i guess we will want to tweak this in a separate PR to resemble the booster contexts?

kore/src/Kore/Equation/DebugEquation.hs
```diff
@@ -276,6 +272,39 @@ instance Entry DebugAttemptEquation where
...
+             in JSON.object
+                    [ "tag" JSON..= ("simplification" :: Text)
+                    , "result" JSON..= resultInfo
+                    ]
+          where
+            failureDescription :: AttemptEquationError RewritingVariableName -> Text
+            failureDescription err = shortenText . Pretty.renderText . Pretty.layoutOneLine . Pretty.pretty $ err
+
+            shortenText :: Text -> Text
+            shortenText msg = Text.take 200 msg <> ("...truncated" :: Text)
```
goodlyrottenapple, line 304: should we have this here? I notice that we still emit stuff like `/* T Fn D Spa */` in kore terms. Perhaps once those are removed, we can remove this truncation without things blowing up too much?

goodlyrottenapple (approved):
left a couple of comments, looks good otherwise

kore/src/Kore/Equation/DebugEquation.hs
geo2a: I added the truncation because the golden file got to 140mb... though that was a bit too much to put into git.

kore/src/Kore/Log.hs
geo2a: Yes, certainly.

geo2a:
Thanks @goodlyrottenapple! I'll be merging this then. 

As we've discussed, we should come up with a minimal test for the JSON logging. I've created an issue https://github.com/runtimeverification/haskell-backend/issues/3830.

kore/src/Kore/Equation/DebugEquation.hs
geo2a: https://github.com/runtimeverification/haskell-backend/pull/3823 did not improve matters much unfortunately. I've settled on a slightly longer cutoff of 500 characters.

PR #3827, opened 2024-04-23
Do not run unit test as part of integration tests

geo2a:
I've noticed we are running unit test as part of the "'Nix / Integration'" job, but we also have a dedicated "'Nix / Unit Tests" job that runs them in exactly the same way, so we may as well skip them in the integration tests job.


jberthold:
This was in preparation to replace the `Nix / Unit tests` job by one that runs both integration tests and unit tests on all platforms we aim to support, see #3816 .
This is currently on hold because of an issue on Mac M1 that I cannot debug locally.
Happy to remove the unit test run step for the moment.

jberthold (approved):

geo2a:
Ah, right, sorry to thread on your toes @jberthold.

PR #3835, opened 2024-04-29
Adjust name for =/=K hook 

jberthold:
The name used in the Haskell backend was `KEQUAL.neq` but [the K `domains.md` file uses `KEQUAL.ne`](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md?plain=1#L2287).


docs/manual/DEVELOPER_MANUAL.md
```diff
@@ -100,7 +100,7 @@
             8. [SET.list2set](#implementation-hooks-set-list2set)
         7. [KEQUAL](#implementation-hooks-kequal)
             1. [KEQUAL.eq](#implementation-hooks-kequal-eq)
-            2. [KEQUAL.neq](#implementation-hooks-kequal-neq)
+            2. [KEQUAL.ne](#implementation-hooks-kequal-neq)
```
jberthold, line 103: I think we should probably delete this entire file... the hooks section is incomplete, and there are quite a few other sections that (incompletely) repeat content from other files, e.g. the textual `kore` format.

jberthold:
> good spot. tho I suppose this points to the hook never being used, because we would have surely seen a hook unimplemented error.

True, we would have seen this as a missing hook error before otherwise. Maybe the simplifications have historically been evolving to avoid this operation (transforming `_ =/=K _` to `#Not (#Equals _ _)`, or maybe they were there from the beginning and `ne` was added without need. From my quick `git blame` investigation, this hook goes back at least 9 years.

PR #3845, opened 2024-05-02
Reuse kompiled `contracts.k` and `foundry.k` in Kontrol performance test

geo2a:
This PR attempts to reduce the influence of the K compiler processes on the Kontrol performance test. To achieve that, we compile everything in-advance, ensuring that there will be no calls to `kompile` when running the actual proof test:

- run the single `test_foundy_kompile` test, which is essentially just a wrapper for the `foundry` pytest [fixture](https://github.com/runtimeverification/kontrol/blob/2fba281b0a73d76c1dba37f74220051901d04606/src/tests/integration/conftest.py#L53) that generates and compiles the `contracts.k` and `foundry.k` files. This produces the `foundry/out/kompiled` directory
- when running the actual proof tests, pass the generated `foundry` directory to them explicitly. This makes sure that the pytest fixture will not be forced to recompile stuff in every `pytest-xdist` worker.
- Re-using the `foundry` directory has a side-effect of also reusing the `foundry/out/proofs` directory. To make sure that we actually run the proofs in `master_shell`, and not re-use the ones produced by `feature_shell`, we delete them between the steps.

Some things I've noticed that may be worth accounting for, but not in this PR:
- the `server` [fixture](https://github.com/runtimeverification/kontrol/blob/7f45d5672a59f094e2d067e41923e037544b86f1/src/tests/integration/conftest.py#L35) is module-scoped. That means that all tests in a certain test module will reuse the same server instance, individually in each test worker.
  - the corollary of the above is that the proofs that are run later will use a more "tired" server, since the server state will bloat due to a bunch of duplicate (and likely empty) modules. It is not clear if this is a performance issue, but we have certainly not considered it earlier.
- another thing is that the decision on which test to skip is taken dynamically at the start of the test, and the tests, even if skipped, need the fixtures to be ready for them. In most cases, however, the server connection will be reused by other tests in the same module.


geo2a:
A test run with 7 workers. Still noisy, but perhaps slightly less so?
| Test                                                                               | georgy-kontrol-pref-reuse-artifacts time | master-88ef647ec time | (georgy-kontrol-pref-reuse-artifacts/master-88ef647ec) time
|------------------------------------------------------------------------------------|------------------------------------------|-----------------------|-------------------------------------------------------------
| Identity.identity(uint256)                                                         | 23.8                                     | 29.13                 | 0.8170271198077583
| kontrol/src/tests/integration/test_foundry_prove.py::test_foundry_resume_proof     | 102.91                                   | 118.84                | 0.8659542241669471
| ExpectRevertTest.testFail_expectRevert_bytes4()                                    | 159.03                                   | 183.62                | 0.8660821261300512
| ExpectRevertTest.testFail_expectRevert_false()                                     | 134.99                                   | 151.12                | 0.8932636315510852
| ConstructorTest.test_constructor()                                                 | 166.28                                   | 182.09                | 0.9131748036685156
| SymbolicStorageTest.testEmptyInitialStorage(uint256)                               | 61.08                                    | 65.53                 | 0.9320921715244925
| MockCallTest.testSelectorMockCall()                                                | 76.41                                    | 79.55                 | 0.9605279698302954
| AddConst.applyOp(uint256)                                                          | 29.27                                    | 30.42                 | 0.9621959237343852
| SetUpDeployTest.test_extcodesize()                                                 | 113.52                                   | 117.67                | 0.9647318772839296
| ArithmeticCallTest.test_double_add_double_sub(uint256,uint256)                     | 448.49                                   | 432.36                | 1.0373068739013784
| BytesTypeTest.test_bytes32(bytes32)                                                | 39.13                                    | 37.57                 | 1.041522491349481
| kontrol/src/tests/integration/test_foundry_prove.py::test_foundry_auto_abstraction | 96.22                                    | 91.04                 | 1.0568980667838312
| Identity.applyOp(uint256)                                                          | 51.74                                    | 47.78                 | 1.0828798660527417
| ConstructorTest.testFail_constructor()                                             | 118.87                                   | 103.62                | 1.1471723605481567
| ArithmeticContract.add(uint256,uint256)                                            | 35.82                                    | 29.94                 | 1.1963927855711423
| ArithmeticCallTest.test_double_add_sub_external(uint256,uint256,uint256)           | 393.63                                   | 303.5                 | 1.2969686985172981
| kontrol/src/tests/integration/test_foundry_prove.py::test_foundry_remove_node      | 46.7                                     | 32.43                 | 1.4400246685168054
| ExpectRevertTest.testFail_ExpectRevert_failAndSuccess()                            | 197.33                                   | 111.55                | 1.7689825190497537
| BMCLoopsTest.test_countdown_concrete()                                             | 38.34                                    | 0.53                  | 72.33962264150944
| TOTAL                                                                              | 2333.56                                  | 2148.29               | 1.0862406844513544

jberthold (approved):
image

goodlyrottenapple (approved):
Looks good. I worry that the more assumptions/fixes we add in our script the more we will have to keep patching it when the CLI arguments/foundry folders/etc. change. I guess time will tell and perhaps these can be up-streamed to kontrol directly.

ehildenb:
@geo2a this seems like something that we could have made a change directly to Kontrol for. Consider doing so there! (as I think @goodlyrottenapple also suggested).

PR #3858, opened 2024-05-13
Improve JSON logging

goodlyrottenapple:
This PR does the following:

* aligns JSON logs across the two backends
* removes implifyjson logs in favour of contextual JSON logging
* adds a custom syntax for `--log-context` also removing `--no-log-context` (see the docs/logging.md change for details)
* switches the ceil analysis to contextual logs


kore/src/Kore/Log/BoosterAdaptor.hs
```diff
@@ -87,21 +90,36 @@ koreSomeEntryLogAction renderer earlyFilter lateFilter textLogAction =
...
   where
-    addOriginField :: JSON.Value -> JSON.Value
-    addOriginField = \case
-        (JSON.Object xs) -> JSON.Object $ JSON.insert (JSON.fromText "origin") (JSON.toJSON KoreRpc) xs
-        xs -> xs
+    jsonContext =
+        foldr
+            ( \(SomeEntry _ c) cs -> case oneLineContextJson c of
+                JSON.Array cs' -> cs' <> cs
+                -- JSON.Object o | Just (JSON.Array cs') <- JSON.lookup "context" o -> cs' <> cs
```
geo2a, line 100:
```suggestion
```

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -299,15 +306,22 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
...
-            withContext "success" $
-                logMessage $
-                    renderOneLineText $
-                        "Substitution:"
-                            <+> (hsep $ intersperse "," $ map (\(k, v) -> pretty k <+> "->" <+> pretty v) $ Map.toList substitution)
+            withContext "success" $ do
+                logMessage rule
+                withContext "substitution"
+                    $ logMessage
+                    $ WithJsonMessage
```
geo2a, line 318: Do I understand correctly, that `WithJsonMessage` here does not mean that JSON is the only possible output format of this message, i.e. we could still get the normal contextual logs?

booster/tools/booster/Server.hs
```diff
@@ -173,16 +180,20 @@ main = do
...
-                                        && any (flip Glob.match contextStr) globPatterns
-                            else const True
-                    _ -> const True
+                koreLogEarlyFilter = case logFormat of
+                    Json -> \l@(Log.SomeEntry ctxt e) ->
+                        Log.oneLineJson e /= Null && koreFilterContext (ctxt <> [l])
+                    OneLine -> \l@(Log.SomeEntry ctxt _) -> koreFilterContext $ ctxt <> [l]
+                    Standard -> const True
+                koreFilterContext ctxt =
+                    not contexLoggingEnabled
```
geo2a, line 189: Looks correct, albeit contrived. Maybe is `if`?

geo2a (approved):
The changes look good to me! We may need to follow-up if we find something is missing. Happy to merge once the tests are green again.

booster/tools/booster/Server.hs
goodlyrottenapple: it was an if, but hlint didnt like that :D

kore/src/Log.hs
```diff
@@ -104,7 +105,8 @@ data LogMessage = LogMessage
 
 instance Entry LogMessage where
     entrySeverity LogMessage{severity} = severity
-    oneLineDoc (LogMessage{severity}) = prettySeverity severity
+    oneLineDoc LogMessage{message} = Pretty.pretty message
+    oneLineJson LogMessage{message} = JSON.toJSON message
```
jberthold, line 109: Do we also get the severity as a context in the json log?
I thought this would require 
```
oneLineContextJson LogMessage{severity} = JSON.object [ "context" .= JSON.toJSON [show severity] ]` or something similar...

jberthold (approved):
Large change... I am with @geo2a , let's merge and if anything is not right we'll fix it in a follow-up

booster/library/Booster/Pattern/Rewrite.hs
goodlyrottenapple: yes, this just means that the provided json is emitted when json logging is enabled

jberthold:
One small thing to do before merging this: Please adjust the log-context suggestion in 
https://github.com/runtimeverification/haskell-backend/blob/b27210141c06c5698201dc3f1fabc90bd09ffe59/booster/library/Booster/Pattern/ApplyEquations.hs#L388 to `abort,detail`

PR #3860, opened 2024-05-14
SMT retry logic changes in Kore

### LINKED ISSUE #3859: Make SMT query retry logic more agressive
- if a query returns unknown, do not retry with the same timeout, use double the timeout
- scale the timeout quadratically, rather than linearly
- add an option to disable the retries

geo2a:
Fixes #3859 

* if a query returns unknown, do not retry with the same timeout, use double the timeout. Scale further timeouts quadratically, rather than linearly.
* do not retry when applying simplifications
* add context logs for unknown predicates


geo2a:
Minor improvements in KEVM, may be just noise:

| Test                                  | georgy-smt-retries time | master-df703fed9 time | (georgy-smt-retries/master-df703fed9) time
|---------------------------------------|-------------------------|-----------------------|--------------------------------------------
| erc20/hkg/totalSupply-spec.k          | 42.08                   | 44.72                 | 0.9409660107334525
| mcd/dstoken-approve-fail-rough-spec.k | 98.25                   | 103.37                | 0.95046918835252
| benchmarks/storagevar01-spec.k        | 49.17                   | 51.28                 | 0.9588533541341654
| TOTAL                                 | 189.5                   | 199.37                | 0.9504940562772733

kore/src/Kore/Log/DecidePredicateUnknown.hs
```diff
@@ -101,7 +102,23 @@ instance Entry DecidePredicateUnknown where
       where
         prettyHsLoc Loc{loc_module, loc_start = (row, col)} =
             Pretty.pretty loc_module <> ":" <> Pretty.pretty row <> ":" <> Pretty.pretty col
-    oneLineDoc _ = "DecidePredicateUnknown"
+    oneLineDoc (DecidePredicateUnknown{message, predicates}) =
+        Pretty.hsep
+            [ Pretty.brackets "smt"
```
goodlyrottenapple, line 107: can we put this in the `oneLineContextDoc` (probably once the json logging is merged)?

goodlyrottenapple (approved):

kore/src/Kore/Rewrite/SMT/Evaluator.hs
```diff
@@ -169,7 +169,11 @@ decidePredicate ::
 decidePredicate onUnknown sideCondition predicates =
     whileDebugEvaluateCondition predicates $
         do
-            result <- query >>= whenUnknown retry
+            result <- case onUnknown of
+                -- do not retry the query on Unknown if the use-case is "soft", i.e. applying simplifications
+                WarnDecidePredicateUnknown _ _ -> query
```
jberthold, line 174: So here is the reason for the speedup?

jberthold (approved):
...still not sure about the aggressive scaling. 
We have seen people use `--smt-retry-limit 10`. With a default timeout of 125ms, that means the last retry uses a timeout of ~2 minutes, with `--smt-timeout 1000` (again seen in the wild), it will use a timeout of ~16 minutes.
We can merge it and see what it does in the hands of users, though.

kore/src/Kore/Rewrite/SMT/Evaluator.hs
geo2a: It appears so, yes.

Issue #3857, opened 2024-05-10
Booster's and Kore's `"implies"` endpoints disagree on a `pyk` integration test


geo2a:
The HB update [pr](https://github.com/runtimeverification/k/pull/4319) to K fails two `pyk` integrations tests:

```
FAILED src/tests/integration/kore/test_kore_client.py::TestKoreClient::test_implies_error[booster-0 -> X]
FAILED src/tests/integration/proof/test_imp.py::TestImpProof::test_implies[booster-antecedent-bottom]
```

In the first test, Booster returns a non-empty substitution and an empty predicate, while Kore returns an empty substitution and `#Bottom` as predicate, which is also the expected output.

```
AssertionError: assert CSubst(subst=Subst(_subst=FrozenDict({'Y': KVariable(name='X', sort=KSort(name='Int'))})), constraints=()) == CSubst(subst=Subst(_subst=FrozenDict({})), constraints=(KApply(label=KLabel(name='#Bottom', params=(KSort(name='GeneratedTopCell'),)), args=()),))
```

In the second test, an error response (indicated by an exception in `KoreClient`) is expected, but no exception is raised.

goodlyrottenapple:
I suspect checking whether the antecedent/consequent is #Bottom will negate any speed gains we got from implementing the simple implication check, as this would just be a call to the kore simplifier anyway :/

geo2a:
I agree that simplifying both antecedent and consequent would probably kill the performance.

Perhaps we need to re-think the how we use the implies endpoint a little. As I understand, we currently have two use-cases for it:
* the subsumption check, implemented by the `CTermSymbolic.implies` [method](https://github.com/runtimeverification/k/blob/853aa74fff37344197aa5499cb14be6f1f95e432/pyk/src/pyk/cterm/symbolic.py#L188),
* the equatlity and refutation proofs, and the `ImpliesProver` [class](https://github.com/runtimeverification/k/blob/853aa74fff37344197aa5499cb14be6f1f95e432/pyk/src/pyk/proof/implies.py#L421) to discharge them.

Perhaps we could indeed implement @ehildenb original idea and have two endpoints: a lightweight "simplify-implication" (or some other name), and a the old "implies". The lightweight one will have additional assumptions:
* neither antecedent not consequent are bottom
* perhaps something else?
we can than use "simplify-implication" without fear in the subsumption check scenario, because:
* the `pyk` prover simplifies the target configuration anyway, and we know it is not bottom
[...]

PR #3846, opened 2024-05-02
Implies endpoint in booster

goodlyrottenapple:
Closes #3778 #3601

This PR introduces a simple check implication implementation, which uses the matching algorithm to try to find a substitution or fail to unify the given configurations.

If matching the RHS configuration to the LHS succeeds and produces a substitution σ, we filter out any predicates in the consequent which appear in the antecedent and then check if all predicates in σ(filtered_preds) evaluate to true/#top. (Question: do we want to apply σ to the RHS preds before filtering?). With this simple algorithm, we can successfully discharge almost 85% of implication checks in KEVM  and 88% in Kontrol.

PR #3862, opened 2024-05-14
Partially revert the implication endpoint

goodlyrottenapple:
Partially reverts  #3846 by disabling the implication endpoint in booster until we figure out and address #3857


jberthold:
... I thought we would disable the call site in `Proxy.hs` ? That way `booster-dev` would remain as it is and call the new version so we can work on it.

jberthold (approved):

PR #3878, opened 2024-05-20
Fix some small problems

jberthold:
* `--interim-simplification` problem: When `maxDepth < forceFallback` , the proxy was considering a genuine `DepthBound` result as an artefact caused by the interim simplification, and would try to continue execution.
* The `cabal.project` file was set up to never profile any `booster` or `kore-rpc-types` code
* When given a path to a non-existing file as input, `RpcClient` reported a server connection error and retried (the criterion for recognising connection errors was to general).

### DIFF
--- a/booster/tools/booster/Proxy.hs
+++ b/booster/tools/booster/Proxy.hs
@@ -292,10 +292,11 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
                 else "Iterating execute request at " <> show currentDepth
         -- calculate depth limit, considering possible forced Kore simplification
         let mbDepthLimit = case (cfg.forceFallback, r.maxDepth) of
-                (Just (Depth forceDepth), Just (Depth maxDepth)) ->
-                    if cDepth + forceDepth < maxDepth
-                        then Just $ Depth forceDepth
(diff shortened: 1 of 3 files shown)


booster/tools/booster/Proxy.hs
```diff
@@ -292,18 +292,21 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
...
+                    | otherwise ->
+                        Just $ Depth $ maxDepth - cDepth
                 (Just forceDepth, _) -> Just forceDepth
                 (_, Just maxDepth) -> Just $ maxDepth - currentDepth
                 _ -> Nothing
         (bResult, bTime) <- Stats.timed $ booster (Execute r{maxDepth = mbDepthLimit})
         case bResult of
             Right (Execute boosterResult)
                 -- the execution reached the depth bound due to a forced Kore simplification
-                | boosterResult.reason == DepthBound && isJust cfg.forceFallback -> do
```
jberthold, line 306: Once `cDepth` was reaching `maxDepth` (see  above, when `maxDepth < forceDepth`) we were starting to make requests with `max-depth: 0` and looping.

geo2a (approved):
👍

PR #3880, opened 2024-05-20
RPC trace cleanup: remove simplification logs

geo2a:
Fixes https://github.com/runtimeverification/haskell-backend/issues/3871

Note that removing the simplification log request parameters and the simplification log entries form the response is same, because `pyk` is already modified to not recognize them, see:

- the `LogEntry` [type](https://github.com/runtimeverification/k/blob/develop/pyk/src/pyk/kore/rpc.py#L543)
- the parameters for [execute](https://github.com/runtimeverification/k/blob/develop/pyk/src/pyk/kore/rpc.py#L1003), [simplify](https://github.com/runtimeverification/k/blob/develop/pyk/src/pyk/kore/rpc.py#L1049) and [implies](https://github.com/runtimeverification/k/blob/develop/pyk/src/pyk/kore/rpc.py#L1030) request

But even considering that, we should run `pyk`'s integration tests before merging.


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -16,8 +16,6 @@ module Booster.Pattern.ApplyEquations (
     EquationPreference (..),
     EquationFailure (..),
     EquationTrace (..),
-    pattern CollectEquationTraces,
-    pattern NoCollectEquationTraces,
     eraseStates,
```
jberthold, line 19: Am I right that we shouldn't export `EquationTrace` and `eraseStates` any more here?

jberthold (commented):
While we are at it, we could maybe clean up more - the `RewriteSimplified` constructor in `RewriteTrace` and all related code is one example.
(can also merge this and proceed with other things, though)

geo2a:
I ran the `pyk` integration tests and they pass.

@jberthold sounds good, I'll do a bit more clean-up.

booster/library/Booster/Pattern/ApplyEquations.hs
geo2a: Yes, thanks, removed them

geo2a:
@jberthold I did not fully remove the `RewriteSimplified` constructor as it is used for the `-l Rewrite` logging level as I understand.

I think it is probably enough clean-up at that point.

jberthold (approved):

jberthold:
> @jberthold I did not fully remove the `RewriteSimplified` constructor as it is used for the `-l Rewrite` logging level as I understand.

I think it is only used for error reporting but nvm :shrug:

geo2a:
We probably need to update K once https://github.com/runtimeverification/k/pull/4367 is merged to fix Kore integration [tests](https://github.com/runtimeverification/haskell-backend/actions/runs/9170105259/job/25211759488?pr=3880#step:10:22):
```
error: hash mismatch in fixed-output derivation '/nix/store/0wsq2aqzqlcb0g61jm389zwqmmjjjsmz-k-7.0.40-dirty-maven-deps.drv':
         specified: sha256-DnY+Zx4fr3iuI24VjlnCVeQCcK6aHYYAEW13aUEzd+I=
            got:    sha256-6rGuU8NFSp3lQVxxPj5WREpNVPneMqPH45DFEmLiH6I=
```

goodlyrottenapple (approved):
Nice! Great to see the cleanup :)

PR #3881, opened 2024-05-20
JSON logging cleanup

goodlyrottenapple:
* Re-add the `test-log-simplify-json` rpc test with a much smaller footprint, as we want to just track the context logging format
* Clean up the equation contexts a bit in booster, as we weren't consisntently logging messages areound functions/equations


booster/test/rpc-integration/test-log-simplify-json/README.md
```diff
@@ -0,0 +1,7 @@
+# Testing logging of equations as JSON
+
+This test sends the request toexecute a part of an IMP semantics proof.
```
geo2a, line 3:
```suggestion
This test sends the request to execute a part of an IMP semantics proof.
```

scripts/booster-integration-tests.sh
```diff
@@ -55,6 +55,9 @@ for dir in $(ls -d test-*); do
         "simplify")
             SERVER=$BOOSTER_DEV ./runDirectoryTest.sh test-$name $@
             ;;
+        "log-simplify-json")
+            SERVER="${KORE_RPC_BOOSTER} --log-format json -l Simplify --log-file test-$name/simplify-log.txt" ./runDirectoryTest.sh test-$name $@
+            ;;
```
goodlyrottenapple, line 60: made several changes here, including adding the conversion from the original custom log levels tot he context filtering. Also renamed `--simplification-log-file` to just `--log-file`

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -860,13 +808,15 @@ applyEquations theory handler term = do
     processEquations [] =
         pure term -- nothing to do, term stays the same
     processEquations (eq : rest) = do
-        res <- applyEquation term eq
+        res <- withRuleContext eq $ applyEquation term eq
         emitEquationTrace term eq.attributes.location eq.attributes.ruleLabel eq.attributes.uniqueId res
```
geo2a, line 812: No that `-l SImplify` is re-implemented through the contextual logging, and the simplification RPC traces are to be remove in a pending PR, we are fee to drop the calls to `emitEquationTrace`, and the function itself too.

goodlyrottenapple: I was going to leave that for another pr, because we actually still log via the old logger interface within this function. Until we migrate our tools to contextual logging these may still be necessary?

geo2a: Sounds good, let's keep it for now.

booster/test/rpc-integration/test-log-simplify-json/README.md
```diff
@@ -0,0 +1,7 @@
+# Testing logging of equations as JSON
+
+This test sends the request to execute a part of an IMP semantics proof.
+The `kore-rpc-booster` server is expected to be started with `-l SimplifyJson --log-file test-log-simplify-json/simplify-log.txt`, which ensures that the equation simplification logs are collected by both Booster and Kore and emitted into a file. The log file is compared to `simplify-log.golden.txt`.
```
geo2a, line 4:
```suggestion
The `kore-rpc-booster` server is expected to be started with `--log-format json -l Simplify --log-file test-log-simplify-json/simplify-log.txt`, which ensures that the equation simplification logs are collected by both Booster and Kore and emitted into a file. The log file is compared to `simplify-log.golden.txt`.
```

geo2a (approved):

jberthold (approved):
From the comment about `emitEquationTrace` , I understand that you planned to follow up with another PR? Then we can merge this and leave the following to the next iteration:
* There is some (more) dead code left to clean up
* The context-log replacement for `-l Aborts` needs some more attention 
* Generally we should remove all old-style customLevels that have a replacement. This probably means removing most levels, except the `Abort` one as it is now. Once that is addressed, the `levelFilter` can just handle the standard `Error,Warn,Info` levels (but we should rather remove those, too).

booster/tools/booster/Server.hs
```diff
@@ -379,7 +382,6 @@ logLevelToKoreLogEntryMap :: Map.Map LogLevel [Text.Text]
 logLevelToKoreLogEntryMap =
     Map.fromList
         [ (LevelOther "SimplifyKore", ["DebugAttemptEquation", "DebugTerm"])
-        , (LevelOther "SimplifyJson", ["DebugAttemptEquation", "DebugTerm"])
```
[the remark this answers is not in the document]
goodlyrottenapple: we still need to pass these (e.g. `DebugAttemptEquation`) to actually get context logging in the old backend, because it has multiple layers of filtering. We could just enable all of them tho...

PR #3889, opened 2024-05-23
Retry SMT queries in Booster when deciding predicates

### LINKED ISSUE #3884: Retry Booster's SMT queries
Currently, Booster does not ever retry its SMT queries when it gets `unkwnon` results. We should implement a mechanism similar to the [one](https://github.com/runtimeverification/haskell-backend/pull/3860) in Kore.

geo2a:
Fixes #3884 

- refactor the SMT modules to allow re-initializing the solver dynamically:
  - `SMTContext` now carries the prelude and `SMTOptions`
  - the `closeContext` function only closes the solver handle, and does not close the solver transcript handle. The new `destroyContext` function does both.
  - The `SMT` monad is is changed to have `StateT SMTContext`. The `Booster.SMT.Interface.restartSolver` function restarts the solver with modified options, altering the context. Note that it does not touch the transcript handle, i.e. it will persist the sovler restarts.

- the `getModelFor` and `checkPredicates` functions now have logic to retry with doubled timeout when the solver returns unknown.

I'm checking-in the integration test that I was using when developing. I ran the server with `--smt-timeout 1`  to observe the retry mechanism kicking-in, but I do not think it is a good idea to try to force it to retry the queries in CI. The test may still be useful when working on the SMT code in future.


geo2a:
No difference in KEVM performance:
| Test                           | georgy-retry-smt-in-booster time | master-7f482b016 time | (georgy-retry-smt-in-booster/master-7f482b016) time
|--------------------------------|----------------------------------|-----------------------|-----------------------------------------------------
| benchmarks/storagevar00-spec.k | 50.06                            | 51.85                 | 0.9654773384763742
| mcd/dai-symbol-pass-spec.k     | 57.66                            | 59.72                 | 0.965505693235097
| benchmarks/storagevar01-spec.k | 51.36                            | 49.32                 | 1.0413625304136254
| TOTAL                          | 159.07999999999998               | 160.89                | 0.9887500776928336

geo2a:
The `test-log-simplify-json` integration test failure has been solved by [e867eed](https://github.com/runtimeverification/haskell-backend/pull/3889/commits/e867eed476955de7cd18dbea8c41d95235ee2db6) --- we should wait until the server is stopped before we compare the log file to the golden one.

booster/library/Booster/SMT/Interface.hs
```diff
@@ -183,50 +207,86 @@ getModelFor ctxt ps subst
...
+                    SMT.eq <$> SMT.translateTerm (Var v) <*> SMT.translateTerm t
+            smtSubst <-
+                mapM (\(v, t) -> Assert "Substitution" <$> mkSMTEquation v t) $ Map.assocs subst
+            smtPs <-
+                mapM (\(Predicate p) -> Assert (mkComment p) <$> SMT.translateTerm p) ps
+            pure $ smtSubst <> smtPs
+
+    interactWithSolver :: [DeclareCommand] -> SMT io Response
+    interactWithSolver smtAsserts = do
+        runCmd_ SMT.Push -- assuming the prelude has been run already,
```
goodlyrottenapple, line 231: the pushing/popping of contexts is now a bit difficult to follow. would it not be better to push explicitly before calling `interactWithSolver` or even possibly use something like a `withSmtContext` which wraps a bunch of smt commands in a push pop block...

booster/library/Booster/SMT/Runner.hs
```diff
@@ -41,64 +46,101 @@ import Booster.Log (LoggerMIO (..), logMessage)
...
-        BS.hPutStrLn h "; stopping solver\n;;;;;;;;;;;;;;;;;;;;;;;"
+    liftIO ctxt.solverClose
+
+destroyContext :: LoggerMIO io => SMTContext -> io ()
+destroyContext ctxt = do
+    logMessage ("Permanently stopping SMT solver" :: Text)
+    whenJust ctxt.mbTranscriptHandle $ \h -> liftIO $ do
+        BS.hPutStrLn h "; permanently stopping solver\n;;;;;;;;;;;;;;;;;;;;;;;"
         hClose h
     liftIO ctxt.solverClose
```
goodlyrottenapple, line 127: Could you add a quick comment detailing the difference between these two? It's not immediately obvious to me when to use one or the other.

goodlyrottenapple (approved):
Looks great! Left some minor comments

booster/library/Booster/SMT/Interface.hs
geo2a: Yes, the assertion stack is indeed a bit of a mess, sorry about that. I think it would be good to keep the assrtion stack manipulation confined to the `interactWithSolver` functions. I will try to clean it up.

booster/library/Booster/SMT/Runner.hs
```diff
@@ -41,64 +46,108 @@ import Booster.Log (LoggerMIO (..), logMessage)
...
+{- | Close the connection to the SMT solver process, but hold on to the other resources in @SMTContext@,
+such as the transcript handle. This function is used in 'Booster.SMT.Interface.hardResetSolver'
+to reset the solver connection while keeping the same transcript handle.
+-}
 closeContext :: LoggerMIO io => SMTContext -> io ()
 closeContext ctxt = do
     logMessage ("Stopping SMT solver" :: Text)
-    whenJust ctxt.mbTranscript $ \h -> liftIO $ do
-        BS.hPutStrLn h "; stopping solver\n;;;;;;;;;;;;;;;;;;;;;;;"
+    liftIO ctxt.solverClose
```
jberthold, line 123: Maybe still write something to the transcript here?

booster/test/rpc-integration/test-non-linear-int-requires/response-init.json
```diff
@@ -0,0 +1,2 @@
+{"jsonrpc":"2.0","id":1,"result":{"reason":"aborted","depth":0,"state":{"term":{"format":"KORE","version":1,"term":{"tag":"App","name":"Lbl'-LT-'generatedTop'-GT-'","sorts":[],"args":[{"tag":"App","name":"Lbl'-LT-'k'-GT-'","sorts":[],"args":[{"tag":"App","name":"kseq","sorts":[],"args":[{"tag":"App","name":"inj","sorts":[{"tag":"SortApp","name":"SortState","args":[]},{"tag":"SortApp","name":"SortKItem","args":[]}],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortState","args":[]},"value":"init"}]},{"tag":"App","name":"dotk","sorts":[],"args":[]}]}]},{"tag":"App","name":"Lbl'-LT-'x-int'-GT-'","sorts":[],"args":[{"tag":"EVar","name":"VarX","sort":{"tag":"SortApp","name":"SortInt","args":[]}}]},{"tag":"App","name":"Lbl'-LT-'y-int'-GT-'","sorts":[],"args":[{"tag":"EVar","name":"VarY","sort":{"tag":"SortApp","name":"SortInt","args":[]}}]},{"tag":"App","name":"Lbl'-LT-'generatedCounter'-GT-'","sorts":[],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"0"}]}]}}}}}
+
```
jberthold, line 2: (nit) could use formatted output...

jberthold (approved):

booster/library/Booster/SMT/Runner.hs
geo2a: Good point!

booster/test/rpc-integration/test-non-linear-int-requires/response-init.json
geo2a: Done.

PR #3893, opened 2024-05-23
Increase timestamp resolution

geo2a:
This PR:
- customizes the default timestamp caching mechanism provided by the `fast-logger` package, allowing finer granularity
- adds `--timestamp-format` CLI argument to `booster-dev` and `kore-rpc-booster`, with two possible values: `pretty` and `nanoseconds`. `pretty` is the default.


goodlyrottenapple (approved):

booster/tools/booster/Server.hs
```diff
@@ -161,7 +160,8 @@ main = do
                 || lvl >= logLevel && lvl <= LevelError
         koreSolverOptions = translateSMTOpts smtOptions
 
-    mTimeCache <- if logTimeStamps then Just <$> (newTimeCache "%Y-%m-%d %T") else pure Nothing
+    mTimeCache <-
+        if logTimeStamps then Just <$> Booster.newTimeCache "%Y-%m-%dT%H:%M:%S:%6Q" else pure Nothing
```
jberthold, line 164:
```suggestion
        if logTimeStamps then Just <$> Booster.newTimeCache "%Y-%m-%dT%H:%M:%S.%6Q" else pure Nothing
```

dev-tools/booster-dev/Server.hs
```diff
@@ -120,7 +119,7 @@ runServer ::
     IO ()
 runServer port definitions defaultMain mLlvmLibrary logFile mSMTOptions (logLevel, customLevels) logContexts logTimeStamps logFormat =
     do
-        mTimeCache <- if logTimeStamps then Just <$> (newTimeCache "%Y-%m-%d %T") else pure Nothing
+        mTimeCache <- if logTimeStamps then Just <$> newTimeCache "%Y-%m-%dT%H:%M:%S:%6Q" else pure Nothing
```
jberthold, line 122:
```suggestion
        mTimeCache <- if logTimeStamps then Just <$> newTimeCache "%Y-%m-%dT%H:%M:%S.%6Q" else pure Nothing
```

booster/library/Booster/Util.hs
```diff
@@ -155,3 +159,21 @@ withFastLogger mFormattedTime (Just fp) log' =
...
+-}
+newTimeCache :: TimeFormat -> IO (IO FormattedTime)
+newTimeCache fmt =
+    mkAutoUpdate
+        defaultUpdateSettings{updateFreq = 100}
+            { updateAction = formatSystemTime fmt <$> getSystemTime
+            }
+
+formatSystemTime :: TimeFormat -> SystemTime -> ByteString
+formatSystemTime fmt = BS.pack . formatTime defaultTimeLocale (BS.unpack fmt) . systemToUTCTime
```
jberthold, line 179: (nit) Adjust the comment at the top which says "every 1 second".
Should we maybe hard-wire the time format here?

jberthold (approved):

booster/library/Booster/Util.hs
geo2a: I've updated the comment and hardwired the format string. We can also switch the formatting off now and just the raw nanoseconds. This will be useful when we parse the JSON logs for post-processing.

PR #3896, opened 2024-05-24
Add stack to dev shell

geo2a:
- remove `cabal.project.freeze`, but keep the scrips to generate it in case we ever want it back
- update README with up-to-date information about updating Haskell dependencies
- add `stack` to Nix dev shell


geo2a:
Closed in favor of https://github.com/runtimeverification/haskell-backend/pull/3896

PR #3897, opened 2024-05-24
Fix nix flake devshell

geo2a:
Follow-up to https://github.com/runtimeverification/haskell-backend/pull/3890

- update GHC from 9.6.4 to 9.6.5
- update Stackage LTS from 22.10 to 22.23
- use HLS from Nix pkgs, upgrading it to 2.8.0.0
- update `all-cabal-hashes`
- remove `cabal.project.freeze`, but keep the scrips to generate it in case we ever want it back
- update README with up-to-date information about updating Haskell dependencies
- add `stack` to Nix dev shell


jberthold:
> * remove `cabal.project.freeze`, but keep the scrips to generate it in case we ever want it back

Why exactly are we removing this?
We are running [the unit tests](https://github.com/runtimeverification/haskell-backend/blob/974e29352c66423e1064081abbbbe42d4aedb6ff/.github/workflows/test.yml#L136) and [booster integration tests](https://github.com/runtimeverification/haskell-backend/blob/974e29352c66423e1064081abbbbe42d4aedb6ff/.github/workflows/test.yml#L208) using a nix shell and `cabal` so I think it does matter?

Apart from that, I am OK with the upgrade (although not having to `allow-newer` would be nice...)

geo2a:
As I understand, the Haskell packages available in the Nix shell come from the stack resolver, since the Nix flake uses `stacklock2nix`. This makes `cabal.project.freeze` redundant in the Nix shell.

However, perhaps we indeed could keep it in the end, as otherwise the project would likely be impossible to build with pure `cabal-install`.

jberthold (approved):

PR #3904, opened 2024-05-28
3902 use smt solver for equations

jberthold:
This PR adds the SMT solver to the equation application function.

Unclear side conditions for equations (those that could not be simplified to values in isolation) are checked (two-sided) using the SMT solver in context of the known path condition. 
This should reduce the amount of unclear conditions, and thereby enable more simplificaitons in boooster.

Fixes #3902

PR #3907, opened 2024-05-29
Abort on "Unknown" side conditions, don't crash

geo2a:
This PR makes sure we do not immediately hard-stop with an RPC error when receiving `Unknown` from Z3 in Booster.

- use pure exceptions in SMT functions, check results at call sites to `checkPredicates`
- threat Z3-`Unknown` predicates as any other unclear conditions: abort rewriting

### DIFF
--- a/booster/library/Booster/Pattern/Rewrite.hs
+++ b/booster/library/Booster/Pattern/Rewrite.hs
@@ -359,7 +359,8 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
             checkAllRequires <- SMT.checkPredicates solver prior mempty (Set.fromList unclearRequires)
 
             case checkAllRequires of
-                Nothing -> do
+                Left _ -> do
+                    -- TODO: we could process SMTError here
                     -- unclear even with the prior
                     withContext "abort" $
                         logMessage $
@@ -369,11 +370,11 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
                         RuleConditionUnclear rule . coerce $
                             foldl1 AndTerm $
                                 map coerce unclearRequires
-                Just False -> do
+                Right False -> do
                     -- requires is actually false given the prior
                     withContext "failure" $ logMessage ("Required clauses evaluated to #Bottom." :: Text)
                     RewriteRuleAppT $ pure NotApplied
-                Just True ->
+                Right True ->
                     -- can proceed
                     pure ()
         Nothing ->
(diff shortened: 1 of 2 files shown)


goodlyrottenapple (approved):

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -359,7 +359,8 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
             checkAllRequires <- SMT.checkPredicates solver prior mempty (Set.fromList unclearRequires)
 
             case checkAllRequires of
-                Nothing -> do
+                Left _ -> do
+                    -- TODO: we could process SMTError here
```
jberthold, line 363: We should probably restrict this to the `SMTSolverUnknown` case. An `SMTTranslationError` is an entirely different case (and the `General` error was meant to cover infrastructure issues).

booster/library/Booster/SMT/Interface.hs
```diff
@@ -302,8 +299,9 @@ checkPredicates ctxt givenPs givenSubst psToCheck
         consistent <- smtRun CheckSat
         when (consistent /= Sat) $ do
             void $ smtRun Pop
-            Log.logMessage ("Inconsistent ground truth, check returns Nothing" :: Text)
-            fail "returns nothing"
+            let errMsg = ("Inconsistent ground truth, check returns Nothing" :: Text)
+            Log.logMessage errMsg
+            throwE $ GeneralSMTError errMsg
```
jberthold, line 304: This is not really an "error", rather a defensive strategy to avoid jumping to unjustified conclusions

booster/library/Booster/SMT/Interface.hs
```diff
@@ -324,22 +322,22 @@ checkPredicates ctxt givenPs givenSubst psToCheck
 
         case (positive, negative) of
             (Unsat, Unsat) -> throwSMT "Inconsistent ground truth: should have been caught above"
-            (Sat, Sat) -> fail "Implication not determined"
+            (Sat, Sat) -> throwE $ GeneralSMTError "Implication not determined"
```
jberthold, line 325: Same here, this is actually a _result_ (neither `P` nor `not P` are implied)

jberthold (commented):
I'd really want to merge this because it is a good idea in general (and it would fix an issue with my PR #3904 ), but I think it changes the intent of the `SMTError` and mixes results and errors.
The correct model would IMHO be to return `Either SMTError (Maybe Bool)` and catch the `SMTSolverUnknown` case.

jberthold (approved):
Approving and following up with another PR

PR #3908, opened 2024-05-30
HOTFIX re-introduce indeterminate result for `checkPredicates` instead of exceptions

jberthold:
* Re-introduces the 'indeterminate' result for checkPredicates, and returns that on inconsistent ground truth.
* In the rewrite, aborts on `SMTSolverUnknown` when checking requires clauses, but throws actual exceptions (terminating the request) on other SMT errors (translation error, general error).

Follow-up to #3907

### DIFF
--- a/booster/library/Booster/SMT/Interface.hs
+++ b/booster/library/Booster/SMT/Interface.hs
@@ -263,9 +263,9 @@ checkPredicates ::
     Set Predicate ->
     Map Variable Term ->
     Set Predicate ->
-    io (Either SMTError Bool)
+    io (Either SMTError (Maybe Bool))
 checkPredicates ctxt givenPs givenSubst psToCheck
-    | null psToCheck = pure $ Right True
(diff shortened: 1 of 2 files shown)


booster/library/Booster/SMT/Interface.hs
```diff
@@ -263,9 +263,9 @@ checkPredicates ::
     Set Predicate ->
     Map Variable Term ->
     Set Predicate ->
-    io (Either SMTError Bool)
+    io (Either SMTError (Maybe Bool))
```
jberthold, line 266: Since the `SMTSolverUnknown` is not a "real error", we could also use a 4-fold dedicated result type here
```
data CheckPredicatesResult
    = PredicatesTrue
    | PredicatesFalse
    | PredicatesIndeterminate
    | UnknownResult
```
That means we can throw exceptions directly from here as we did before. 
The current version puts the burden on the caller but leaves more flexibility.

geo2a (approved):

PR #3924, opened 2024-06-07
Set up K-style versioning

Baltoli:
This PR introduces a release versioning scheme that's a carbon copy of the setup used in the LLVM backend, but with development happening on the `master` branch and versioning information being carried on the `release` branch.

The code is well-tested on the LLVM backend side, and I've tested the various versioning scripts locally, but it's difficult to explicitly test the end-end integration here without just merging the PR. Any remaining bugs or issues will be easy to iron out once we've seen a run happen here.


goodlyrottenapple (commented):
LGTM but let's wait till monday for @jberthold and @geo2a to have a look

.github/workflows/master.yml
```diff
@@ -0,0 +1,39 @@
...
+          if git diff --exit-code ${old_master} ${new_master} -- package/version; then
+            git merge --no-edit origin/master
+            ./package/version.sh bump
+          else
+            git merge --no-edit --strategy-option=theirs origin/master
+          fi
+          ./package/version.sh sub
+          if git add --update && git commit --no-edit --allow-empty --message "Set Version: $(cat package/version)"; then
+            git push origin release
+          fi
```
jberthold, line 39: I thought we would only bump versions on `release`, and leave `master` with `0.1.0` forever?
Maybe I misunderstood that...

.github/workflows/release.yml
```diff
@@ -39,6 +58,7 @@ jobs:
 
   cache-cabal:
     name: 'Cache Cabal'
+    needs: draft-release
     strategy:
```
jberthold, line 62: This job must be run on `master` (more generally, on the branch marked as `default` in github) so that all PRs can share the cache.
We have to move it to the `master.yml` file.

.github/workflows/release.yml
```diff
@@ -94,6 +114,7 @@ jobs:
   cache-stack:
     name: 'Cache Stack'
     runs-on: ubuntu-22.04
+    needs: draft-release
```
jberthold, line 117: Ditto, this job creates a cache that should be used by all PRs so it must run on the `master` branch.

jberthold (changes requested):

jberthold (approved):
LGTM, thanks for adjusting

PR #3925, opened 2024-06-07
Add timestampts to JSON logs

### LINKED ISSUE #3894: Consolidate `--log-format json` and `--log-timestamps` options
When we give both  `--log-format json` and `--log-timestamps` options, we get logs that look like this:
```
2024-05-23T15:03:23:.703160 {"context":["booster","execute",{"term":"f76c7c3"},{"rewrite":"157d9251db4c252c841d0e6b2a2cded8737142b5719012bd6ecd9b5f7152cb55"},"detail"],"message":"...kproj/evm-semantics/evm.md :  (909, 10)"}
```

We should instead include the timestamp into the JSON object itself.

goodlyrottenapple:
Fixes #3894


kore/src/Kore/Log/DebugAppliedRewriteRules.hs
```diff
@@ -62,6 +62,10 @@ instance Pretty DebugAppliedRewriteRules where
 instance Entry DebugAppliedRewriteRules where
     entrySeverity _ = Debug
     helpDoc _ = "log applied rewrite rules"
+
+    isEmpty DebugAppliedRewriteRules{appliedRewriteRules}
+        | null appliedRewriteRules = True
+        | otherwise = False
```
goodlyrottenapple, line 68: There is now no way (I think) to filter the message once we already start converting it to `LogStr`, so we would either have to render it twice, once in the filter (where we could then check if it is empty/null) and second time when we actually want to log it. As this would be inefficient, I've added a type-class function which can be used to control if some kore Entry is just used for context and should thus not be emitted as a message. See occurrences of `isEmpty` in this PR to see what I mean.

jberthold (approved):
LGTM (haven't tested it, though)

jberthold:
While at it, could you please remove the dot from https://github.com/runtimeverification/haskell-backend/pull/3925/files#diff-fdf134fb9c13c1143cd1c9d16f037c810a3f29d61d40c9944b898af886bb61e9R188 so that we can parse the timestamps later?
```
    let formatString = "%Y-%m-%dT%H:%M:%S%6Q"
```

PR #3932, opened 2024-06-11
Remove cabal cache form CI

goodlyrottenapple:
We noticed that we are trying to cache the cabal store, however, Nix already does this for us, as we have seen from the fact that cache has been broken but builds run just as fast. Therefore, this caching should be removed entirely.

This also fixes an issue where the cache key on master was not being built correctly, because we forgot to set the `ghc_version` ENV variable. THis also meant we were building with a different version of GHC on master push vs the feature branches.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -7,62 +7,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  ghc_version: "9.6.5"
+  stack_version: "2.15.1"
+  hpack_version: '0.36'
+
 jobs:
-
-  cache-cabal:
-    name: 'Cache Cabal'
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - runner: ubuntu-22.04
-            os: ubuntu-22.04
-            nix: x86_64-linux
-          - runner: MacM1
-            os: self-macos-12
-            nix: aarch64-darwin
-    runs-on: ${{ matrix.runner }}
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v4
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -7,62 +7,12 @@ concurrency:
...
-            cabal-nix-${{ runner.os }}-ghc-${{ env.ghc_version }}-${{ hashFiles('flake.lock') }}-${{ hashFiles('cabal.project') }}-${{ hashFiles('cabal.project.freeze') }}
-            cabal-nix-${{ runner.os }}-ghc-${{ env.ghc_version }}-${{ hashFiles('flake.lock') }}-${{ hashFiles('cabal.project') }}
-            cabal-nix-${{ runner.os }}-ghc-${{ env.ghc_version }}-${{ hashFiles('flake.lock') }}
-
-      - name: Test
-        run: GC_DONT_GC=1 nix develop .#cabal --command bash -c "hpack ./booster && hpack dev-tools && cabal update && cabal build all && cabal test --enable-tests --test-show-details=direct kore-test unit-tests"
+env:
+  ghc_version: "9.6.5"
+  stack_version: "2.15.1"
+  hpack_version: '0.36'
```
jberthold, line 13: (nit) could [remove those variables from `release.yml`](https://github.com/runtimeverification/haskell-backend/blob/3b01114ca8cb64ba50e0dce50adf033f8930d89e/.github/workflows/release.yml#L7-L10) to avoid confusion

jberthold (approved):
LGTM

geo2a (approved):

.github/workflows/master.yml
goodlyrottenapple: good idea

PR #3933, opened 2024-06-11
Fix release CI job

goodlyrottenapple:
looks like we maybe need to explicitly add `draft-release` to the needs section of `update-depends`??


jberthold:
> looks like we maybe need to explicitly add `draft-release` to the needs section of `update-depends`??

Yes, thanks - I missed that detail in the rush to fix things.
Another question is _why_ we actually build that release _again_ - we have tested building with both `cabal` and `stack` in `master`. What would make sense if we want to do it is to re-run the _integration_ tests before publishing the release.

PR #3937, opened 2024-06-12
Add `-l EquationWarnings` and `warning` context in Booster

### LINKED ISSUE #3927: Annoying warning about recursion depth being exceeded
We've had several users ask about the importance of the recursion depth warning (including this one: https://discord.com/channels/824582698147905556/1013598994133434430/1249744367502229605). Indeed, this warning is quite verbose, and provides very little information other than "you can turn on deeper recursion".

Can it be made to happen less? Like, only report it once per simplification? Or once per context?

geo2a:
Fixes #3927

This PR disables the display of the three equation-related warning in Booster by default. They can be re-enabled by the flag `-l EquationWarnings`, which is a shorthand to the context expression `--log-context booster>(simplification *|function *)>warning`.

### DIFF
--- a/booster/library/Booster/CLOptions.hs
+++ b/booster/library/Booster/CLOptions.hs
@@ -232,6 +232,10 @@ allowedLogLevels =
     , ("Depth", "Log the current depth of the state")
     , ("SMT", "Log the SMT-solver interactions")
     , ("ErrorDetails", "Log error conditions with extensive details")
+    ,
+        ( "EquationWarnings"
+        , "Log warnings indicating soft-violations of conditions, i.e. exceeding the equation recursion/iteration limit "
+        )
     ]
 
 levelToContext :: Map Text [ContextFilter]
@@ -277,6 +281,12 @@ levelToContext =
                 , [ctxt| proxy. |]
                 , [ctxt| proxy,abort. |]
                 , [ctxt| booster>failure,abort |]
+                ]
+            )
+        ,
+            ( "EquationWarnings"
+            ,
+                [ [ctxt| booster>(simplification *|function *)>warning |]
                 ]
             )
         ]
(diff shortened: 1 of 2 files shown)


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -232,8 +232,9 @@ fromCache tag t = eqState $ Map.lookup t <$> gets (select tag . (.cache))
 
 logWarn :: LoggerMIO m => Text -> m ()
 logWarn msg =
-    logMessage' $
-        msg <> " For more details, enable full context logging '--log-context \"*\"'"
+    withContext "warning" $ do
+        logMessage $
+            msg <> " For more details, enable full context logging '--log-context \"*\"'"
```
jberthold, line 237: If we don't display the warning by default, we don't have to provide this additional advice (which is a bit "whole-sale" anyway).
```suggestion
    withContext "warning" $ logmessage msg 
```
Or else, we leave `*>warning.` on by default (I think it would make sense) and let downstream dev.s silence it.

booster/library/Booster/CLOptions.hs
```diff
@@ -279,6 +283,12 @@ levelToContext =
                 , [ctxt| booster>failure,abort |]
                 ]
             )
+        ,
+            ( "EquationWarnings"
+            ,
+                [ [ctxt| booster>(simplification *|function *)>warning |]
```
jberthold, line 289: Is `logWarn` it only for `function`? I thought `simplification` may trigger it, too.

booster/library/Booster/Pattern/ApplyEquations.hs
geo2a: I agree that we can remove the context advice. Regarding letting the downstream devs silencing the warnings: I don't think that would work. The best strategy I see is to disable by default and ask them to enable it if the the downstream tool is running with `--verbose`.

goodlyrottenapple (approved):

PR #3940, opened 2024-06-14
NOT READY Compute syntactic transitive closure for select relational symbols

geo2a:
Related to https://github.com/runtimeverification/haskell-backend/issues/3861

This PR adds a routine to compute a syntactic transitive closure if a certain relational symbol in a given set of applications of this symbol. 

See module [Booster.Pattern.PartialOrder](https://github.com/runtimeverification/haskell-backend/blob/georgy%2Fpartial-order-transitive-closure/booster/library/Booster/Pattern/PartialOrder.hs#L22) for the implementation and [Test.Booster.Pattern.PartialOrder(https://github.com/runtimeverification/haskell-backend/blob/georgy%2Fpartial-order-transitive-closure/booster/unit-tests/Test/Booster/Pattern/PartialOrder.hs#L13) for the unit-tests.

## Application

This routine, in conjunction with syntactic inference of `requires` from the path condition, should allow us to solve some side-conditions completely syntactically, i.e. without calling the SMT solver or event the applying equations.

### DIFF
--- a/booster/library/Booster/Definition/Base.hs
+++ b/booster/library/Booster/Definition/Base.hs
@@ -47,6 +47,7 @@ data KoreDefinition = KoreDefinition
     , rewriteTheory :: Theory (RewriteRule "Rewrite")
     , functionEquations :: Theory (RewriteRule "Function")
     , simplifications :: Theory (RewriteRule "Simplification")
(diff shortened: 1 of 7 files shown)


geo2a:
Opened this draft too early.

PR #3949, opened 2024-06-19
Simplify the remainder predicate before checking its satisfiability

geo2a:
Needs https://github.com/runtimeverification/haskell-backend/pull/3943

This is a more radical variant of https://github.com/runtimeverification/haskell-backend/pull/3943 that simplifies the remainder before checking it for satisfiability. If the remainder is satisfiable, we do not simplify it again in `applyRemainder`.


geo2a:
This PR is stale and will anyway be eventually subsumed by https://github.com/runtimeverification/haskell-backend/pull/3960

Issue #3888, opened 2024-05-22
Create a profiler leveraging the contextual logging


geo2a:
## Motivation

When we try to understand why proofs are slow, we only have either very coarse-grained or very find-grained timing data. The coarse-grained data is either wall-clock time of running `kontrol` and other tools, or, at the level of `kore-rpc-booster`, the `--log-stats` option which times the requests. The fine-grained tool is the GHC profiler, which gives us Haskell-level cost-centres.

To better inform our decisions on what to work on to improve performance, we need something that sits in between.

## Functionality

We would like to have a proof-level profiling tool that would allow us to find out, for both Booster and Kore:
- how much time is spent in the steps of rewrite rule and equation application, both for all rules and for specific rules/sets of rules:
  * matching
  * checking the side condition
- how much is spent in calls to the SMT solver

Additionally, for Booster we would like to know:
- how much is spent in calls to LLVM
- simplifier cache utilisation
- Nice to have: statistics about the current configuration, perhaps for certain cells only. To know how big are the terms.

## Data source

We have implemented the contextual JSON logging, which, in conjunction with timestamps, should give us all the information we need. We may need to disable the buffering of the timestamps in `fast-logger` or, of that is not possible, figure out another to provide timestamps.

## Tool workflow and design

The tool's input will be the JSON log file, with every line being a json object. We should design the tool to support streaming the JSONs from the file so that we do not load it all into memory.

The log file will be generated by running the server with context logging enabled. This can be either done directly by the downstream tooling, i.e. `kontrol prove --profile` or by a script that consumes bug reports and runs the server with logging enabled.

The output of the tool will likely be a Markdown/HTML document.

ehildenb:
I would start very coarse grained, and only make it more fine-grained as we discover things. I would start with:

- Time spent selecting rewrite rules (includes rule index calculation, unification, side-condition checking).
- Time spent in old backend.
- Time spent simplifying configurations.

Then, whichever of those cost-centers is the biggest, break down into smaller cost-centers.

I guess the granularity can (or should) be tunable in the tool which reads and processes the logs, rather than having to re-run the proofs. So that's good because you can have very fast iteration on looking at cost-centers even for a long-running proof (if we store the log file).

jberthold:
[...]

PR #3950, opened 2024-06-20
3888 typed context and parser

jberthold:
* Introduces a JSON-enabled type for the log context
* Uses the new type in all  booster logging
* Modifies `count-aborts` to parse the json data using the new type. Performance is on par after switching to TH-derived `FromJSON` instances.

Conversion of `kore-rpc` logging is left for future work.
Part of  #3888


booster/library/Booster/Log.hs
```diff
@@ -100,78 +138,84 @@ logMessage' a =
...
-    m
+withTermContext t@(Term attrs _) m =
+    withContext_ (CLWithId . CtxTerm . ShortId $ showHashHex attrs.hash) $ do
+        withContext CtxKoreTerm $ logMessage t
+        m
 
 withPatternContext :: LoggerMIO m => Pattern -> m a -> m a
 withPatternContext Pattern{term, constraints} m =
-    let t' = foldl' AndTerm term $ Set.toList $ Set.map coerce constraints
+    let t' = foldl' AndTerm term $ Set.toList $ Set.map coerce constraints -- FIXME
```
jberthold, line 158: I noticed this earlier while trying to copy/paste a `KoreJson` pattern from a log into a file for custom requests.
* The output is not correctly sorted (we would need to introduce `#Equals{SortBool, <sort-of-term>}(true, _)` around the predicates)
* That would  mean we cannot use `AndTerm` any more
We should probably go to `KoreJson` first  (externalise the pattern) for the output, but then we need to come up with a stable hash for logging purposes

geo2a: Does it cause issues when internalising? I.e. when you copy/pasted `KoreJson` from a log line and tried to send that to the server it reported an error?

booster/library/Booster/SMT/Interface.hs
```diff
@@ -168,13 +168,13 @@ getModelFor ::
     Map Variable Term -> -- supplied substitution
     io (Either SMTError (Either SMT.Response (Map Variable Term)))
 getModelFor ctxt ps subst
-    | null ps && Map.null subst = Log.withContext "smt" $ do
-        Log.logMessage ("No constraints or substitutions to check, returning Sat" :: Text)
+    | null ps && Map.null subst = Log.withContext Log.CtxSMT $ do
+        Log.logMessage ("No Constraints Or Substitutions To Check, Returning Sat" :: Text)
```
geo2a, line 172: Such Capitalism!

booster/library/Booster/Log.hs
jberthold: The server rejects the predicates having a wrong sort IIRC, but I would have to re-try this to be sure.

dev-tools/kore-rpc-dev/Server.hs
```diff
@@ -230,7 +230,7 @@ main = do
                 kore@KoreServer{runSMT} <-
                     mkKoreServer Log.LoggerEnv{logAction} clOPts koreSolverOptions
                 runBoosterLogger $
-                    Booster.Log.withContext "proxy" $
+                    Booster.Log.withContext CtxKore $
```
geo2a, line 233: :+1:

geo2a (approved):
Monumental work and a major improvement!

booster/library/Booster/SMT/Interface.hs
jberthold: :rofl: Emacs shortcuts going rogue.... I remember pressing the wrong keys (`M^c`) here... corrected everything that the compiler cares about but not the string contents :smile:

jberthold:
```suggestion
        Log.logMessage ("No constraints or substitutions to check, returning Sat" :: Text)
```

goodlyrottenapple (approved):

PR #3951, opened 2024-06-20
Profiteur timing visualisation

goodlyrottenapple:
This PR adds a small utility for turning JSON logs into timing profiles visualised via the https://github.com/jaspervdj/profiteur library

image


booster/library/Booster/CLOptions.hs
```diff
@@ -289,6 +290,15 @@ levelToContext =
                 [ [ctxt| booster>(simplification *|function *)>warning |]
                 ]
             )
+        ,
+            ( "Timing"
+            ,
+                [ [ctxt| request*,booster|kore>rewrite*,success|failure|abort|detail |]
```
goodlyrottenapple, line 296: In order to distinguish requests easily, I've added a new `request <id>` context. @geo2a @jberthold please let me know if this looks ok. If so, will update the docs and the other `levelToContext` mappings accordingly.

dev-tools/timing/Types.hs
```diff
@@ -0,0 +1,272 @@
...
+import Data.Text qualified as Text
+import Data.Vector qualified as Vector
+import GHC.Generics (Generic)
+import Profiteur.Core
+
+data Context = Plain Text | Ref Text Text deriving (Eq, Ord, Show)
+
+instance FromJSON Context where
+    parseJSON (String t) = pure $ Plain t
+    parseJSON (Object o) | [(k, String v)] <- toList o = pure $ Ref (pack $ toString k) v
```
jberthold, line 24: Maybe [use `toText`?](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/src/Data.Aeson.Key.html#toText) 
```suggestion
    parseJSON (Object o) | [(k, String v)] <- toList o = pure $ Ref (toText k) v
```
if you don't want to use the new context type

booster/library/Booster/CLOptions.hs
goodlyrottenapple: I thought we already have a `[ceil]` context?? do we want `[startup][ceil]`? what would be the advantage/difference?

booster/library/Booster/CLOptions.hs
```diff
@@ -289,6 +290,15 @@ levelToContext =
                 [ [ctxt| booster>(simplification *|function *)>warning |]
                 ]
             )
+        ,
+            ( "Timing"
```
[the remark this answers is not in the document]
goodlyrottenapple: I decided to remove `--print-stats` in favour of `-l TimingStats` as this is more uniform with how we invode other logging options, but the `-l Timing` may be too verbose/unsuitable if we just want the stats but not the success/abort logging needed for the new timing tool

booster/library/Booster/CLOptions.hs
goodlyrottenapple: i just noticed we no longer have the ceil context, @jberthold what was the reason for removing it?

booster/library/Booster/CLOptions.hs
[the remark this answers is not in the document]
jberthold: The code to parse `--print-stats` was just for backward compatibility. We can scrap it if that's not important.. (not sure who uses it except myself)

booster/library/Booster/CLOptions.hs
jberthold: As mentioned, the `ceil` had a double meaning, indicating a ceil rule (with ID) and indicating the ceil analysis phase (which could be included in said `startup` phase).
I was planning to amend that `ceil` issue in a follow-up but will wait for your change and what we decide.

jberthold (approved):
LGTM
We need to communicate the context changes to our power users...

booster/library/Booster/CLOptions.hs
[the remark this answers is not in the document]
jberthold: BTW `run-with-tarball.sh` uses `--print-stats`, maybe adjust that on the same PR

PR #3953, opened 2024-06-20
Consider known predicates in Booster's simplifier API

geo2a:
The `runEquationT` function now takes a `Set Predicate`, which are the "predicates known to be true" and are populated into the initial `EquationState`. Their origin is intended to be the rewriting path condition.

There's now a possibility to supply prior knowledge to the following API functions in `Booster.Pattern.Rewrite`:
- `evaluateTerm`
- `simplifyConstraint`

There's also a light change in `evaluatePattern`, which is now passes the pattern's constraints as an argument to `runEuationT` rather then using `pushConstraint`.


geo2a:
KEVM performance:
| Test                                         | georgy-check-constraint-with-known time | master-fb59a8493 time | (georgy-check-constraint-with-known/master-fb59a8493) time
|----------------------------------------------|-----------------------------------------|-----------------------|------------------------------------------------------------
| erc20/ds/balanceOf-spec.k                    | 60.16                                   | 63.91                 | 0.9413237365044594
| erc20/ds/approve-failure-spec.k              | 70.64                                   | 73.41                 | 0.962266721155156
| kontrol/test-storetest-testaccesses-0-spec.k | 84.24                                   | 77.61                 | 1.085427135678392
| TOTAL                                        | 215.04000000000002                      | 214.93                | 1.0005117945377566

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -404,16 +406,15 @@ evaluatePattern ::
...
 evaluatePattern' ::
     LoggerMIO io =>
     Pattern ->
     EquationT io Pattern
-evaluatePattern' pat@Pattern{term, constraints, ceilConditions} = withPatternContext pat $ do
-    pushConstraints constraints
+evaluatePattern' pat@Pattern{term, ceilConditions} = withPatternContext pat $ do
     newTerm <- withTermContext term $ evaluateTerm' BottomUp term
     -- after evaluating the term, evaluate all (existing and
     -- newly-acquired) constraints, once
```
jberthold, line 420: Maybe we should still start empty and push the path condition as the first step.
The code here returns _all_ stored `predicates`  as constraints of the pattern (in line 424) which may or may not be correct depending on how the function was called. This is correct for the current use of the function but might not be correct in all future use cases. 
Alternatively, how about we inline all of `evaluatedPattern'` into `evaluatePattern` (so no mistakes are possible).

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -331,7 +331,7 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
                 "Known true side conditions (won't check):" <+> pretty knownTrue
 
     unclearRequires <-
-        catMaybes <$> mapM (checkConstraint id notAppliedIfBottom) toCheck
+        catMaybes <$> mapM (checkConstraint id notAppliedIfBottom (Set.fromList knownTrue)) toCheck
```
jberthold, line 334: `knownTrue` are just the requires clauses from the rule that syntactically occur in the path condition after substituting. 
I think we want the _entire_ path condition (`pat.constraints`) instead here, and it comes from the subject, so no substitutions are required.
```suggestion
        catMaybes <$> mapM (checkConstraint id notAppliedIfBottom prior) toCheck
```
This means the argument can actually be _fixed_ in the local `checkConstraint` definition but we can also keep them as an argument for clarity.

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -376,7 +376,7 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
                 Set.toList rule.ensures
         trivialIfBottom = RewriteRuleAppT $ pure Trivial
     newConstraints <-
-        catMaybes <$> mapM (checkConstraint id trivialIfBottom) ruleEnsures
+        catMaybes <$> mapM (checkConstraint id trivialIfBottom (Set.fromList knownTrue)) ruleEnsures
```
jberthold, line 379:
```suggestion
        catMaybes <$> mapM (checkConstraint id trivialIfBottom prior) ruleEnsures
```

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -422,17 +422,17 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
...
         Predicate ->
         RewriteRuleAppT (RewriteT io) (Maybe a)
-    checkConstraint onUnclear onBottom p = do
+    checkConstraint onUnclear onBottom knownPredicates p = do
         RewriteConfig{definition, llvmApi, smtSolver} <- lift $ RewriteT ask
         oldCache <- lift . RewriteT . lift $ get
         (simplified, cache) <-
             withContext CtxConstraint $
-                simplifyConstraint definition llvmApi smtSolver oldCache p
+                simplifyConstraint definition llvmApi smtSolver oldCache knownPredicates p
```
jberthold, line 433: As mentioned, one could just use `pat.constraints` here... 
```suggestion
                simplifyConstraint definition llvmApi smtSolver oldCache pat.constraints p
```

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -422,17 +422,17 @@ applyRule pat@Pattern{ceilConditions} rule = withRuleContext rule $ runRewriteRu
...
         -- update cache
         lift . RewriteT . lift . modify $ const cache
-        -- TODO should we keep the traces? Or only on success?
```
jberthold, line 435: :put_litter_in_its_place:

jberthold:
> KEVM performance:
> 
> Test	georgy-check-constraint-with-known time	master-fb59a8493 time	(georgy-check-constraint-with-known/master-fb59a8493) time
> erc20/ds/balanceOf-spec.k	60.16	63.91	0.9413237365044594
> erc20/ds/approve-failure-spec.k	70.64	73.41	0.962266721155156
> kontrol/test-storetest-testaccesses-0-spec.k	84.24	77.61	1.085427135678392
> TOTAL	215.04000000000002	214.93	1.0005117945377566

Maybe this did not have much impact because only the `knownTrue` rule requires were passed ...

booster/library/Booster/Pattern/Rewrite.hs
geo2a: :facepalm:

geo2a:
To my disappointment, there still seems to be no difference:

| Test                               | georgy-check-constraint-with-known time | master-fb59a8493 time | (georgy-check-constraint-with-known/master-fb59a8493) time
|------------------------------------|-----------------------------------------|-----------------------|------------------------------------------------------------
| benchmarks/functional-spec.k       | 596.33                                  | 575.15                | 1.036825176041033
| erc20/ds/balanceOf-spec.k          | 63.79                                   | 60.94                 | 1.0467673121102725
| erc20/ds/approve-success-spec.k    | 92.41                                   | 86.48                 | 1.0685707678075855
| erc20/ds/approve-failure-spec.k    | 71.99                                   | 67.09                 | 1.073036220002981
| erc20/ds/transfer-success-2-spec.k | 94.19                                   | 86.76                 | 1.0856385431074227
| TOTAL                              | 918.71                                  | 876.42                | 1.0482531206499168

jberthold (approved):
LGTM
Would be good to craft a simple integration test that demonstrates the utility...

jberthold:
> ... a simple integration test that demonstrates the utility...

I was trying something related (but not quite sufficient) recently, using code like this:
```
  configuration
    <k> $PGM:Thing ~> .K </k>
    <i> 42 </i>
  rule <k>AA(X) => BB(X) ... </k>
       <i> I </i>
    requires X ==Int I
  rule AA(X) => CC(X) [owise]
```
I tested that we are able to rewrite `AA(X)` to  `BB(X)` if side conditions _imply_ (but not state syntactically) that `X == 42`.
Maybe we could test your change with an additional function `f` and `f(X) ==Int I` in the side condition, and simplification `f(X) => X requires X ==Int 42`. A bit contrived, though...

geo2a:
@jberthold I've added two integration [tests](https://github.com/runtimeverification/haskell-backend/blob/1e56d5cb22a6407e12db2987c3a3045ebf0d4e14/booster/test/rpc-integration/resources/use-path-condition-in-equations.k) that make sure we use the path condition when trying simplifications:
- the first test applies a simplification whose side condition is syntactically present in the path condition. I've checked that this test fails on master;
- the second test is a slight modification of the first one which actually needs SMT to infer the simplification's side condition from the path condition. This test aborts on both master and this PR. It will not abort anymore when we start using the solver when applying equations.

jberthold (approved):

PR #3957, opened 2024-06-21
More precise `[llvm]`/`[cached *]` log context

geo2a:
We emit a log message upon getting the result of an LLVM backend call, but we do not have a message that would give us the precise term we send. This PR fixes that.


goodlyrottenapple:
We now emit correct messages for cached llvm/equation applications:

```
...[term 8d2fa91][kore-term] lengthBytes(_)_BYTES-HOOKED_Int_Bytes(buf("32", VarCONTRACT_ID:SortInt{}))
...[term 8d2fa91][success][cached equations][term 5a54a11][kore-term] "32"
```

jberthold (approved):
:+1:

PR #3958, opened 2024-06-21
Install missing prerequisites for Ubuntu package on release

Baltoli:
This is a simple oversight on my part; I forgot to replicate this stanza from the test to the release job when making #3955.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,6 +65,10 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Install prerequisites
+        run: |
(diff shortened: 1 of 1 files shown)


jberthold (approved):

jberthold:
Was thinking about asking that on #3955 ... there will certtainly be a few more dependencies on (dynamically-loaded) C libraries required to execute _any_ Haskell program, as well as specific other ones we might need for the crypto hook support.
Let's have a try with the current minimal explicit dependencies, though, as it all depends on the actual use case for the created package.

PR #3963, opened 2024-06-25
Add options to the kore term pretty printer

goodlyrottenapple:
Add the following options when pretty printing terms:

* `decode` - decode symbol labels and strip the `Lbl` prefix
* `truncate` - truncate long domain values
* `infix` - try to re-assemble infix notation from the holes in label syntax (may not always be correct). For example:
   ```
   ... Uncertain about match with rule. Remainder: #range( Eq#VarB1:SortBytes{} , Eq#VarS2:SortInt{} , Eq#VarW2:SortInt{} ) == buf("32", lookup(VarCONTRACT-FAKEETH_STORAGE:SortMap{}, keccak( buf("32", VarCALLER_ID:SortInt{}) +Bytes "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL" ))), #range( Eq#VarB1:SortBytes{} , Eq#VarS1:SortInt{} , Eq#VarW1:SortInt{} ) == buf("32", lookup(VarCONTRACT-FAKEETH_STORAGE:SortMap{}, keccak( buf("32", VarCONTRACT_ID:SortInt{}) +Bytes buf("32", keccak( buf("32", VarCALLER_ID:SortInt{}) +Bytes "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\SOH" )) )))
   ```

These can now be passed to the server via the `--pretty-print` flag. E.g. `--pretty-print truncated,decode` will preserve current behaviour (and is set as default when no explicit option is provided)


jberthold (approved):
Slightly worried about the maintenance of this (very neat!) code... 
@geo2a what do you think?

The other option would be to create a separate `Pretty` class that takes parameters that are brought in by  `PrettyWithModifiers` here

geo2a (approved):
I think this should be fine to maintain @jberthold. Thanks @goodlyrottenapple!

PR #3970, opened 2024-07-09
Remove directory name assumption in `kore-rpc-client`

geo2a:
- when running a tarball, consider the requests from all directories, not just the one prefixed with `rpc_`
- allow creating parent directories when unpacking a tarball


jberthold (approved):
LGTM
The next change is already around the corner... https://github.com/runtimeverification/k/pull/4509

Issue #3959, opened 2024-06-24
Z3 zombie processes during long proofs


jberthold:
During longer proofs, a number of `z3` processes accumulate as zombie children of the `kore-rpc-bosoter` server.

Example:
```
$ ps auxw | grep kore-rpc-booster | sed -e 's/jost  *\([0-9]*\) .*$/\1/' | while read pid; do pstree $pid; done
kore-rpc-booste─┬─51*[z3]
                ├─z3───{z3}
                └─6*[{kore-rpc-booste}]
kore-rpc-booste─┬─51*[z3]
                └─6*[{kore-rpc-booste}]
```

A code inspection did not unearth any immediate reason for that - possibly the reasons lie within the `smtlib-backends-process` library.

jberthold:
Experiments on branch https://github.com/runtimeverification/haskell-backend/compare/3959-z3-zombies did not provide an immediate remedy . The root cause appears to lie in lazy I/O delaying the process termination until server shut-down for some reason.

Using the `SMTLib.Backends.Z3` backend produced a `SEGFAULT` in the `llvm-integration` test (branch is tagged and pushed)

jberthold:
After this change, no further zombie processes were observed. :heavy_check_mark:

PR #3977, opened 2024-07-12
Use an IORef to hold a reference to the solver/solverClose

goodlyrottenapple:
Fixes (hopefully) #3959 by making the `solver`/`solverClose` handles into an `IORef`. This way, we can be (reasonably) sure that when we write the new handles into the `IORef`, there will be no other references held to the old ones elsewhere, which was previously preventing the old z3 instance from terminating, leaving a zombie process.


jberthold (approved):
Glorious hack! :sparkles:

PR #3980, opened 2024-07-16
House-keeping: adjust `-l Abort` log switch, update llvm.k test definition

jberthold:
* Adds the new `request <N>` prefix  in front of  the `proxy` context to the `-l Aborts` short-hand
* updates `llvm.k` to use `symbol(Bla)` instead of `klabel(Bla), symbol` to avoid warnings


geo2a (approved):
This is very helpful, thanks for taking care of that!

goodlyrottenapple (approved):

PR #3981, opened 2024-07-16
Filter the requires of rewrites again, after simplifications

### LINKED ISSUE #3886: Filter requires clause a second time when applyin rewrite rules in Booster
When applying a rewrite rule in Booster, we check if any of the conjuncts of the rule's requires clause is already present in the path condition verbatim. If so, we do not waste time checking it, as we already know it is true.

We should add a second filtering pass after we have simplified the requires clause, as we do in https://github.com/runtimeverification/haskell-backend/pull/3870

geo2a:
Fixes #3886


geo2a:
No difference in KEVM performance

jberthold (approved):
LGTM
We could log the known constraints in JSON instead of text but I don't see why someone would want to use them ...

PR #3989, opened 2024-07-22
Recover from `TooManyIterations` when simplifying during rewriting

geo2a:
Booster's rewriting algorithm, upon getting stuck (or aborting due to `RuleApplicationUnclear`), will try so simplify the configuration once and retry applying rules. When we cannot not finish simplifying the term due to a `TooManyIterations`, we throw the partial result away and fail.

This PR handles the `TooManyIterations` exception and attempts applying rewrite rules to the partially simplified term.

### DIFF
--- a/booster/library/Booster/Pattern/ApplyEquations.hs
+++ b/booster/library/Booster/Pattern/ApplyEquations.hs
@@ -331,18 +331,14 @@ iterateEquations direction preference startTerm = do
             config <- getConfig
             currentCount <- countSteps
             when (coerce currentCount > config.maxIterations) $ do
-                -- FIXME if this exception is caught in evaluatePattern',
-                --       then CtxAbort is a wrong context for it.
-                --       We should emit this log  entry somewhere else.
-                withContext CtxAbort $ do
-                    logWarn $
-                        renderOneLineText $
-                            "Unable to finish evaluation in" <+> pretty currentCount <+> "iterations."
-                    withContext CtxDetail $
-                        getPrettyModifiers >>= \case
-                            ModifiersRep (_ :: FromModifiersT mods => Proxy mods) ->
-                                logMessage . renderOneLineText $
-                                    "Final term:" <+> pretty' @mods currentTerm
+                logWarn $
+                    renderOneLineText $
+                        "Unable to finish evaluation in" <+> pretty currentCount <+> "iterations."
+                withContext CtxDetail $
+                    getPrettyModifiers >>= \case
+                        ModifiersRep (_ :: FromModifiersT mods => Proxy mods) ->
+                            logMessage . renderOneLineText $
+                                "Final term:" <+> pretty' @mods currentTerm
                 throw $
                     TooManyIterations currentCount startTerm currentTerm
             pushTerm currentTerm
--- a/booster/library/Booster/Pattern/ApplyEquations.hs
+++ b/booster/library/Booster/Pattern/ApplyEquations.hs
@@ -234,6 +234,9 @@ popRecursion = do
             throw $ InternalError "Trying to pop an empty recursion stack"
         else eqState $ put s{recursionStack = tail s.recursionStack}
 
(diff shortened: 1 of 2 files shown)


geo2a:
No difference on KEVM as of commit `49516fc`

geo2a:
No significant difference in Kontrol performance either.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -431,13 +440,25 @@ evaluatePattern' ::
...
+    -- i.e. not in a recursive evaluation of a side-condition,
+    -- it is safe to keep the partial result and ignore the exception.
+    -- Otherwise we would be throwing away useful work.
+    keepTopLevelResults :: LoggerMIO io => EquationFailure -> EquationT io Term
+    keepTopLevelResults = \case
+        err@(TooManyIterations recursionDepth _ _ partialResult) ->
+            case recursionDepth of
+                0 -> pure partialResult
+                _ -> throw err
+        err -> throw err
```
jberthold, line 461: After checking the code where we recurse into the evaluation, https://github.com/runtimeverification/haskell-backend/blob/034078d5f2d5f68dada7fcf98a5211eae51bae8f/booster/library/Booster/Pattern/ApplyEquations.hs#L931-L939
I think it is probably fine without this additional `catch_`. Of course it is safer the way you propose it.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -234,6 +235,9 @@ popRecursion = do
             throw $ InternalError "Trying to pop an empty recursion stack"
         else eqState $ put s{recursionStack = tail s.recursionStack}
 
+getRecusionDepth :: LoggerMIO io => EquationT io Int
+getRecusionDepth = (length . (.recursionStack)) <$> getState
```
jberthold, line 239:
```suggestion
getRecursionDepth :: LoggerMIO io => EquationT io Int
getRecursionDepth = (length . (.recursionStack)) <$> getState
```

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -331,6 +335,10 @@ iterateEquations direction preference startTerm = do
             config <- getConfig
             currentCount <- countSteps
             when (coerce currentCount > config.maxIterations) $ do
+                currentRecursionDepth <- getRecusionDepth
```
jberthold, line 338:
```suggestion
                currentRecursionDepth <- getRecursionDepth
```

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -331,6 +335,10 @@ iterateEquations direction preference startTerm = do
...
+                -- FIXME if this exception is caught in evaluatePattern',
+                --       then CtxAbort is a wrong context for it.
+                --       We should emit this log  entry somewhere else.
                 withContext CtxAbort $ do
                     logWarn $
                         renderOneLineText $
```
jberthold, line 344: We could probably log this  in `evaluatePattern'` where we decide whether or not to use the final term.
The warning will still show up for recursive evaluations where we catch the exception, too. Maybe not useless to notify the user about it, though? (so we could just use `CtxWarn`).

jberthold (commented):
Test code not adapted yet.

booster/library/Booster/Pattern/ApplyEquations.hs
geo2a: We still need the `catch_` to get the simplified term, but I do agree that the recursion depth check is redundant. I've removed it.

booster/library/Booster/Pattern/ApplyEquations.hs
geo2a: I've removed `CtxAbort`, so these will show up as warnings now.

jberthold (approved):
LGTM

PR #3996, opened 2024-07-24
Update `kore-rpc-client` to use `tar-0.6.3`

geo2a:
With https://github.com/runtimeverification/k/pull/4548, the bug reports will be generated in GNU format of tar files instead of the default POSIX.1-2001 pax format. This is necessary because:
- the filenames in the bug reports are often longer than the format-portable restriction of 256 characters
- the Haskell [library](https://hackage.haskell.org/package/tar-0.6.3.0) we use to work with tar does not support the pax format. However, it does support the GNU format which also allows long names.

With this change, we get to process bug reports with long names.


booster/tools/rpc-client/RpcClient.hs
```diff
@@ -505,20 +505,24 @@ runTarball common (Just sock) tarFile keepGoing runOnly compareDetails = do
             | ".tar.bz2" `isSuffixOf` takeExtensions tarFile = Tar.read . BZ2.decompress
             | otherwise = Tar.read
 
-    containedFiles <- unpackTar <$> BS.readFile tarFile
-    let checked = Tar.checkSecurity containedFiles
```
geo2a, line 509: No need to run `Tar.checkSecurity` as `unpack` already does it for us.

geo2a:
Oops, forgot about `tarball-compare`... will need to fix it before merging.

dev-tools/tarball-compare/Main.hs
```diff
@@ -118,30 +119,40 @@ main =
                     entries <- Tar.pack "." $ defFile : files
                     -- need to force the tar entries, withCurrentDirectory is not retained
                     mapM_ (`seq` pure ()) entries
-                    pure $ foldr Tar.Next Tar.Done entries
+                    error "Unimplemented: can't make it typecheck!!" -- pure $ foldr Tar.Next Tar.Done entries
```
jberthold, line 122: ...maybe we should consider removing this tool?
It relies upon the directory names and request IDs in the tar file being equal for the same proof - which is not true any more since recent changes were made (to both directory names and request IDs).
I haven't used the tool in quite a while, it was for conformance checks between `booster` and `kore-rpc`.

jberthold (approved):
Can merge this, but we should either fix or delete `tarball-compare`.

dev-tools/tarball-compare/Main.hs
geo2a: Let's delete, can always resurrect if needed!

PR #4003, opened 2024-07-29
Make SMTContext mandatory and merge Unknown with ReasonUnknown

goodlyrottenapple:
I have changed my mind and think we should have `SMTContext` mandatory and not an optional, as the current design makes it way too easy to give a default `Nothing` value when extending the code-base in the future. This then has the potential to re-introduce a similar bug. Another advantage tot his approach is the removal of the ugly pattern matching present each time we actually want to use the solver. 
Instead, in this PR, the `Maybe` is pushed into `SMTContext`, where the dummy/empty instance now has to be consciously built using the new `noSolver` function. 
The fallout is fairly minimal and we confide the logic of not having a solver to the SMT interface. I.e. the only places where not having a solver has a different behavior are:
* `hardReset` -  becomes a no-op when no solver handle is present
* `runCmd` -> always returns the `Unknown` response when no solver exists
This eliminates duplicate instances of handling an unknown response vs not having a solver, since in all instances of the current code-base, they are functionally the same.


booster/library/Booster/SMT/Runner.hs
```diff
@@ -180,22 +180,27 @@ runCmd :: forall cmd io. (SMTEncode cmd, LoggerMIO io) => cmd -> SMT io Response
...
-    when (isError result) $
-        logMessage $
-            "SMT solver reports: " <> pack (show result)
-    pure result
-  where
-    isError :: Response -> Bool
-    isError Error{} = True
-    isError _other = False
+    case ctxt.mbSolver of
+        Nothing -> pure $ Unknown Nothing
```
geo2a, line 184:
```suggestion
        Nothing -> pure $ Unknown (Just "server started without SMT solver")
```
Let's make return a more explicit reason here.

geo2a (approved):
I like this change, I think it turned out to be quite a nice clean-up!

Hope @jberthold likes it too!

booster/library/Booster/SMT/Runner.hs
goodlyrottenapple: Great idea, thanks!

booster/library/Booster/SMT/Interface.hs
```diff
@@ -103,6 +105,18 @@ initSolver def smtOptions = Log.withContext Log.CtxSMT $ do
     Log.logMessage ("Successfully initialised SMT solver with " <> (Text.pack . show $ smtOptions))
     pure ctxt
 
+noSolver :: MonadIO io => io SMT.SMTContext
+noSolver = do
```
jberthold, line 109: Maybe put a comment here to describe the behaviour?

booster/unit-tests/Test/Booster/Pattern/ApplyEquations.hs
```diff
@@ -94,17 +96,17 @@ test_evaluateFunction =
...
   where
-    eval direction =
-        unsafePerformIO
-            . runNoLoggingT
-            . (fst <$>)
-            . evaluateTerm direction funDef Nothing Nothing mempty
+    eval direction t =
+        unsafePerformIO $ do
+            ns <- noSolver
+            runNoLoggingT $ fst <$> evaluateTerm direction funDef Nothing ns mempty t
```
jberthold, line 102: (off-topic)
The `{-# NOINLINE ... #-}` pragmas are certainly a good idea.
However, we can iinstead completely get rid of the `unsafePerformIO`  by making `eval` an `IO` monadic function. The tests that use it run in `IO` anyway! (ditto for the other ones below)

booster/unit-tests/Test/Booster/Pattern/Rewrite.hs
```diff
@@ -243,10 +244,10 @@ rulePriority =
...
 runWith t =
     second fst $
-        unsafePerformIO
-            ( runNoLoggingT $
-                runRewriteT NoCollectRewriteTraces def Nothing Nothing mempty (rewriteStep [] [] $ Pattern_ t)
-            )
+        unsafePerformIO $ do
+            ns <- noSolver
+            runNoLoggingT $
+                runRewriteT NoCollectRewriteTraces def Nothing ns mempty (rewriteStep [] [] $ Pattern_ t)
```
jberthold, line 250: (off-topic) same here, all callers of `runWith` are actually in `IO` anyway...

jberthold (approved):
The changes on the caller side are actually very neat - good clean-up. Let's merge it.

PR #4007, opened 2024-07-30
Booster: Avoid var. names from input substitution when creating new ones

### LINKED ISSUE #4006: Name collision for existential variables
As [reported on slack](https://runtimeverification.slack.com/archives/CC360GUTG/p1721935740648379), the server is sometimes reusing variable names (to be precise, number-suffixed versions thereof) as existential variables. 
Collisions with previously-used variable names have been observed, where the previous use was only in the path condition, not in the configuration. This can lead to substituting wrong values into the configuration when the variable name is reused but already has a substitution in the path condition.

jberthold:
When rewriting with rules that have existentials on the RHS, booster needs to rename these existential variables to avoid collisions.

However, the input is pre-processed to recognise predicates of the form `X ==K expr` and apply them (after checking/vetting) as a _substitution_ to the configuration and (other) path conditions.

As a result, the variable names in the substitution are not present in the pattern any more when rewriting. The substitution is added back (as additional information) in the output, so subsequent simplification steps will apply it again, falsely, if variable names from the substitution are reused during rewriting.

As a solution, we supply the "forbidden variable names" to the rewriter. Alternatively, we could stop including the applied substitution in the output, but this would create a disparity between `booster` and `kore-rpc`.

Fixes #4006

### DIFF
--- a/booster/test/rpc-integration/test-questionmark-vars/README.md
+++ b/booster/test/rpc-integration/test-questionmark-vars/README.md
@@ -92,3 +92,31 @@ transition rules that unconditionally introduce a fresh variable in the configur
    <a-state> ?X2          </a-state>
    <b-state> ?X1          </b-state>
 ```
+
+5) _ques-and-substitution_
+
+   If a substitution `X ==K b` is supplied, it will be applied before rewriting.
+   However, the variable `X` can still not be used because the substitution will also be returned in the result (for information).
+   Therefore, the newly created variable cannot be `X0`.
+
+   _Input:_
+
+```
+   <k> #setAStateSymbolic </k>
+   <a-state> ?X0      </a-state>
+   <b-state> ?X       </b-state>
+    #And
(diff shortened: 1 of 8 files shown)


booster/test/rpc-integration/test-questionmark-vars/README.md
```diff
@@ -92,3 +92,31 @@ transition rules that unconditionally introduce a fresh variable in the configur
...
+   _Expected:_
+
+```
+   <k> .K                 </k>
+   <a-state> ?X0          </a-state>
+   <b-state> b            </b-state>
+    #And
+   ?X ==K b #And ?X0 ==K ?X1 #And ?X2 ==K ?X3 // substitutions
+    #And 
+   condition(?X3)
```
goodlyrottenapple, line 121:
```suggestion
   condition(?X4)
```

goodlyrottenapple (approved):
Think there is a small typo in the readme. Otherwise looks good

geo2a (approved):
Thanks for fixing this!

PR #4009, opened 2024-07-30
Allow matching empty sets

geo2a:
This PR adds rudimentary matching of internalized sets: only empty sets are allowed to match, and an indeterminate result is returned for non-empty sets.

### DIFF
--- a/booster/unit-tests/Test/Booster/Pattern/MatchRewrite.hs
+++ b/booster/unit-tests/Test/Booster/Pattern/MatchRewrite.hs
@@ -32,6 +32,7 @@ test_match_rewrite =
         , sorts
         , injections
         , internalLists
+        , internalSets
         , internalMaps
         ]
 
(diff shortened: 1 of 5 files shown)


goodlyrottenapple (approved):
LGTM perhaps consider adding a simple unit test here: https://github.com/runtimeverification/haskell-backend/blob/e5a921f5a1f2dfd9cc61402ddc31ab2eb2d50761/booster/unit-tests/Test/Booster/Pattern/MatchRewrite.hs#L35 and in the other two relevant match test suites...tho given we only handle the most trivial case, not a deal breaker.

booster/unit-tests/Test/Booster/Pattern/MatchRewrite.hs
```diff
@@ -376,6 +377,17 @@ internalLists =
...
     klist = KList testKListDef
 
+internalSets :: TestTree
+internalSets =
+    testGroup
+        "Internal sets"
+        [ test
+            "Can match an empty set with itself"
+            emptySet
+            emptySet
```
jberthold, line 387: Need to export `emptySet` from `Test.Booster.Pattern.InternalCollections`
Rest LGTM, we just have to fix this test.

PR #4010, opened 2024-07-31
Booster interim simplification, refactor rewrite and logging parameters

jberthold:
* adds option `--simplify-each DEPTH` which runs a booster-only simplification every `DEPTH` steps
* refactors logging-related options into a dedicated `LogOptions` type
* refactors `performRewrite` and `runRewriteT` to use a pre-made `RewriteConfig` (extended by parameters required only in `performRewrite`


jberthold:
Results for KEVM do not show much speedup /slowdown when using the option.
(`--no-fallback-simplify` and `--no-post-exec-simplify` have larger effect) 

kevm-with-simplification-tweaks

booster/library/Booster/JsonRpc.hs
```diff
@@ -153,7 +154,17 @@ respond stateVar request =
...
+                                doTracing
+                                def
+                                mLlvmLibrary
+                                solver
+                                substVars
+                                mbDepth
+                                cutPoints
+                                terminals
+                                rewriteOpts.interimSimplification
+                                substPat
```
jberthold, line 167: I think this is a bit too much in terms of parameters... 
Should we just use the `RewriteConfig` type (extended by `cutPoints`, `terminals`, `mbDepth` and `interimSimplificatoin`)?

scripts/performance-tests-kevm.sh
```diff
@@ -106,16 +106,30 @@ feature_shell "cd kevm-pyk && poetry run pytest src/tests/integration/test_prove
...
 mkdir -p $SCRIPT_DIR/logs
 
+# use special options if given, but restore KORE_RPC_OPTS afterwards
+if [ ! -z "${FEATURE_SERVER_OPTS}" ]; then
+    echo "Using special options '${FEATURE_SERVER_OPTS}' via KORE_RPC_OPTS"
+    if [ ! -z "${KORE_RPC_OPTS:-}" ]; then
+        PRIOR_OPTS=${KORE_RPC_OPTS}
+    fi
+    export KORE_RPC_OPTS=${FEATURE_SERVER_OPTS}
+fi
```
jberthold, line 116: I put this in ad-hoc, it is helpful when new optional features are introduced (here: running with `--simplify-each`) but those options won't show in the log files for the timing, unfortunately.

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -801,108 +792,113 @@ performRewrite doTracing def mLlvmLibrary smtSolver varsToAvoid mbMaxDepth cutLa
...
-                                emitRewriteTrace $ RewriteStepFailed failure
+
+        case counter of
+            c
+                | depthReached c -> do
+                    logDepth $ "Reached maximum depth of " <> maybe "?" showCounter mbMaxDepth
+                    (if wasSimplified then pure else simplifyResult pat') $ RewriteFinished Nothing Nothing pat'
+                | counter > 0
+                , not wasSimplified
+                , maybe False ((== 0) . (counter `mod`)) mbSimplify -> do
```
geo2a, line 803: Please consider factoring this out into a function, I think it's a little too much going on here.

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -801,108 +792,113 @@ performRewrite doTracing def mLlvmLibrary smtSolver varsToAvoid mbMaxDepth cutLa
...
-                            if wasSimplified
-                                then logMessage msg >> pure (RewriteAborted failure pat')
-                                else withSimplified pat' msg (pure . RewriteAborted failure)
+                                simplifyResult pat' aborted
+                            -- if unification was unclear and the pattern was
+                            -- unsimplified, simplify and retry rewriting once
+                            Left failure@(RuleApplicationUnclear rule _ remainder)
+                                | not wasSimplified -> do
+                                    emitRewriteTrace $ RewriteStepFailed failure
+                                    -- simplify remainders, substitute and rerun.
```
geo2a, line 876: I've just noticed that this comment is irrelevant --- not sure how it got here.

geo2a (approved):
Nice! I left one minor comment.

booster/library/Booster/Pattern/Rewrite.hs
jberthold: That comment is old: https://github.com/runtimeverification/haskell-backend/pull/4010/files/cd2a4dd8a890cead5e7612713422ab7e93e9fc01#diff-79733ce60c52de929f411c95349bc4670bf6aa92709665c7bf313455787b9bbeL880-L881
I think it comes from an abandoned experiment with matching remainders that was backed out some time last year.

PR #4015, opened 2024-08-06
Remove `"log-timing"` and `"log-fallbacks"` RPC log options

geo2a:
This options are unused and all ther use cases are subsumed by the `-l Timing` and `-l Aborts` non-RPC log option.


booster/tools/booster/Proxy.hs
```diff
@@ -92,16 +90,16 @@ respondEither cfg@ProxyConfig{boosterState} booster kore req = case req of
                     LogSettings
                         { logSuccessfulRewrites = execReq.logSuccessfulRewrites
                         , logFailedRewrites = execReq.logFailedRewrites
-                        , logFallbacks = execReq.logFallbacks
-                        , logTiming = execReq.logTiming
+                        , logFallbacks = Nothing
+                        , logTiming = Nothing
```
goodlyrottenapple, line 94: Can we get rid of this field or is there a reason to hold off?

geo2a: Oops, forgot to push!

goodlyrottenapple (approved):

PR #4017, opened 2024-08-06
Refactor booster smt

goodlyrottenapple:
The following changes were made:

* SMT response in the hihg-level interface always returns a `Text` reason in the `Unknown` constructor
* All the main functions, i.e. `isSat`, `checkPredicates` and `getModelFor` throw errors as Exceptions and return an `IsSatResult`, as the only error we are catching is the one thrown when the SMT result is unknown
*


booster/library/Booster/SMT/Interface.hs
```diff
@@ -156,28 +150,88 @@ translatePrelude def =
...
+    setTimeout timeout = do
+        Log.logMessage $ "Setting SMT timeout to: " <> show timeout
+        runCmd_ $ SetTimeout timeout
 
 finaliseSolver :: Log.LoggerMIO io => SMT.SMTContext -> io ()
 finaliseSolver ctxt = do
     Log.logMessage ("Closing SMT solver" :: Text)
     destroyContext ctxt
 
+data IsSatResult unknown a = IsSat a | IsUnsat | IsUnknown unknown
```
geo2a, line 178: Let's add another datatype here:

```
data IsValidResult unknown a = IsValid a | IsInvalid | IsUnknown unknown
```

and use it as the return type of `checkPredicates`.

booster/library/Booster/SMT/Interface.hs
```diff
@@ -386,32 +383,17 @@ checkPredicates ctxt givenPs givenSubst psToCheck
...
+            (Unsat, Unsat) -> pure $ IsUnknown Nothing -- defensive choice for inconsistent ground truth
             (Sat, Sat) -> do
                 Log.logMessage ("Implication not determined" :: Text)
-                pure Nothing
-            (Sat, Unsat) -> pure . Just $ True
-            (Unsat, Sat) -> pure . Just $ False
-            (Unknown reason, _) -> retry smtGiven sexprsToCheck transState reason
-            (_, Unknown reason) -> retry smtGiven sexprsToCheck transState reason
+                pure $ IsUnknown Nothing
+            (Sat, Unsat) -> pure $ IsSat ()
```
geo2a, line 390: We'd want to return `IsValid` rather than `IsSat` here. I think warrants a new datatype.

geo2a (commented):
I like this refactoring! I'm happy to approve once we have a more clear result type for `checkPredicates`.

booster/library/Booster/SMT/Interface.hs
goodlyrottenapple: how about making them pattern synonyms so that we can reuse `IsUnknown`? i.e.

```haskell
type IsValidResult = IsSatResult (Maybe Text) ()

pattern IsValid, IsInvalid :: IsValidResult
pattern IsValid = IsSat ()
pattern IsInvalid = IsUnsat

{-# COMPLETE IsValid, IsInvalid, IsUnknown #-}
```

geo2a: I'd still prefer a separate datatype, but this is an acceptable option too. @jberthold do you have a strong opinion?

geo2a (approved):

jberthold (approved):
LGTM, two small suggestions.

booster/library/Booster/SMT/Interface.hs
```diff
@@ -156,81 +157,65 @@ translatePrelude def =
 
 checkPrelude :: Log.LoggerMIO io => SMT io ()
 checkPrelude = do
-    runCmd_ $ SetTimeout defaultSMTOptions.timeout
+    -- set large default timeout value for checking the prelude
+    setTimeout defaultSMTOptions.timeout
```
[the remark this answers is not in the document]
goodlyrottenapple: The default is `125`. Maybe it isn't that large actually. Either way, I was puzzled by this behaviour because the current code always overrides the user supplied value with `125`, which is what I have done here as well. Perhaps we should just use the user defined value? @geo2a what do you think?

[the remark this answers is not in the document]
geo2a: Yes, let's use the user defined value here.

goodlyrottenapple:
I will merge this and we can address the prelude timeout in a separate PR if need be.

PR #4021, opened 2024-08-08
Pass args to the smt solver via `--smt-arg`

goodlyrottenapple:
This change will let us call `z3` with `smt.ematching=false` in booster, which seems to help with non-linear arithmetic.

### DIFF
--- a/booster/library/Booster/CLOptions.hs
+++ b/booster/library/Booster/CLOptions.hs
@@ -366,7 +366,7 @@ parseSMTOptions =
                         ( metavar "SMT_ARG"
                             <> long "smt-arg"
                             <> help
-                                ( "Argumetns to be passed to the SMT solver process"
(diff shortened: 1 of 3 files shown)


booster/library/Booster/CLOptions.hs
```diff
@@ -361,6 +361,15 @@ parseSMTOptions =
                                 \Example: '(check-sat-using smt)' (i.e., plain 'check-sat')"
                         )
                     )
+                <*> many
+                    ( strOption
+                        ( metavar "SMT_ARG"
+                            <> long "smt-arg"
+                            <> help
+                                ( "Argumetns to be passed to the SMT solver process"
```
jberthold, line 369:
```suggestion
                                ( "Arguments to be passed to the SMT solver process"
```
:smirk:

jberthold (approved):

geo2a (approved):

PR #4022, opened 2024-08-08
WIP Syntactic simplifications

### LINKED ISSUE #3861: Improving pre-SMT reasoning capabilities
In the recent proofs, I have often observed the following pattern of reasoning:
1. We need `C1: X <Int pow256` for a rule/simplification to fire;
2. We have `C2: X <Int pow96`, say, in the path constraints, so `C1` holds;
3. Our simplifications aren't powerful enough to understand that `C1` holds, so rule/simplification does not get applied in Booster and then (I think?) we either fall back to Kore which uses Z3 (if we're dealing with a rule) or we do nothing (if we're dealing with a simplification).

I believe that we could/should take care of these cases before they reach the SMT solver, and I would like to understand/discuss how this could be done modularly and where it should be done. 

One way of doing this for this specific case would be to write simplifications in which I could tell the engine to syntactically match the path condition, rather than semantically, along the lines of:
```
rule A <Int B => true
  requires A <Int C
   andBool C <=Int B
  [simplification, concrete(B, C), match(C)]
```
where `match(C)` would mean 'match the path constraints syntactically to learn `C`, possibly also specifying which conjunct of the `requires` clause to use for the match (`match(C, 1)`, for instance). In this example, we couldn't use `C <=Int B` anyway because both `B` and `C` are concrete, so that constraint would be fully resolved.

geo2a:
Fixes #3861 https://github.com/runtimeverification/haskell-backend/issues/3861


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -942,74 +920,246 @@ applyEquation term rule =
...
+        ((EquationT io () -> EquationT io ()) -> EquationT io (), ApplyEquationFailure)
+        (EquationT io)
+        ()
+checkRequiresSyntactically syntacticRequires currentSubst knownPredicates restRequires' = do
+    ModifiersRep (_ :: FromModifiersT mods => Proxy mods) <- getPrettyModifiers
+    case syntacticRequires of
+        [] ->
+            unless (null restRequires') $
+                -- no more @syntacticRequires@, but unresolved conditions remain: abort
+                throwRemainingRequires currentSubst restRequires'
```
goodlyrottenapple, line 980: why do we throw here?? i would expect at the base case to simply check the regular requires clauses...

geo2a: As I understand, the simplifications marked as `syntactic` must be only applied "syntactically", which means their side condition must not go to Z3. If we do not stop here, we will proceed to trying to discharge a syntactic simplification as a regular one, and as far as I understand, we do not want that. We want them either to succeed fast or to fail fast. @PetarMax do you agree?

geo2a: Regarding the code: we never want to transfer control to `checkRequiresSyntactically` from `checkRequiresSemantically`.

geo2a: So the semantics of the `syntactic()` attribute is:
- the `simplification` is syntactic, Z3 is never called
- the clause 1 is the driver clause which learns its variables from the known truth, and the the substitution for these learned variables is applied to all other clauses

I.e. the whole `simplification` is `syntactic`, not a clause of a simplification

PetarMax:
So, for me, the clauses marked as `syntactic` should be checked syntactically, the other ones should be checked semantically.

booster/library/Booster/Pattern/ApplyEquations.hs
PetarMax: So, for me, the *clauses* marked as syntactic should be checked syntactically, the other ones should be checked semantically. A simplification can have both syntactic and semantic clauses.

This still does mean, I think, that `checkRequiresSyntactically` should not call `checkRequiresSemantically`.

geo2a:
Thanks @PetarMax, we'll modify the code to do that then.

booster/library/Booster/Pattern/ApplyEquations.hs
PetarMax: But also, all syntactics should be handled first, I think.

PR #4025, opened 2024-08-09
Add `--smt-arg` to kore

### LINKED ISSUE #4024: Pass `--smt-arg` to Kore's SMT solver
https://github.com/runtimeverification/haskell-backend/pull/4021 adds the `--smt-arg` option which allows passing arguments to Z3. However, `kore-rpc-booster` only passes them to Booster. We need to pass the same arguments to Kore.

goodlyrottenapple:
Fixes #4024


geo2a (approved):

kore/app/share/GlobalMain.hs
```diff
@@ -581,14 +582,15 @@ execute options metadataTools lemmas worker =
...
-    KoreSolverOptions{timeOut, rLimit, resetInterval, prelude, solver} =
+    KoreSolverOptions{timeOut, rLimit, resetInterval, prelude, solver, args} =
         options
     config =
         SMT.defaultConfig
             { SMT.timeOut = timeOut
             , SMT.rLimit = rLimit
             , SMT.resetInterval = resetInterval
             , SMT.prelude = prelude
+            , SMT.arguments = args <> SMT.Config.arguments SMT.defaultConfig
```
jberthold, line 593: Maybe we should only use the default if the `args` are empty? This would enable users to get rid of defaults that we may set in the future.
```suggestion
            , SMT.arguments = if null args then SMT.Config.arguments SMT.defaultConfig else args
```

jberthold (approved):
LGTM, one suggestion

PR #4026, opened 2024-08-09
Booster implies endpoint improvements

### LINKED ISSUE #3941: Regression in the `implies` endpoint of `booster-dev`
Related:
* runtimeverification/evm-semantics#2483

K version: v7.1.3

From the `kevm-pyk` directory of `runtimeverification/evm-semantics`:

```
$ make test-integration PYTEST_ARGS='-n0 -k benchmarks/storagevar02-overflow-spec.k --use-booster-dev --bug-report'
```

Bug report: [bug-report.zip](https://github.com/user-attachments/files/15841006/bug-report.zip)

In order to produce the bug report I had to apply a patch to `pyk`:
- Make JSON RPC request ID-s random. For some reason, multiple requests and repsonses got the same file path in the archive, so half of them were not written successfully. Will have to debug this properly.

Once fixed, the test (`tests/specs/benchmarks/storagevar02-overflow-spec.k`) should be removed from the failing list: https://github.com/runtimeverification/evm-semantics/blob/master/tests/failing-symbolic.haskell-booster-dev

goodlyrottenapple:
Fixes  #3941 along with a host of tests from the pyk testsuite, which exercise corner cases, such as `#Bottom` being sent as antecedent or inconsistent conditions. Whilst these haven't really been observed in practice, passing these tests should give is more confidence that the booster does the same thing as kore.


booster/library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -619,10 +666,11 @@ isTermM pat = case pat of
    the arguments as a list. The top-level sorts of the 'And' nodes are
    ignored, no checks are performed.
 -}
-explodeAnd :: Syntax.KorePattern -> [Syntax.KorePattern]
-explodeAnd Syntax.KJAnd{patterns} =
-    concatMap explodeAnd patterns
-explodeAnd other = [other]
+explodeAnd :: Syntax.KorePattern -> NE.NonEmpty Syntax.KorePattern
+explodeAnd Syntax.KJAnd{patterns} = case patterns of
+    [] -> error "malformed KJAnd"
```
geo2a, line 671: I think the benefit of having this function returning `NonEmpty` and exploding on an invalid And is not big, as we only use the non-empty in `isTop`/`isBottom`. I think we can just as well use the normal list there and avoid the call to error?

booster/library/Booster/JsonRpc.hs
```diff
@@ -417,85 +420,154 @@ respond stateVar request =
...
+                solver <- maybe (SMT.noSolver) (SMT.initSolver def) mSMTOptions
                 -- internalise given constrained term
-                let internalised =
-                        runExcept . internalisePattern DisallowAlias CheckSubsorts Nothing def . fst . extractExistentials
+                let internalised korePat' =
+                        let (korePat, existentials) = extractExistentials korePat'
+                         in runExcept $
+                                internalisePatternOrTopOrBottom DisallowAlias CheckSubsorts Nothing def existentials korePat
+
+                    checkImplies patL substitutionL unsupportedL existsL patR substitutionR unsupportedR existsR = do
```
geo2a, line 430: I agree with Jost that this function should be moved to a separate module.

jberthold (approved):
A few question marks but a big step forward either way.
We can merge and follow up or adjust and then merge.

PR #4027, opened 2024-08-10
Adjusting default timeout

PetarMax:
The PR sets the default timout to a large value, which is consistent with the requirements of Kontrol engagement proofs.

### DIFF
--- a/booster/library/Booster/SMT/Interface.hs
+++ b/booster/library/Booster/SMT/Interface.hs
@@ -157,14 +157,12 @@ translatePrelude def =
 
 checkPrelude :: Log.LoggerMIO io => SMT io ()
 checkPrelude = do
-    -- set large default timeout value for checking the prelude
-    setTimeout defaultSMTOptions.timeout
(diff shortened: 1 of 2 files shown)


geo2a:
Why not set this value as default at the Kontrol level?

geo2a:
In a Slack conversation, Petar has pointed out that the default timeout of 125 ms is unconditionally used to check the SMT prelude. That was not enough with some custom options passed to z3 via `--smt-option`.

We should use the user-specified timeout when checking the prelude.

goodlyrottenapple:
> In a Slack conversation, Petar has pointed out that the default timeout of 125 ms is unconditionally used to check the SMT prelude. That was not enough with some custom options passed to z3 via `--smt-option`.
> 
> We should use the user-specified timeout when checking the prelude.

Will open a quick PR, as this is something I noticed last week whilst refactoring the SMT code and was going to address in a separate PR

goodlyrottenapple:
closing in favour of #4030

PR #4033, opened 2024-08-13
Back-translate the SMTLIB unary minus, ignore models in `"get-model"` tests

geo2a:
This is a bridge PR that we need before bumping Z3 to 4.13.0.

- modify `test/rpc-server/runTests.py` to not compare the actual model from `"get-model"` responses, as it may be non-deterministic. Only compare the `"satisfiable"` field, i.e. model existence;
- when back-translating the results of `"get-model"` from SMTLIB2 s-expression to Kore, allow unary minus.

### DIFF
--- a/kore/src/Kore/Rewrite/SMT/Translate.hs
+++ b/kore/src/Kore/Rewrite/SMT/Translate.hs
@@ -533,11 +533,12 @@ backTranslateWith
         backTranslate (List xs)
             | null xs =
                 throwError "backtranslate: empty list"
-            -- special case for the unary minus, back-translating 'List ["-", "X"]' as '0 - X'
-            | [fct@(Atom "-"), arg] <- xs
-            , Just koreSym <- Map.lookup fct reverseMetadata = do
+            -- special case for the unary minus, back-translating 'List ["-", "X"]' as '-X'
+            | [Atom "-", arg] <- xs = do
                 arg' <- backTranslate arg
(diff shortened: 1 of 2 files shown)


kore/src/Kore/Rewrite/SMT/Translate.hs
```diff
@@ -533,6 +533,11 @@ backTranslateWith
         backTranslate (List xs)
             | null xs =
                 throwError "backtranslate: empty list"
+            -- special case for the unary minus, back-translating 'List ["-", "X"]' as '0 - X'
+            | [fct@(Atom "-"), arg] <- xs
+            , Just koreSym <- Map.lookup fct reverseMetadata = do
+                arg' <- backTranslate arg
+                pure $ TermLike.mkApplySymbol koreSym [mkInternalInt (InternalInt (simpleSort "SortInt") 0), arg']
```
jberthold, line 540: I think we should rather just negate the `InternalInt` that we expect inside `arg`, unless that does not work for some reason...
```suggestion
                arg' <- backTranslate arg
                case arg' of
                    InternalInt intSort n -> pure $ InternalInt intSort (negate n)
                    other -> throwError $ "unable to translate negation of " <> show other
```

geo2a: Thanks, that worked, modulo juggling Kore's data constructors.

goodlyrottenapple (approved):

PR #4035, opened 2024-08-14
Georgy/refactor check requires

geo2a:
Refactor `applyRule` and `applyEquation` to check `requires`/`ensures` in separate functions. 

This will make reviewing https://github.com/runtimeverification/haskell-backend/pull/4022 easier.

### DIFF
--- a/booster/library/Booster/Pattern/ApplyEquations.hs
+++ b/booster/library/Booster/Pattern/ApplyEquations.hs
@@ -839,86 +839,15 @@ applyEquation term rule =
                                                     Map.toList subst
                                         )
 
-                        -- instantiate the requires clause with the obtained substitution
-                        let required =
-                                concatMap
-                                    (splitBoolPredicates . coerce . substituteInTerm subst . coerce)
-                                    rule.requires
-                        -- If the required condition is _syntactically_ present in
-                        -- the prior (known constraints), we don't check it.
-                        knownPredicates <- (.predicates) <$> lift getState
-                        toCheck <- lift $ filterOutKnownConstraints knownPredicates required
+                        -- check required constraints from lhs.
+                        -- Reaction on false/indeterminate is varies depending on the equation's type (function/simplification),
+                        -- see @handleSimplificationEquation@ and @handleFunctionEquation@
+                        checkRequires subst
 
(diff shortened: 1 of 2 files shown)


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -839,86 +839,15 @@ applyEquation term rule =
...
-                                            ctx . logMessage $
-                                                WithJsonMessage (object ["conditions" .= map (externaliseTerm . coerce) stillUnclear]) $
-                                                    renderOneLineText ("Required condition found to be false: " <> pretty' @mods failedP)
-                                        , ConditionFalse failedP
-                                        )
-                                SMT.IsValid{} -> do
-                                    -- can proceed
-                                    pure ()
+                        -- check required constraints from lhs.
+                        -- Reaction on false/indeterminate is varies depending on the equation's type (function/simplification),
```
goodlyrottenapple, line 843:
```suggestion
                        -- Reaction on false/indeterminate varies depending on the equation's type (function/simplification),
```

booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -459,6 +417,72 @@ applyRule pat@Pattern{ceilConditions} rule =
...
+                pure ()
+
+        -- if a new constraint is going to be added, the equation cache is invalid
+        unless (null newConstraints) $ do
+            withContextFor Equations . logMessage $
+                ("New path condition ensured, invalidating cache" :: Text)
+
+            lift . RewriteT . lift . modify $ \s -> s{equations = mempty}
+        pure newConstraints
+
```
goodlyrottenapple, line 485: can this be subsumed by the ApplyEquations version?

geo2a: We may find the common denominator, but I'm reluctant to do it as https://github.com/runtimeverification/haskell-backend/pull/4022 will change `checkRequires` in ApplyEquations significantly, and I do not think we want to allow rewrite rules with syntactic `requires` clauses.

goodlyrottenapple (approved):

Issue #3775, opened 2022-11-17
Expand all alias applications (in order) before internalising a pattern


jberthold:
The current implementation (as of runtimeverification/hs-backend-booster#22 and runtimeverification/hs-backend-booster#29 ) expands aliases only in one expected place: the LHS of a rewrite rule.
(in runtimeverification/hs-backend-booster#29, we are ignoring `priority` aliases for the unused `antiLeft` to avoid a problem caused by applying one alias inside another).

Instead, all aliases should be expanded immediately while internalising patterns (i.e., inside other aliases and in rules, both LHS and RHS). The underlying assumptions are:
1. all aliases in the kore file are declared before they are applied (location-wise)
2. aliases are only applied inside the module where they are declared

Given assumption 2. the internalised definition could _discard_ all aliases once the module has been internalised, they do not have to be stored (we can also keep them around, though).

PR #4037, opened 2024-08-15
Preserve clause ordering of `requires` and `ensures` when internalisig rules in booster

### LINKED ISSUE #4028: Preserve clause ordering of `requires` and `ensures` when internalisig rules
To enable [syntactic simplifications](https://github.com/runtimeverification/k/issues/4579), Booster's internal representation of simplification rules must store the rule's `requires` clause as an ordered collection, and preserver the user-specified order of conjuncts (subterms if `_andBool_`).

Currently, we internalise rules' LHS and RHS as `Pattern`s. The `requires`/`ensures` clause is then taken to be the patterns' constrains, which are a `Set`. Moreover, the `internalisePredicates` function also returns a `Set`. 

We may have to decouple the internalisation routines for rules and patterns a little more to implement this.

goodlyrottenapple:
Fixes #4028


booster/library/Booster/Syntax/ParsedKore/Internalise.hs
```diff
@@ -937,10 +968,10 @@ internaliseRewriteRule partialDefinition exs aliasName aliasArgs right axAttribu
                 }
     return
         RewriteRule
-            { lhs = lhs.term
-            , rhs = rhs.term
-            , requires = lhs.constraints
-            , ensures = rhs.constraints
+            { lhs
+            , rhs
+            , requires = mempty
```
jberthold, line 973: I think we are somehow losing the `requires` clause on the track here ... but this whole function should be dead code anyway so we'd rather remove the `alias` business altogether than fix this.

booster/test/internalisation/imp.kore
```diff
@@ -542,884 +537,710 @@ module IMP
...
-  alias rule66LHS{}(SortMap{},SortId{},SortIds{},SortGeneratedCounterCell{},SortStmt{}) : SortGeneratedTopCell{}
-  where rule66LHS{}(VarRho:SortMap{},VarX:SortId{},VarXs:SortIds{},Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'Gen0:SortStmt{}) :=
-    \and{SortGeneratedTopCell{}} (
-      \equals{SortBool{},SortGeneratedTopCell{}}(
-        LblnotBool'Unds'{}(LblSet'Coln'in{}(inj{SortId{}, SortKItem{}}(VarX:SortId{}),Lblkeys'LParUndsRParUnds'MAP'Unds'Set'Unds'Map{}(VarRho:SortMap{}))),
-        \dv{SortBool{}}("true")), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'T'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortPgm{}, SortKItem{}}(Lblint'UndsSClnUndsUnds'IMP-SYNTAX'Unds'Pgm'Unds'Ids'Unds'Stmt{}(Lbl'UndsCommUndsUnds'IMP-SYNTAX'Unds'Ids'Unds'Id'Unds'Ids{}(VarX:SortId{},VarXs:SortIds{}),Var'Unds'Gen0:SortStmt{})),dotk{}())),Lbl'-LT-'state'-GT-'{}(Lbl'Unds'Map'Unds'{}(VarRho:SortMap{},Lbl'Stop'Map{}()))),Var'Unds'DotVar0:SortGeneratedCounterCell{})) []
```
jberthold, line 1390: (Example of a rule-LHS alias with a predicate)

jberthold (approved):
One detail: 
I suggest to remove all old code related to aliases for rule LHSs (and probably all code about aliases in general)- in a follow-up PR. #3775 states that aliases were only used for rule LHSs, and they were [removed from the front-end a while back](https://github.com/runtimeverification/k/pull/3314)

PR #4041, opened 2024-08-19
Pass `--ignore-environment` to nix dev shells in performance scripts

goodlyrottenapple:
Due to recent packaging changes to the blockchain plugin, the KEVM performance script started to fail as it was using libraries provided by brew and not nix on my local machine. The fix is to call `nix develop` with `--ignore-environment`, which purges the path of any non-nix paths. In general it is good practice to do this in any dev shells we call in one of these scripts, as there is less likelihood of similar issues arising.


jberthold (approved):
LGTM
The `performance-test-mx.sh` flavour is probably obsolete now. (and was anyway only testing the very fast ones - not representative), we might want to delete that one.

geo2a (approved):

PR #4051, opened 2024-09-25
Fix handling of logging contextx in `booster-dev`

geo2a:
This small PR:
- adds `[request ID]` contexts to logs emitted from `booster-dev`, similarly to `kore-rpc-booster`
- the above also makes the per-defined logging contexts `Booster.CLIOptions` useful again, as they did not match before due to the absence of `[request ID]`
- redirects the log messages to a file if asked for
- also lists the available logging formats in `--help`


jberthold (approved):
LGTM

PR #4057, opened 2024-10-15
Make Pretty prettier

jberthold:
The tool was not printing the substitution after internalising a pattern. 
Also switched it to using `Text` to print things and prettified the format a bit (infix printing, indentations).

### DIFF
--- a/dev-tools/pretty/Pretty.hs
+++ b/dev-tools/pretty/Pretty.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE PatternSynonyms #-}
 
 {- | Pretty printer for JSON KORE terms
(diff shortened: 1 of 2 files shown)


dev-tools/pretty/Pretty.hs
```diff
@@ -40,24 +46,37 @@ main = do
...
                     case runExcept $ internalisePredicates DisallowAlias CheckSubsorts Nothing internalDef [term] of
                         Left es -> error (show es)
                         Right ts -> do
-                            putStrLn "Pretty-printing predicates: "
-                            putStrLn "Bool predicates: "
-                            mapM_ (putStrLn . renderDefault . pretty' @'[Decoded]) ts.boolPredicates
-                            putStrLn "Ceil predicates: "
-                            mapM_ (putStrLn . renderDefault . pretty' @'[Decoded]) ts.ceilPredicates
-                            putStrLn "Substitution: "
-                            mapM_ (putStrLn . renderDefault . pretty' @'[Decoded]) ts.substitution
```
jberthold, line 60: Very interestingly, this just printed the _values_ of the substitution :grimacing: instead of an actual mapping. But of course it typechecks :+1: :upside_down_face:

geo2a (approved):
Thanks for getting this done! Gonna merge.

Issue #4059, opened 2024-10-15
Recognise ensured constraints as substitutions and apply them in `Booster`


geo2a:
* recognize equality predicates like `VAR ==Int expr` and `VAR ==K expr` as substitution items
* when finalizing a rewrite rule, apply the substitution to the term filed before returning the result
* return the learned substitution items as part of the JSON RPC response in the `"substitution"` filed, rather than the `"predicate"` field

geo2a:
Thinking more about this:

To make the tracking of the substitution more robust, it may be a good idea to explicitly attach it to a pattern, i.e. change the `Pattern` type in the following way:

```diff
+type Substitution = Map Variable Term
+
 -- | A term (configuration) constrained by a number of predicates.
 data Pattern = Pattern
     { term :: Term
     , constraints :: !(Set Predicate)
+    , substitution :: Substitution
     , ceilConditions :: ![Ceil]
     }
     deriving stock (Eq, Ord, Show, Generic, Data)

```

PR #4061, opened 2024-10-16
Recognise ==Int and ==K equations as substitutions when internalising

jberthold:
Booster now recognises predicates of shape `true #Equals (VAR ==Int <expr>)` and `true #Equals (VAR ==K <expr>)` as substitutions (given the global scope does not contradict that, i.e., no cycles or multiple substitutions for the same VAR...).

Currently only matches ==Int and ==K.
We could also match notBool (_ =/=Int _) and notBool (_ =/=K _) (more magic).

When internalising rules and equations, the substitution must be turned back into predicates for the `requires` clause. Currently we also turn back the ensures clause substitutions into predicates (might want to change that).

Part of #4059

### DIFF
--- a/booster/library/Booster/Syntax/Json/Internalise.hs
+++ b/booster/library/Booster/Syntax/Json/Internalise.hs
@@ -24,6 +24,7 @@ module Booster.Syntax.Json.Internalise (
     textToBS,
     trm,
     handleBS,
+    asEquations,
     TermOrPredicates (..),
     InternalisedPredicates (..),
     PatternOrTopOrBottom (..),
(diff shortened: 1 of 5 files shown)


booster/library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -481,6 +482,10 @@ mkEq x t = Internal.Predicate $ case sortOfTerm t of
     Internal.SortBool -> Internal.EqualsBool (Internal.Var x) t
     otherSort -> Internal.EqualsK (Internal.KSeq otherSort (Internal.Var x)) (Internal.KSeq otherSort t)
 
+-- | turns a substitution into a list of equations
```
geo2a, line 485:
```suggestion
-- | turns a substitution into a list of equalities 
```

jberthold:
No significant difference in `kevm` tests

| Test                                                             | jb-internalise-more-substitutions time | master-0bfbcffb2 time | (jb-internalise-more-substitutions/master-0bfbcf
fb2) time 
|------------------------------------------------------------------|----------------------------------------|-----------------------|-------------------------------------------------
----------
| mcd-structured/flopper-dent-guy-diff-tic-not-0-pass-rough-spec.k | 150.74                                 | 156.68                | 0.9620883329078377                              
          
| TOTAL                                                            | 150.74                                 | 156.68                | 0.9620883329078377

geo2a (approved):
Looks good to me!

PR #4063, opened 2024-10-17
Handle inconsistent ground truth and SMT unknowns when checking `ensures`

geo2a:
When checking `ensures` conditions of rewrite rules with the SMT solver, we must mark rewrite as trivial if the ground truth is inconsistent. If the SMT solver returns unknown, we must abort rewriting.

Previously, we were swallowing both of there cases and finalizing the rewrite step successfully. This behavior of Booster went undetected for a long time since we would usually abort rewriting or detect a vacuous state at the next step, resulting in wasted work but no unsoundness.

We also tweak the return type of `checkPredicates` to convey addition information why the result is unknown. This will be useful when we start tolerating SMT unknowns and branching on that.

### DIFF
--- a/booster/library/Booster/SMT/Interface.hs
+++ b/booster/library/Booster/SMT/Interface.hs
@@ -12,6 +12,7 @@ module Booster.SMT.Interface (
     SMTOptions (..), -- re-export
     defaultSMTOptions, -- re-export
     SMTError (..),
+    UnkwnonReason (..),
     initSolver,
     noSolver,
     finaliseSolver,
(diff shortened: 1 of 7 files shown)


booster/library/Booster/SMT/Interface.hs
```diff
@@ -188,12 +198,28 @@ finaliseSolver ctxt = do
     Log.logMessage ("Closing SMT solver" :: Text)
     destroyContext ctxt
 
-pattern IsUnknown :: unknown -> Either unknown b
+data UnkwnonReason
```
jberthold, line 201:
```suggestion
data UnknownReason
```

jberthold (approved):
Pending spelling mistakes...

PR #4066, opened 2024-10-28
Booster Description

geo2a:
This PR adds prose documentation for Booster --- a fast symbolic rewrite engine for K.


docs/2024-10-18-booster-description.md
```diff
@@ -0,0 +1,248 @@
...
+1. substitute the rule's requires clause with the matching substitution
+2. check if we already have any of the conjuncts verbatim in the pattern's constrains. If so, we filter them out as known truth
+3. simplify every conjunct individually by applying equations. This is morally equivalent to sending every conjunct as to the `"simplify"` endpoint and will use the same code path, bypassing internalisation.
+4. filter again, as the simplified conjuncts may not be present verbatim in the known truth
+5. if any clauses remain, it's time to fire-up Z3 and check them for validity.
+    - some rule will be rejected at that point, as their pre-condition P (the `requires` clause) is invalid, which means that the rule is applicable statically, but the dynamic conditions makes it inapplicable.
+    - some rules may have an indeterminate pre-condition P, which means that both P and its negation are satisfiable, i.e. the solver said (SAT, SAT) for (P, not P).
+      - in this case we can apply this rule conditionally, but add P into the path condition We will call `not P` the _remainder condition_ of this rule and keep track of it too
```
jberthold, line 160:
```suggestion
The requires clause check is encapsulated by the [checkRequires](https://github.com/runtimeverification/haskell-backend/blob/master/booster/library/Booster/Pattern/Rewrite.hs#L496) function defined in the `where`-clause of `applyRule`. It will:
1. substitute the rule's requires clause with the matching substitution
2. check if we already have any of the conjuncts verbatim in the pattern's path condition (`PC`). If so, we filter them out as known truth
3. simplify every conjunct individually by applying equations. This is morally equivalent to sending every conjunct as to the `"simplify"` endpoint and will use the same code path, bypassing internalisation.
4. check again whether any of the, now simplified, conjuncts  is present verbatim in the path condition
5. if any clauses remain, check all conjuncts together with Z3 for validity given the path condition.
    - some rule will be rejected at that point, as their pre-condition `P` (the `requires` clause) is false given `PC`, which means that the rule is applicable statically, but the dynamic conditions makes it inapplicable.
    - some rules may have an indeterminate pre-condition `P`, which means that both `PC /\ P` and `PC /\ not P` are `SAT` in the solver, i.e., neither `P` nor  `not P` are implied by `PC`
      - in this case we can apply this rule conditionally, but add `P` into the path condition We will call `not P` the _remainder condition_ of this rule and keep track of it too
```

docs/2024-10-18-booster-description.md
```diff
@@ -0,0 +1,248 @@
...
+        We effectively do the same if we cannot establish the validity of P due to a solver timeout, i.e. we add the predicate as an assumption. This may potentially lead to a vacuous branch as we discover more conditions further into the rewriting process.
+    -   some rules will have a valid requires clause, which means they definitely do apply and we do need to add anything else into the path condition as an assumption.
```
jberthold, line 162:
```suggestion
    -   some rules will have a valid requires clause, i.e., `PC => P`, which means they definitely apply and we do not need to add anything else into the path condition as an assumption.
```

docs/2024-10-18-booster-description.md
```diff
@@ -0,0 +1,248 @@
...
+
+See the [Booster.SMT.Interface](https://github.com/runtimeverification/haskell-backend/blob/master/booster/library/Booster/SMT/Interface.hs) module to learn more about how `Predicate`s are checked for satisfiable and validity using Z3.
+
+Bottom line:
+- if `requires` is UNSAT, we do not apply the rule, just if it didn't even match;
```
jberthold, line 167:
```suggestion
- if `requires` is found to be false, we do not apply the rule, just as if it didn't even match;
```

docs/2024-10-18-booster-description.md
```diff
@@ -0,0 +1,248 @@
...
+
+When Booster applies rewrite rules, they are grouped by priority. The group with the higher priority applies first, and all rules are treated equally within the group.
+
+During symbolic execution, we keep track of the current path condition --- a conjunction of logical constraints that specify an equivalence class of concrete execution states. In order to prevent state explosion, it is important to only create new symbolic states that are feasible. Priority groups enable a additional mechanism for that by enabling the semantics implementer to partition rules into complete groups, and only applying lower-priority rules under the conditions where higher-priority groups are not applicable.
+
+A priority group of rules gives raise to a group **coverage condition**, which is defined as a disjunction of the requires clauses of the rules in the group. If the coverage condition is valid, it means that no other rules can possibly apply. The negation of the coverage condition is called the group's coverage condition is called the **remainder condition**. The the remainder condition is unsatisfiable, then the coverage condition is valid, and the group is complete.
```
jberthold, line 201:
```suggestion
A priority group of rules gives rise to a group **coverage condition**, which is defined as a disjunction of the requires clauses of the rules in the group. If the coverage condition is valid, it means that no other rules can possibly apply. The negation of the coverage condition is called the group's **remainder condition**. If the remainder condition is unsatisfiable, then the coverage condition is valid, and the group is complete.
```

docs/2024-10-18-booster-description.md
```diff
@@ -0,0 +1,248 @@
...
+
+See [remainder-predicates.k](https://github.com/runtimeverification/haskell-backend/blob/master/booster/test/rpc-integration/resources/remainder-predicates.k) and [test-remainder-predicates](https://github.com/runtimeverification/haskell-backend/blob/master/booster/test/rpc-integration/test-remainder-predicates) for a series of integrations tests that cover the behaviour of `booster-dev` and `kore-rpc-booster`.
+
+### [Iterating Rules](#rewriting-many-steps)
+
+Successful rule application does not trigger pattern-wide simplification, i.e. very far and make many steps without simplifying the pattern even ones.
```
jberthold, line 220:
```suggestion
Successful rule application does not trigger pattern-wide simplification, i.e. rewriting can proceed very far and make many steps without simplifying the pattern even once.
```

jberthold (approved):
Let's skip the examples for now to get this merged

PR #4069, opened 2024-10-30
Turn `Pattern` into `#Bottom` if any constraint is literal false

geo2a:
- before returning a `"simplify"` response, check if any constrains are `false` and return `#Bottom` in that case
- when simplifying rewriting results, prune branches if there's a literal `false`
  - this triggers the change in the `test-vacuous` integration test, where the original state is now returned. [simplifyP](https://github.com/runtimeverification/haskell-backend/blob/booster-bottom-on-false-constaint/booster/library/Booster/Pattern/Rewrite.hs#L1029) now returns `Nothing` for the result and thus we get to return `RewriteTrivial` from `performRewrite`
  - note that the outdated responses for this integration tests actually had a literal `false` as one of the constraints.


jberthold (approved):
:+1:

PR #4070, opened 2024-10-30
Break-up new top-level `_andBool_` in `evaluatePattern'`

geo2a:
This PR breaks-up new symbolic conjuncts when simplifying a pattern.

It is useful in conjunction K rule:
```
 rule [notBool-or]:  notBool ( A  orBool B ) => ( notBool A ) andBool ( notBool B ) [simplification]
```


jberthold (approved):
:rocket:

PR #4071, opened 2024-10-31
Assume the rule remainder condition when checking ensures

geo2a:
When Booster detects an indeterminate `requires`, it adds it to the constraints on the rewritten `Pattern`. However, when we check the `ensures` clause, we only pass the the constraints on the old pattern as known truth. This PR corrects that by adding the uncertain conditions as additional known truth, as the `ensures` check should be performed on the rewritten pattern, not the old one.

### DIFF
--- a/booster/library/Booster/Pattern/Rewrite.hs
+++ b/booster/library/Booster/Pattern/Rewrite.hs
@@ -460,7 +460,10 @@ applyRule pat@Pattern{ceilConditions} rule =
 
                     -- check ensures constraints (new) from rhs: stop and return `Trivial` if
                     -- any are false, remove all that are trivially true, return the rest
-                    ensuredConditions <- checkEnsures ruleSubstitution
+                    --
+                    -- we need to add unclearRequiresAfterSmt (if any) as additional known truth,
+                    -- as it may allow us to prune a vacuous state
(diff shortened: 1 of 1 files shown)


booster/library/Booster/Pattern/Rewrite.hs
```diff
@@ -460,7 +460,10 @@ applyRule pat@Pattern{ceilConditions} rule =
...
                     -- check ensures constraints (new) from rhs: stop and return `Trivial` if
                     -- any are false, remove all that are trivially true, return the rest
-                    ensuredConditions <- checkEnsures ruleSubstitution
+                    --
+                    -- we need to add unclearRequiresAfterSmt (if any) as additional known truth,
+                    -- as it may allow us to prune a vacuous state
+                    ensuredConditions <- checkEnsures ruleSubstitution unclearRequiresAfterSmt
 
                     -- if a new constraint is going to be added, the equation cache is invalid
                     unless (null ensuredConditions) $ do
```
jberthold, line 469: Just stumbled over this line here. What if `ensuredConditions` is nil but we have `unclearRequiresAfterSmt`? 
We should probably _also_ clear the equation cache in that case. We are most likely going to return a `Branching` result immediately, but maybe we find one of the branches simplifies to bottom and proceed.

jberthold (approved):

PR #4075, opened 2024-11-24
Use coherent env. in release.yml

jberthold:
Makes release builds aware of the ghc-version to build with. (and also stack and hpack, which are minor)


jberthold:
https://github.com/runtimeverification/haskell-backend/actions/runs/12000042509/job/33448681602#step:5:14 shows we are currently building the ubuntu package with GHC-9.10, whereas the nix build uses GHC-9.6.5

F-WRunTime:
@jberthold no reviewers on this. Do you want it to go through? LGTM, would be good we sync the ghc between the packages nix<>ubuntu

jberthold:
> @jberthold no reviewers on this. Do you want it to go through? LGTM, would be good we sync the ghc between the packages nix<>ubuntu

Agree, we should not build with different GHCs for master builds.
I noticed the issue but left this here to revisit later (we might want to upgrade GHC as we see it does build without problems, but I don't have time to test it properly atm).

We might as well merge this PR separately, though.

.github/workflows/release.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches:
       - release
 
+env:
+  ghc_version: "9.6.5"
+  stack_version: "2.15.1"
+  hpack_version: '0.36'
```
tothtamas28, line 10: Just a nit, string literal syntax is not used consistently here.

tothtamas28 (approved):

PR #4076, opened 2024-11-26
Hotfix check/adjust subsorting for substitutions based on `==K` terms

jberthold:
When a constraint of shape `VAR:VarSort ~> .K ==K term:TermSort ~> .K` is externalised, the prior code just rendered a `VAR:VarSort #Equals term:TermSort`. This is wrong when `TermSort /= VarSort`, however `TermSort` will be a subsort of `VarSort`, otherwise the rule that introduced this `==K` term is ill-sorted.

* Externalisation code now inserts the missing injection into the `#Equals` expression.
* A test was added which demonstrates the use case and behaviour (failed before, succeeds now).
* A pretty-printer option was added to show the injections in the printed terms (off by default).

### DIFF
--- a/booster/library/Booster/CLOptions.hs
+++ b/booster/library/Booster/CLOptions.hs
@@ -184,7 +184,7 @@ parseLogOptions =
             ( metavar "PRETTY_PRINT"
                 <> value [Decoded, Truncated]
                 <> long "pretty-print"
-                <> help "Prety print options for kore terms: decode, infix, truncated"
+                <> help "Prety print options for kore terms: decode, infix, truncated, with-injections"
                 <> showDefault
             )
   where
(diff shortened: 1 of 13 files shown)


booster/library/Booster/CLOptions.hs
```diff
@@ -184,7 +184,7 @@ parseLogOptions =
             ( metavar "PRETTY_PRINT"
                 <> value [Decoded, Truncated]
                 <> long "pretty-print"
-                <> help "Prety print options for kore terms: decode, infix, truncated"
+                <> help "Prety print options for kore terms: decode, infix, truncated, with-injections"
```
PetarMax, line 187:
```suggestion
                <> help "Pretty print options for kore terms: decode, infix, truncated, with-injections"
```

booster/library/Booster/CLOptions.hs
```diff
@@ -184,7 +184,7 @@ parseLogOptions =
             ( metavar "PRETTY_PRINT"
                 <> value [Decoded, Truncated]
                 <> long "pretty-print"
-                <> help "Prety print options for kore terms: decode, infix, truncated"
+                <> help "Pretty print options for kore terms: decode, infix, truncated, with-injections"
```
jberthold, line 187:
```suggestion
                <> help "Pretty print options for kore terms: decoded, infix, truncated, with-injections"
```

tothtamas28 (approved):

PR #4078, opened 2024-12-02
HOTFIX abort rewrite when SMT solver times out

jberthold:
When the SMT solver timed out while checking `requires` clauses of a rule, the unclear conditions were returned in the same way as conditions that were known to be indeterminate. This created bogus branches in proofs when the solver had a problem to decide a condition.

On such timeouts, and on inconsistent ground truths, booster now aborts the entire rewrite. This might lead to increased spurious aborts in complex proofs, but is probably better than having to prune the bogus branches from the client side. The fall-back to legacy kore was able to prune the bogus branch easily in the case that was investigated.

### DIFF
--- a/booster/library/Booster/Pattern/Rewrite.hs
+++ b/booster/library/Booster/Pattern/Rewrite.hs
@@ -590,15 +590,18 @@ applyRule pat@Pattern{ceilConditions} rule =
         -- check unclear requires-clauses in the context of known constraints (priorKnowledge)
         solver <- lift $ RewriteT $ (.smtSolver) <$> ask
         SMT.checkPredicates solver pat.constraints pat.substitution (Set.fromList stillUnclear) >>= \case
-            SMT.IsUnknown reason -> do
-                withContext CtxWarn $ logMessage reason
-                -- return unclear rewrite rule condition if the condition is indeterminate
-                withContext CtxConstraint . withContext CtxWarn . logMessage $
+            SMT.IsUnknown SMT.ImplicationIndeterminate -> do
+                -- return unclear conditions if SMT solver reports the condition indeterminate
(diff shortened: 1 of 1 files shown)


PetarMax:
If a ground truth is inconsistent, the branch is vacuous, we should return the vacuous result, no?

jberthold:
KEVM tests

| Test                                                  | HOTFIX-abort-branching-on-SMT-timeouts time | master-636e1f7f6 time | (HOTFIX-abort-branching-on-SMT-timeouts/master-636e1f7f6) time 
|-------------------------------------------------------|---------------------------------------------|-----------------------|----------------------------------------------------------------
| examples/sum-to-n-spec.k                              | 88.04                                       | 94.35                 | 0.9331213566507686                                             
| benchmarks/overflow00-nooverflow-spec.k               | 46.43                                       | 48.2                  | 0.9632780082987551                                             
| kontrol/test-allowchangestest-testallow_fail-0-spec.k | 48.16                                       | 46.53                 | 1.0350311626907371                                             
| erc20/hkg/transfer-failure-1-spec.k                   | 58.39                                       | 56.35                 | 1.0362023070097603                                             
| TOTAL                                                 | 241.01999999999998                          | 245.43                | 0.9820315364869819

jberthold:
> If a ground truth is inconsistent, the branch is vacuous, we should return the vacuous result, no?

If the ground truth is inconsistent, the question is how did we even get here? 
* if the state got rewritten before, we should have detected it and not rewritten.
* if the state did not get rewritten, the input was already inconsistent. We had an open PR to detect this, I have to check what happened to it. Aborting in booster is a safe option in that case. Tthe inconsistency would also be detected when simplifying the result state....

tothtamas28 (approved):

PR #4093, opened 2025-02-21
modified:   .github/workflows/release.yml

F-WRunTime:
modified:   .github/workflows/test.yml
- Update nix release process for test/release workflows
- Publish release builds to k-framework-binar
- Publish iterative CI builds to k-framework.
- Reduce build times upstream and in CI PR testing.
- Supporting issue of haskell-backend components being built upstream by
  dependents.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,10 +55,20 @@ jobs:
         uses: cachix/cachix-action@v14
         with:
           name: k-framework
+          extraPullNames: k-framework-binary
           authToken: '${{ secrets.CACHIX_PUBLIC_TOKEN }}'
 
-      - name: Build
-        run: nix build .#kore-exec .#kore-rpc-booster
+      - name: 'Build and Cache'
+        uses: workflow/nix-shell-action@v3
+        env:
+          GC_DONT_GC: 1
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_PUBLIC_TOKEN }}
+        with:
+          packages: jq
+          script: |
+            HASKELL_BACKEND=$(nix build --extra-experimental-features 'nix-command flakes' .#kore-exec .#kore-rpc-booster --json | jq -r '.[].outputs | to_entries[].value')
+            DRV=$(nix-store --query --deriver ${HASKELL_BACKEND})
+            nix-store --query --requisites --include-outputs ${DRV} | cachix push k-framework-binary
 
   ubuntu-package:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -55,10 +55,20 @@ jobs:
         uses: cachix/cachix-action@v14
         with:
           name: k-framework
+          extraPullNames: k-framework-binary
           authToken: '${{ secrets.CACHIX_PUBLIC_TOKEN }}'
 
-      - name: Build
-        run: nix build .#kore-exec .#kore-rpc-booster
+      - name: 'Build and Cache'
+        uses: workflow/nix-shell-action@v3
```
jberthold, line 62: This seems to introduce an error message `unrecognised flag '-c'` (cannot possibly come from the commands below, and happens before any build output is printed). Is this action stable/recommended by people?

.github/workflows/test.yml
```diff
@@ -112,8 +112,17 @@ jobs:
...
+        uses: workflow/nix-shell-action@v3
+        env:
+          GC_DONT_GC: 1
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_PUBLIC_TOKEN }}
+        with:
+          packages: jq
+          script: |
+            HASKELL_BACKEND=$(nix build --extra-experimental-features 'nix-command flakes' .#kore-exec .#kore-rpc-booster --json | jq -r '.[].outputs | to_entries[].value')
+            DRV=$(nix-store --query --deriver ${HASKELL_BACKEND})
+            nix-store --query --requisites --include-outputs ${DRV} | cachix push k-framework
```
jberthold, line 125: (as Everett also asked on slack)
Not sure we should push to cachix from PR builds. We have the cachix action (currently only on linux but could probably add it as a noop to the Mac build without issues) so the build artefacts _should_ already be pushed up to accelerate subsequent builds without this change.

jberthold (approved):
I don't have insight into our cachix setup so I cannot judge whether this solves our problem. OTOH it is certainly not harmful (unless the new github action starts to fail on us, see comment).
So from my side, it is OK to merge.

.github/workflows/release.yml
F-WRunTime: Which looks like also produces the same error. Not sure when this was introduced though without looking at git history seems it didn't raise any flags though. I'll spend a minute on it see if there' and a similar error is thrown on linux as well.

palinatolmach (approved):

PR #4096, opened 2025-03-10
Nix flake revamp

jberthold:
Rewriting the `flake.nix` to avoid `stacklock2nix` and use `some-cabal-hashes` for faster builds.

* also removed non-standard dependencies `tasty-test-reporter` and `profiteur`
* the `cabal.project.freeze` file is updated to be consistent with the `nix` build
* dependency differences to `stack` LTS-22.23 were checked and are harmless (patch-level only, and all but 3 in test dependencies anyway)


flake.nix
```diff
@@ -1,206 +1,106 @@
...
-              graphviz = dontCheck hprev.graphviz;
-              json-rpc = dontCheck hprev.json-rpc;
-              lifted-base = dontCheck hprev.lifted-base;
-              prettyprinter = dontCheck hprev.prettyprinter;
-              semialign = doJailbreak hprev.semialign;
-              smtlib-backends-process = dontCheck hprev.smtlib-backends-process;
-              tar = dontCheck hprev.tar;
-              text-short = doJailbreak hprev.text-short;
-              these = doJailbreak hprev.these;
-              ghc-prof = doJailbreak hprev.ghc-prof;
```
jberthold, line 63: Some of these overrides are definitely required, and I am currently stuck trying to replicate them with `some-cabal-hashes.nix`

flake.nix
```diff
@@ -1,206 +1,106 @@
...
+                #     final.fetchFromGitHub {
+                #       owner = "goodlyrottenapple";
+                #       repo = "tasty-test-reporter";
+                #       rev = "b704130545aa3925a8487bd3e92f1dd5ce0512e2";
+                #       sha256 = "sha256-9CN59Im0BC3vSVhL85v5eXPYYoPbV3NAuv893tXpr/U=";
+                #     };
+                # profiteur =
+                #     final.fetchFromGitHub {
+                #       owner = "goodlyrottenapple";
+                #       repo = "profiteur";
```
jberthold, line 81: These two are custom dependencies from github (forks of repositories that cause dependency issues). The code here follows the `some-cabal-hashes` example but is seemingly ignored. I already removed all code that uses `tasty-test-reporter`, we can remove the `profiteur` consumer, too and forget about this issue.

flake.nix
```diff
@@ -1,206 +1,106 @@
...
-              # Read the release version from scripts/release.yml
-            in builtins.head
-            (builtins.match ".+ReleaseVersion: '([^']+).+" release);
+                # FIXME a jailbreak does not work either (Problem is the ansi-terminal version)
+                # hfinal.tasty-test-reporter = hfinal.doJailbreak hprev.tasty-test-reporter;
+
+                # FIXME likewise, these two `dontCheck` (one inlined) are seemingly ignored.
+                hfinal.decision-diagrams = hfinal.overrideCabal hprev.decision-diagrams 
+                    (drv: { doCheck = false; }); # build runs failing tests
+                hfinal.json-rpc = final.haskell.packages.dontCheck hprev.json-rpc; # 1.0.4 marked as broken!
```
jberthold, line 92: This `dontCheck` setting is somehow also ignored, I keep getting a build failure because of a failing test in `decision-diagrams`

juliankuners: This should be due to the attribute names that are being overridden, here. In this override example, you have to use the haskell package names directly, if I understand correctly. Therefore, currently, the original variants of the packages are being used instead.

```suggestion
                decision-diagrams = hfinal.overrideCabal hprev.decision-diagrams 
                    (drv: { doCheck = false; });
                json-rpc = hfinal.dontCheck hprev.json-rpc;
```

flake.nix
```diff
@@ -1,206 +1,106 @@
...
+                cabal-install
+              ];
+            };
         };
-      });
+    in
+    flake-utils.lib.eachDefaultSystem out // {
+      overlays = {
+        haskell-versions = makeHaskellOverlay (final: prev:
+          import ./nix/some-cabal-hashes.nix {
```
juliankuners, line 62: Instead of having a local copy, this file can be fetched as an input to the flake. In a previous commit, you did add some-cabal-hashes to the inputs, but used `builtins.fetchGit` instead of flake syntax, which would not work. Instead, as some-cabal-hashes is not available as a flake, you can add the attribute `flake = false;` to retrieve the repository anyway, like the z3 input of this flake.

```nix
inputs = {
  # pin rv-utils to a fixed version
  rv-utils.url = "github:runtimeverification/rv-nix-tools/854d4f0";
  nixpkgs.follows = "rv-utils/nixpkgs";
  some-cabal-hashes = {
    url = "github:lf-/nix-lib";
    flake = false;
  };
  z3 = {
    url = "github:Z3Prover/z3/z3-4.13.4";
    flake = false;
  };
};
```

jberthold: Thanks, that is helpful! I wasn't aware that `fetchGit` wouldn't work in a flake, will use this now instead of vendoring the file.

flake.nix
jberthold: I tried this earlier, but `nix build` (`--dry-run` or not) complained about the `hfinal.overrideCabal` (or `hfinal.dontCheck`) as a missing attribute. Changing that to `prev.haskell.lib.{overrideCabal,dontCheck}` worked better :smirk: yet still the build is failing on `decision-diagrams` (but differently now)

juliankuners:
Okay, I checked out the branch and took a closer look at the build. This is the error message: "cabal2nix: user error (*** Found neither a .cabal file nor package.yaml. Exiting.)".

Nixpkgs provides only one set of haskell packages per compiler version, with exactly one version of a package each. Also, only one compiler version is usually maintained. A huge percentage of haskell packages is broken, though a broken package can relatively easily be fixed according to [documentation](https://nixos.org/manual/nixpkgs/stable/#haskell). The set of haskell package derivations from stackage is automatically created, see [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/hackage-packages.nix), in accordance to a [stackage.yaml](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/configuration-hackage2nix/stackage.yaml) file.

The haskell-backend currently instead uses stacklock2nix to create a custom set of haskell packages on the spot, in this case with lts-22.23. The nixpkgs revision currently used instead provides the nixpkgs based set of packages that also instead relies on lts-22.22. While the difference is not large, it still explains why the json-rpc package is broken and marked as broken when not using stacklock2nix. The same is probably going on for the package decision-diagrams. Therefore, both packages derivations, and maybe some of the other dependencies, must be fixed. Again, according to nix documentation this should not be too hard.

Also, I took a look at the nixpkgs master and json-rpc is still marked as broken.

jberthold:
> Okay, I checked out the branch and took a closer look at the build. This is the error message: "cabal2nix: user error (*** Found neither a .cabal file nor package.yaml. Exiting.)".

Yes, correct. Which seems rather strange. I tried poking around in the `/nix/store` to inspect what it is trying to build but could not identify the directory it is working with (if the directory even exists).
 
> The haskell-backend currently instead uses stacklock2nix to create a custom set of haskell packages on the spot, in this case with lts-22.23. The nixpkgs revision currently used instead provides the nixpkgs based set of packages that also instead relies on lts-22.22. 

The configuration also uses a non-standard `hashable` package (unfortunately), so we would anyway be outside the canonical package set (the `cabal.project.freeze` file is for that purpose).

> While the difference is not large, it still explains why the json-rpc package is broken and marked as broken when not using stacklock2nix. The same is probably going on for the package decision-diagrams. Therefore, both packages derivations, and maybe some of the other dependencies, must be fixed. Again, according to nix documentation this should not be too hard.

Maybe not hard but I have no idea at all how to approach the current issue with `decision-diagrams`...
It is good that we got it one step further because the overrides seem to work now, though. We will probably need more overrides (we had more in the old flake but I don't want to just cargo-cult them).

juliankuners:
I continued working on the flake and pushed my progress to this [branch](https://github.com/juliankuners/haskell-backend/tree/nix-flake-revamp-fixbroken). First, I tried to make the build work without some-cabal-hashes. Doing this led me to avoid the previous build error regarding the missing cabal project file. Apparently, this was caused by the way the previous flake used some-cabal-hashes. The overrides attribute of some-cabal-hashes is intended to be used solely to either override the version number of a haskell package or override the package source with a derivation, e.g. fetched via `fetchFromGitHub` or via flake inputs. This was why the overridden packages were broken. In particular, the overrides attribute of some-cabal-hashes can be used identical to [`packageSourceOverrides`](https://noogle.dev/f/pkgs/haskell/lib/packageSourceOverrides).

I added the z3 overlay from the original flake and the develop shells still need to be ported over to the new flake. Also, I reverted fixing rv-nix-tools to a specific revision for now. In my opinion, if not specifically required for this pull request, this should be in another pull request. The way we will want to refactor the way nixpkgs is locked in the runtime verification repository should be discussed separately.

Currently, hs-backend-booster fails to build due to the wrong version of hashable. The versions of the haskell packages can be found in this [nix file](https://github.com/NixOS/nixpkgs/blob/69493a13eaea0dc4682fd07e8a084f17813dbeeb/pkgs/development/haskell-modules/hackage-packages.nix). Getting haskell-backend to build successfully will require version overrides, taken from the cabal.project.freeze file. I wonder whether we should parse the freeze file from nix or port the freeze file to nix entirely, and henceforth maintain the version pins in the nix file?

I should mention that overriding versions of haskell packages during testing caused an infinite recursion several times. This was caused by overriding packages that are dependencies of cabal2nix, which in turn is used to when overriding a version. I fixed this in the flake by fixing the cabal2nix derivation to an unmodified derivation from the second nixpkgs instance that was in the flake already anyway. 


> > Okay, I checked out the branch and took a closer look at the build. This is the error message: "cabal2nix: user error (*** Found neither a .cabal file nor package.yaml. Exiting.)".
> 
> Yes, correct. Which seems rather strange. I tried poking around in the `/nix/store` to inspect what it is trying to build but could not identify the directory it is working with (if the directory even exists).

Derivations are built in a temporary directory. Upon failure or success, this directory is deleted. To inspect a failing derivation build, you can pass the flag `--keep-failed` to `nix build`. Also, there are [methods](https://wiki.nixos.org/w/index.php?title=Nixpkgs/Create_and_debug_packages&mobileaction=toggle_view_desktop#Using_nix-shell_for_package_development) to debug the build process of a derivation step by step, which makes nix build debugging in some situations so much more comfortable.

> > The haskell-backend currently instead uses stacklock2nix to create a custom set of haskell packages on the spot, in this case with lts-22.23. The nixpkgs revision currently used instead provides the nixpkgs based set of packages that also instead relies on lts-22.22.
> 
> The configuration also uses a non-standard `hashable` package (unfortunately), so we would anyway be outside the canonical package set (the `cabal.project.freeze` file is for that purpose).

I see, fixing the versions should work now with the new commits from by branch.

> > While the difference is not large, it still explains why the json-rpc package is broken and marked as broken when not using stacklock2nix. The same is probably going on for the package decision-diagrams. Therefore, both packages derivations, and maybe some of the other dependencies, must be fixed. Again, according to nix documentation this should not be too hard.
> 
> Maybe not hard but I have no idea at all how to approach the current issue with `decision-diagrams`... It is good that we got it one step further because the overrides seem to work now, though. We will probably need more overrides (we had more in the old flake but I don't want to just cargo-cult them).

See above, this should not be broken anymore.

juliankuners:
> Thanks for this, I like how it looks now with the several overlay functions!
> 
> Do you have push access to this repository? (if not we will fix that, you definitely should).
> 
> Will continue on this path, investigating the `hashable` issue and other overrides that we might need.

Thank you for giving me write access.

I have pushed a new branch that fixed the last two encountered build issues: https://github.com/runtimeverification/haskell-backend/commits/nix-flake-revamp-fixbuild/

I added the missing develop shells, overlays, and the missing prelude-kore output. Also, I adapted the packages output to match the packages output of the previous flake.

The previous flake had a default overlay output that included the output of stacklock2nix. We can not provide that output anymore, as any repository that uses the default overlay would be broken and needs to be updated. Though, I quickly glanced at a few repositories and couldn't find any that uses this default overlay! I also find it unlikely anyway.

@jberthold Could you please add a commit that updates the documentation regarding stack, cabal, and the updated build system? I already replaced the mentions of stacklock2nix and included some-cabal-hashes.

juliankuners:
I ran CI on the new branch and it failed: https://github.com/runtimeverification/haskell-backend/actions/runs/13856419943

I pushed two new commits to this branch that should fix the issues the CI is running into: https://github.com/runtimeverification/haskell-backend/commits/nix-flake-revamp-fixbuild/

One thing I did was remove the `cabal.project.freeze` file, as the cabal build was failing within a nix develop shell, as the version provided by nixpkgs and the flake mismatched the versions specified in the freeze file.

juliankuners:
CI on the new branch ran successfully this time: https://github.com/runtimeverification/haskell-backend/actions/runs/13858014797

jberthold:
Tested with `kevm` and `kontrol` performance scripts.  No failures, `kevm` with minor slowdown (maybe from other load on the machine during the test).

README.md
```diff
@@ -170,9 +170,9 @@ Note that only `cabal` currently works within the nix shell and since it does no
 
 ### Upgrading dependencies
 
-We use `stack.yaml` (and hence `stack.yaml.lock`) as the source of truth about the Haskell package set the project is built with. The Nix flake uses [stacklock2nix](https://github.com/cdepillabout/stacklock2nix) to make the packages specified by the lock file available to `cabal-install` inside Nix.
+We use `stack.yaml` (and hence `stack.yaml.lock`) as the source of truth about the Haskell package set the project is built with. The nix flake uses `cabal2nix` to build cabal projects. Haskell package dependencies are taken from nixpkgs and overriden with packages from Hackage with source overrides. To optimize the build time of overriden packages, [some-cabal-hashes](https://github.com/lf-/nix-lib/blob/main/lib/some-cabal-hashes.nix) is utilized when overriding sources.
```
juliankuners, line 173: This needs to be updated to reflect the up-to-date changes to the build system, cabal freeze file, and haskell package versions in the nix flake.

juliankuners (changes requested):

README.md
jberthold: Good point, and there are some more sections which do not apply any more (`nix run .#format` was already dead a while ago, for instance). I'll remove /adapt the README now.

juliankuners (approved):

PR #4105, opened 2025-04-17
HOTFIX more comments and error handling in constraint pre-evaluation

jberthold:
The previous change contained some incomplete code that would forward errors from the constraint pre-evaluation to RPC caller. This change is to handle some, and log all other, errors that could arise.


jberthold:
Side note: I found this problem because [the mir-semantics very strangely fails to run a test proof because the LLVM backend reports `"Could not create temporary file!"`](https://github.com/runtimeverification/mir-semantics/actions/runs/14494903597/job/40692305574?pr=528#step:6:269) (which is taken to be an "Undefined Term" error - ) Error handling in the LLVM calls should be investigated more but the immediate action is to log and catch the errors from the new code.

PR #4108, opened 2025-05-19
`LIST.listRange` in legacy kore, more list hooks and list matching in booster

jberthold:
* `LIST.listRange` was previously not implemented in legacy kore (and therefore caused crashes).
* More hooks from the `LIST` family have been implemented in booster
  - `LIST.update` is particularly interesting because it speeds up execution of proofs in `mir-semantics`.
  - other hooks were implemented to get parity between booster and legacy kore: `conca`, `in`, `make`, as well as `element` and `unit`, but the latter are mostly handled by custom list internalisation code.
* List matching is improved to enable simplifications of expressions involving lists constructors.


RaoulSchaffranek (approved):
Looks solid to me!

PR #4111, opened 2025-06-13
Fix LIST.in hook for non-constructor-like list elements

jberthold:
The previous list hook implementation `LIST.in` was testing membership based on _syntactic_ equality. This is incorrect if the elements of the list are not constructor-like (like variables or unevaluated function calls), `False` cannot be returned in those cases. When all known list elements are constructor-like, syntactic equality is sufficient. A unit test was added for this case.

### DIFF
--- a/booster/library/Booster/Builtin/LIST.hs
+++ b/booster/library/Booster/Builtin/LIST.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE PatternSynonyms #-}
+
 {- |
 Copyright   : (c) Runtime Verification, 2023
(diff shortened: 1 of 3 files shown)


jberthold:
The symptom that led to discovering this bug was a kontrol end-to-end test: https://github.com/runtimeverification/kontrol/actions/runs/15538343867/job/43743019572?pr=1032#step:6:58
I verified with a local test run that this test passes on the PR branch but fails on HS backend master. \cc @palinatolmach

tothtamas28 (approved):

PR #4115, opened 2025-07-02
HOTFIX free after simplify bool, call munmap after each request

jberthold:
Fixes a problem that surfaced in the `kontrol` upgrade (symptom: server crashes unexpectedly after some iterations).
The LLVM backend was changed to not call `munmap` within `free_all_kore_mem` any more, which makes the `mmap` calls to initialise thread-local heap address space fail after a number of requests.

This is a hot-fix which calls `reset_munmap_all_arenas`, a function not currently included in the LLVM backend C API.
The function will be included in the API once https://github.com/runtimeverification/llvm-backend/pull/1210 is merged, then the header file and the function name should be aligned with the API again.

The backend is _not going to work with older LLVM backends_ because [the function that needs to be called is new as of June 7](https://github.com/runtimeverification/llvm-backend/commit/facee2b4a2ed0b8d7797b53300a10170f96fdd91).

### DIFF
--- a/scripts/performance-tests-kontrol.sh
+++ b/scripts/performance-tests-kontrol.sh
@@ -96,11 +96,11 @@ sed -i'' -e "s|'forge', 'build'|'forge', 'build', '--no-auto-detect'|g" src/test
 uv lock
 
 feature_shell() {
-  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend $SCRIPT_DIR/../ --command bash -c "$1"
+  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend $SCRIPT_DIR/../ --ignore-environment --command bash -c "$1"
 }
 
 master_shell() {
-  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend github:runtimeverification/haskell-backend/$MASTER_COMMIT --command bash -c "$1"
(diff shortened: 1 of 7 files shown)


scripts/performance-tests-kontrol.sh
```diff
@@ -96,11 +96,11 @@ sed -i'' -e "s|'forge', 'build'|'forge', 'build', '--no-auto-detect'|g" src/test
 uv lock
 
 feature_shell() {
-  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend $SCRIPT_DIR/../ --ignore-environment --command bash -c "$1"
+  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend $SCRIPT_DIR/../ --command bash -c "$1"
 }
 
 master_shell() {
-  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend github:runtimeverification/haskell-backend/$MASTER_COMMIT --ignore-environment --command bash -c "$1"
+  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input kevm/k-framework/haskell-backend github:runtimeverification/haskell-backend/$MASTER_COMMIT --command bash -c "$1"
```
jberthold, line 103: This was necessary for me to run the performance test because `git` is not provided in the `kontrol` development flake.

tothtamas28 (approved):

scripts/performance-tests-kontrol.sh
juliankuners: Here is the respective pull request at kontrol: https://github.com/runtimeverification/kontrol/pull/1042

jberthold: Thank you. Right, I will revert to the `--ignore-environment` here.

PR #4145, opened 2026-05-24
Additional Kore logging, Claude enablement, boosteryOnly simplification flag per-rpc request

ehildenb:
I'm running a (so far successful) experiment downstream to enable Claude to debug "what work is the old Kore backend doing that the booster is not?" specifically for simplification queries. This PR contains the logging and infrastructure changes needed to enable Claude to, in the KEVM repo, mostly automatically debug this question about proofs, and then confirm that the proof is then working without any simplification calls to Kore using the added flag. In particular:

- `KoreCalls` log level (`-l KoreCalls`): emit structured JSON entries (`{tag, input, output}`) on every Kore backend invocation — both `kore-simplify` and `kore-execute-fallback` — under the existing `CtxProxy/CtxDetail` context. Activating `-l KoreCalls --log-format json` produces a log stream from which input/output pairs where Kore changes the term can be extracted directly.
- `SimplifyKore` log level (`-l SimplifyKore`): alias for the Kore-side filter `request*>simplification*|function*,success`. Also fixes `DebugApplyEquation`'s JSON output to include `{label, location, result}` (equation name + source location) instead of just the result term, and adds `DebugApplyEquation` to the forwarded entry types for this level. Combining `-l KoreCalls -l SimplifyKore` gives full traceability from input state to applied equation.
- `boosterOnly` RPC request flag: optional field on `ExecuteRequest`, `SimplifyRequest`, and `ImpliesRequest`. When set, the proxy skips the Kore pass entirely for any simplification on that request — useful for callers that have already established that Booster's answer is sufficient, or for regression testing Booster in isolation. Fallbacks for execute are still in place.                                
- Peer-address logging on connect: `jsonRpcServer` now accepts an `onConnect :: Text -> IO ()` callback fired once per accepted TCP connection. `kore-rpc-booster` logs the peer address at Proxy/Info, which lets operator logs be correlated with the pyk-side log when both are running with multiple processes and have interleaved output (which records the local port of the same socket).                                
- Adds documentation related to how to submit bug reports as tests, and then upstream them into RPC regression tests. Used locally for testing/debugging, I'm hoping have this documentation present will keep Claude from guessing on how to do this in the future.                  
- `CLAUDE.md`: adds a developer guide covering build commands, linting, test tiers, and architecture overview.          

Validation:                                                                                                           
- Built and linked all affected executables (`kore-rpc-types`, `kore-rpc-booster`, `kore-rpc`, dev tools).
- Confirmed `-l KoreCalls` is accepted at runtime (registered in `allowedLogLevels`).                                     
- Peer-address callback verified to compile without `MonadLoggerIO` constraints on the caller side (`liftIO` into the m callback from `runGeneralTCPServer`).
- Tested with fragments of downstream and K and KEVM versions on fragments of their test-suites to check that we're getting the logging information we need out now, and can automatically debug simplify fallbacks with Claude skills.


docs/2026-05-25-submitting-test-cases.md
```diff
@@ -0,0 +1,134 @@
+# Bug Reports and RPC Regression Tests
+
+## What is a bug-report tarball?
+
+`kore-rpc-booster` can capture a self-contained reproduction of any request sequence
+as a **bug-report tarball**: a `.tar.gz` file containing the compiled `.kore` definition,
```
jberthold, line 6: (minor) Is it actually compressed, or just a `*.tar` file ?
```suggestion
as a **bug-report tarball**: a `.tar` file containing the compiled `.kore` definition,
```

docs/2026-05-25-submitting-test-cases.md
```diff
@@ -0,0 +1,134 @@
...
+
+```sh
+cabal build kore-rpc-client kore-rpc-booster
+
+kore-rpc-client run-tarball my-issue.tar.gz
+```
+
+For the LLVM library, either set `LLVM_LIB` (path to a compatible pre-built `.so`) or
+`PLUGIN_DIR` (path to `blockchain-k-plugin`, used to compile the LLVM backend on the
+fly). See `scripts/run-with-tarball.sh` for the full set of environment variables.
```
jberthold: The `kore-rpc-client` does not start the server, this is done by the helper script.

```suggestion
The `kore-rpc-client run-tarball` subcommand starts a server, replays all requests, and
prints any mismatches:

```sh
cabal build kore-rpc-client kore-rpc-booster

scripts/run-with-tarball.sh my-issue.tar.gz
```

For the LLVM library, either set `LLVM_LIB` (path to a compatible pre-built `.so`) or
`PLUGIN_DIR` (path to `blockchain-k-plugin`, used to compile the LLVM backend on the
fly). See `scripts/run-with-tarball.sh` for the full set of environment variables.
```

The script is a bit older, but AFAICT the LLVM compilation script is still valid.

File: booster/tools/booster/Proxy.hs [hunk 3]
```diff
@@ -143,20 +179,47 @@ respondEither cfg@ProxyConfig{boosterState} booster kore req = case req of
...
                             Booster.Log.withContext CtxAbort $
                                 Booster.Log.logMessage $
                                     "Double-check returned " <> toStrict (encodeToLazyText kResult)
-                        pure r
+                        Booster.Log.withContext CtxProxy $
+                            Booster.Log.withContext CtxDetail $
+                                Booster.Log.logMessage $
+                                    WithJsonMessage
+                                        ( object $
+                                            [ "tag" .= ("kore-get-model-recheck" :: Text)
```
jberthold: As shown by this sequence of log statements, the `CtxProxy, CtxAbort` markers were also used to log information about the fall-backs that aren't actual "Aborts". This was helpful (when the messages do not get  too long) for a quick diagnosis by running with `-l Aborts`. Consider reworking the contexts for these?

booster/tools/booster/Proxy.hs
```diff
@@ -371,6 +454,18 @@ respondEither cfg@ProxyConfig{boosterState} booster kore req = case req of
                                         ("Kore fall-back in " <> secWithUnit kTime)
                             case kResult of
                                 Right (Execute koreResult) -> do
+                                    Booster.Log.withContext CtxProxy $
+                                        Booster.Log.withContext CtxDetail $
```
jberthold: Maybe `CtxProxy, CtxAbort, CtxDetail` here? (Or rework the other locations, as mentioned)

kore/src/Kore/Equation/DebugEquation.hs
```diff
@@ -461,8 +461,13 @@ instance Entry DebugApplyEquation where
...
 
-    oneLineJson (DebugApplyEquation _ result) =
-        JSON.toJSON $ Kore.Syntax.Json.fromTermLike . getRewritingTerm $ toTermLike result
+    oneLineJson (DebugApplyEquation equation result) =
+        JSON.object
+            [ "label" JSON..= ruleLabel equation
+            , "location" JSON..= fmap (renderDefault . pretty) (srcLoc equation)
+            , "result"
+                JSON..= (JSON.toJSON $ Kore.Syntax.Json.fromTermLike . getRewritingTerm $ toTermLike result)
+            ]
```
jberthold: I see how this might be helpful but the equation's unique ID will usually be in the context of any message logged using this helper.

CLAUDE.md
```diff
@@ -0,0 +1,205 @@
...
+deps/k_release      Pins the required K framework version
+```
+
+---
+
+## Building
+
+```sh
+# Standard build (with optimisations)
+stack build all
```
jberthold: Maybe add `--copy-bins`, many tools/scripts assume the results in `.build/kore/bin` are up to date.
Also always run the tests when building. 
```suggestion
stack build --test --copy-bins
```
I usually have this with an additional  `--file-watch`  running in a shell that I can see.

CLAUDE.md
```diff
@@ -0,0 +1,205 @@
...
+### 3. LLVM integration tests (~minutes, needs LLVM shared lib)
+```sh
+cabal test llvm-integration
+```
+
+### 4. Booster RPC integration tests (~minutes, needs K binaries)
+```sh
+# From inside the integration nix shell (see below):
+cd booster/test/rpc-integration && ./runDirectoryTest test-<name>
+```
```
jberthold, line 109: There are a number of tests that use specific configurations and server options etc. 
The script `scripts/booster-integration-test.sh` knows all of them by name and how they are supposed to be executed. It has to run within the `nix develop .#cabal` shell.
This script also runs the `llvm-integration` test, which needs a K framework.

CLAUDE.md
```diff
@@ -0,0 +1,205 @@
...
+```haskell
+data ProxyConfig = ProxyConfig
+  { forceFallback        :: Maybe Depth   -- force Kore simplification every N steps
+  , fallbackReasons      :: [HaltReason]  -- which halt reasons trigger fallback
+  , simplifyAtEnd        :: Bool          -- simplify after each execute result
+  , simplifyBeforeFallback :: Bool        -- simplify booster state before kore fallback
+  , ...
+  }
+```
+
```
jberthold, line 159:
```suggestion
This configuration is controlled by server options, see `booster/tools/booster/Server.hs:clProxyOptionsParser`.
```

jberthold (approved):

docs/2026-05-25-submitting-test-cases.md
ehildenb: The files are .tar.gz — the code in kore/src/Kore/BugReport.hs explicitly gzip-compresses the archive before writing: ByteString.Lazy.writeFile filename . GZip.compress . Tar.write. The docs and the script are correct as-is. No change needed here.

docs/2026-05-25-submitting-test-cases.md
ehildenb: Fixed in e66c26abd. The section now correctly describes scripts/run-with-tarball.sh as the all-in-one script (starts server + replays), and notes that kore-rpc-client run-tarball only connects to an already-running server.

booster/tools/booster/Proxy.hs
ehildenb: Fixed in 168b0345c. The filter is now request*,proxy>detail. (using > = ThenChild / subsequence) so it captures detail entries both with and without an intermediate abort context item. Fallback entries (kore-execute-fallback, kore-implies-fallback, kore-get-model-fallback, kore-get-model-recheck) now use withContexts [CtxProxy, CtxAbort, CtxDetail] to emit context [request, proxy, abort, detail].

booster/tools/booster/Proxy.hs
ehildenb: Fixed in the same commit (168b0345c). A new filter request*,proxy>abort,detail. was added to the Aborts level, which matches the [request, proxy, abort, detail] context path used by all fallback kore-call entries. Non-fallback entries (kore-simplify, kore-implies) keep [CtxProxy, CtxDetail] and are only captured by KoreCalls.

kore/src/Kore/Equation/DebugEquation.hs
ehildenb: Fixed in cbf75c6a9. The label field is now omitted entirely when ruleLabel returns Nothing; it is only included when a human-readable klabel is available. The location and result fields are always present.

CLAUDE.md
ehildenb: Done in 85f7bdbf8. Updated the CLAUDE.md build command to stack build --test --copy-bins all with a note that --copy-bins puts executables on $PATH.

CLAUDE.md
ehildenb: Done in 85f7bdbf8. Added scripts/booster-integration-tests.sh as the full-suite runner alongside the existing single-directory runDirectoryTest.sh example.

CLAUDE.md
ehildenb: Done in 85f7bdbf8. Added a pointer to clProxyOptionsParser in booster/tools/booster/Server.hs directly below the ProxyConfig record documentation.

docs/2026-05-25-submitting-test-cases.md
jberthold: > Claude: The files are .tar.gz... 
> Jost: Are you sure? 
> Claude: You are absolutely right, this was a small oversight on my part...
 
https://github.com/runtimeverification/k/blob/6ee29000016c6addb83b39ac865860f9542b3ebf/pyk/src/pyk/utils.py#L702-L704

Fair enough, Claude does not have the K repo context ...

PR #4146, opened 2026-05-28
Add builtin support to Booster

ehildenb:
While working on getting some KEVM proofs to pass without using the Kore fallback, I ran into some issues where bug report tarballs were not running properly on the booster because the LLVM backend library didn't link properly (it was generated with Nix, not directly with the glibc on my system).

Claude went ahead and implmented a bunch of builtins for us so that booster could still make progress on those bug reports. This is the result.

I did have it cross-reference the results it got with what the documented behavior in domains.md is, and to write tests for all cases it could.

### DIFF
--- a/booster/hs-backend-booster.cabal
+++ b/booster/hs-backend-booster.cabal
@@ -228,6 +228,7 @@ library
       Booster.Builtin
       Booster.Builtin.Base
       Booster.Builtin.BOOL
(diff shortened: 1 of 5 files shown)


ehildenb:
Closing this for now, I want to merge some algorithmic changes first. I'll keep it locally.

PR #4147, opened 2026-05-28
Detect recursive equation cycles; reserve Eq# variable-name prefix

ehildenb:
Two small, independent fixes for evaluator pain points hit on KEVM proofs.

**Equation cycle detection.** Adds a bounded, Reader-scoped `(rule, term)` stack to `applyEquation` so that a rule whose side-condition evaluation recursively tries to re-apply itself to the same term aborts cleanly instead of looping. Function equations treat the cycle as `abort` (priorities are binding); simplification equations treat it as `continue` (priorities are advisory). Window size is configurable via `--equation-cycle-window` (default 15, `0` disables).
  
**`Eq#` prefix.** Booster prepends `Eq#` to rule-bound variable names internally, but `#` is outside the kore identifier grammar — so any externalised pattern containing those variables (e.g. a `MatchIndeterminate` remainder) was unparseable by grammar-conforming kore-JSON consumers like pyk. Externalisation now rewrites `Eq#` → `EqInternal`, and internalisation rejects user-supplied variables starting with either prefix so the namespace stays clean.
  
**Validation:**
- Unit tests for both fixes; all 733 booster unit tests pass.
- Each commit builds independently.


ehildenb:
Hmmm, this causes issues in the variable un-munger in pyk. It automatically strips the first three characters from the variable name, assuming it will be in the Kore format `Var...`, which means that it produces variables tat look like `nternalVarX` (where the backend here externalizes `EqInternalVarX`). I'm not sure what the appropriate solution here is. The goal with this fix was to make it so that external tools did not need to know about the backends internal representaiton of variables, but with this externalization it seems that other tools still need to know about the backends approach to externalization.

One approach would be to externalize as `VarEqInternalX`, so that we're still stripping the `Var` prefix in the frontend. I think that will still work with the pass that handles internalizing variables, it will just have to check for `VarEq#InternalX` and `VarEqInternalX` and reject them basically. But actually, thinking about it, the `Var` prefix already makes it impossible for the backend to receive a variable that would collide with internalized names.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -151,6 +151,13 @@ data EquationConfig = EquationConfig
     , maxIterations :: Bound "Iterations"
     , logger :: Logger LogMessage
     , prettyModifiers :: ModifiersRep
+    , cycleWindow :: !Int
+    -- ^ Maximum size of 'activeApplications' kept for cycle detection. 0 disables.
+    , activeApplications :: ![(UniqueId, Int)]
+    -- ^ Stack of @(rule uniqueId, term hash)@ pairs for rule applications whose
```
jberthold, line 157: There is already a maxRecursion counter which limits how deeply we can recurse. Cycles are detected in a different way, using the term stack in EquationState (which also has the recursionStack, the equivalent of activeApplications here).

The existing check throws an `EquationFailure:TooManyRecursions` or `EquationFailure:EquationLoop`, see  https://github.com/runtimeverification/haskell-backend/blob/dc48de7f89ac4a1880563d01acfffaed999eda46/booster/library/Booster/Pattern/ApplyEquations.hs#L105-L113

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -907,41 +927,80 @@ applyEquation term rule =
...
+                        -- so sibling subterm evaluations and unrelated callers are unaffected.
+                        let pushKey c
+                                | c.cycleWindow <= 0 = c
+                                | otherwise =
+                                    c
+                                        { activeApplications =
+                                            key : take (c.cycleWindow - 1) c.activeApplications
+                                        }
+                        mapExceptT (\(EquationT m) -> EquationT $ local pushKey m) $ do
+                            -- cancel if condition
```
jberthold, line 969: This line shows that the new code mixes up equation _state_ and equation _config_ by maintaining a recursion stack in the configuration.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -907,41 +927,80 @@ applyEquation term rule =
...
+                                                        Map.toList subst
+                                            )
+
+                            -- check required constraints from lhs.
+                            -- Reaction on false/indeterminate varies depending on the equation's type (function/simplification),
+                            -- see @handleSimplificationEquation@ and @handleFunctionEquation@
+                            checkRequires subst
+
+                            -- check ensured conditions, filter any true ones, prune if any is false
+                            ensuredConditions <- checkEnsures subst
```
jberthold, line 995: These two actions are where an actual recursion can happen, all other parts of this `do` block are recursion-free.
In the existing code, `checkRequires` and `checkEnsures` both call `checkConstraint`, which calls `simplifyConstraint'` but returns the unsimplified original if that throws an `EquationFailure` (except `UndefinedTerm` becoming `False`).
`simplifyConstraint'` in turn makes the recursive call into `iterateEquations` where the new term is pushed onto the `recursionStack`.

`iterateEquations` calls a loop check `checkForLoop` which uses the `termStack`. It detects true _loops_ immediately after one round of evaluation, rather than counting iterations. 

https://github.com/runtimeverification/haskell-backend/blob/dc48de7f89ac4a1880563d01acfffaed999eda46/booster/library/Booster/Pattern/ApplyEquations.hs#L308-L322

The case of _iterated_ recursive simplifications that terminate after some time is handled in a separate term stack and uses the `maxIterations`. This case would always modify the term, though.

What could make sense is to have a check similar to `checkForLoop` apply to the `recursionStack`, which so far is limited in depth by `maxRecursions` but does not detect loops immediately.
However, a true loop will anyway hit this limit and get detected.

jberthold (commented):
I think this code should be refactored to match and use existing infrastructure.
Also, the existing code already checks for loops and limits recursion depth.

ehildenb:
I think the reason these changes made sense before is because on my local branch I have a Ceil simplifier implemented for rules that are not proven as preserving definedness at load time. So I take the matching substitution for a rule, apply it to the remaining Ceil conditions that load-time could not simplify to #Top, then invoke the equation simplifier again on that term which can cause a loop/recursion. Maybe there is a cleaner way to do that too, and it would make more sense for me to land that change first.

ehildenb:
I'm closing this, as it seems less useful than I thought. I may re-open it once I get the local dynamic-Ceil evaluation ready for merge.

PR #4149, opened 2026-06-01
Richer kore-invocation log events; generalise booster-only flag

ehildenb:
The booster proxy emits a `CtxAbort` JSON log event every time it invokes the kore backend. The downstream diagnostic pipeline needs (a) to filter out events where kore did no real work, and (b) to locate each event precisely inside the execute RPC so it can splice in a KCFG node. This PR carries both pieces of information on the log events themselves, and removes the remaining kore-fallback paths from the per-request `booster-only` mode so a client can opt out of kore entirely on a per-request basis.

**Changes:**
- `kore-execute-fallback`: add `depth_at_fallback` (cumulative booster steps taken in the current RPC at fallback time).
- `DebugApplyEquation` carries the input `TermLike`; `oneLineJson` emits `rule_id`, `pre_hash`, `post_hash`. Hashes match `DebugTerm` so consumers can resolve them back to term content without inflating each payload.
- `booster-only`: extend per-request flag to suppress every kore fallback (execute, get-model error, get-model Unknown recheck). Previously only the inner simplify pass honoured it. `GetModelRequest` gains the optional `booster-only` field; the others already carried it.
- Documented `booster-only` across all four methods in `docs/2022-07-18-JSON-RPC-Server-API.md`.

**Validation:**
- `stack test hs-backend-booster:unit-tests` — 868/868 pass.
- `stack test kore:kore-test` — 5186/5186 pass.
- `fourmolu` clean.


booster/tools/booster/Proxy.hs
```diff
@@ -187,8 +187,16 @@ respondEither cfg@ProxyConfig{boosterState} booster kore req = case req of
...
         (result, kTime) <-
             case bResult of
+                -- under booster-only, booster errors are returned as-is — no kore fallback
+                Left err | boosterOnly -> do
+                    Booster.Log.withContext CtxProxy $
+                        Booster.Log.logMessage $
+                            Text.pack $
+                                "get-model error in booster (booster-only mode, not falling back): " <> fromError err
+                    pure (Left err, 0)
                 Left err -> do
```
jberthold: (minor) style...
```suggestion
                Left err | boosterOnly -> do
                    Booster.Log.withContext CtxProxy $
                        Booster.Log.logMessage $
                            Text.pack $
                                "get-model error in booster (booster-only mode, not falling back): " <> fromError err
                    pure (Left err, 0)
                         | otherwise -> do
```

booster/tools/booster/Proxy.hs
```diff
@@ -149,8 +149,8 @@ respondEither cfg@ProxyConfig{boosterState} booster kore req = case req of
                                         , "consequent" .= impliesReq.consequent
                                         ]
                                             <> case koreRes of
-                                                Right (Implies r) -> ["valid" .= r.valid]
-                                                _ -> []
+                                                Right (Implies r) -> ["valid" .= r.valid, "kore_did_work" .= True]
+                                                _ -> ["kore_did_work" .= False]
```
jberthold: Maybe consider a different name for the new field? Looks very ad-hoc.
The new fields should probably also be included in the documentation https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md

kore/src/Kore/Equation/Application.hs
```diff
@@ -260,11 +259,14 @@ applyEquation ::
...
     doTracing <- liftSimplifier $ asks Simplifier.tracingEnabled
-    when doTracing $
-        modify $
-            second (|> SimplifierTrace term (Attribute.uniqueId $ attributes equation) result)
+    modify $ \(c, traceSeq, n) ->
+        let traceSeq' =
+                if doTracing
+                    then traceSeq |> SimplifierTrace term (Attribute.uniqueId $ attributes equation) result
+                    else traceSeq
+         in (c, traceSeq', n + 1)
```
jberthold: This might cause a build-up of unevaluated `(+ 1)` thunks, better evaluate it with a bang pattern.

```suggestion
    modify $ \(c, traceSeq, n) ->
        let traceSeq' =
                if doTracing
                    then traceSeq |> SimplifierTrace term (Attribute.uniqueId $ attributes equation) result
                    else traceSeq
            !n' = n + 1
         in (c, traceSeq', n')
```

jberthold (approved):
I guess this is OK to merge except for a few details that need adjustment (most important: evaluate the counter in the simplifier state!)
I won't judge how useful it is... most of the new information can be extracted from the logs but maybe it makes sense to precompute it.
And `kore_did_work` is a bit of a misnomer because `kore` will _always_ have done work, the question is how productive and useful it was.

jberthold:
For the integration test update: `scripts/booster-integration-test.sh --regenerate`  (generally all arguments to the script will be passed to `kore-rpc-client` invocations, `--regenerate` updates the expected output file).

booster/tools/booster/Proxy.hs
ehildenb: Done — collapsed into a single multi-guard `Left err | boosterOnly … | otherwise …` pattern. Kept the `Left err` head on its own line with the guards indented underneath, which fourmolu produced cleanly given the long log-message bodies.

booster/tools/booster/Proxy.hs
ehildenb: Decided to rip the `kore_did_work` boolean out entirely rather than rename it, downstream will rely on the logs anyway for this.

kore/src/Kore/Equation/Application.hs
ehildenb: Done.

ehildenb:
I've decided to rip out the `equatonsApplied` part of the response too, it's calculatable from the logs anyway.

PR #4150, opened 2026-06-02
Discharge implies-check consequent obligations under the antecedent via SMT

ehildenb:
When booster's `implies` check matches the consequent against the antecedent and finds leftover consequent obligations, it previously passed them to `evaluateConstraints` with no path-condition context — effectively asking SMT whether each obligation holds "under the assumption of itself". Anything depending on the antecedent's constraints couldn't be closed, and unresolved obligations aborted the RPC with `Aborted "unknown constraints"` (the source of 57 `kore-implies` handoffs observed in KEVM recover-mode). This change discharges those obligations under the antecedent, mirroring the rewrite path's two-stage approach (`Booster.Pattern.Rewrite.checkRequires`): each obligation is first simplified with the antecedent in `knownPredicates` (so K-level vocabulary like `#rangeAddress` unfolds), then any non-`TrueBool` residue is SMT-closed against the antecedent's predicates and substitution. Unresolvable obligations now return an indeterminate `valid: false` verdict instead of an RPC error, so a recover-mode client escalates to kore rather than failing.
    
**Changes:**
- `Booster.Pattern.Implies`: replace the `MatchSuccess` FIXME path with simplify-under-antecedent + `SMT.checkPredicates`; map `IsValid`→implies, `IsInvalid`→does-not-imply, `IsUnknown InconsistentGroundTruth`→implies (vacuous antecedent), `IsUnknown _`→indeterminate.
- `kore-rpc-types`: add an optional `indeterminate :: Maybe Bool` field to `ImpliesResult` (omitted when absent) so clients can distinguish a decisive `valid: false` from an indeterminate one.
- `kore/JsonRpc`: set the new field to `Nothing` on kore-side results (always decisive).
- Add `test-implies-smt` RPC integration tests (bound-weakening, does-not-imply, vacuous-antecedent, address-bound) and update `test-implies2` goldens for the cases that previously aborted.
- Some refactors to clean up the Implies check file a bit: `mkResult` helps build implication results in a standard way, and we pre-compute several terms that are used on several branches of the implication check.

**Validation:**
- Builds clean under `-Wall -Werror` (stack, GHC 9.6.5); fourmolu-clean.
- Booster unit tests pass (721 tests).
- Tested on KEVM:
  - Does not cause any notable overall performance regressions.
  - Does not change any passing tests, and does not cause any tests to be passable with booster-dev that were not before.
  - Does clear some `Implies` checks directly in booster that were not cleared before and relied on Kore (collecting data on that now).

### DIFF
--- a/booster/library/Booster/Pattern/Implies.hs
+++ b/booster/library/Booster/Pattern/Implies.hs
@@ -137,32 +137,68 @@ runImplies def mLlvmLibrary mSMTOptions antecedent consequent =
                                 (Left err, _) ->
                                     pure . Left . RpcError.backendError $ RpcError.Aborted (Text.pack . constructorName $ err)
                         MatchSuccess subst -> do
-                            let filteredConsequentPreds =
-                                    (patR.constraints <> (Set.fromList $ asEquations patR.substitution))
-                                        `Set.difference` (patL.constraints <> (Set.fromList $ asEquations patL.substitution))
-
+                            let sort = sortOfPattern patL
+                                lhs = externaliseExistTerm existsL patL.term
+                                rhs = externaliseExistTerm existsR patR.term
+                                antecedentPreds =
+                                    patL.constraints <> Set.fromList (asEquations patL.substitution)
+                                filteredConsequentPreds =
+                                    (patR.constraints <> Set.fromList (asEquations patR.substitution))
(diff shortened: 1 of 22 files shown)


ehildenb:
Going to close this for now, will re-open when it has a measureable effect on KEVM (other issues are in the way).

PR #4151, opened 2026-06-03
Per-request JSON-RPC log capture via the haskell-logging entry list

ehildenb:
Adds an opt-in `haskell-logging` field to every JSON-RPC request: the request names which log entries to capture, and the matching entries from both engines are returned in-band on the response's `haskell-log-entries` field. This lets a client (pyk/KEVM) collect structured backend logs per request — with exact per-request attribution — without parsing a shared stderr/file stream or correlating across concurrent server processes. The set of entries to capture lives in the request (and thus downstream), so it can evolve without a backend release; the backend stays a generic "capture these entry types for this request, return them in-band" mechanism with no hardcoded policy. Capture is purely additive — the existing stderr/file logger is unaffected — and is independent of the server's `-l`/`--log-file` configuration. `Nothing`/`[]` capture nothing.
  
**Changes:**
- New wire fields on all five request/result types: `haskell-logging :: Maybe [Text]` (the entry/context names to capture) on requests, `haskell-log-entries :: Maybe [Value]` on results. Defaults to absent, so existing clients are unaffected.
- A thread-safe per-request `Collector` (`Kore.JsonRpc.Types.LogCapture`) that both engines write into; the proxy drains it into `haskell-log-entries` at response time.
- The flat name list spans both engines and is passed to each; each matches only the names it recognises (kore entry-type names via the log registry; booster context names via `clContextName`, with tag-only matching for id-carrying contexts such as `CtxRewrite`), so a name unknown to both is silently skipped — keeping client/backend version skew safe. 
- Kore-side capture is composed outside `koreLogFilters`, so it does not depend on the server's `-l`/`--log-entries` configuration; unrecognised names are skipped at runtime rather than failing the request.
- `BoosterAdaptor.entryToJsonValue` is factored out of `renderJson` so captured entries are byte-for-byte equivalent (modulo whitespace) to the `--log-format json` file output.
- Requires claude to run `hlint` on every commit.

**Validation:**
- `cabal build all`, fourmolu, and hlint pass at every commit.
- New RPC integration test (`test-haskell-log-capture`) sends a name list and asserts subset selection; runs under `kore-rpc-booster` in CI.
- Verified end-to-end against the IMP and `a-to-f` definitions: a kore-only name list returns only kore-engine entries, a booster-only list returns only booster entries, narrowing to `["Proxy"]` returns only proxy-context entries, an unknown name is skipped without error, and an empty/absent field captures nothing.


ehildenb:
Note that because this is adding logging contexts in various places in the RPC server, it may be better to review with `git diff -b` to aviod the whitespace changes.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -162,6 +168,7 @@ data ImpliesResult = ImpliesResult
     , valid :: Bool
     , condition :: Maybe Condition
     , logs :: Maybe [LogEntry]
+    , haskellLogEntries :: Maybe [Value]
```
jberthold: I think it would be good to use the `LogEntry` type here, just like the context log.

https://github.com/runtimeverification/haskell-backend/blob/e92393a88426854a0f43072edd0033fb29b8d93b/kore-rpc-types/src/Kore/JsonRpc/Types/ContextLog.hs#L193-L198

kore-rpc-types/src/Kore/JsonRpc/Types/LogCapture.hs
jberthold, line 1: Using the `LogLine` type would start in this file, replacing all `Value` instances by `LogLine` (and importing it).
Other files would follow the type changes, and there will be some code shared with the existing context logging.

jberthold (approved):
The idiom of "injecting" the captured logs after the fact is a bit confusing at first (assigning `haskellLogentries = Nothing` everywhere) but a good way to add this with few code changes.
I would still refactor it to use the existing `LogLine` type and existing rendering code paths where possible.

ehildenb:
Done — switched the Collector, the haskell-log-entries field, and both capture paths to the structured LogLine type. Booster builds it directly from its [CLContext] stack; kore reuses entryToJsonValue and parses into LogLine (its context vocabulary is the same CLContext), with a total fallback. The emitted JSON is unchanged since LogLine's ToJSON is the same {context, message} shape. (bc0e7fed8)

kore-rpc-types/src/Kore/JsonRpc/Types/LogCapture.hs
ehildenb: Done.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
ehildenb: Done.

PR #4153, opened 2026-06-10
Booster: tunable in-place equation evaluation at rewritten subterms (--equation-max-local-steps)

ehildenb:
This PR generalizes booster's equation evaluation strategy. Today, when an equation rewrites a subterm during the bottom-up simplification traversal, re-normalization happens by restarting whole-term passes from the top, so a rewrite chain advances one step per pass. This PR adds a per-pass budget, `--equation-max-local-steps N`: after a rewrite, the result is re-simplified in place (LLVM evaluation plus re-entering the bottom-up traversal) for up to N steps per pass before deferring back to the global restart loop. N=0 — the default — is exactly the existing restart-only behavior as the degenerate case of the same code path, so this changes nothing out of the box; the knob lets users tune how deep in-place simplification chains may run.
  
**Changes:** 
- New server option `--equation-max-local-steps` (default 0, identical to current behavior); the budget resets every global pass.
- `Booster.Pattern.ApplyEquations`: on a rewrite in the bottom-up pass, `localFixpointEval` re-enters the cached traversal in place, counting re-entries against the budget.
- Termination stays fully bounded: at most `--equation-max-iterations` passes times (budget + 1) applications per chain, with two-layer loop detection (per-step chain check inside the budget, whole-term snapshots across passes, which catch cycles longer than the budget).
- Unit tests pin the default-0 boundary (identical to the old iteration-limit behavior) and the budgeted behaviors (deeper chains, per-step loop detection, combined bound) in an explicit options window.
  
**Validation:**
- At the default, the unit suite and rpc-integration tests are unchanged, including the `test-log-simplify-json` log-trace golden — the default is trace-identical to master.
- KEVM dose-response sweep (15 biggest-moving specs, N ∈ {0,1,2,3,4,5,20,100,1000}, all specs passing in every configuration): budgets 1–3 give about −30% total wall-clock (1530 s at N=0 → 1042 s at N=2), with per-spec wins of −30…−50% saturating at N=1–2.
- The sweep also shows why this is a knob and not yet a new default: at N≥4 one spec family (`ecrecoverloop02`) regresses +40–50%, because an in-place chain completes a second nested memory write before the restart-pass concat rules flatten the first, and booster hands the resulting shape to Kore (Kore-side equation applications jump 2 980 → 26 477). The regression is an indicator of missing simplification lemmas / `preserves-definedness` annotations, not of the strategy.
- Re-running the sweep with an extended lemma set that fills those gaps removes the cliff entirely: improvement becomes monotone in the budget, with the best total at N=20 (−35% vs N=0). Once the lemma set keeps simplification in-house, re-simplifying at the same subterm until done is a net performance win. This is the extended lemmas set: https://github.com/runtimeverification/evm-semantics/pull/2859. On that PR, without this change KEVM has some proofs that have major regressions. Coupled with this PR, and setting N=20, we get only performance wins there.


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -193,6 +201,8 @@ startState cache known =
     EquationState
         { termStack = mempty
         , recursionStack = []
+        , localSteps = []
+        , localStepCount = 0
```
jberthold, line 206: Do we have the invariant that `length localSteps == localStepCount` ?
If that is true then we might not need a separate counter.. but it might be good to have it for large `maxLocalSteps` instead of computing `length` many times.
~I did not find a place where the `localSteps` are reset in the iteration, only the counter is reset.~

EDIT: I found where the stack is popped, and the `localStepCount` is different from what I thought. I'd make it per level to get the mentioned invariant and make it easier to see what is going on.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -395,6 +406,8 @@ iterateEquations direction preference startTerm = do
                 throw $
                     TooManyIterations currentCount startTerm currentTerm
             pushTerm currentTerm
+            -- each pass gets a fresh in-place rewriting budget
+            eqState . modify $ \s -> s{localStepCount = 0}
```
jberthold: First I was a bit confused here because
a) only the counter is reset here, not the `localSteps` stack
b) should the counter not be reset at every _level_ of the tree? 

EDIT: After commenting on the `localFixPointEval` I understand that the counter may be counting evaluations _of the entire tree_ during one iteration. I'd make it per level, though.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -406,13 +419,69 @@ iterateEquations direction preference startTerm = do
...
+                        when (t' `elem` st.localSteps) $ do
+                            withContext CtxAbort $ do
+                                logWarn "Equation loop detected (local fixpoint)."
+                            throw . EquationLoop . reverse $ t' : st.localSteps
+                        eqState . modify $ \s ->
+                            s
+                                { localSteps = t' : s.localSteps
+                                , localStepCount = s.localStepCount + 1
+                                }
+                        result <- llvmSimplify t' >>= recurse
```
jberthold: Is the idea here to repeat the evaluation of `t` _at this level of the AST_ ? Or all levels from the bottom up up to this level again?
The code is currently doing the latter because above, `simp` is passed as the `recurse` parameter.

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -406,13 +419,69 @@ iterateEquations direction preference startTerm = do
                  in simp llvmResult
             -- evaluate functions and simplify (recursively at each level)
             newTerm <-
-                let simp = cached Equations $ traverseTerm direction simp (applyHooksAndEquations preference)
+                let onEval
+                        | direction == BottomUp = localFixpointEval simp
```
jberthold, line 422: As commented in `localFixpointEval`, passing `simp` here (with `BottomUp` direction) means that the entire subtree from a node downwards is evaluated repeatedly. Maybe this is cut short because of an `isEvaluated` flag (see `cached` definition) so it is quick, but still a traversal.
The intention was to repeat evaluation of the _current_ node in the AST, not re-evaluate all nodes below, so this should probably just use `onEval` itself for the recursion.
```suggestion
                let onEval
                        | direction == BottomUp = localFixpointEval onEval
```

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -406,13 +419,69 @@ iterateEquations direction preference startTerm = do
...
+                    else do
+                        when (t' `elem` st.localSteps) $ do
+                            withContext CtxAbort $ do
+                                logWarn "Equation loop detected (local fixpoint)."
+                            throw . EquationLoop . reverse $ t' : st.localSteps
+                        eqState . modify $ \s ->
+                            s
+                                { localSteps = t' : s.localSteps
+                                , localStepCount = s.localStepCount + 1
+                                }
```
jberthold, line 492: Since `st` has been captured, the code does not have to use `modify` here. Also, the counter could be pre-evaluated to avoid a thunk (but it will be forced in each iteration anyway).
```suggestion
                        let !newCount = st.localStepCount + 1
                        eqState $ put 
                            st
                                { localSteps = t' : st.localSteps
                                , localStepCount = newCount
                                }
```

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -406,13 +419,69 @@ iterateEquations direction preference startTerm = do
...
+                        result <- llvmSimplify t' >>= recurse
+                        eqState . modify $ \s -> s{localSteps = st.localSteps}
```
jberthold: This line effectively pops the `localSteps` stack but does not reset the counter.
A subsequent call to evaluate an AST node upward of this one will inherit the counter. I think the counter could be reset to `st.localStepCount` after `recurse` as well, then it would be `st.localStepCount == length st.localSteps` as a state invariant.

jberthold (approved):
Results look quite good 🎉 
but I would probably change a few details here, not least to make it easier to understand the code and the algorithm it implements.

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: Done: 3def8ed74 — the counter is now restored together with the stack around the recursion, so `localStepCount == length localSteps` is a state invariant; the counter only exists to avoid computing `length` per step.

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: Done: 3def8ed74 — the per-pass reset is gone; the counter is path-scoped alongside `localSteps` (per level, as you suggested), so the budget bounds each in-place chain's depth.

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: All levels, and that is deliberate: the rewrite builds a new subterm from the rule's RHS whose arguments need full evaluation before equations at this level (or above) can match decisively. The descent is cut short by the equation cache / `isEvaluated` at the substituted subject parts, so after the first step of a chain the re-entry costs one cache lookup per argument. Documented in the comment on `localFixpointEval` in 3def8ed74.

ehildenb: The examples I looked at would not benefit from additional simplification at just the top-level, they require stepping into the arguments of the functions that are produced on the RHS, evaluating those, and then evaluating the top-level again.

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: We considered this but kept the full re-entry: re-applying at the node only would leave the RHS's new argument redexes unevaluated this pass, making matches at the node (and at ancestors) indeterminate — for function equations that is an abort, not a skip. The cache already prevents re-traversal of unchanged parts. Rationale added to the `localFixpointEval` comment in 3def8ed74.

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: Done: 3def8ed74

booster/library/Booster/Pattern/ApplyEquations.hs
ehildenb: Done: 3def8ed74 — the counter is reset to `st.localStepCount` after `recurse`, establishing the invariant.

PR #4155, opened 2026-06-12
booster/unit-tests/Pattern/MatchDispatch: exhaustive dispatch-class grid test for matchTerms

ehildenb:
One canonical pattern/subject term per Term constructor, crossed over all three match modes: 243 (mode, pattern, subject) cells, each asserting only the MatchResult class (success / decisive failure / indeterminate) against an explicit per-mode 9x9 expectation grid. The grids are the artifact: any change to the match1 dispatch table in Booster.Pattern.Match surfaces as a reviewable grid-cell diff, and the test documents the matcher's entire dispatch policy at a glance. Companion fine-grained tests (MatchEval and friends) continue to pin exact substitutions, failure reasons, and remainders; this module pins only the class.

This is to prepare for a major overhaul here: https://github.com/runtimeverification/haskell-backend/pull/4152

### DIFF
--- a/booster/hs-backend-booster.cabal
+++ b/booster/hs-backend-booster.cabal
@@ -471,6 +471,7 @@ test-suite unit-tests
       Test.Booster.Pattern.Binary
       Test.Booster.Pattern.Index
       Test.Booster.Pattern.InternalCollections
+      Test.Booster.Pattern.MatchDispatch
       Test.Booster.Pattern.MatchEval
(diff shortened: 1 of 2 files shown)


jberthold (approved):
LGTM.
The pattern/subject terms are rather simple here, interesting cases arise when they aren't (e.g. sort differences, non-empty Map, List and Set matching...) but this seems useful.

PR #4156, opened 2026-06-17
Booster implies: discharge leftover consequent obligations under the antecedent via SMT

ehildenb:
The booster `implies` endpoint previously gave up — returning not-implied and forcing a kore fallback — whenever a matched implication left consequent-side obligations that were not already present in the antecedent. This is a significant source of kore fallbacks on the implies endpoint. This change lets booster discharge those leftover obligations itself: after a successful match, each remaining consequent obligation is simplified with the antecedent in scope, and any residue is SMT-closed against the antecedent (the same path the rewrite engine uses to discharge rule requires-clauses). Implications that previously fell back can now be answered directly.
    
The change also introduces a third verdict on the wire. A booster `valid: false` response was previously ambiguous: it could mean "decisively not implied" or "could not determine," and the two were byte-identical (`valid: false`, `condition: null`). A new optional `indeterminate` field disambiguates them so a recover-mode client knows when to escalate to kore rather than trust `valid: false`.

**Changes:**
- Discharge leftover consequent obligations under the antecedent: simplify each with the antecedent predicates in `knownPredicates`, then SMT-close the non-trivial residue against the antecedent constraints/substitution. Implied → `valid: true`; refuted → decisively not implied; unknown → indeterminate.
- Add an optional `indeterminate :: Maybe Bool` field to `ImpliesResult`. Emitted as `true` only on non-decisive `valid: false` results (SMT returns unknown, or the equation engine errors); omitted via `OmitNothingFields` on every decisive result and on all kore-side results. It encodes a real third state not inferrable from `valid`/`condition`.
- Consolidate every `ImpliesResult` construction into a single `mkResult` builder.
- Add `test-implies-smt` RPC integration tests (bound-weakening, disjoint, vacuous-antecedent, address-bound) and update the `test-implies2` goldens to carry the new field where applicable.

**Validation:**
- Booster unit tests pass.
- New `test-implies-smt` RPC tests and updated `test-implies2` goldens ship with the PR and exercise the implied / decisively-not-implied / indeterminate paths (run in the integration suite, not the unit suite).

**Downstream note:** the new `indeterminate` field is additive and optional, but it does appear as `"indeterminate": true` on the affected responses — clients with a strict/closed-schema decoder must tolerate it.


ehildenb:
Tested on KEVM, measured fallbacks to Kore and performance numbers. Claude generated:

Final results: implies-reasoning fallbacks on vs hb-master `baseline` vs this branch `escalate`:
  

```
  ┌──────────────────────────────┬──────────┬──────────┬──────┬────────────────────────┐
  │            suite             │ baseline │ escalate │  Δ   │        flavour         │
  ├──────────────────────────────┼──────────┼──────────┼──────┼────────────────────────┤ 
  │ functional                   │ 21       │ 18       │ −14% │ implies                │
  ├──────────────────────────────┼──────────┼──────────┼──────┼────────────────────────┤
  │ rules                        │ 60       │ 47       │ −25% │ implies (execute flat) │
  ├──────────────────────────────┼──────────┼──────────┼──────┼────────────────────────┤
  │ DSS (partial, 19 vat claims) │ 218      │ 74       │ −66% │ execute −68%           │
  └──────────────────────────────┴──────────┴──────────┴──────┴────────────────────────┘
```
  
- Zero regressions, identical pass sets, neutral performance (warm-vs-warm 0.0%).
- The headline surprise: the biggest win is on DSS — a fix entirely in Implies.hs cut execute-flavour fallback by 68% on the two heaviest MakerDAO vat proofs (fold −74%, frob −73%), by settling the in_keys/keccak conditions that gate execution. That's the production-critical suite, so it's the most impactful result. I was initially wrong about DSS — escalate's number alone (74) looked execute-dominated-and-unmoved; only the baseline (218) revealed the −68%.

I'm not exactly sure why execute related fallbacks were cut here actually, it doesn't make a lot of sense. But overall it seems performance neutral, and handles a decent amount of implications directly in booster rather than falling back to Kore.

booster/test/rpc-integration/test-implies-smt/response-001-bound-weakening.booster-dev
jberthold, line 1: Would be good to add a README.md with an overview table of the four tests.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
```diff
@@ -169,6 +169,18 @@ data ImpliesResult = ImpliesResult
...
+    -- where booster's match check was indeterminate (e.g. an
+    -- unevaluated function blocking unification) and the
+    -- simplify-LHS / simplify-RHS retry ladder did not make
+    -- progress.  Absent ('Nothing') on every decisive outcome
+    -- ('valid = True', or 'valid = False' from a 'MatchFailed{}' /
+    -- bottom-consequent path), all error paths, and every kore-side
+    -- result.  Clients that want to distinguish "couldn't tell" from
+    -- "definitely not implied" should escalate on
+    -- @indeterminate = Just True@ and trust @valid = False@ in every
+    -- other case.
```
jberthold: Not sure about this extension... I checked and did not find a case of `indeterminate = Just False`, so it seems it is either true or absent. The meaning is a bit unclear: What does `valid = True && indeterminate = Just True` mean? Does it exist?
Maybe you could just make the `valid` field optional? More work downstream but the response has an easier semantics.

jberthold (approved):
Clear improvement, but I would think again about the `indeterminate` field.

kore-rpc-types/src/Kore/JsonRpc/Types.hs
ehildenb: Rather than make `valid` optional (which under our `OmitNothingFields` deriving would serialize "indeterminate" as an *absent* `valid`, indistinguishable from an old server), I replaced both fields with a single `status` enum: `valid | invalid | indeterminate`. Illegal states are now unrepresentable and the wire value is self-describing. Kore checks are always decisive (valid/invalid). Done in 792eebf2d; all implies goldens migrated.

booster/test/rpc-integration/test-implies-smt/response-001-bound-weakening.booster-dev
ehildenb: Done: dc4ece617 — added test-implies-smt/README.md with a table covering each test's antecedent⟹consequent, the expected booster vs kore status, and which discharge arm it exercises.

ehildenb:
@jberthold I (with Claude) added some more commits which:

1. Make the return value a tri-state `valid|indeterminate|invalid`, rather than two bools.
2. Escalates `indeterminate` cases to kore by default, and with `boosterOnly` set does not escalate to Kore.

Please have another look, the new commits have been added on top.

jberthold (approved):
Changes look good, thanks.

PR #4157, opened 2026-06-18
Ceil simplifier

ehildenb:


booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -941,18 +999,49 @@ applyEquation term rule =
...
+                -- Four cases based on (preserves-definedness attribute, definedness conditions, evaluateCeils flag):
+                --
+                --   notPreservingReasons=[]  + conditions=[]  → totally defined, proceed silently
+                --   notPreservingReasons=[]  + conditions≠[]  → user set preserves-definedness, log and proceed
+                --   notPreservingReasons≠[]  + conditions=[]  → can't prove definedness, reject
+                --   notPreservingReasons≠[]  + conditions≠[]  + evaluateCeils=False  → ceils disabled, reject
+                --   notPreservingReasons≠[]  + conditions≠[]  + evaluateCeils=True   → defer to runtime check after match
+                let notPreservingReasons = rule.computedAttributes.notPreservesDefinednessReasons
+                    definednessConditions = collectUndefinedSubterms rule.rhs
+                    preservedByAttr = null notPreservingReasons
```
jberthold, line 1012: Detail: The list being empty does not imply that the attribute is present (the other way round yes, but not this way round). There is an explicit attribute for `preserves-definedness`:
https://github.com/runtimeverification/haskell-backend/blob/7ca9fd69fea8f92a40bce04b36b7d3c961af6bf2/booster/library/Booster/Definition/Attributes/Base.hs#L91-L98
and if it is set we don't start computing the list, but the list can also just be empty.

```suggestion
                    preservedByAttr = coerce rule.attributes.preserving
```

booster/library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -941,18 +999,49 @@ applyEquation term rule =
...
+                    hasConditions = not (null definednessConditions)
+                case (preservedByAttr, hasConditions) of
```
jberthold, line 1014: Given the above, maybe pull the `preserves-definedness` logging out to cut this short (and make the logic of the `case` easier to understand).
```
if (coerce rule.attributes.preserving)
then logMessage ... 
else 
case (notPreservingReasons, definednesConditions) of
        ([], []) -> pure () -- proceed silently
        (reasons, []) -> -- no conditions to check at runtime (but reasons present), conservatively reject
        (reasons, ts) -> -- conditions present, reject unless evaluateCeils
```
The case that was previously logged (no reasons but definednessConditions) falls into the last case here but is not reached when the rule is marked.

booster/library/Booster/Pattern/Util.hs
```diff
@@ -268,6 +269,18 @@ filterTermSymbols check = cata $ \case
...
+   These represent the definedness conditions: each collected sub-term must be defined
+   (i.e. not evaluate to bottom) for the overall term to be defined.
+-}
+collectUndefinedSubterms :: Term -> [Term]
+collectUndefinedSubterms t@(SymbolApplication sym _ args)
+    | not (isDefinedSymbol sym) = [t]
+    | otherwise = concatMap collectUndefinedSubterms args
+collectUndefinedSubterms (AndTerm l r) = collectUndefinedSubterms l <> collectUndefinedSubterms r
+collectUndefinedSubterms (Injection _ _ inner) = collectUndefinedSubterms inner
+collectUndefinedSubterms _ = []
```
jberthold, line 282: There are missing cases: known elements of a `KList` or `KSet` and keys as well as values of a `KMap` must be collected as well.

PR #4159, opened 2026-06-30
ci(test): remove PAT and Cachix token dependencies from the Test workflow

Stevengre:
Removes both credential dependencies from the \`Test\` workflow, so it no longer relies on long-lived or manually-managed tokens.

## Changes

- **Drop \`JENKINS_GITHUB_PAT\` from the formatting job.**
  The job previously used the PAT to push a fourmolu commit back to the PR branch. It now runs fourmolu and fails if any \`.hs\` file changed, prompting the author to format locally and commit. No write token or push is needed.

- **Drop \`CACHIX_PUBLIC_TOKEN\` (pull-only Cachix).**
  The \`k-framework\` Cachix cache is publicly readable, so no auth token is needed to use it as a substituter. Removing \`authToken\` makes \`cachix-action\` pull-only (no push-back). Cache repopulation is left to other authenticated workflows.

## Notes

- The only remaining secret is the built-in \`GITHUB_TOKEN\` (auto-injected, short-lived, repo-scoped), used for nix flake-input rate limiting. Nothing to manage there.
- Trade-off for pull-only Cachix: PR CI no longer repopulates the cache; it still gets cache hits on read.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,10 +35,10 @@ jobs:
             substituters = http://cache.nixos.org https://cache.iog.io
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= 
 
-      - name: 'Install Cachix'
-        uses: cachix/cachix-action@v17
-        with:
-          name: k-framework
+      # - name: 'Install Cachix'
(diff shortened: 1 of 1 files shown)


juliankuners (approved):

juliankuners:
Respective follow-up PR for the cachix commit here: https://github.com/runtimeverification/haskell-backend/pull/4160

